PR #6, opened 2026-05-08
Use llc in Rust toolchain

ur4t:
Use `llc` in Rust toolchains' component `llvm-tools`, automatically installed via `rustup`.
Rust `nightly-2026-04-03` is built with LLVM 22 and `llc` here is built with NVPTX enabled.
This provides more controlled automatical environments for novices.


younes-io:
Could you describe the purpose of this PR and what it does? It'll help everyone else.

ur4t:
> Could you describe the purpose of this PR and what it does?

@younes-io Some context:

- This project uses `llc` to translate generated LLVM IR into NVPTX.
- Before: tell users to manually install LLVM, detect installed `llc-22` or `llc-21` and check their feature completeness.
- After: use `llc` in Rust toolchains' component `llvm-tools`, automatically installed via `rustup`. Rust `nightly-2026-04-03` is built with LLVM 22 and `llc` here is built with NVPTX enabled.
- Pros: more controlled automatical environment for novices.

I'll update PR description later.

nihalpasham:
Thanks for looking into the setup path here. I don’t think we can take this change as-is: the assumption that Rust’s llvm-tools component provides a usable `llc` is not true in practice, and this PR also removes the existing `llc-22` / `llc-21` PATH probing which would make `cargo oxide doctor` report success even when no usable `llc` is available, and could break PTX generation for users who currently have a working LLVM install.

I’m going to leave this open for now in case I’m missing something, but I’d ideally want this reworked around a verified, reliable `llc` discovery path before we consider merging. Though, if the goal is smoother onboarding, I’d prefer a smaller PR that improves diagnostics/docs without removing the current working fallback.

ur4t:
> the assumption that Rust’s `llvm-tools` component provides a usable `llc` is not true in practice

I have been using it locally and it works well, at least with `nightly-2026-04-03` and `1.96.0-beta`.

> this PR also removes the existing `llc-22` / `llc-21` PATH probing which would make `cargo oxide doctor` report success even when no usable `llc` is available, could break PTX generation for users who currently have a working LLVM install.

Yes, this PR assumes that `rustup` works well to install the usable `llc` as expected and users who want to use other LLVM installs are supposed to specify `CUDA_OXIDE_LLC` intentionally.

> I’d prefer a smaller PR that improves diagnostics/docs without removing the current working fallback.

Argee, this PR might be too radical. I will bring `llc-22` / `llc-21` PATH probing back for smooth transition. I thought it is pure maintenance burden (too much issue space) hah.

ur4t:
@nihalpasham This PR is minimalized and ready. Feel free to suggest further modifications.

nihalpasham (approved):
Thanks for the PR @ur4t, LGTM 👍🏾

Just two additions: pushed two follow-up commits on top:

- **`ab64415`** Merge `origin/main` — auto-clean, picks up 20 commits since fork.
- **`86c96dc`** Closes the doctor + docs gap: `cargo oxide doctor` was still probing only `[llc-22, llc-21, llc]` on PATH, so a user on just `rustup component add llvm-tools` (the exact onboarding case this PR enables) saw a confusing `✗ llc not found` even though the pipeline would have built fine. Doctor now mirrors the pipeline's resolution order exactly, and the book's lowering-pipeline / building-from-source / architecture-overview chapters plus the atomics README all reflect rustup as the primary source.

PR #7, opened 2026-05-08
  Fix CUDA event binding compatibility and arch forwarding

mohamedsamirx:
## Summary

This PR fixes two setup/runtime issues found while validating cuda-oxide on a CUDA 12.4 system with an sm_75 NVIDIA T1200 GPU.

  - Add a CUDA Driver API compatibility wrapper for event elapsed timing.
  - Use `cuEventElapsedTime_v2` when the installed CUDA headers expose it.
  - Fall back to `cuEventElapsedTime` for older CUDA 12.x headers such as CUDA 12.4.
  - Forward explicit `cargo oxide --arch` values to the normal PTX generation path through `CUDA_OXIDE_TARGET`, while preserving backend auto-target selection when `--arch` is omitted.

## Motivation
CUDA 12.4 headers generate a `cuEventElapsedTime` binding, not `cuEventElapsedTime_v2`, which caused `cuda-core` to fail to compile.

Also, `cargo oxide run vecadd --arch sm_75` accepted the `--arch` flag but did not apply it for normal PTX output. The generated PTX still targeted the auto-selected architecture, causing PTX JIT failures on sm_75 hardware. With this change, `--arch sm_75` correctly generates `.target sm_75`.

The change preserves the existing default behavior when `--arch` is omitted: the backend still performs auto-target selection.

## Testing

  ```bash
  cargo fmt --check
  cargo check
  cargo clippy -p cuda-bindings -p cuda-core -p cargo-oxide -- -D warnings
  cargo test -p cuda-bindings -p cuda-core -p cargo-oxide
  cargo oxide doctor
  unset CUDA_OXIDE_TARGET && cargo oxide build vecadd
  unset CUDA_OXIDE_TARGET && cargo oxide run vecadd --arch sm_75

  When --arch is omitted, the generated vecadd.ptx keeps the backend-selected default:

  .target sm_80

  With --arch sm_75, the generated vecadd.ptx contains:

  .target sm_75

  and the example prints:

  SUCCESS: All 1024 elements correct!

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -149,7 +149,7 @@ pub fn codegen_run(
     forward_env_var(&mut cmd, "CUDA_OXIDE_DUMP_MIR");
     forward_env_var(&mut cmd, "CUDA_OXIDE_DUMP_LLVM");
 
-    apply_output_mode(&mut cmd, dlto, emit_nvvm_ir, target_arch);
+    apply_output_mode(&mut cmd, dlto, emit_nvvm_ir, arch, target_arch);
     apply_ld_library_path(&mut cmd);
 
     if let Some(bin) = bin {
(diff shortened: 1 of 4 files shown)


nihalpasham:
Landed via #15 with a verified maintainer signature while preserving your authorship. For future PRs, it would be great if you could use verified commit signatures so branch protection can merge them directly. Thanks again for the fix!

mohamedsamirx:
Sure, thank you

Issue #8, opened 2026-05-08
Build error on aarch64-based systems


Barrow099:
Running the provided install command (`cargo install  --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide`), results in the following error:
```
error[E0308]: mismatched types
    --> crates/cuda-core/src/context.rs:225:44
     |
 225 |             cuda_bindings::cuDeviceGetName(buf.as_mut_ptr(), buf.len() as c_int, self.cu_device)
     |             ------------------------------ ^^^^^^^^^^^^^^^^ expected `*mut u8`, found `*mut i8`
     |             |
     |             arguments to this function are incorrect
     |
     = note: expected raw pointer `*mut u8`
                found raw pointer `*mut i8`
```

The root cause seems to be in the way the FFI types are defined inside the Rust core library, as the referenced function binding uses `std::os::raw::c_char`. The definition of `c_char` can be traced back to `core/src/ffi/primitives.rs` file. The following code snippet shows the definition based on the target platform (os and architecture):
```rust
all(
            not(windows),
            not(target_vendor = "apple"),
            not(target_os = "vita"),
            any(
                target_arch = "aarch64",
                target_arch = "arm",
                target_arch = "csky",
                target_arch = "hexagon",
                target_arch = "msp430",
                target_arch = "powerpc",
                target_arch = "powerpc64",
                target_arch = "riscv32",
                target_arch = "riscv64",
                target_arch = "s390x",
                target_arch = "xtensa",
            )
        ) => {
            pub(super) type c_char = u8;
        }
        // On every other target, c_char is signed.
        _ => {
            pub(super) type c_char = i8;
        }
```

Based on this, on aarch64 based platforms, `c_char` is `u8`, not `i8`, thus resulting in the error above.

sempervictus:
Ran into something similar with nccl in vllm.rs - seems a somewhat common problem with rust code on Spark of all things :-)

PR #9, opened 2026-05-08
Fix buffer type in CudaContext::device_name

Barrow099:
Fixes the problem described in https://github.com/NVlabs/cuda-oxide/issues/8, by leaving the type of the array to be inferred by the compiler to the correct type on the target platform.

### DIFF
--- a/crates/cuda-core/src/context.rs
+++ b/crates/cuda-core/src/context.rs
@@ -220,7 +220,7 @@ impl CudaContext {
     /// decoded UTF-8 string with any trailing NULs stripped.
     pub fn device_name(&self) -> Result<String, DriverError> {
         self.bind_to_thread()?;
-        let mut buf = [0i8; 256];
+        let mut buf = [0; 256];
         unsafe {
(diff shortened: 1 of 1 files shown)


nihalpasham (approved):
Thanks for the fix, and for the clear issue writeup. I checked this locally: the existing `i8` buffer shape reproduces the `c_char` mismatch on `aarch64-unknown-linux-gnu`, while the inferred `[0; 256]` version lets Rust choose the target-correct `c_char` type. 

- `cargo check -p cuda-core` and `cargo test -p cuda-core --no-run` also pass on the host target.

This is a small, correct portability fix. I’m going to run the repo CI and merge once it’s green.

nihalpasham:
@Barrow099 — For future PRs, it would be great if you could use verified commit signatures. I landed this via #14 with a verified maintainer signature while preserving your authorship. Thanks again for the fix!

Barrow099:
> @Barrow099 — For future PRs, it would be great if you could use verified commit signatures. I landed this via #14 with a verified maintainer signature while preserving your authorship. Thanks again for the fix!

Sure, thank you for the suggestion and the quick merge.

PR #17, opened 2026-05-09
fix(cuda-bindings): resolve build failure by checking CUDA_HOME

### LINKED ISSUE #16: build.rs does not check CUDA_HOME (only checks CUDA_TOOLKIT_PATH)
The build script fails in some standard environments because it ignores CUDA_HOME. PR incoming to add fallback detection.

JonatanLindh:
Resolves #16 

### Problem
In environments where `CUDA_TOOLKIT_PATH` is not explicitly set but standard environment variables like `CUDA_HOME` are, `build.rs` silently falls back to `/usr/local/cuda`. If this directory doesn't exist or contains a different CUDA version, `std::fs::read_to_string` silently fails. 

This causes the build script to skip enabling the `cuda_has_cuEventElapsedTime_v2` feature flag, resulting in a cryptic `cannot find function` compilation error downstream on CUDA 13 (when CUDA_HOME is the correct location).

### Changes
- `cuda_toolkit_dir()` now checks `CUDA_HOME before falling back to `/usr/local/cuda`.
- Added `cargo:rerun-if-env-changed=CUDA_HOME`
- Replaced the silent `.unwrap_or(false)` with a `cargo:warning`. If the script cannot read `cuda.h`, it now emits a visible warning during `cargo build` explaining exactly which path failed.

### Checklist
- [x] I have signed my commits (DCO).
- [x] Code is formatted with `cargo fmt`.
- [x] Checked with `cargo clippy`.

### DIFF
--- a/crates/cuda-bindings/build.rs
+++ b/crates/cuda-bindings/build.rs
@@ -5,10 +5,13 @@
 
 use std::{env, error::Error, path::Path, path::PathBuf, process::exit};
 
-/// Returns the CUDA toolkit install root: `CUDA_TOOLKIT_PATH` if set, otherwise `/usr/local/cuda`.
-/// Used for include paths, library search paths, and bindgen’s Clang configuration.
+/// Returns the CUDA toolkit install root: `CUDA_TOOLKIT_PATH` or `CUDA_HOME` if set,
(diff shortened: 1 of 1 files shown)


nihalpasham (approved):
Thanks for the fix. Approved as-is

PR #19, opened 2026-05-09
Make DeviceBuffer transfers require DeviceCopy

mohamedsamirx:
## Summary

This PR adds an unsafe `DeviceCopy` marker trait for types that can be safely copied between host and device memory as raw bytes, then constrains `DeviceBuffer<T>` transfer APIs to `T: DeviceCopy`.

The change prevents safe Rust code from using `DeviceBuffer` with host-owned or non-POD types such as `String` or `Vec<T>`.

## Motivation

`DeviceBuffer<T>` currently copies raw bytes to and from CUDA device memory. Some APIs then materialize host values from those bytes using `Vec::set_len`.

That is only sound for plain device-copyable data. Before this change, the safe API accepted any `T`, including types with Rust ownership or drop semantics.

For example, this could compile:

```rust
let _ = DeviceBuffer::<String>::zeroed(stream, 1);
```
That is unsound because zeroed or device-written bytes are not a valid String. If such a value is copied back and dropped, Rust may attempt to free an invalid pointer.

## Failure Demonstration

I first added a compile-fail doctest showing that DeviceBuffer<String> should be rejected:

 
```compile_fail
/// # use cuda_core::{CudaStream, DeviceBuffer};
/// # fn rejects_non_device_copy(stream: &CudaStream) {
/// let _ = DeviceBuffer::<String>::zeroed(stream, 1);
/// # }
```

Before the fix, the doctest failed because the invalid code still compiled:
Test compiled successfully, but it's marked `compile_fail`.

After adding the DeviceCopy bound, the same doctest passes because String does not implement DeviceCopy.

## Implementation

  - Add unsafe marker trait DeviceCopy: Copy.
  - Implement DeviceCopy for primitive POD-like numeric types, arrays, tuples, raw pointers, and half-precision float types.
  - Move DeviceBuffer<T> allocation and transfer helpers behind impl<T: DeviceCopy>.
  - Keep raw-parts and metadata accessors available for all T.
  - Document the safety contract, including the requirement that all-zero initialization is valid because DeviceBuffer::zeroed writes zero bytes.
  - Add a compile-fail doctest covering DeviceBuffer<String>.

## Testing

  cargo fmt --check
  cargo clippy -p cuda-core -p cuda-host -p cuda-async -- -D warnings
  cargo test -p cuda-core
  cargo check
  cargo oxide build f16_stress --arch sm_75

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -332,6 +332,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "cuda-bindings",
+ "half",
  "oxide-artifacts",
 ]
 
--- a/crates/cuda-core/Cargo.toml
+++ b/crates/cuda-core/Cargo.toml
@@ -10,6 +10,7 @@ repository.workspace = true
 [dependencies]
 cuda-bindings = { path = "../cuda-bindings" }
 anyhow = { workspace = true }
+half = { workspace = true }
 oxide-artifacts = { workspace = true, features = ["object-read"] }
 
 [dev-dependencies]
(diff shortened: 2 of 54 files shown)


nihalpasham:
Thanks for the PR! The `DeviceCopy` bound is the right direction here: `DeviceBuffer<T>`'s safe transfer helpers are byte-copying device memory, so they should only accept element types that can safely make that round trip as raw bytes.

I pushed a small follow-up commit on top of your branch to clarify the `DeviceCopy` safety docs and refresh the generated standalone-example lockfiles after the new `cuda-core -> half` dependency.

Validation passed locally:

- `just smoketest` (`49/49` examples)
- `cargo clippy --all-targets -- -D warnings` in `crates/rustc-codegen-cuda`
- clippy over every standalone example workspace
- unit-test CI matrix from `unit-tests.yml`
- targeted `cargo test -p cuda-core` / `cargo clippy -p cuda-core -- -D warnings`

One note from review: cuda-oxide already does dynamic struct layout matching for Rust host/device structs, so `DeviceCopy` should stay focused on value safety rather than requiring `#[repr(C)]` for normal cuda-oxide kernels.

nihalpasham (approved):
Approved

PR #20, opened 2026-05-09
Propagate cuda-async synchronization errors

mohamedsamirx:
## Summary

  This PR changes `cuda-async` synchronous execution and async deallocation paths to report CUDA errors instead of panicking.

  - Convert `DeviceOperation::sync_on` stream synchronization failures into `DeviceError::Driver`.
  - Avoid panicking from `DeviceBox::drop` when deallocator stream lookup or `cuMemFreeAsync` enqueue fails.
  ## Motivation

  `DeviceOperation::sync_on` previously called:

  ```rust
  stream.synchronize().expect("Synchronize failed.");
```
  That means a CUDA stream synchronization failure could unwind through a safe API instead of being returned through the existing Result<_, DeviceError> surface.

  Before this change, the failure behavior was effectively:

  thread panicked at 'Synchronize failed.'

  This is especially undesirable because synchronization is where asynchronous CUDA errors are often surfaced.

  DeviceBox::drop also panicked if the async deallocator stream could not be accessed. Drop paths cannot return errors, so they should avoid panicking and report cleanup failures best-effort instead.

  ## Implementation

  - Add a small internal finish_sync helper that combines the operation result with the stream synchronization result.
  - Return Err(DeviceError::Driver(_)) when stream synchronization fails.
  - Preserve the existing behavior of synchronizing after operation execution.
  - Change DeviceBox::drop to log deallocator lookup or async free enqueue failures instead of panicking.

  ## Failure Demonstration

  A regression test covers the synchronization failure path directly:

  let driver_error = cuda_core::DriverError(cuda_bindings::cudaError_enum_CUDA_ERROR_INVALID_VALUE);
  let result = finish_sync::<u32>(Ok(7), Err(driver_error));
  assert_eq!(result, Err(DeviceError::Driver(driver_error)));

  Before the fix, the equivalent sync_on path used .expect("Synchronize failed."), so this error case would panic instead of returning DeviceError.

  ## Testing

  cargo fmt --check
  cargo test -p cuda-async
  cargo clippy -p cuda-async -- -D warnings
  cargo check

### DIFF
--- a/crates/cuda-async/src/device_box.rs
+++ b/crates/cuda-async/src/device_box.rs
@@ -20,6 +20,7 @@ use crate::error::DeviceError;
 use crate::launch::{AsyncKernelLaunch, KernelArgument};
 use cuda_bindings::CUdeviceptr;
 use cuda_core::memory::free_async;
+use std::io::{self, Write};
 use std::marker::PhantomData;
 
 /// Non-owning, `Copy` handle to a typed device allocation.
@@ -104,7 +105,9 @@ impl<T: Send + ?Sized> Drop for DeviceBox<T> {
         match result {
             Ok(Ok(())) => {}
             Ok(Err(err)) | Err(err) => {
-                eprintln!(
+                let mut stderr = io::stderr().lock();
+                let _ = writeln!(
+                    stderr,
                     "cuda-async: failed to enqueue async free for device pointer on device_id={}: {}",
                     self.device_id, err
                 );
--- a/crates/cuda-async/src/device_box.rs
+++ b/crates/cuda-async/src/device_box.rs
@@ -16,6 +16,7 @@
 //! streams that reference the allocation before dropping the box.
 
 use crate::device_context::with_deallocator_stream;
+use crate::error::DeviceError;
 use crate::launch::{AsyncKernelLaunch, KernelArgument};
 use cuda_bindings::CUdeviceptr;
 use cuda_core::memory::free_async;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/cuda-async/src/device_operation.rs [hunk 1]
```diff
@@ -236,11 +236,18 @@ pub trait DeviceOperation:
...
-        res
+        finish_sync(res, stream.synchronize())
     }
 }
 
+fn finish_sync<T>(
+    operation_result: Result<T, DeviceError>,
+    synchronize_result: Result<(), cuda_core::DriverError>,
+) -> Result<T, DeviceError> {
+    synchronize_result.map_err(DeviceError::Driver)?;
```
nihalpasham, line 248: I think `finish_sync` should preserve `operation_result` if execution already failed. With the current ordering, an operation error followed by a synchronization error returns the synchronization error and masks the first known failure.

Could we either make operation errors take precedence, or add a test documenting that sync errors intentionally win when both fail?

For example:

```rust
operation_result?;
synchronize_result.map_err(DeviceError::Driver)
```

That gives:

```text
op error + sync error -> op error
op ok + sync error    -> sync error
```

_One nuance:_ that code is correct inside `finish_sync` because `stream.synchronize()` has already been called by the caller before entering the helper. So it changes only which already-collected result wins, not whether synchronization happens.

mohamedsamirx: I updated `finish_sync` so the already-collected operation result takes precedence over a synchronization error when both fail.

  The stream synchronization still happens before entering `finish_sync`; this I also added a regression test for the combined failure case:

  - operation error + sync error -> operation error
  - operation ok + sync error -> sync error

  Validation:

  ```bash
  cargo fmt --check
  cargo test -p cuda-async
  cargo clippy -p cuda-async -- -D warnings
  cargo check
```
@nihalpasham

File: crates/cuda-async/src/device_box.rs [hunk 2]
```diff
@@ -87,23 +89,29 @@ unsafe impl<T: Send + ?Sized> Send for DeviceBox<T> {}
...
-                    "Failed to free device pointer on device_id={}",
-                    self.device_id
-                )
+                free_async(self.cudptr, stream.cu_stream()).map_err(DeviceError::Driver)
             })
+        };
+
+        match result {
+            Ok(Ok(())) => {}
+            Ok(Err(err)) | Err(err) => {
```
nihalpasham, line 107: To avoid using `eprintln!` in `DeviceBox::drop`, I added this. Since this drop path is explicitly trying not to panic, the new version writes to locked stderr with `writeln!` and ignores any stderr write failure as best-effort cleanup reporting.

nihalpasham (approved):
Merged. 

Thanks for the PR :)

PR #22, opened 2026-05-09
cargo-oxide: allow overriding RUSTFLAGS via env variable

tdortman:
I noticed in the process of writing a nix devShell for `cuda-oxide` that there doesn't seem to be any way to influence the flags that are passed to rustc.

`cuda-bindings` seems to have hardcoded paths for the various stub libraries that should be used for builds, allowing users to add things to `RUSTFLAGS` would in turn let everyone who is using individual redistributables (I know NixOS encourages this, not sure about other distros) to just point to the stub in `cuda_cudart` directly and not have to eat the cost of using the entire SDK.

I'm sure there are other reasons why someone might want to add things to `RUSTFLAGS` but this is one at least.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -799,8 +799,18 @@ fn resolve_example_dir(ctx: &Context, example: &str) -> PathBuf {
 ///
 /// Always includes `-Z codegen-backend`, `-C opt-level=3`, disabled debug
 /// assertions, suppressed JumpThreading (prevents barrier duplication), and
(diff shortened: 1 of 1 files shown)


tdortman:
I actually don't need this merged anymore as it turns out I can just make my own little toolkit with the few required parts but it probably doesn't hurt to still include this functionality.

nihalpasham (approved):
Approved. 

Thanks, this is a useful quality-of-life improvement. I added a tiny follow-up to document the appended `RUSTFLAGS` behavior and cover the ordering with tests.

PR #23, opened 2026-05-09
Respect nvcc host compiler overrides in FFI examples

lindestad:
## Summary

Some FFI examples call `nvcc` directly from shell helper scripts. On systems where the default host compiler is newer than the installed CUDA Toolkit supports, those examples fail unless users know to rely on `nvcc`-specific escape hatches like `NVCC_PREPEND_FLAGS`.

This makes the FFI scripts honor `NVCC_CCBIN`, falling back to `CUDAHOSTCXX`, and passes the selected compiler to `nvcc` as `-ccbin=...`. The selected compiler is printed when set, so the build log makes it clear which host compiler `nvcc` is using.

This updates the direct `nvcc` call sites in:

- `device_ffi_test`
- `cpp_consumes_rust_device`
- `mathdx_ffi_test`

It also documents the override in the matching READMEs and uses `CUDA_HOME` for `device_ffi_test`'s CCCL include path instead of hardcoding `/usr/local/cuda`.

## Test plan

Tested on RTX 5080 (`sm_120`), CUDA 13.2, Fedora 44, with `NVCC_CCBIN=/usr/bin/g++-15`.

- `NVCC_CCBIN=/usr/bin/g++-15 scripts/smoketest.sh --keep-logs -o '^(device_ffi_test|cpp_consumes_rust_device|mathdx_ffi_test)$'`
- Force rebuilt `device_ffi_test/tools`, then:
  `NVCC_CCBIN=/usr/bin/g++-15 cargo oxide run device_ffi_test --emit-nvvm-ir --arch=sm_120`
- `cd crates/rustc-codegen-cuda/examples/cpp_consumes_rust_device/cuda-caller && NVCC_CCBIN=/usr/bin/g++-15 ./run_test.sh sm_120`
- `bash -n` on all edited scripts
- `cargo fmt --check`

### DIFF
--- a/crates/rustc-codegen-cuda/examples/cpp_consumes_rust_device/README.md
+++ b/crates/rustc-codegen-cuda/examples/cpp_consumes_rust_device/README.md
@@ -95,6 +95,13 @@ Device: NVIDIA GeForce RTX 5090 (sm_120)
 - LTOIR tools from `device_ffi_test/tools/` (built automatically by `run_test.sh`)
 - Blackwell+ GPU (sm_100+) — LTOIR requires NVVM 20 dialect
 
+If your default host compiler is newer than the CUDA Toolkit supports, set
+`NVCC_CCBIN` or `CUDAHOSTCXX` before running the example:
+
+```bash
+NVCC_CCBIN=/usr/bin/g++-15 cargo oxide run cpp_consumes_rust_device --emit-nvvm-ir --arch=sm_120
+```
+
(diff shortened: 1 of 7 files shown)


nihalpasham (approved):
Approved as-is

nihalpasham:
Thanks, this looks good. I confirmed the `NVCC_CCBIN` path locally on the runnable FFI examples.

PR #26, opened 2026-05-10
Embedded Kernels & typed launch interface

josephglanville:
## Summary

This PR makes `cuda-oxide` host binaries self-contained by embedding generated device artifacts into
the final executable, then adds a typed host API for loading and launching those embedded kernels.

From a user perspective, CUDA kernels no longer need to be loaded from loose `.ptx` files at runtime,
and launches can now be written as ordinary Rust method calls with type-checked arguments,
completion, and refactoring support.

- Adds `#[cuda_module]` for inline kernel modules.
- Generates a typed `LoadedModule` with one launch method per `#[kernel]`.
- Loads embedded device artifacts with `kernels::load(&ctx)` or `kernels::load_async(device_id)`.
- Supports sync, borrowed async, and owned async launch APIs.
- Moves generated PTX plus entry metadata into an embedded `.oxart` section, so examples and apps do not need to find sidecar `.ptx` files at runtime.

### Syntax highlighting and completions in IDEA

Screenshot 2026-05-10 at 5 48 22 am

### Sync Launch
```rust
use cuda_host::cuda_module;

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    pub fn vecadd(a: &[f32], b: &[f32], mut c: DisjointSlice<f32>) {
        let idx = thread::index_1d();
        if let Some(c_elem) = c.get_mut(idx) {
            *c_elem = a[idx.get()] + b[idx.get()];
        }
    }
}

let a = DeviceBuffer::from_host(&stream, &a_host)?;
let b = DeviceBuffer::from_host(&stream, &b_host)?;
let mut c = DeviceBuffer::<f32>::zeroed(&stream, a_host.len())?;

module.vecadd(
    &stream,
    LaunchConfig::for_num_elems(a_host.len() as u32),
    &a,
    &b,
    &mut c,
)?;
```
### Async Launch
```rust
use cuda_async::device_operation::DeviceOperation;

let module = kernels::load_async(0)?;

module
    .vecadd_async(
        LaunchConfig::for_num_elems(n as u32),
        &a_dev,
        &b_dev,
        &mut c_dev,
    )?
    .sync()?;

// or `.await?`
```
### Owned Async Launch
```rust
let (a_dev, b_dev, c_dev) = module
    .vecadd_async_owned(
        LaunchConfig::for_num_elems(n as u32),
        a_dev,
        b_dev,
        c_dev,
    )?
    .await?;
```


nihalpasham:
@josephglanville - thanks for this, this is a great addition.

_One process note:_ while the PR is a strong addition, I’d appreciate it if future PRs are broken into smaller pieces. This one combined embedded artifacts, typed launch APIs, async launch updates, example migrations, and launch ergonomics, which made review and conflict resolution harder than it needed to be.

A few pieces were missing and I filled them in while resolving the branch:

- Updated the book and crate/example READMEs so `#[cuda_module]` is documented as the primary launch path.
- Updated `cargo oxide new` so both sync and async scaffolds use the typed module API by default.
- Kept `load_kernel_module` / `cuda_launch!` documented as lower-level escape hatches rather than deprecating them.
- Added a couple more examples to strengthen coverage around typed launches, generics, and the lower-level manual launch path.
- Cleaned up the CI surface so the expanded checks pass locally.

Also, thanks for enhancing the async API, that is a great addition; I had been meaning to fortify that area but hadn’t had time.

**Important point on the design boundary:** `#[cuda_module]` should own embedding/loading ergonomics and typed launch generation only. It should not decide PTX vs LTOIR vs fatbin or multi-arch packaging policy. That policy belongs in compiler/runtime/artifact layers.

_Follow-up I think we can handle separately:_ typed support for cooperative/cluster launch, so those examples can eventually move to the typed API too as this PR is already too big.

nihalpasham (approved):
approved

Issue #21, opened 2026-05-09
PTX Codegen Failure: StepBy struct lowering (invalid cast opcode) exposed by missing Pat type translation


JonatanLindh:
**Description**
While experimenting with CUDA Oxide, I attempted to use a standard `(start..end).step_by(step)` iterator inside a kernel. This initially failed due to a missing translation arm for `TyKind::Pat` in the MIR importer.

After adding a local patch to strip the pattern metadata and translate the base type, the frontend passed, but it revealed a deeper LLVM PTX code generation failure related to how the `StepBy` struct is lowered.

**1. The "Hidden" Bug (`Pat` missing)**
Initially, compiling a `step_by` loop fails because the MIR importer doesn't handle the `Pat` types used by the iterator's internal state. I patched this locally in `crates/mir-importer/src/translator/types.rs`:

```rust
// ...
rustc_public::ty::TyKind::RigidTy(rustc_public::ty::RigidTy::Array(elem_ty, len_const)) => { /* ... */ }

// ADDED THIS ARM
rustc_public::ty::TyKind::RigidTy(rustc_public::ty::RigidTy::Pat(base_ty, _pat)) => {
    // Patterns don't have a distinct runtime type, so we translate to the base type.
    translate_type(ctx, &base_ty)
}

rustc_public::ty::TyKind::RigidTy(rustc_public::ty::RigidTy::Slice(elem_ty)) => { /* ... */ }
// ...

```

(Note: I have the frontend Pat translation patch ready locally. Should I open a PR for that, or do you prefer to keep the Unsupported construct frontend error in place as a guardrail until the backend StepBy lowering is fully implemented?)

**2. The Revealed Error (PTX Codegen Failure)**
With the frontend satisfied, the compiler proceeds to LLVM IR generation, where `llc` crashes. It appears the optimizer fails to lower the `StepBy` state machine struct correctly, resulting in an illegal `bitcast` between a scalar `i64` and an aggregate struct:

```text
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: PTX generation failed: CUDA_OXIDE_LLC=/nix/store/.../bin/llc failed:
/nix/store/.../bin/llc: error: /nix/store/.../bin/llc: [...].ll:126:18: error: invalid cast opcode for cast from 'i64' to '{ i8, { { i64 } } }'
         %v8 = bitcast i64 %v7 to { i8, { { i64 } } }

```

**Minimal Reproducible Example**
[...]

PR #27, opened 2026-05-10
fix(codegen): convert three silent miscompiles to hard build errors

alejandro-soto-franco:
Three paths through the cuda-oxide codegen pipeline silently dropped operations from the generated PTX rather than emitting a diagnostic. A user kernel exercising any of them compiled cleanly, the PTX loaded, the kernel ran, and the dropped operation produced a wrong result with no warning. This commit converts each to a `pliron::input_err!` so the build fails at codegen with a named diagnostic and adds a regression test crate per fix.

1. wgmma.mma_async lowering (mir-lower) `crates/mir-lower/src/convert/intrinsics/wgmma.rs::convert_mma` previously emitted `// wgmma.mma placeholder` as an inline-asm comment and erased the op. Calls from a `#[kernel]` to `cuda_device::wgmma::wgmma_mma_*` produced PTX where every multiply-accumulate was zero without warning. The companion sync ops (`wgmma.fence`, `wgmma.commit_group`, `wgmma.wait_group`, `make_smem_desc`) lower correctly and are unaffected. Regression test: `examples/error_wgmma_mma_unimplemented/`.

2. Unhandled MIR statement kinds (mir-importer) `crates/mir-importer/src/translator/statement.rs` only handled four `StatementKind` variants and silently `Ok(prev_op)`-d everything else. The catch-all swallowed two variants with observable runtime effect — `Intrinsic(NonDivergingIntrinsic::CopyNonOverlapping(_))` (the user-visible memcpy emitted by `core::ptr::copy_nonoverlapping`) and `SetDiscriminant`. The match is now exhaustive on the 12-variant enum: codegen-irrelevant variants (`FakeRead`, `Retag`, `PlaceMention`, `AscribeUserType`, `Coverage`, `ConstEvalCounter`, `Intrinsic(Assume)`) skip; the two with runtime effect error out. New variants added upstream now produce a non-exhaustive-match compile error in the importer rather than another silent fall-through. Regression test: `examples/error_copy_nonoverlapping_unhandled/`.

3. Drop terminator on non-Copy types (mir-importer) `translate_drop` in `crates/mir-importer/src/translator/terminator/mod.rs` lowered every `TerminatorKind::Drop` to a goto, regardless of whether the dropped place's type had drop glue. The function's own doc-comment claimed "On GPU, Drop is a no-op because: Only Copy types are supported" but nothing enforced that. A `#[kernel]` (or any function reachable from one) that owned a non-`Copy` type with a `Drop` impl had its destructor body silently elided. The fix threads `body` into `translate_drop` (as `translate_assert` and friends already take it), queries the dropped place's type via `place.ty(body.locals())`, and returns `TranslationErr::unsupported(...)` naming the type. The error message points at the `core::mem::ManuallyDrop` workaround for users who genuinely want to suppress drop glue on a Copy-shaped value. Regression test: `examples/error_drop_glue/`.

The three regression-test example crates follow the existing `error/` example template and are classified into a new `ERROR_EXAMPLES` array in `scripts/smoketest.sh` (mirroring `TCGEN05_EXAMPLES` / `WGMMA_EXAMPLES` / `LTOIR_EXAMPLES`) so the existing `error` build-must-fail verdict rule applies.

Verified locally on RTX 5060 / sm_120 host with CUDA 13.1.115, nightly-2026-04-03, LLVM 21.1.8: a representative subset of existing examples (`atomics`, `array_index`, `barrier`, `cluster`, `compiler_features`, `coop_groups_demo`, `cross_crate_kernel`, `debug`, `device_global`, `dynamic_smem`, `host_closure`, `async_mlp`, `wgmma`) all still build clean; each new error example fails to build with the expected diagnostic; classify() routes all four `error*` example names to the `error` category.

Adjacent: Issue #21 has the same "mir-importer fall-through silently miscompiles" pattern.


nihalpasham (approved):
Approved. 

Thanks, this looks good to me. I added a small follow-up test on top of the PR: `error_set_discriminant_unhandled`, which directly exercises the `SetDiscriminant` fail-loud path via `custom_mir`.

Validated with:

```sh
scripts/smoketest.sh -x -v '^error_set_discriminant_unhandled$'

PR #28, opened 2026-05-10
during cargo oxide new adjust the name passed to load_kernel_module by converting - to _, to prevent NoArtifact

mutantbob:
`cargo oxide new a-b` generates a main.rs containing the code `load_kernel_module(&ctx, "a-b")` .  This fails at runtime because the generated .ptx file will be named a_b.ptx.

This adjustment seems to solve the problem for me.  I do not know if similar problems exist elsewhere in the code.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -1002,6 +1002,8 @@ cuda-core = {{ git = "{GIT_REPO}" }}
         )
     };
 
+    let adjusted_file_name = name.replace("-", "_");
+
     let main_rs = if async_mode {
         format!(
             r#"use cuda_device::{{kernel, thread, DisjointSlice}};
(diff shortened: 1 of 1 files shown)


mutantbob:
I like the concept from https://github.com/NVlabs/cuda-oxide/pull/26 as a more comprehensive fix for my problem.

nihalpasham:
Thanks for catching this. 

This is now fixed on `main` through PR #26. The `cargo oxide new` templates no longer generate direct `load_kernel_module("{name}")` calls; they now use the typed `#[cuda_module]` loader for both sync and async projects. The embedded artifact producer/loader path was also aligned around `CARGO_PKG_NAME`, so hyphenated package names no longer hit this generated-scaffold mismatch.

Closing this as superseded by #26, but the issue you identified was valid.

PR #33, opened 2026-05-11
Dedupe SharedArray/DynamicSharedArray warnings by DefId

menaman123:
## Summary
When running `cargo oxide run gemm_sol` I get 140+ warnings. When looking into it function warn once uses span which takes into account the line location to create a unique key instance of the operand. However when running some tests on the actual unique MLIR of SMEM_A0 which is the MLIR causing the warning there is only 55 unique instances. By using DefId which assigns a unique id which would be perfect for this case. If warn once is genuinely for warning once at each unique instance of this operation then we should use DefId. Makes the logs less noisey as well.

### DIFF
--- a/crates/rustc-codegen-cuda/src/collector.rs
+++ b/crates/rustc-codegen-cuda/src/collector.rs
@@ -775,6 +775,7 @@ impl<'tcx> DeviceCollector<'tcx> {
             let static_name = self.tcx.def_path_str(static_def_id);
             self.warn_once(
                 "shared-array-static",
+                static_def_id,
                 span,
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. I believe this has been addressed on `main` already by commit `54f45e6` (`Fix index_2d soundness: stride-consistency gap`), which now emits the shared-memory launch warning only once per program.

I’m going to close it as superseded.

menaman123:
Thanks @nihalpasham quick question. So for warn once it's once per program or once per unique instance of the MLIR per program?

PR #38, opened 2026-05-11
docs: add LLVM 21 installation guide in Debian/Ubuntu system because default package repositories do not contain LLVM 21

vito-lin-dev:
The default Ubuntu package repositories do not contain LLVM 21, causing the apt-get install command to fail with a 'no installation candidate' error.

Because some case it shows:
```
/content# sudo apt install llvm-21
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package llvm-21 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'llvm-21' has no installation candidate
/content# llc-21 --version | grep nvptx
-bash: llc-21: command not found
```

and install the official packages from the LLVM repository could solve the problem

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,12 @@ nvcc --version
 # Ubuntu/Debian
 sudo apt install llvm-21
 
+For most Debian-based systems, you can install the official packages from the LLVM repository:
+# Setup the official LLVM repository
+sudo apt-get install -y lsb-release wget software-properties-common gnupg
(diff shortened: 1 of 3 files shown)


nihalpasham (approved):
Merged 

Thanks, this is useful. I made some edits to keep the explanatory text outside the shell code fences, so the install commands can be copied directly.

Issue #36, opened 2026-05-11
commit c91cf17f9c5962ffb844cdbedf1163c716d12cd2 fails to compile the template application


mutantbob:
I switched my cargo-oxide tool to commit c91cf17f9c5962ffb844cdbedf1163c716d12cd2 using `CUDA_TOOLKIT_PATH=/opt/cuda cargo install --path ~/vendor/cuda-oxide/crates/cargo-oxide cargo-oxide`

I rebuilt my toy app and it exploded.

I rebuilt the template app 
```
cargo oxide new a-b
(cd a-b && CUDA_TOOLKIT_PATH=/opt/cuda CUDA_OXIDE_LLC=/usr/lib/llvm/21/bin/llc cargo oxide run)
```
and it malfunctions:
```
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Verification failed for 'cuda_device::__internal::index_1d':
         [name: "Span { id: 47, repr: "/home/thoth/.cargo/git/checkouts/cuda-oxide-6d394bb007f5e114/f420aa7/crates/cuda-device/src/thread.rs:214:9: 219:10" }", loc: (?)] Compilation error: verification failed.
       MirConstructStructOp result must be a struct type
         Failed operation:
       op188v1_res0 = mir.construct_struct (op184v1_res0, op185v1_res0, op186v1_res0, op187v1_res0) [] []: <(builtin.integer ui64, mir.struct <PhantomData,[],[],[],[],0>, mir.struct <PhantomData,[],[],[],[],0>, mir.struct <PhantomData,[],[],[],[],0>) -> (builtin.integer ui64)> !0
```

If I back up to an earlier version of the cargo-oxide tool (aef2b4fefab6230bfea46d483c0b8bcbb63c089d) and generate the test app it fails due to a different problem:

```
error[E0382]: borrow of moved value: `idx`
 --> src/main.rs:9:21
  |
7 |     let idx = thread::index_1d();
  |         --- move occurs because `idx` has type `ThreadIndex<'_>`, which does not implement the `Copy` trait
8 |     if let Some(c_elem) = c.get_mut(idx) {
  |                                     --- value moved here
9 |         *c_elem = a[idx.get()] + b[idx.get()];
  |                     ^^^ value borrowed here after move
```

My initial suspicion is API drift given that `ThreadIndex` is now a lot more complicated than it was a day or two ago.

mutantbob:
I iterated through a couple of git commits.

[...]

PR #39, opened 2026-05-11
  Auto-detect local GPU target for cargo oxide run

mohamedsamirx:
## Summary

  While investigating #36, I could not reproduce the exact Pliron verifier failure on current `main`.

  However, I found a separate reproducible template runtime issue: `cargo oxide run` can generate PTX for the backend default target even when that target is not runnable on the local GPU.

  On this machine, with an sm_75 GPU and `CUDA_OXIDE_TARGET` unset, the generated template produced `.target sm_80` and failed during module load with:

  ```text
  Failed to load embedded CUDA module: Driver(DriverError(218, "a PTX JIT compilation failed"))
```
  This PR updates cargo oxide run to auto-detect device 0 compute capability and forward it as CUDA_OXIDE_TARGET when no explicit target was provided.

  ## Behavior

  - Keeps explicit --arch behavior unchanged.
  - Keeps explicit CUDA_OXIDE_TARGET behavior unchanged.
  - Keeps normal cargo oxide build behavior unchanged.
  - Only changes cargo oxide run, where the output should be runnable on the local machine by default.

  ## Testing

  cargo fmt --check
  cargo test -p cargo-oxide
  cargo clippy -p cargo-oxide -- -D warnings
  cargo check
  cargo build -p cargo-oxide

  End-to-end generated template test:

  env -u CUDA_OXIDE_TARGET cargo oxide run

  Before this change, the generated PTX targeted sm_80 and failed to load on sm_75 hardware.

  After this change, the generated PTX contains:

  .target sm_75

  and the generated template runs successfully:

  PASSED: all 1024 elements correct

  ## Note on #36

  This came from investigating #36, but I could not reproduce the exact verifier error reported there on current main. This PR fixes the separate target-selection/runtime failure found during that investigation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,6 +191,7 @@ name = "cargo-oxide"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "cuda-core",
  "cuda-host",
  "libnvvm-sys",
  "nvjitlink-sys",
--- a/crates/cargo-oxide/Cargo.toml
+++ b/crates/cargo-oxide/Cargo.toml
@@ -16,3 +16,4 @@ clap = { version = "4", features = ["derive"] }
 libnvvm-sys = { workspace = true }
 nvjitlink-sys = { workspace = true }
 cuda-host = { workspace = true }
+cuda-core = { workspace = true }
--- a/crates/cargo-oxide/README.md
+++ b/crates/cargo-oxide/README.md
@@ -51,6 +51,12 @@ cargo oxide setup                   # explicitly build the codegen backend
 
 Builds the codegen backend, compiles the example with the custom backend, and runs it. This is the primary command for day-to-day development.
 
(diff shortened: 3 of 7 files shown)


nihalpasham (approved):
Thanks for the PR. 

Just a note: Added the following on top:

1. Rebased onto current main; dropped dead `--dlto` parameter from the helper.
2. Expanded `detect_run_target_arch` docs with the full precedence chain and run-only rationale.
3. Added 3 unit tests covering the early-exit branches.
4. cargo oxide run now prints the auto-detected arch instead of being silent.
5. Updated clap help for `Run::arch` and the two relevant book chapters.

mohamedsamirx:
Anytime. Great additions, thanks. Also, if you have something in mind for CUDA‑Oxide and would like someone to work on it, please feel free to reach out to me.

PR #40, opened 2026-05-11
ran `cargo clippy --allow-dirty --fix -- -Wclippy::use_self` to make use of `Self` uniform

adamnemecek:


nihalpasham:
Thanks for the PR. I appreciate the cleanup, but I think we’ll skip this one for now.

`clippy::use_self` isn’t a lint we currently enforce in CI, and this PR is mostly mechanical style churn across many crates. I’d prefer not to merge that unless we first decide as a project that we want this style and add the lint to our checked policy.

But thanks again for taking a look at the codebase.

PR #41, opened 2026-05-11
fix(codegen): unify architecture handling and add target fallback warnings

winshaurya:
### **Description**
Replaces a silent fallback to `sm_100` with a proper `rustc` warning. 

**Why:**
LTOIR is architecture-specific. If `CUDA_OXIDE_ARCH` is missing, the compiler previously defaulted to Blackwell (`sm_100`) without telling the user. This fix ensures that users are alerted to the fallback, preventing silent hardware mismatch bugs that cause GPU crashes at runtime.

**Changes:**
- Uses `tcx.dcx().warn()` to emit a native compiler warning.
- Only triggers if `CUDA_OXIDE_EMIT_LTOIR` is enabled to avoid unnecessary noise.

PS: 1st PR in the Repo , I am Shaurya, Creating a small game on Cuda-Oxide, will report if I ran into some issue in Future too.


nihalpasham:
Thanks for the PR. Checking the current LTOIR paths I think this particular fix lands on an older direct-`--dlto` path.

`cargo oxide --dlto` already sets `CUDA_OXIDE_ARCH`, while the newer LTOIR/cubin examples mostly go through `--emit-nvvm-ir` plus libNVVM/nvJitLink and use `CUDA_OXIDE_TARGET`. So this warning would only help a narrow manual-backend case.

I’m going to hold off on merging this as-is. We should probably clean this up as part of the ongoing LTOIR/NVVM IR/cubin embedding work and unify the arch handling around `--arch` / `CUDA_OXIDE_TARGET`.

winshaurya:
Thanks for the excellent context, @nihalpasham! You are absolutely correct, indeed the first implementation was limited to the previous `--dlto` pipeline and had a gap in the new `--emit-nvvm-ir` pipeline.

I did implement your suggestion for **unification of the architecture handling using `--arch`/`CUDA_OXIDE_TARGET`** in both `frontend_wrapper` and `codegen_backend` and squashed the whole change into a single commit.

The changes consist of the following parts:
1. **Canonical Env Var (`cargo-oxide`)**: `apply_output_mode` now consistently sets `CUDA_OXIDE_TARGET` in case of any arch specified by `--arch`, without respect of the mode. `CUDA_OXIDE_ARCH` remains as a backward-compatible env var that may be used in case of custom build script.
2. **Unified Reader & Universal Warning (`device_codegen`)**: Backend now attempts to read `CUDA_OXIDE_TARGET` first and then uses `CUDA_OXIDE_ARCH` only as fallback. Moreover, missing architecture warning is made universal to cover both `emit_ltoir` and `emit_nvvm_ir` cases in order to catch silent falling-backs to sm_100.
3. **Docs**: Modified the `README.md` list of environment variables to note that `CUDA_OXIDE_TARGET` is now the primary one.

Please let me know whether the direction is consistent with your future plans,
And I'll be happy to help with the further cleanup process, do share your vision on the "unification".

winshaurya:
Updated the name of PR, defining the aim of the changes

nihalpasham (approved):
Thanks for the PR, @winshaurya. I layered a few edits on top of your PR and pushed one cleanup commit.

It keeps the architecture handling unified around `CUDA_OXIDE_TARGET`, removes the dead direct `--dlto / CUDA_OXIDE_ARCH` path and adds focused tests for the remaining PTX and NVVM IR flows.

PR #42, opened 2026-05-11
Add pinned host buffer support

mohamedsamirx:
## Summary

  Add `PinnedHostBuffer<T>` to `cuda-core` for CUDA page-locked host memory.

  This provides a safe RAII wrapper around `cuMemAllocHost` / `cuMemFreeHost` and integrates pinned host buffers with `DeviceBuffer<T>` transfers.

  ## Motivation

  CUDA pinned/page-locked host memory is important for host-device transfer performance and is required when copies need true asynchronous overlap with GPU work.

  `cuda-core` already exposes device allocation and host-device copy helpers, but it did not have a safe owned pinned-host-memory type. This PR adds that missing runtime building block.

  ## Changes

  - Add `PinnedHostBuffer<T: DeviceCopy>`.
  - Allocate pinned host memory with `cuMemAllocHost`.
  - Free pinned host memory on drop with `cuMemFreeHost`.
  - Support initialized buffers via `new` and `from_slice`.
  - Support empty buffers without calling CUDA allocation.
  - Add `DeviceBuffer::from_pinned_host`.
  - Add `DeviceBuffer::copy_to_pinned_host`.
  - Document pinned host memory in the main README and `cuda-core` README.
  - Add tests for pinned allocation, empty allocation, device roundtrip, and non-`DeviceCopy` rejection.

  ## Safety

  `PinnedHostBuffer<T>` requires `T: DeviceCopy`, matching the safety model used by `DeviceBuffer<T>` transfer helpers.

  This rejects host-owned values like `String`, which cannot safely round-trip through raw byte copies.

  Drop follows the existing cuda-core RAII pattern: cleanup errors are recorded on the context instead of panicking.

  ## Testing

  ```bash
  cargo fmt --check
  cargo test -p cuda-core
  cargo clippy -p cuda-core -- -D warnings
  cargo check
  cargo clippy --workspace -- -D warnings
  cargo oxide run vecadd --arch sm_75

  The pinned-memory tests include a real CUDA roundtrip through:

  PinnedHostBuffer::from_slice
  DeviceBuffer::from_pinned_host
  DeviceBuffer::copy_to_pinned_host

  The real vecadd example completed successfully:

  SUCCESS: All 1024 elements correct

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The workspace combines:
 - single-source compilation -- host and device code live in the same file, built with one `cargo oxide build`
 - a rustc codegen backend that compiles `#[kernel]` functions to CUDA PTX
 - device-side abstractions (type-safe indexing, shared memory, scoped atomics, barriers, TMA, warp/cluster ops)
-- a host-side runtime for memory management and kernel launching (`cuda-core`, `cuda-async`)
+- a host-side runtime for memory management, pinned host transfers, and kernel launching (`cuda-core`, `cuda-async`)
 - a rust-native compilation pipeline using [Pliron](https://github.com/vaivaswatha/pliron), an MLIR-like IR framework in Rust (Rust → Rust MIR → Pliron IR → LLVM IR → PTX)
 
 ## Project Status
@@ -247,7 +247,7 @@ cargo oxide run gemm_sol
 | `cuda-host`         | Typed module loading, launch helpers, LTOIR loader                        |
 | `cuda-macros`       | Proc macros (`#[cuda_module]`, `#[kernel]`, `gpu_printf!`)                |
 | `cuda-bindings`     | Raw `bindgen` FFI bindings to `cuda.h`                                    |
-| `cuda-core`         | Safe RAII wrappers (`CudaContext`, `CudaStream`, `DeviceBuffer<T>`)       |
+| `cuda-core`         | Safe RAII wrappers (`CudaContext`, `CudaStream`, `DeviceBuffer<T>`, `PinnedHostBuffer<T>`) |
 | `cuda-async`        | Async execution layer (`DeviceOperation`, `DeviceFuture`, `DeviceBox<T>`) |
 | `libnvvm-sys`       | `dlopen` bindings to libNVVM (used by `cuda-host::ltoir`)                 |
 | `nvjitlink-sys`     | `dlopen` bindings to nvJitLink (used by `cuda-host::ltoir`)               |
@@ -289,7 +289,7 @@ cargo oxide run gemm_sol
 - LTOIR generation for Blackwell+ (device-side LTO)
 - Device FFI: Rust <-> C++/CCCL interop via LTOIR
 - MathDx integration: cuFFTDx thread-level FFT, cuBLASDx block-level GEMM
-- Host runtime: `cuda-core` (explicit control) and `cuda-async` (composable async operations)
+- Host runtime: `cuda-core` (explicit control, pinned host transfers) and `cuda-async` (composable async operations)
 - GEMM SoL: 868 TFLOPS (58% cuBLAS SoL) on B200 with cta_group::2, CLC, 4-stage pipeline
 
 ## Documentation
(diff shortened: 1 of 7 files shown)


nihalpasham:
Thanks, this is a nice addition. A few things before merging:

### To fix: 
- `copy_to_pinned_host` internally syncs, which contradicts the "true async overlap" benefit the docs promise. Either add an `_async` variant or tone down the doc claims; I'd prefer the former.
- `from_pinned_host` doesn't check that `data.context() == stream.context()`. `cuMemAllocHost_v2` (no `PORTABLE` flag) only pins in the allocating context, so a mismatch silently loses the pinning win. Suggest a `debug_assert!(Arc::ptr_eq(...))` plus a doc note on the single-context limitation.
- `HtoD`/`DtoH` asymmetry is worth a doc note: `from_pinned_host` returns after enqueue, `copy_to_pinned_host` blocks. Not obvious from the names.

### Nits
- The constructor is called `new`, but `DeviceBuffer`'s equivalent is `zeroed`. Pick the same name across both types so users don't have to remember which is which.
- The two `Drop` impls handle the same error case differently (one records it, one drops it). Make them consistent or add a one-line comment saying why they differ.
- `PinnedHostBuffer` doesn't print with `{:?}`. Every other public type in cuda-core does, so adding `Debug` keeps it uniform.
- `num_bytes()` recomputes `len * size_of::<T>()` from scratch even though we already computed and validated it at allocation time. Just store it.
- The safety doc on `malloc_host` says passing zero is undefined behavior, but CUDA actually just returns an error for that. Reword so people aren't scared off.
- The zero-init loop in `new` writes each element one at a time. A single slice fill (or `write_bytes`) does the same thing and reads more cleanly.
- No tests for: zero-sized types, `from_slice` with an empty input. Both work today, worth a test so they keep working.
- The `compile_fail` doctest proves `String` is rejected, but doesn't say whether it's because of `DeviceCopy` or `Default`. This is fine to leave, just confirming.

mohamedsamirx:
Thanks for the detailed review. I pushed an update addressing the feedback.

## Changes

- Added `DeviceBuffer::copy_to_pinned_host_async`.
  - `copy_to_pinned_host` remains the blocking convenience method.
  - The async variant enqueues the DtoH copy and lets the caller synchronize later.

- Documented the HtoD/DtoH asymmetry.
  - `from_pinned_host` returns after enqueueing the HtoD copy.
  - `copy_to_pinned_host` synchronizes before returning.
  - `copy_to_pinned_host_async` is available for non-blocking DtoH.

- Added context checks for pinned transfers.
  - `from_pinned_host` now has a `debug_assert!(Arc::ptr_eq(...))`.
  - `copy_to_pinned_host_async` has the same context assertion.
  - Docs note that `cuMemAllocHost` is used without `PORTABLE`, so the pinned allocation is tied to the allocating context.

- Renamed `PinnedHostBuffer::new` to `PinnedHostBuffer::zeroed` for consistency with `DeviceBuffer::zeroed`.

- Made `PinnedHostBuffer::zeroed` byte-zero the allocation directly and removed the `T: Default` bound.
  - This matches `DeviceBuffer::zeroed` semantics and the existing `DeviceCopy` all-zero validity contract.

- Made drop error handling consistent.
  - `DeviceBuffer` now records `bind_to_thread()` errors like `PinnedHostBuffer`.

- Added `Debug` for `PinnedHostBuffer<T>`.

- Stored `num_bytes` at allocation time instead of recomputing it.

- Reworded the `malloc_host` zero-byte safety docs.

- Added tests for:
  - zero-sized types
  - `from_slice` with empty input
  - async pinned DtoH copy with later synchronization
  - `num_bytes`
  - `Debug`

## Testing

```bash
cargo fmt --check
cargo test -p cuda-core
cargo clippy -p cuda-core -- -D warnings
cargo check
cargo clippy --workspace -- -D warnings
cargo oxide run vecadd --arch sm_75

nihalpasham (approved):
Looks good. 

Added 2 additions on top:

Marked stream-ordered pinned host transfers as unsafe
- `from_pinned_host` / `copy_to_pinned_host_async` return as soon as the copy is enqueued, but the GPU keeps using the pinned pointer after the Rust borrow ends.
- dropping the buffer before `stream.synchronize()` frees pinned memory mid-flight, which is UB. Marked both unsafe 
- `copy_to_pinned_host` stays safe (it syncs internally). Tests now wrap the async calls in unsafe {} with safety comments.

Added `DeviceBuffer::copy_from_pinned_host_async` to refill in place
- removes asymmetry: DtoH had blocking + async, but HtoD only had `from_pinned_host`, which allocates a fresh device buffer per call.
- new unsafe instance method enqueues `cuMemcpyHtoDAsync` into an existing `DeviceBuffer`, so rotating pinned stagers can feed one device buffer without per-iteration allocs. Same safety contract as the others.
- new test drives three refill -> readback rounds against the same buffers.

PR #44, opened 2026-05-12
Extend embedding support to IR and binary payloads

josephglanville:
# Summary
This PR extends the support for embedded artifacts added in [#26](https://github.com/NVlabs/cuda-oxide/pull/26) from PTX to NVVM IR and binary artifacts so kernels that trigger those outputs can also benefit from embedded loading and typed launchers.

## Example
```rust
use cuda_device::{cuda_module, kernel, thread, DisjointSlice};

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    pub fn sqrt_kernel(input: &[f32], mut output: DisjointSlice<f32>) {
        if let Some(out) = output.get_mut(thread::index_1d()) {
            // Calls libdevice (`__nv_sqrtf`), so cuda-oxide emits embedded NVVM IR.
            *out = input[i].sqrt();
        }
    }
}

    let input = [1.0f32, 4.0, 9.0, 16.0];
    let input_dev = DeviceBuffer::from_host(&stream, &input)?;
    let mut output_dev = DeviceBuffer::<f32>::zeroed(&stream, input.len())?;

    // Loads the embedded IR payload, builds it to cubin with libNVVM/nvJitLink,
    // then returns typed launch methods.

    module.sqrt_kernel(
        &stream,
        LaunchConfig::for_num_elems(input.len() as u32),
        &input_dev,
        &mut output_dev,
    )?;

    let output = output_dev.to_host_vec(&stream)?;
    assert_eq!(output, [1.0, 2.0, 3.0, 4.0]);

    Ok(())
}
```


crates/cuda-host/tests/cuda_module_api.rs
```diff
@@ -13,6 +13,9 @@ use cuda_host::cuda_async::device_operation::DeviceOperation;
 use cuda_host::cuda_module;
 use cuda_macros::kernel;
 
+#[cfg(feature = "async")]
+type TwoF32Buffers = (DeviceBox<[f32]>, DeviceBox<[f32]>);
```
josephglanville, line 17: Clippy was complaining about this and local type alias seemed to be the cleanest fix.

josephglanville:
I guess another design thing to call out was the `cuda-core` to `cuda-host` shuffling was mostly to avoid making `cuda-core` depend on the NVVM toolchain.

nihalpasham (approved):
Thanks. LGTM

Only change: pushed 3 commits on top to unblock the merge:

- **`0b28c2b`** Merge `main`. Three mechanical conflicts in the NVVM IR target path (your branch and main's `b6c0cc1` both touched it). Notable: dropped the obsolete `if config.emit_ltoir` branch (main removed the field), and the NVVM-IR bundle target now reads `CUDA_OXIDE_TARGET` with main's `"nvvm-ir"` sentinel as fallback.
- **`f304123`** Doc + test follow-ups: list `embedded`/`ltoir` in the cuda-host module docs, test `target_arch_for_bundle` against both legacy sentinels (`libdevice`, `nvvm-ir`), and track the missing `manual_launch_libdevice/Cargo.lock`.
- **`1215842`** Drive-by: gitignore `.mcp/`.

PR #48, opened 2026-05-12
Add flake.nix and CI for it

JonatanLindh:
### Motivation

Currently, setting up the required environment for `cuda-oxide` (specifically aligning Nightly Rust, LLVM 21, and the correct CUDA toolchain so `CUDA_HOME` resolves properly) can be a bit manual.

This PR introduces a `flake.nix` to provide a fully reproducible development environment for anyone using the Nix package manager. Running `nix develop` drops the user into a shell with the dependencies required to compile the compiler and run the PTX backend.

### What this does

* Uses `symlinkJoin` to pull necessary CUDA 13 packages (`nvcc`, `cudart`, `libnvvm`, `libnvjitlink.lib`).
* LLVM 21 and the pinned Nightly Rust toolchain (via `rust-toolchain.toml`).
* Exports `CUDA_HOME`, `LIBCLANG_PATH`, and `CUDA_OXIDE_LLC`.
* Includes a dynamic `shellHook` (inspired by `cutile-rs`) that symlinks the host machine's NVIDIA driver into the Nix environment. Ensures that drivers works across distros (NixOS or otherwise).
* Supports both `x86_64-linux` and `aarch64-linux`.

### Note on CI and Maintenance

I have included a `.github/workflows/nix-flake.yml` to prove that this environment builds cleanly across both x86_64 and aarch64 Ubuntu runners without issues.

However, I completely understand if you prefer not to officially "support" Nix or add it to your required CI checks to avoid future maintenance burdens. If you'd prefer to just merge the `flake.nix` without the CI, just let me know and I will happily drop the workflow file from this PR!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,6 +110,10 @@ sign-off will not be merged.
 cuda-oxide requires the Rust nightly toolchain with `rustc_private` support.
 See the [README](README.md) for setup instructions.
 
+The repository includes a `flake.nix` that provides a fully reproducible development
+environment (CUDA 13, LLVM 22, Clang, pinned Rust nightly). If you have Nix with
+flakes enabled, `nix develop` is the quickest way to get everything in place.
+
 ### Formatting and Style
 
 - Run `cargo fmt` before submitting. All code must be formatted with
--- a/README.md
+++ b/README.md
@@ -136,6 +136,15 @@ cargo install --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide
 
 On first run, `cargo-oxide` will automatically fetch and build the codegen backend.
 
+#### Nix (alternative)
+
+If you have Nix with flakes enabled, `nix develop` in the repo gives you a reproducible shell with CUDA 13, LLVM 22, Clang, and the pinned Rust nightly — no manual apt installs. The shellHook auto-discovers host NVIDIA drivers on NixOS and non-NixOS systems.
+
+```bash
+nix develop                                       # full dev shell in this repo
+nix run github:NVlabs/cuda-oxide#new my-project   # bootstrap a project
+```
+
 #### Rust
 
 ```bash
--- a/cuda-oxide-book/getting-started/installation.md
+++ b/cuda-oxide-book/getting-started/installation.md
@@ -62,6 +62,37 @@ Rust setup sections below.
 
 ---
 
+## Nix / flake.nix
+
+The repository also ships a `flake.nix` providing a reproducible dev shell
+(CUDA 13, LLVM 22, Clang, pinned Rust nightly). Requires
+[Nix](https://nixos.org/download/) with flakes enabled, an NVIDIA driver on
+the host, and Linux (x86\_64 or aarch64).
+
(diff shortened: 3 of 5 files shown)


nihalpasham:
Thanks for putting this together, I think the flake itself is well scoped and the cross-distro driver shim is pretty useful.

**A couple of points though:**
- We'd rather not take the per-PR CI workflow. Cold Nix runs pull multi-GB of CUDA each time, and nixos-unstable drift would gate unrelated PRs on Nix-only failures.
- Quick note: #6 merged yesterday adds llvm-tools to rust-toolchain.toml, so rustup auto-installs an NVPTX-enabled llc (LLVM 22) that the pipeline picks up by default.
- Could you drop the `CUDA_OXIDE_LLC` export from the flake? It would silently pin Nix users to LLVM 21, but the new atomics example needs LLVM 22 for correct syncscope.
- Letting rustup's llc win also means future LLVM bumps in rust-toolchain.toml flow through automatically, no parallel flake.nix edit needed.
- Small stuff: bump cachix/install-nix-action to v27+, add an SPDX header to flake.nix, and a one-line README / CONTRIBUTING pointer so contributors can find the shell.

JonatanLindh:
Thanks for the review @nihalpasham, pushed a revision addressing the feedback:

  Your points:
  - Dropped the per-PR CI workflow.
  - Removed CUDA_OXIDE_LLC and bumped llvmPackages_21 → llvmPackages_22. Changes in #6 work perfectly.
  - Added the SPDX header, a one-line pointer in CONTRIBUTING.md, and a Nix subsection in the README + book

  Added beyond the feedback (happy to drop any of this if you'd rather keep the PR strictly to the dev shell):
  - Build cargo-oxide as a derivation via crane. Lets shells have cuda-oxide automatically and allows creating a new project without a clone of this repo or a  bootstrap shell (packages.cargo-oxide, apps.default, apps.new, and templates.default fall out of that).
  - The user-facing template flake uses inputsFrom = [ cuda-oxide.devShells.${system}.default ] so downstream projects inherit the full CUDA + driver setup instead of duplicating it. CUDA_HOME / LIBCLANG_PATH moved into
  the shellHook so they actually propagate through inputsFrom (mkShell attrs don't).

  One known impurity, out of scope for this PR I imagine: on first run, cargo-oxide compiles librustc_codegen_cuda.so and writes it outside the Nix store. The Nix-pure fix is to build the backend as part of packages.cargo-oxide and point cargo-oxide at the store path at runtime.
  
  Also, I kept cargo-oxide in the in-repo dev shell; the alias-shadow warning is noisy but harmless today, and downstream projects inheriting via inputsFrom need it. Might change as rust-lang/cargo#10049 is resolved.

JonatanLindh:
Force-pushed to squash and add signed-off-by

nihalpasham (approved):
LGTM. 

added one small commit on top: it hardens the nix run .#new wrapper to pick the first non-flag argument as the project directory (a leading --async previously sent the template copy to the wrong path)

The backend-build impurity you flagged is fine to leave for a follow-up. Approving.

PR #52, opened 2026-05-13
intrinsics: add globaltimer

josephglanville:
## Summary

Adds a CUDA device-side `globaltimer()` debug intrinsic that reads PTX `%globaltimer` as a 64-bit timer value. This complements `clock64()` with a global timer source that is better suited for measurements that may span multiple SMs.

## Example

```rust
use cuda_device::debug;

let start = debug::globaltimer();

// ... operation to measure ...

let end = debug::globaltimer();
let ticks = debug::elapsed_ticks(start, end);
```

## Implementation

- Adds `cuda_device::debug::globaltimer() -> u64` and `cuda_device::debug::elapsed_ticks(start, end) -> u64`.
- Adds the `nvvm.read_ptx_sreg_globaltimer` op to `dialect-nvvm`, including registration and result-type verification.
- Wires `cuda_device::debug::globaltimer` through MIR intrinsic dispatch.
- Lowers the NVVM op to convergent inline asm:

```text
mov.u64 $0, %globaltimer;
```

- Updates debug and dialect documentation to include the new timer intrinsic.
- Updates the debug example to exercise `globaltimer()` alongside `clock64()`.

## Validation

- Adds `test_globaltimer_lowers_to_inline_asm`, which verifies that `ReadPtxSregGlobaltimerOp` lowers to the expected inline asm template and is marked convergent

### DIFF
--- a/crates/cuda-device/README.md
+++ b/crates/cuda-device/README.md
@@ -129,7 +129,9 @@ pub fn debug_kernel(data: &[f32]) {
 }
 ```
 
-`gpu_printf!` compiles to device-side `vprintf` with C vararg promotion. `gpu_assert!` traps on failure.
+`gpu_printf!` compiles to device-side `vprintf` with C vararg promotion.
+`gpu_assert!` traps on failure. The `debug` module also exposes GPU timing
+register reads such as `clock64()` and `globaltimer()`.
 
 ## Proc-Macro Re-exports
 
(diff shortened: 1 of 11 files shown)


nihalpasham (approved):
Thanks for adding this 

I've layered a few changes on top before merging:

- **Lower `nvvm.read_ptx_sreg_globaltimer` via the named LLVM intrinsic
  rather than inline asm.** `llvm.nvvm.read.ptx.sreg.globaltimer`
  is a real NVPTX intrinsic supported well below our LLVM 21
  (verified with `llc-21.1.8`: emits exactly the same `mov.u64 %rdN, %globaltimer;`).
- Going through the named intrinsic also drops the unnecessary `convergent`
  attribute; SREG reads don't require warp-synchronous execution, only
  primitives like `shfl` / `vote` / `barrier` / `mbarrier` / `cp.async.bulk` do.
- **Updated the unit test** to mirror the existing `test_intrinsic_insertion`
  pattern (assert the `declare` + the `call`, plus a regression guard that
  globaltimer must not lower to inline asm).

PR #55, opened 2026-05-13
Fix `llvm.addressof` export symbols

josephglanville:
## Summary

The LLVM exporter could emit an undefined SSA value for `llvm.addressof` when a block used an addressof result before the block containing the `AddressOfOp` was printed.

With the old exporter, a source-level `SharedArray` repro failed during NVVM IR compilation with:

```text
/usr/bin/llc-21: .../addressof_sharedarray_repro.ll:15:56: error: use of undefined value '%v4'
  %v2 = getelementptr inbounds float, ptr addrspace(3) %v4, i64 0
                                                       ^
```

## Cause

`AddressOfOp` is a virtual global-symbol reference in textual LLVM IR. It should print as `@global`; it should not define an SSA register.

The exporter naming pre-pass assigned every ordinary op result a `%vN` name. Since `AddressOfOp` was not special-cased, an earlier-printed use could emit that temporary. When the `AddressOfOp` block printed later, export rewrote the mapping to `@global`, but no instruction ever defined the earlier `%vN`.

## Fix

Pre-register `AddressOfOp` results as `@{global_name}` in the naming pre-pass and skip assigning them SSA temporaries.

This is the right model because LLVM global addresses are symbol operands, not instruction results. It matches the existing constant handling and makes addressof uses independent of CFG block print order.

## Regression Tests

Added:

- A focused `dialect-llvm` exporter test where an `AddressOfOp` is in a later-printed block and a GEP using it is in an earlier-printed block.
- A user-level `SharedArray` LTOIR repro `addressof_sharedarray_repro`.

With the fix, the repro emits:

```llvm
%v2 = getelementptr inbounds float, ptr addrspace(3) @__shared_mem_0, i64 0
```

and the IR is accepted by LLVM.


nihalpasham (approved):
This looks good. Added a couple more pieces on top 

- Op-emit `AddressOfOp` arm now `debug_assert!`s the pre-pass invariant instead of duplicating it.
- Test walks the IR and asserts every `%vN` has a definition (catches the bug class, not just the GEP).
- Fixture now launches through libNVVM and validates on GPU; a regression fails at the verifier.

Issue #49, opened 2026-05-12
stale librustc_codegen_cuda.so may cause problems after updating cuda-oxide


mutantbob:
`~/.cargo/cuda-oxide/librustc_codegen_cuda.so` can cause problems if it does not match the `~/.cargo/bin/cargo-oxide` .

`cargo uninstall cuda-oxide` does not remove `librustc_codegen_cuda.so` .

`cargo install --path ~/vendor/cuda-oxide/crates/cargo-oxide cargo-oxide` does not replace `librustc_codegen_cuda.so` with a fresh version based on the code in the --path .

**Also**

`~/.cargo/cuda-oxide/src` is pulled from github instead of the `--path` used to install cargo-oxide.  This can cause a mismatch as well.  I do not know if switching this will break something else.  I have not yet found a section in the book that outlines the workflow for developing and testing modifications to the codegen and other cuda-oxide tools, but so far I think that 
```
rm ~/.cargo/cuda-oxide/librustc_codegen_cuda.so 
cargo install --path ~/vendor/cuda-oxide/crates/cargo-oxide cargo-oxide
```
might be part of it until I find a better solution.

PR #56, opened 2026-05-13
suggest rm librustc_codegen_cuda.so before cargo install cargo-oxide

mutantbob:
This is a placeholder/interim "solution" for part of https://github.com/NVlabs/cuda-oxide/issues/49 .  By documenting the existence and relevance of `librustc_codegen_cuda.so` the few folks who actually modify or update their cuda-oxide installation will know how to remove a stale artifact.

A more long-term solution would be to automate the replacement of `librustc_codegen_cuda.so` when a new cuda-oxide tool is installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Inside the cuda-oxide repo, `cargo oxide` works out of the box via a workspace a
 For use outside the repo (your own projects):
 
 ```bash
+rm -f $HOME/.cargo/cuda-oxide/librustc_codegen_cuda.so # stale artifact can cause error
 cargo install --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide
 ```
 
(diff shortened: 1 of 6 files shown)


nihalpasham (approved):
Thanks for the PR

A couple of additions/changes on layered on top:

- Replaced the docs-side `rm -f` workaround with a real fix in `backend.rs`.
- Step 3 now drops the cached `.so` and `src/` when the cargo-oxide binary is newer, so step 4 re-clones fresh on next upgrade.
- Three unit tests cover stale, fresh, and missing-cache branches; closes #49.

So here's what a before - after looks like 

```text
Before:
  user upgrades cargo-oxide  →  step 3 returns OLD .so  →  binary dlopens old backend  →  crash

After:
  user upgrades cargo-oxide  →  step 3 sees binary mtime > .so mtime
                              →  drops both .so and src/ from cache
                              →  step 4 re-clones fresh and rebuilds
                              →  binary dlopens matching backend
```

Issue #30, opened 2026-05-10
replace HashMap with FxHashMap?


hsqStephenZhang:
I noticed that many data structured are using [std HashMap](https://github.com/NVlabs/cuda-oxide/blob/aef2b4fefab6230bfea46d483c0b8bcbb63c089d/crates/mir-lower/src/context.rs#L19) now.

Should they be replaced by FxHashMap for better performance? I'd like to help if it's wanted.

fbrv:
I’d benchmark this first. After looking at the code, I don’t expect changing only the hasher to be a major speedup here.
`SharedGlobalsMap` is keyed by `String`, and the current code materializes an owned `String` before lookup:
```rust
.map(|s| String::from((*s).clone()))
```
So the cost is not only hashing. It may be more useful to avoid constructing an owned `String` for lookup-only cases, and only clone when inserting

nihalpasham:
@hsqStephenZhang -- thanks for bringing this up.

@fbrv -- I agree, we should benchmark this first. To my mind, the easier perf gains are more likely lurking in the avoidable allocations we currently do. If you're interested, I’d recommended starting with `mir-lower` first and comparing borrowed lookups / entry APIs against `FxHashMap`.

Suggested order:

1. `std::HashMap` as-is
2. Avoid lookup-only `String` allocation, using borrowed lookup where possible
3. Maybe `rustc_hash::FxHashMap` after that

hsqStephenZhang:
fair enough, i'll benchmark it first.

vaivaswatha:
In general, for compilers, `FxHashMap` (and `FxHashSet`) is a better choice because they are deterministic, as against the standard library hash maps, which are randomized. Deterministic makes it easier to debug, when, for example, we're debugging iterating over a hash map's elements.

That said, it is only deterministic per version of the library / rust and platform. (That's still good enough though).

hsqStephenZhang:
here is the  [code](https://gist.github.com/hsqStephenZhang/a495b0f85092efd05579c0f4461b8e6a) and outcome of the benchmark

## Benchmark setup

* use `String` keys
* 100,000 key-value pairs
* 100,000 lookups
* lookup hit rates are exactly 25% / 50% / 75%
* bench both lookup and insert

## Result

* Lookup (25% hit)

  * `std::HashMap`: 2.9553 ms
  * `FxHashMap`: 2.4458 ms
  * `FxHashMap` ~1.21× faster

* Lookup (50% hit)

[...]

PR #57, opened 2026-05-13
chore: cleaner alloc_key usage

hsqStephenZhang:
## What this PR does.

While reading https://github.com/NVlabs/cuda-oxide/issues/30, i found that the usage of `alloc_key` could be simplified by more rusty grammar


nihalpasham (approved):
Looks good. 

A few small additions on top:

- Drop the `is_some()` + `unwrap()` dance; bind the key once with `as_ref()`.
- Cache miss still moves the key into the helper, no extra clone.
- Updated the helper's doc to say it takes ownership of the key when caching.

PR #60, opened 2026-05-14
Further fixes to closure launch ABI

josephglanville:
## Summary

Recently landed fixes to pass aggregate types byval did a lot to unify and fix underlying issues in the closure problem but were not the complete fix, some types of closures aren't getting closure specific handling they need, some problems with ZST and also the pre-existing problem of layout issues.

This PR addresses these problems and extends the test coverage to cover all APIs across a broader range of closure shapes.

## Test Coverage

`host_closure` now runs the same closure cases across all four launch surfaces. The matrix covers:

- zero-capture closures
- single scalar capture
- multiple scalar captures
- polynomial/four-capture closures
- `where`-clause `Fn` bounds
- mixed-size captures
- struct captures with internal padding and mixed field types
- non-`move` stack-reference captures


nihalpasham (approved):
Thanks, this is excellent 🚀 and LGTM.

Pushed 2 commits on top to layer some extra defensive coverage and bring the branch up to date:

- **`ba047a5`** Merge `origin/main` — no conflicts, picks up PR #44 and PR #52 since fork.
- **`fecec55`** Three additions to `host_closure`:
  - Tests 10/11: `FnMut` and `FnOnce` kernels exercising the device-side dispatch paths the macro already recognized via `find_closure_generic`. They pass with or without the layout fix (same-size captures) — pure defensive coverage of the dispatch trait.
  - Test 12: `#[repr(C)]` counterpart to test 8. Locks down that the layout fix handles identity-order layouts with alignment padding too. Adversarial check confirms it ALSO fails 1024/1024 when `types.rs` is reverted.
  - Test 9: one-line note that non-move ref captures depend on HMM (sm_75+ on Linux); GPUs without HMM surface this as `CUDA_ERROR_ILLEGAL_ADDRESS`, which `verdict_standard` flags as FAIL.

PR #62, opened 2026-05-14
Add f32::max / f32::min support via libdevice fmax/fmin

nyoki-mtl:
# Add `f32::max` / `f32::min` support via libdevice fmax/fmin

## Summary

`f32::max` / `f32::min` (and the f64 forms) lower in MIR to
`core::intrinsics::maximum_number_nsz_f32` / `minimum_number_nsz_f32`
(and f64). `RustFloatMathIntrinsic::from_core_path` did not match those
four names, so the calls fell out of the rustc float-math placeholder
pipeline added in 61028e6 and propagated as unresolved intrinsics into
`mir-lower`. Wire them in as a thin extension of the existing
`sqrt` / `sin` / `fma` machinery:

| Public API | MIR intrinsic | libdevice |
| --- | --- | --- |
| `f32::max` | `core::intrinsics::maximum_number_nsz_f32` | `__nv_fmaxf` |
| `f64::max` | `core::intrinsics::maximum_number_nsz_f64` | `__nv_fmax`  |
| `f32::min` | `core::intrinsics::minimum_number_nsz_f32` | `__nv_fminf` |
| `f64::min` | `core::intrinsics::minimum_number_nsz_f64` | `__nv_fmin`  |

## Implementation

Five small edits, mirroring `sqrt` / `fma` / `fabs`:

- **`dialect-mir::rust_intrinsics`** — four new placeholder constants.
- **`mir-importer::translator::terminator::intrinsics::float_math`** —
  four enum variants, `from_core_path` arms (both `core::` and `std::`
  aliases), `placeholder_callee` mappings.
- **`mir-lower::convert::ops::call`** — symmetric variants,
  `from_placeholder_callee` arms, libdevice symbol table entries,
  `arg_count = 2`.

Unit tests on both sides lock the `from_core_path` ↔
`placeholder_callee` ↔ `libdevice_name` chain so a future rustc rename
of `maximum_number_nsz_*` or a typo in any single matcher surfaces as a
unit-test failure rather than a runtime "intrinsic not lowered" error
after a long compile cycle. The importer test also explicitly rejects
the NaN-propagating `maximumf*` / `minimumf*` family (backing
`f32::maximum` / `f32::minimum`) so the deferred scope is enforced.

`examples/fmaxmin_smoke/` exercises the full chain end-to-end. It
mirrors the `primitive_stress` shape — the libdevice auto-detection
picks the `__nv_fmax*` / `__nv_fmin*` calls up and
[...]

PR #63, opened 2026-05-14
Emit NaN float literals as hex bit patterns, not bare `nan`

nyoki-mtl:
# Emit NaN float literals as hex bit patterns, not bare `nan`

## Summary

`format_float_literal` in `dialect-llvm::export` produced the bare token
`nan` for NaN inputs:

```text
store float nan, ptr %v9
```

That is not part of the LLVM IR float-constant grammar. Both `llc`
and libNVVM reject it:

```
llc-21: error: expected value token
  store float nan, ptr %v9
              ^

Nvvm(Call { operation: "nvvmCompileProgram", code: 9,
  log: Some("...ll (...): parse expected value token") })
```

Any kernel that holds an `f32::NAN` / `f64::NAN` constant as a `float` /
`double` SSA value — e.g. storing through a `*mut f32` write, or passing
as a float argument — fails to compile. Integer-bit views via
`to_bits()` happen to fold to integer constants in MIR and bypass this
helper, which is why this has not bitten more users yet.

## Fix

Emit a canonical quiet NaN as a 16-hex-digit double bit pattern:

- positive NaN → `0x7FF8000000000000`
- negative NaN → `0xFFF8000000000000`

LLVM converts the f64 hex pattern down to the destination type, so the
same helper output works for `float` (lands as f32 qNaN `0x7FC00000`)
and `double` (lands as f64 qNaN `0x7FF8000000000000`). The sign bit is
preserved so the helper does not silently flip a `-f32::NAN` to a
positive NaN; the payload bits below the leading qNaN marker are
canonicalized.

Existing `+Inf` / `-Inf` arms were already emitted as hex and are
unchanged.

This also removes the need for the NaN-literal workaround used in the
`fmaxmin_smoke` example from #62 (where the NaN argument is currently
passed in from the host); a follow-up can simplify that example once
this lands.

## Tests

Five new inline `float_literal_tests` cover NaN, -NaN, +Inf, -Inf, and a
finite control with decimal-point preservation. The new smoke example
`examples/fp_special_literal_smoke/` stores `f32::NAN` / `f32::INFINITY`
/ `f32::NEG_INFINITY` (and the f64 forms) directly through a
`DisjointSlice<f32>` so the constants survive as `float` SSA values
rather than being folded into integer `to_bits()` constants by rustc.

## Verification

```
cargo oxide fmt --check                                 ✓ clean
cargo clippy -p dialect-llvm --lib --tests -- -D warnings ✓ clean
cargo test -p cuda-host -p cuda-macros -p dialect-llvm
   -p dialect-mir -p dialect-nvvm -p mir-lower
   -p mir-importer --lib --tests                        ✓ 74 passed
cargo oxide build fp_special_literal_smoke              ✓ build succeeded
```

Generated `.ll` (after fix):

```llvm
store float  0x7FF8000000000000, ptr %v9
store float  0x7FF0000000000000, ptr %v10
store float  0xFFF0000000000000, ptr %v11
store float  1.5, ptr %v12
store double 0x7FF8000000000000, ptr %v9
store double 0x7FF0000000000000, ptr %v10
store double 0xFFF0000000000000, ptr %v11
store double 1.5, ptr %v12
```

Generated `.ptx` correctly converts the f64 hex to f32 qNaN:

```
st.b32 [%rd1],    2143289344  // 0x7FC00000 — f32 qNaN
st.b32 [%rd1+4],  2139095040  // 0x7F800000 — f32 +Inf
st.b32 [%rd1+8],  -8388608    // 0xFF800000 — f32 -Inf
st.b32 [%rd1+12], 1069547520  // 0x3FC00000 — f32 1.5
```

Device launch on my local host fails with CUDA driver error 209 on
existing libdevice examples too (WSL2 + CUDA 13.1 driver / 12.9
toolkit on RTX 3070 Ti / sm_86), unrelated to this PR. A reviewer
with a working device path is welcome to confirm the smoke prints
`SUCCESS`.

## Before / after

Without the fix, `cargo oxide build fp_special_literal_smoke` fails
during `llc` with `expected value token`. With the fix, the same
kernel reaches a cubin through the standard `llc → PTX` path. Both
backends (`llc` and libNVVM, the latter via the libdevice
auto-detection path) are affected by the original bug, so the
single-helper fix repairs both.

## DCO

Single commit, signed off.


nihalpasham:
Thanks for this fix @nyoki-mtl! PR #63 was based on the pre-migration `dialect-llvm` crate, which was renamed to `llvm-export` in #114 after you filed this. We ported your fix to the new location with your co-author credit preserved and landed it in #116.

PR #64, opened 2026-05-15
 fix(codegen): compile step_by by translating Pat types and rebuilding niched Option aggregates

### LINKED ISSUE #21: PTX Codegen Failure: StepBy struct lowering (invalid cast opcode) exposed by missing Pat type translation

g-ampo:
Fixes [#21](https://github.com/NVlabs/cuda-oxide/issues/21).

`for i in (a..b).step_by(s)` compiles and runs after this. Verified on a real GPU: the `step_by` kernel and an equivalent `while`-loop kernel produce identical sums for all 32 threads and match host-computed expected values.

## Issue

Two compiler gaps, exposed by `StepBy::next` after MIR inlining:

1. `TyKind::Pat` was unsupported in the importer. `core` encodes `NonZeroUsize` as `Pat<usize, 1..=usize::MAX>`, so importer bailed before reaching lowering.
2. After teaching the importer about `Pat`, mir-lower emitted `bitcast i64 to { i8, { { i64 } } }` for rustc's `Transmute usize -> Option<NonZeroUsize>`. rustc treats that `Option` as a niche-optimised 8-byte scalar (`0 = None`, non-zero = `Some(x)`); `MirEnumType` lowers it to the un-niched aggregate. A `bitcast` across the scalar/aggregate boundary is invalid LLVM and showed up as `invalid cast opcode` from llc.

## Fix

* `mir-importer/.../types.rs`: lower `RigidTy::Pat(base, _)` as the base type, with an in-source note on the layout assumption.
* `mir-importer/.../rvalue.rs`: when a Transmute targets a niche-optimised enum, attach the niche encoding (`niche_start`, `niche_variant_idx`, `untagged_variant_idx`) as attributes on the cast op, read straight from `ty.layout()`.
* `mir-lower/.../cast.rs`: for `scalar -> struct` casts, rebuild the un-niched aggregate explicitly using those attributes: `icmp eq val, niche_start` -> `select` for the discriminant, plus a nested `insertvalue` that lands the payload at the deepest scalar slot of the variant struct.  if the importer did not classify the destination as niched, we error out rather than fall through to an invalid bitcast.
* `dialect-llvm/.../export.rs`: emit `SelectOp` in the LLVM IR exporter. The op was registered but unprinted, so the new lowering needed it.
* `examples/step_by/`: positive regression with the `step_by` kernel and a `while`-loop control.

The discriminant computation is driven by the niche encoding rustc hands us, so this is not limited to `niche_value = 0`: any single-niche enum that fits in a scalar (`Option<NonZeroT>`, `Option<&T>`, `Option<NonNull<T>>`, `Option<Box<T>>`, `Option<bool>`, `Option<char>`, ...) takes the same path.


nihalpasham (approved):
Thanks for the PR. It needed some additions, which I've layered on top of your branch 

### The importants ones:

- **Pointer-source niche transmutes:** Your fix handled the integer-source case (`Option<NonZeroUsize>` -> aggregate) but missed the pointer-source case (`Option<&T>`, `Option<NonNull<T>>`, `Option<Box<T>>`). I added a sibling example (`option_ref_transmute`) that reproduced the bug as a pliron verifier error, reordered the dispatch in `emit_pointer_cast` so the niche reconstruction runs first whenever the importer attached a niche encoding, regardless of whether the source is an integer or a pointer. The helper now also accepts pointer sources directly (`icmp eq ptr %src, inttoptr i64 niche_start`).

- **Typed `NicheEncodingAttr` instead of string-keyed attributes:** The three loose `IntegerAttr` entries (`niche_start`, `niche_variant_idx`, `untagged_variant_idx`) glued together by a "keep in sync" string contract between importer and lowering are replaced with a proper `#[pliron_attr]` struct declared in `dialect-mir`, wired onto `MirCastOp` via its `attributes = (...)` slot. Both sides now use the macro-generated `get_attr_niche_encoding` / `set_attr_niche_encoding`. You can check how `cast_kind` did this before. Removed the cross-crate string contract, the hand-rolled `read_niche_info` / `set_niche_transmute_attrs` helpers, and the "keep in sync" comment.

- **Size guard on struct -> scalar:** The reverse direction (`emit_struct_to_scalar`) was doing an unchecked alloca + store + load, which would silently truncate or pull padding bytes on any size mismatch. It now walks the single-field newtype chain, compares widths, and errors loudly with a "size mismatch" diagnostic if they differ. Same loud-failure as the scalar -> struct direction.

### Also just some nits that were addressed in this PR

- `niche_start` is now carried as `u64` instead of `i64`, so we drop the bogus overflow rejection on niches with the top bit set.
- The magic `path.len() > 8` cap in `deep_scalar_index_path` is now a named `MAX_NEWTYPE_DEPTH = 8` constant with a comment explaining the choice.
- `read_niche_info` is a one-liner now, it just forwards to the typed getter.
- `MirCastOp::verify` rejects `niche_encoding` on non-Transmute casts, so the contract is enforced in one place.

PR #69, opened 2026-05-18
feat: support pre-Ampere GPUs (sm_61+) and std::sys::cmath transcendentals

heath-hunnicutt-ruach-tov:
## Summary

This PR enables cuda-oxide to compile and run kernels using transcendental math functions on **all sm architectures**, including pre-Ampere GPUs like Tesla P4 (sm_61, ~$50 on eBay).

## Problem

1. Rust math methods that route through `std::sys::cmath` (`.tanh()`, `.sinh()`, `.erf()`) cause the collector to panic with "FORBIDDEN CRATE IN DEVICE CODE"
2. Even for intrinsics like `.exp()` that compile, the pipeline skips `llc` and emits NVVM IR, requiring nvJitLink at runtime — which fails with `ModuleNotFound` on pre-Ampere GPUs

## Solution

**4 files, 223 insertions, 21 deletions:**

| File | Change |
|---|---|
| `rust_intrinsics.rs` | 16 placeholder constants for cmath functions |
| `call.rs` | 16 enum variants + `__nv_*` libdevice mappings |
| `collector.rs` | Intercept `std::sys::cmath::*` before forbidden-crate panic |
| `pipeline.rs` | Link `libdevice.10.bc` before `llc` instead of skipping `llc` |

The pipeline change replaces the "skip llc, emit NVVM IR" path with:
```
llvm-link kernel.ll libdevice.10.bc → merged.bc
opt -passes='internalize,globalopt,...' → optimized.bc  (28K lines → 126)
llc -march=nvptx64 -mcpu=sm_61 → kernel.ptx
```

This produces compact PTX with libdevice math functions inlined.

## Empirical Results (Tesla P4, sm_61)

**11 arithmetic kernels: BIT-IDENTICAL** (XOR = 0x00000000 vs CPU)
```
k_add, k_sub, k_mul, k_div, k_neg, k_abs, k_sqr,
k_relu, k_leaky_relu, k_hardsigmoid, k_softsign
```

**Transcendental kernels via libdevice link:**
- `k_silu`: ≤2 ULP vs CPU, **0 ULP vs nvcc** (bit-identical with nvcc on same hardware)
- `k_sigmoid`: ≤2 ULP vs CPU

## New functions unlocked

`tanh`, `sinh`, `cosh`, `asin`, `acos`, `atan`, `atan2`, `erf` — all mapped to `__nv_*` libdevice calls, available on every sm architecture from Pascal (sm_61) through Blackwell (sm_100+).

## Environment

- LLVM 21.1.0-rc1 (NVPTX backend)
- CUDA Toolkit 12.8
- Rust nightly-2026-04-03
- Hardware: Tesla P4 (Pascal, sm_61, 8GB GDDR5)
- `CUDA_OXIDE_LIBDEVICE` env var for libdevice path override

## Notes

- The `--emit-nvvm-ir` flag is preserved for users who want the existing nvJitLink path
- `find_libdevice()` searches `$CUDA_OXIDE_LIBDEVICE` then `$CUDA_TOOLKIT_PATH/nvvm/libdevice/`
- `llvm-link` and `opt` are found adjacent to the configured `llc` binary

## Blog Post

For the full story — why we built this, how we tested it, and what's next — see our write-up:

**[Bringing Rust GPU Kernels to Abandoned Hardware: Our CUDA-oxide Patch](https://ruachtov.ai/blog/bringing-rust-gpu-kernels-to-abandoned-hardware-our-cuda-oxide-patch.html)**


sempervictus:
Having support for hardware which is not "going forward" with CUDA but is commonly proliferated across the world is key IMO for faster and broader adoption. Thank you. Hopefully this can be used for things like a unified attention implementations across the range of viable hardware

heath-hunnicutt-ruach-tov:
Thank you @sempervictus — that's exactly the motivation. There are millions of Pascal and Turing GPUs still in active service, and they're perfectly capable hardware for inference workloads.

Your point about unified attention implementations is where this is heading. The libdevice linking approach in this PR means the same Rust kernel source compiles to correct PTX for any sm architecture — the `llvm-link + opt + llc` pipeline handles architecture-specific math library resolution automatically. A single attention kernel written in Rust could target sm_61 through sm_100+ without source changes.

We've verified bit-identical output (0 ULP) between nvcc and CUDA-oxide on the same Tesla P4 across 23 kernels including transcendentals like exp, tanh, and erf. The hardware produces the same bits regardless of compilation path.

kitty-eu-org:
I'm really looking forward to this feature and hope it can be merged as soon as possible.

sempervictus:
@nihalpasham: since this doesnt actually permit CUDA13+ to work with these older devices, there's no extension of the maintenance burden on NV staff for those older ISAs by merging this but far as adoption of this interface modality it should yield much faster ecosystem proliferation given the prevalence of kit covered out there (we build/run clusters for GPUaaS clouds - tend to be the first with access to new kit and internally _still_ have V100s doing some of our work because they simply work). ~30y ago i had a similar discussion w/ ATI's folks about unified driver design (that, they did; but only for newer generations of kit at the time) and user-accessible GPGPU API because OpenCL wasn't any better back then... that turned out as predicted and while its not as dire for NV to leave people out in the cold, doing so would add impetus to solving the inter-hardware disparity issue outside of NV's direct purview... :innocent:

nihalpasham:
Thank you for the PR. There's welcome work in here, and I want to be clear about which part we're taking and which we'd rather not; reasoning below. 

The per-function `libdevice` math support (tanh, atan, max/min, and friends) is useful, and we're bringing those in through the existing float-math placeholder pattern in smaller, focused PRs: #62 (f32::max/min) and #78 (atan/atan2) are examples. That pattern keeps each function's MIR-intrinsic ↔ `libdevice` mapping unit-tested and individually reviewable, and we'll keep extending it to cover the rest of this list.

The part that we can't take as-is is the change to how `libdevice` is consumed. Today a kernel that calls __nv_* math is routed through `NVVM IR` + `libNVVM` + nvJitLink, with `libdevice.10.bc` linked there. This PR instead links the `libdevice` bitcode with `llvm-link` + `opt` and emits `PTX` through `llc`, which is a second, parallel device-compilation strategy. Supporting older architectures (to a limited extent) on the `libNVVM` path is already on our roadmap (it needs typed-pointer NVVM IR for pre-Blackwell `libNVVM`, landing on top of an in-progress migration to upstream `pliron-llvm`), so we'd rather extend that one path than maintain two side by side. Combined with the 11 bundled examples and ~4000-line footprint, this is too broad to review and own as a single change.

So I'm going to close this PR, but the underlying goal isn't being dropped: the math-function coverage is landing piece by piece, and broader-target support is coming through the `libNVVM` path. 

**Note:** A focused proposal scoped to that path (separate from a new llc-based `libdevice` pipeline) would be very welcome. Thanks again for the thorough write-up and benchmarks.

PR #70, opened 2026-05-20
fbrv:refactor export dialect llvm

### LINKED ISSUE #59: Proposal: refactor `dialect-llvm` `export.rs` into fundamental building blocks
`crates/dialect-llvm/src/export.rs` is currently a large and difficult to navigate file. From a new contributor perspective, it appears to mix several responsibilities:
- export/backend configuration
- device extern declarations
- module and metadata emission
-  exporter state and kernel bookkeeping
- value naming and symbol normalization
- type/literal formatting
- function, block, and operation emission
- lowering-adjacent logic, such as block argument to `phi` handling

This makes the file hard to understand, test and modify.

This issue proposes a conservative structural refactor: split `export.rs` into a small number of fundamental modules while preserving the generated LLVM IR output.

A possible first-step structure:
```
crates/dialect-llvm/src/export/
  mod.rs        // public API / high-level orchestration
  config.rs     // backend configuration
  externs.rs    // device extern declarations
  state.rs      // exporter state and kernel bookkeeping
  names.rs      // value names, block labels, symbol normalization
  module.rs     // module-level export flow
  function.rs   // function and basic block emission
  ops.rs        // operation emission
  types.rs      // LLVM type printing
  literals.rs   // constant/literal formatting
  metadata.rs   // nvvm annotations and version, llvm.used
```
Initial goals:
- preserve current output exactly
- reduce the size and scope of `export.rs`
- make responsibilities clearer
- avoid introducing a large abstraction layer
- make future refactors easier to review

This should be a structural split rather than a semantic rewrite. Follow-ups can handle deeper changes, such as splitting operation emission by category or moving block argument to `phi` conversion out of the final textual printer.
CC: @nihalpasham

fbrv:
Closes #59.
Splits crates/dialect-llvm/src/export.rs into a module tree (config, externs, state, names, module, function, ops, types, literals, metadata) with no semantic changes and no new abstraction layers.

`ops.rs` replaces the `op_id == X::get_opid_static()` chain with a typed `LlvmOp<'op>` enum dispatched via `TryFrom<&dyn Op>`, populated by the classify_op! macro. Adding an op requires one macro entry and one match arm.


nihalpasham (approved):
Looks good. The split reads cleanly and `LlvmOp` + `classify_op!` is a tidy replacement for the long op_id chain.

I layered a small commit on top (c4f9803) that brings the value-name and predecessor-map pre-passes in `function.rs` into the same idiom: swapping the five `op_id == X::get_opid_static()` sites to `downcast_ref::<X>()`, the same primitive your macro uses internally. 

Also added a maintenance-contract docstring on `LlvmOp` so future contributors know which four places to touch when adding an op.

Ps: Just to be sure, also smoketested this - to check `.ll` parity holds against main after normalizing for monomorphization ordering artifacts and it does.

PR #73, opened 2026-05-21
fix(docs): clarify that installation requires nightly toolchain

vito-lin-dev:
**Description:**
The current installation steps often fail because cargo install defaults to the stable toolchain, causing #![feature] errors. This PR clarifies that cargo-oxide must be built in nightly mode and provides instructions on using the +nightly flag. This approach ensures a successful installation without requiring users to change their global default toolchain.

**Error Messages**

```
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> crates/cuda-core/src/lib.rs:39:1
   |
39 | #![feature(f16)]
   | ^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0554`.
error: could not compile `cuda-core` (lib) due to 1 previous error
error: failed to compile `cargo-oxide v0.1.0 (https://github.com/NVlabs/cuda-oxide.git#c8b3103c)`, intermediate artifacts can be found at `/tmp/cargo-installCSlgWm`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_BUILD_BUILD_DIR` to that path.
```

**Changes:**
Updated the installation command to enforce the specific nightly toolchain required by the project:

**From** 
```
cargo install --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide
```

**To**
```
cargo +nightly-2026-04-03 install --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,12 +130,12 @@ cargo oxide debug vecadd --tui
 
 #### cargo-oxide
 
-Inside the cuda-oxide repo, `cargo oxide` works out of the box via a workspace alias.
+Inside the cuda-oxide repo, `cargo oxide` works out of the box via a workspace alias, and you must build cuda-oxide in Rust nightly mode.
 
 For use outside the repo (your own projects):
 
 ```bash
-cargo install --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide
+cargo +nightly-2026-04-03 install --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide
 ```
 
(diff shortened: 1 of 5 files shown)


nihalpasham:
LGTM, thanks,  I pushed a small wording tweak on top: it now reads "install cargo-oxide with the pinned nightly toolchain" rather than "build cuda-oxide in Rust nightly mode", and moves the note next to the external-install step instead of the in-repo workspace-alias line. 

Good to merge.

PR #78, opened 2026-05-23
mir: lower f32::atan{,2} / f64::atan{,2} via libdevice

### LINKED ISSUE #77: Add support for {f32,f64}::atan{,2}
Currently, when using atan or atan2, kernel compilation fails

paulzhng:
Closes #77

After this change, kernels can now write `y.atan2(x)` / `x.atan()` directly.
Before that, using these functions lead to a kernel compilation failure.


nihalpasham:
Thanks for this, @paulzhng, LGTM. 

Once #62 merged it overlapped the same `RustFloatMathIntrinsic` block, and your fork didn't have maintainer-edit enabled, so I rebased your commit onto main, verified it (GPU `math_atan` passes within 2 ULP; reverting the collector change reproduces the FORBIDDEN CRATE: `std::sys::cmath::atanf` failure), and brought it in as #109 with your commit preserved as-authored. 

Closing in favor of #109. Appreciate it.

PR #80, opened 2026-05-24
add missing dependency on llvm-tools component

workingjubilee:
Subtle difference: "${rustc_sysroot}/lib/rustlib/${host_tuple}/lib", which is part of the library search path during linking, will now include links to libLLVM if llvm-tools is a component.

This allows `cargo test -p mir-importer` to pass without further fussing.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "nightly-2026-04-03"
-components = ["rust-src", "rustc-dev", "rust-analyzer", "clippy"]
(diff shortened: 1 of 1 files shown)


workingjubilee:
I guess this was as reasonable an excuse as any to finish learning commit signing.

Anyway, there's certainly other ways to address this, but this one seemed simplest.

nihalpasham:
Hi @workingjubilee, thanks for sending this in (and for the diagnosis).

I merged this into main yesterday as part of #6, which needed some changes. This also bumped the inner `crates/rustc-codegen-cuda/rust-toolchain.toml` to match. 

Going to close this as superseded #6  

Also: thank you for taking the time to get signed commits setup. Please keep the PRs coming.

PR #82, opened 2026-05-25
feat: CUDA constant memory via #[constant] / Constant<T>

rparolin:
Adds first-class support for CUDA constant memory (`__constant__` / PTX `.const`, address space 4) to cuda-oxide. Kernels can now declare module-scope statics that live in constant memory and read them as ordinary values; the host populates them between launches via macro-generated typed setters.

## Usage

```rust
use cuda_device::{constant, cuda_module, kernel, thread, Constant, DisjointSlice};

#[cuda_module]
mod kernels {
    use super::*;

    #[constant]
    static COEFFS: Constant<[f32; 4]> = Constant::UNINIT;

    #[kernel]
    pub fn apply(mut out: DisjointSlice<f32>) {
        let c = COEFFS.get();             // value-typed read from .const memory
        if let Some(e) = out.get_mut(thread::index_1d()) {
            *e = c[0] + c[1] * (i as f32);
        }
    }
}

// Host
module.set_COEFFS(&stream, &[10.0, 20.0, 30.0, 40.0])?;          // stream-ordered
module.set_COEFFS_blocking(&[10.0, 20.0, 30.0, 40.0])?;          // one-shot, sync
```

## Design

- **`Constant<T>` wrapper (`cuda-device`).** `#[repr(transparent)] Constant<T>(UnsafeCell<T>)`. The `UnsafeCell` is load-bearing: without it, rustc would constant-fold the initializer into every kernel body and the host's `cuMemcpyHtoD` would be invisible. `Constant<T>` is `Sync` (constant memory is single-host-writer / many-device-reader, with no in-kernel writes), unlike `SharedArray` which is per-block and `!Sync`.

- **`#[constant]` attribute (`cuda-macros`).** Validates the static is typed `Constant<T>`, rejects reserved names, and stamps it with `#[unsafe(export_name = "cuda_oxide_const_246e25db_<NAME>")]` so the PTX symbol is resolvable from the host.

- **`#[cuda_module]` expansion.** For every `#[constant]` static, the macro adds a private `ConstantHandle` field on `LoadedModule`, initializes it via `cuModuleGetGlobal` + a load-time `size_of::<T>()` assertion against the driver-reported size, and generates two typed setters: `set_<NAME>(&stream, &value)` (stream-ordered) and `set_<NAME>_blocking(&value)` (sync). Lookups happen once at load time; each write is a single `cuMemcpyHtoD[Async]` with no symbol resolution on the hot path.

- **`cuda-core`.** New `ConstantHandle` (a thin device-pointer wrapper) and `CudaModule::get_global` / `copy_bytes_to_device_global_sync`. Adds `memcpy_htod_sync` alongside the existing async variant.

- **MIR / lowering.** `MirGlobalAllocOp`'s verifier now accepts both global (`addrspace(1)`) and constant (`addrspace(4)`) result pointers (shared `addrspace(3)` remains rejected). The importer detects the mangled `cuda_oxide_const_*` export name when translating an ordinary Rust `static` reference and routes it to `addrspace(4)`; `mir-lower` propagates the address space through to LLVM globals. `reserved-oxide-symbols` exposes `constant_symbol()` so importer and macros agree on the mangling.

## Restrictions

- Initializer must be `Constant::UNINIT` (or another all-zeros value). Honoring arbitrary non-zero initializers in codegen is not yet implemented; populate from the host before any kernel reads.
- `T` must be valid as all-zeros bits (true for integers, floats, raw pointers, SIMD, and arrays thereof; **not** true for `NonZero*`, `&T`, or niched enums — using `UNINIT` with those and reading before populating is UB).
- `#[constant]` must appear inside a `#[cuda_module]`; placed elsewhere it produces an unreachable symbol with no setter.
- Export names are global — two `#[constant]` statics with the same identifier (even in different `#[cuda_module]`s in the same crate) will collide at link time.

## Tests / examples

- `dialect-mir/tests/ops_test.rs::test_mir_global_alloc_verify` — verifier accepts addrspace 1 and 4, rejects 3, rejects missing attrs.
- `examples/constant_memory_simple` — minimal end-to-end (`Constant<[f32; 4]>`, single kernel, host writes once).
- `examples/constant_memory` — larger surface, including stream-ordered updates between launches.
- `examples/constant_memory_coeffs` — matches the canonical CUDA C++ constant-memory coefficient pattern, side-by-side comparable with the reference.


nihalpasham (approved):
Thanks @rparolin for this. LGTM  .. 🚀 

Just needed the following changes/additions:
- Renamed `Constant<T>` to `ConstantMemory<T>` for clearer UX and CUDA terminology.
- Added unsafe marker trait `ConstantMemoryValue`, so - `UNINIT` is only safe for byte-copyable, all-zero-valid types.
- Generated setters are now idiomatic Rust names: `set_coeffs`, `set_coeffs_blocking`, `set_scale`, etc.
- Safe async setters now stage bytes into owned host memory and drop it with a stream callback, so temporaries like `&3.0` are safe with `cuMemcpyHtoDAsync`.
- Staging uses `Box<[MaybeUninit<u8>]>` so copied padding bytes are treated as raw storage, not initialized Rust u8 values.
- If callback enqueue fails and stream sync also fails, the staging buffer is intentionally leaked rather than possibly freed while CUDA may still read it.
- Constant symbols now include module/source-location context, avoiding collisions for same-named constants in different modules.
- LoadedModule resolves constant symbols lazily on first setter call, so unused declared constants no longer make `load()` fail.
- MIR detection now requires `cuda_device::ConstantMemory`, avoiding accidental special handling of unrelated user types with the same name.
- Added an unused `#[constant]` in `constant_memory_simple` as a regression guard for lazy resolution.

Note: The staging buffer overhead is acceptable for the normal constant-memory use case: small values populated once or a few times before launches. An unsafe borrowed/pinned fast path can be added later if needed.

PR #83, opened 2026-05-26
fix(docs): Adding missing sphinx-autobuild dependency

huymq1710:
## Description
This change adds `sphinx-autobuild` to requirements.txt

`make livehtml` currently fails with:
```
/bin/sh: sphinx-autobuild: command not found
```

After this update, the live docs server command works in a fresh venv after installing requirements


nihalpasham:
Thanks for this! It's a duplicate of #81, which makes the same one-line addition and was opened first, so I'm going to land that one and close this.

PR #88, opened 2026-05-26
cuda-bindings: handle CUDA header discovery errors

### LINKED ISSUE #87: cargo oxide doctor cannot diagnose missing CUDA headers because cuda-bindings panics first
## Summary

On a machine without CUDA headers installed, `cargo oxide doctor` cannot reach the doctor checks. The cargo alias first builds `cargo-oxide`, which pulls `cuda-bindings`; that build script falls through to bindgen and panics when `cuda.h` is missing.

## Repro

From a fresh clone on a host without `/usr/local/cuda/include/cuda.h`:

```bash
cargo oxide doctor
```

Observed failure:

```text
warning: cuda-bindings@0.1.0: cuda-bindings: Could not read cuda.h at /usr/local/cuda/include/cuda.h: No such file or directory
wrapper.h:1:10: fatal error: 'cuda.h' file not found
thread 'main' panicked at crates/cuda-bindings/build.rs:70:10:
Unable to generate CUDA bindings: ClangDiagnostic(...)
```

## Expected

The build script should report missing CUDA headers directly and avoid a bindgen panic. Related: CUDA redistributable/cross layouts may place headers under target-specific directories such as `targets/x86_64-linux/include` or `targets/sbsa-linux/include`; the build script should not hardcode a single target directory.

## Notes

PR #88 updates the build script to:

- discover `cuda.h` in both `include/` and `targets/*/include/`
- use the discovered header for version/cfg probing
- pass all discovered include directories to bindgen
- add matching target-layout library paths for discovered target roots
- return a direct build-script error for missing/unreadable headers instead of panicking in bindgen

yagna-1:
## Summary
- Make the `cuda-bindings` build script report missing or unreadable CUDA headers directly instead of falling through to a bindgen panic.
- Discover CUDA headers in both the standard `include/` layout and target-specific redistributable layouts such as `targets/x86_64-linux/include/` and `targets/sbsa-linux/include/`.
- Pass discovered include paths to bindgen and matching target-layout library paths to rustc.

Fixes #87.

## Test plan
- `cargo fmt --check --all`
- `cargo test -p cuda-device -p oxide-artifacts -p reserved-oxide-symbols --lib --tests`
- `cargo test -p cargo-oxide --all-targets` on a host without CUDA headers now fails with the direct `cuda-bindings: could not find cuda.h...` build-script error instead of a bindgen panic.

### DIFF
--- a/crates/cuda-bindings/build.rs
+++ b/crates/cuda-bindings/build.rs
@@ -32,22 +32,28 @@ fn run() -> Result<(), Box<dyn Error>> {
     println!("cargo::rustc-check-cfg=cfg(cuda_has_cuEventElapsedTime_v2)");
 
     let toolkit = cuda_toolkit_dir();
-    let cuda_h = Path::new(&toolkit).join("include/cuda.h");
+    let include_paths = collect_include_paths(&toolkit);
+    let cuda_h = include_paths
+        .iter()
+        .map(|include| include.join("cuda.h"))
+        .find(|path| path.is_file())
+        .ok_or_else(|| {
+            format!(
+                "cuda-bindings: could not find cuda.h under {}. Set CUDA_TOOLKIT_PATH or CUDA_HOME to a CUDA Toolkit install that contains include/cuda.h or targets/x86_64-linux/include/cuda.h.",
+                toolkit
+            )
+        })?;
     println!("cargo:rerun-if-changed={}", cuda_h.display());
 
-    match std::fs::read_to_string(&cuda_h) {
-        Ok(contents) => {
-            if contents.contains("cuEventElapsedTime_v2") {
-                println!("cargo:rustc-cfg=cuda_has_cuEventElapsedTime_v2");
-            }
-        }
-        Err(err) => {
-            println!(
-                "cargo:warning=cuda-bindings: Could not read cuda.h at {}: {}",
-                cuda_h.display(),
-                err
-            );
-        }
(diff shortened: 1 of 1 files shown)


fbrv:
I think this is not enough: in sbsa environment this wouldn’t work, even if cuda is installed.
Also it doesn’t take in consideration environments where cuda is not installed in the default path but I think it’s a project decision to support or not default paths (@nihalpasham)

yagna-1:
Updated to scan CUDA target-layout directories via `targets/*` instead of hardcoding `targets/x86_64-linux`, so SBSA layouts such as `targets/sbsa-linux` are covered as well.

fbrv:
I think this introduces another problem: if multiple CUDA targets are installed, we may end up mixing dependencies and likely cause build/link errors. We should probably first discover the CUDA host target, and only then look for cuda.h under that target.

nihalpasham:
Thanks for the PR, the diagnosis is right and the error-message UX is a nice improvement. We can't accept external contributions to `cuda-bindings` though (it's governed by an NVIDIA license), so I've re-landed the idea NVIDIA-authored as #172 

Two changes from your version:
- discovery picks the single targets/<dir> matching the build target instead of a sorted glob (on multi-target installs the glob puts sbsa-linux ahead of x86_64-linux)
- rebased onto the post-#152 build.rs

The doctor half of #87 stays open. Closing as already-fixed. (Your #95 merged today, thanks for both.)

PR #89, opened 2026-05-26
fix(cuda-core): use malloc_sync for DeviceBuffer allocations

winshaurya:
### **Description**
Pairs `DeviceBuffer` constructors with the correct matching allocation strategy to fix undefined behavior.

**Why:**
I ran into this bug while developing a small game on `cuda-oxide`. As entity buffers were constantly allocated and dropped in the game loop, the game would crash or render corrupted overlapping meshes after a few minutes. 

`DeviceBuffer` was allocating memory via `cuMemAllocAsync` (stream-ordered) but freeing it via `cuMemFree_v2` (legacy synchronous) on drop. Passing a stream-allocated pointer to `cuMemFree_v2` violates the CUDA API contract, causing the driver to silently corrupt the virtual memory pool. This leads to silent memory corruption, premature OOMs, or runtime crashes.

The Impact: It corrupts the graphics card's memory pool. After a few minutes of creating and deleting buffers 
1. Random crashes.
2. Fake "Out of Memory" errors.
3. Silent data corruption (objects overlapping in memory and overwriting each other).

**Changes:**
Switched `malloc_async` to `malloc_sync` in `DeviceBuffer::from_host` and `DeviceBuffer::zeroed` to correctly pair allocations with the `free_sync` call in `Drop`.

PS: CUDA rules state that you must use the same type for both (async-with-async, or sync-with-sync). Mixing them is not allowed.

### DIFF
--- a/crates/cuda-core/src/device_buffer.rs
+++ b/crates/cuda-core/src/device_buffer.rs
@@ -212,7 +212,7 @@ impl<T: DeviceCopy> DeviceBuffer<T> {
         let len = data.len();
         let num_bytes = std::mem::size_of_val(data);
 
-        let ptr = unsafe { crate::memory::malloc_async(stream.cu_stream(), num_bytes)? };
+        let ptr = unsafe { crate::memory::malloc_sync(num_bytes)? };
         unsafe {
             crate::memory::memcpy_htod_async(ptr, data.as_ptr(), num_bytes, stream.cu_stream())?;
         }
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks, this is the right fix. `DeviceBuffer` is the synchronous RAII type, so its allocation should use `cuMemAlloc` to match the `cuMemFree` on drop. The old `cuMemAllocAsync` was pulling in the stream-ordered pool allocator, which is what corrupted the pool. Pairing both as sync is correct.

**One tip for your game-loop case:** since you're allocating and freeing every frame, `DeviceBox` in cuda-async is the better fit. It's the pool-backed, stream-ordered type (frees via `cuMemFreeAsync`), so high alloc/free churn stays cheap, whereas `DeviceBuffer` trades that for simple synchronous ownership. 

But thanks for the catch here!

PR #91, opened 2026-05-26
docs: update README examples count and highlights (#84)

### LINKED ISSUE #84: README example count is outdated (says 46, actually 56)
In README.md:

> **46 examples** in `crates/rustc-codegen-cuda/examples/`.

But `find crates/rustc-codegen-cuda/examples -mindepth 1 -maxdepth 1 -type d | wc -l` reports 56.

The table of highlights may also need updating if new examples should be listed.

This is a simple doc fix.

edengilbertus:
updated the example count from 46 to the current actual count of 56 and added the 'device_ffi_test' and 'mathdx_ffi_test' examples to the highlight tables in both README files. Closes #84.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ compiles a Rust kernel to PTX, launches it on the GPU, and prints
 
 ## Examples
 
-**46 examples** in `crates/rustc-codegen-cuda/examples/`. Highlights:
+**56 examples** in `crates/rustc-codegen-cuda/examples/`. Highlights:
 
 | Example              | Description                                                              |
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks,  I changed the count to "60+ examples" instead of a hard number, since the exact value keeps drifting (46 → your 56 → main is already at 62), and "60+" stays accurate without a re-edit every time someone adds an example. 

Good to merge; closes #84.

nihalpasham:
closes #84.

PR #92, opened 2026-05-26
Fix/GPU memory leak on DeviceBuffer allocation failure

winshaurya:
# Description 
There is a silent GPU memory leak during DeviceBuffer construction when initialization fails.

Specifically, in DeviceBuffer::from_host and DeviceBuffer::zeroed, the code allocates raw GPU memory (malloc_sync wrapping cuMemAlloc_v2), which successfully returns a CUdeviceptr handle. The code then enqueues an asynchronous operation (memcpy_htod_async or memset_d8_async) and uses the ? operator to bubble up errors.

If this asynchronous step fails, the function returns early. Because the DeviceBuffer struct was never fully constructed, its standard Drop implementation (which executes free_sync wrapping cuMemFree_v2) never triggers. The raw CUdeviceptr is discarded, and the VRAM allocation is permanently leaked.

## Why
Allocating device memory is a synchronous step, but copying data to it is asynchronous. If the async copy fails mid-transit, we must manually free the allocated VRAM.

## Impact
 Any long-running server or retry loop that hits transient stream errors will bleed GPU memory rapidly, eventually leading to an Out-of-Memory crash that forces a host process restart to clear the VRAM.
 
## Changes
Added an internal RAII cleanup helper DevicePtrGuard that takes ownership of the CUdeviceptr and its Arc<CudaContext> immediately after allocation.
If any subsequent async operation fails, the guard drops, binds the context to the active host thread (ctx.bind_to_thread()), and calls free_sync(ptr).
If construction succeeds, we call guard.disarm() which invokes std::mem::forget to bypass the destructor at zero runtime cost.
Created integration tests in a new file crates/cuda-core/tests/device_buffer.rs to verify normal buffer roundtrips, zeroed allocations, and zero-length boundaries.

PS : pointers with provenance vid. of compiler bits, helped me to solve this cleanly after finding the issue.


nihalpasham:
Thanks. LGTM. 

A couple of things:
- Merged origin/main: your branch carried the same `malloc_sync` hunks #89 already landed, which is why GitHub showed conflicts
- Replaced `DevicePtrGuard` with constructing the owning `DeviceBuffer` right after `malloc_sync`, so the buffer's own `Drop` is the single failure-path cleanup. The disarm via `mem::forget` also forgot the guard's `Arc<CudaContext>` clone, leaking one context strong count per successful construction
- Special-cased `num_bytes == 0` in both constructors: `cuMemAlloc` rejects zero-byte requests since #89, so your `device_buffer_supports_empty_allocations` test now passes on hardware
- Your tests are unchanged (cargo fmt only); added tests/device_buffer_leaks.rs with strong-count, `cuMemGetInfo` VRAM, and zero-length regression probes

Note: the constructors allocate in whatever CUDA context the calling thread currently has bound, not necessarily the stream's context. In multi-context programs, alloc and free could run under different contexts. This would be a nice but separate follow-up

winshaurya:
Hey @nihalpasham sir,

thanks for refining and merging the PR! I really appreciate the refinements you added. Constructing the owning DeviceBuffer immediately after malloc_sync is definitely much cleaner than my DevicePtrGuard workaround great catch on mem::forget leaking the Arc<CudaContext> strong count. 

I have a couple of thoughts on next steps to make sure we're fully covered:

Since the constructors allocate in whichever CUDA context is currently active, should we explicitly validate or enforce the target context during from_host and zeroed to prevent allocations from landing in the wrong place?

I’d be happy to take on the multi-context validation issue next , please provide me a direction to approach it.

Thanks

nihalpasham:
Hey @winshaurya, 

Just so that we aim at the right fix - it's not really a "validate the user's context" problem. Only one context matters here which is the stream's. The real gap is where the sync allocation lands:

- `cuMemAlloc` (our `malloc_sync`) allocates in whatever context is currently bound to the calling thread.
-` from_host` / `zeroed` / `from_host_async_unchecked` clone `stream.context()` for bookkeeping and drop, but never bind it before allocating.
- So if the thread has a different context current, the buffer gets allocated in context A, recorded as context B, and Drop frees it under B. That's a cross-context free, which the driver rejects (or worse).

The async constructor already shows the correct model: `uninitialized_async` uses `cuMemAllocAsync(stream, …)`, which is stream-ordered and inherits the stream's context automatically, so it can't drift.

So the fix is small and local, not a new validation layer:

In `from_host`, `rom_host_async_unchecked`, and `zeroed`, bind the stream's context right after cloning it and before `malloc_sync`:
```rust
let ctx = stream.context().clone();
ctx.bind_to_thread()?; // alloc must land in the stream's context, not the thread's current one
```
- That makes alloc context == recorded context == free context. Done.
- Skip the `debug_assert!(Arc::ptr_eq(...))` style check. That pattern (in `from_pinned_host`) is for a different invariant, where two user objects must share a context. You don't need an extra arg or assert here.

One testing heads-up:
- `CudaContext::new(n)` retains the device primary context, so two `new(0)` calls hand back the same context. A true two-context regression test needs two GPUs today (`new(0)` + `new(1)`), or we'd first have to expose a secondary (`cuCtxCreate)` context.
- So land the fix as a correctness change modeled on the async path, and add a GPU-gated multi-GPU test in the style of tests/device_buffer_leaks.rs if hardware allows.

Process: open a short tracking issue, then a small PR against `cuda-core` (label `host-apis`). Ping me on the issue and I'll point at anything else that needs the same bind.

Issue #99, opened 2026-05-29
Owned async launches don't seem cancellation-safe after first poll


fallintoplace:
I think the owned async launch path introduced in `04d7631` (`Generate typed CUDA module launch APIs`, 2026-05-09) has a cancellation-safety hole.

From reading the current code, `*_async_owned(...)` is documented as the safe way to move device buffers into a `'static` future and spawn it, but once the future has been polled the owned resources can still be dropped before the stream finishes.

Why I think this is a bug:

- `OwnedAsyncKernelLaunch<R>::execute` launches the kernel and immediately returns `resources`.
- `DeviceFuture::execute` stores that `R` in `self.result` while the future is still pending.
- If the future is then dropped before the callback fires, `self.result` is dropped immediately.
- For the common case where `R` contains `DeviceBox<_>`, that is not just "dropping the output early": `DeviceBox::drop` enqueues `cuMemFreeAsync` on the deallocator stream, and its own docs say all streams using the allocation must be synchronized before drop.
- I don't see any ordering between the launch stream and the deallocator stream here.

So this safe pattern seems wrong:

```rust
let op = module.vecadd_async_owned(config, a_dev, b_dev, c_dev)?;
let fut = op.into_future();

let handle = tokio::spawn(async move {
    let _ = fut.await;
});

handle.abort();
```

If that task is aborted, or if the future is otherwise dropped after the first poll but before completion, the owned buffers look like they can be freed while the kernel is still in flight.

There also seems to be a second bad path:

- `DeviceFuture::poll` does `execute()` first and `register_callback()` second.
- If the kernel launch succeeds but callback registration fails, `poll()` returns `Err(...)`.
- At that point the owned resources are dropped even though GPU work may already be running.

The reason this stood out to me is that the docs currently promise the opposite. The README says the owned form "keeps device buffers alive until the CUDA stream reaches the kernel completion callback" and shows `tokio::spawn(op.into_future())` as the intended pattern.

Relevant spots:

- `crates/cuda-async/src/launch.rs`
- `crates/cuda-async/src/device_future.rs`
- `crates/cuda-async/src/device_box.rs`
- `crates/cuda-async/README.md`

I haven't run a GPU repro for this yet, so if I'm missing an ordering guarantee somewhere I'd love to know. But from the source-level ownership path, this looks like a real bug.

Possible fix directions:

- make `DeviceFuture` hold onto owned results until stream completion even if the future is dropped
[...]

PR #100, opened 2026-05-29
Keep in-flight async future results alive on drop

fallintoplace:
Partially addresses #99.

This keeps the owned async launch path from dropping an in-flight stored result as soon as the future is dropped after its first poll.

What changed:
- add `DeviceFuture::Drop` cleanup that synchronizes the assigned stream before releasing a stored in-flight result
- if that cleanup cannot prove the stream is idle, log the cleanup failure and leak the stored result instead of risking an early drop
- add small unit tests around cleanup ordering and the sync-failure behavior
- document the blocking drop behavior in the `cuda-async` README

Why:
- `OwnedAsyncKernelLaunch::execute` moves owned resources into `DeviceFuture::result` while the future is still pending
- dropping that future after submission could then drop those resources while stream work was still in flight
- for `DeviceBox`, that is especially risky because drop enqueues `cuMemFreeAsync` on a separate deallocator stream

Tradeoff:
- dropping an in-flight future can now block long enough to synchronize the assigned stream

Scope note:
- this PR is intentionally narrow and only covers the normal drop/cancellation path after first poll
- it does not try to solve the separate "launch succeeded, callback registration failed" path yet


nihalpasham:
Thanks for this. I verified it on main with a repro where a cancelled in-flight kernel overwrote `4096/4096` elements of a freshly allocated buffer through stream-ordered allocator reuse. 

A couple of things: layered the following commits on top
- 41f78f4 makes the `Drop` design compile: explicit field construction at both scheduling sites (functional update is E0509 once the type has Drop) and the stream `Arc` cloned out before the `&mut` cleanup call (E0502)
- fec3f83 adds a non-blocking `CudaEvent::query` (cuEventQuery) to `cuda-core`
- 3803bc0 swaps the blocking `cuStreamSynchronize` in `Drop` for deferred reclamation: drop records an event on the assigned stream and parks the result; later polls/drops sweep parked entries once the event has passed. Drop never blocks executor threads; your synchronize-then-leak-loudly path survives as the fallback when the event cannot be recorded
- 2c60a07 covers the second path from #99: launch Ok + register_callback `Err` left a complete future still holding live resources; cleanup now gates on submitted-with-result and routes through the same reclamation
- b3638a5 keeps your three unit tests unchanged and adds the callback-failure shape plus limbo sweep/drain coverage
- 43b7707 updates the README: cancelling never cancels GPU work, it defers host-side reclamation until the device timeline passes the recorded event

One API consequence: `DeviceOperation::Output` is now `Send + 'static`, since a parked result can outlive any borrow it carries.

PR #101, opened 2026-05-30
Support typed NVVM IR for pre-Blackwell libNVVM targets

### LINKED ISSUE #98: nvvmCompileProgram fails with "parse expected type" when libdevice calls (__nv_expf) trigger NVVM IR mode
# nvvmCompileProgram fails with "parse expected type" when libdevice calls trigger NVVM IR mode

## Environment

| Component | Version |
|-----------|---------|
| Rust | `1.96.0-nightly (55e86c996 2026-04-02)` |
| LLVM | `22.1.2` |
| CUDA Toolkit | `13.2 (V13.2.78)` |
| GPU Driver | `595.58.03` (Open Kernel Module) |
| GPU | `NVIDIA GeForce RTX 3050 Ti Laptop GPU` (sm_86) |
| OS | `Linux 6.19.10-arch1-1 x86_64` |
| cuda-oxide commit | `396c76a` (NVlabs/cuda-oxide) |

```
cargo oxide doctor
cargo-oxide environment check
==============================

Rust nightly toolchain... ✓ rustc 1.96.0-nightly (55e86c996 2026-04-02)
rust-toolchain.toml... ✓ present
Codegen backend... ✓ /home/moyi/.cargo/cuda-oxide/librustc_codegen_cuda.so
CUDA toolkit (nvcc)... ✓ Cuda compilation tools, release 13.2, V13.2.78
nvJitLink (libnvJitLink.so)... ✓ nvJitLink 13.2
libdevice (libdevice.10.bc)... ✓ /opt/cuda/nvvm/libdevice/libdevice.10.bc
llc (LLVM)... ✓ LLVM version 22.1.2 (llc)
clang / libclang resource dir... ✓ /usr/lib/clang/22
cuda-gdb (optional)... ✓ exec: /opt/cuda/bin/cuda-gdb-minimal --version

✅ Environment looks good!
```

## Minimal Reproduction

`crates/rustc-codegen-cuda/examples/libdevice_ir_compat/`

**Cargo.toml:**
```toml
[package]
name = "libdevice_ir_compat"
version = "0.1.0"
edition = "2024"

[workspace]

[dependencies]
cuda-device = { git = "https://github.com/NVlabs/cuda-oxide.git" }
cuda-host = { git = "https://github.com/NVlabs/cuda-oxide.git" }
cuda-core = { git = "https://github.com/NVlabs/cuda-oxide.git" }
```

**src/main.rs:**
```rust
use cuda_device::{cuda_module, kernel, thread, DisjointSlice};

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    pub fn swiglu_libdevice(x: &[f32], y: &[f32], mut out: DisjointSlice<f32>) {
        let idx = thread::index_1d();
        let i = idx.get();
        if let Some(out_elem) = out.get_mut(idx) {
            let xi = x[i];
            let yi = y[i];
            let sigmoid = 1.0f32 / (1.0f32 + (-xi).exp());
            *out_elem = xi * sigmoid * yi;
        }
    }
}

    let host_x = vec![-2.0f32; 4];
    let host_y = vec![1.0f32; 4];
    let dev_x = DeviceBuffer::from_host(&stream, &host_x)?;
    let dev_y = DeviceBuffer::from_host(&stream, &host_y)?;
    let mut dev_out = DeviceBuffer::from_host(&stream, &vec![0.0f32; 4])?;

    module.swiglu_libdevice(
        &stream,
        LaunchConfig::for_num_elems(4),
        &dev_x,
        &dev_y,
        &mut dev_out,
    )?;

    let actual = dev_out.to_host_vec(&stream)?;
    let expected: Vec<f32> = host_x
        .iter()
        .zip(host_y.iter())
        .map(|(&x, &y)| {
            let s = 1.0f32 / (1.0f32 + (-x).exp());
            x * s * y
        })
        .collect();
    for i in 0..4 {
        assert!((actual[i] - expected[i]).abs() < 1e-5);
    }
    println!("PASS");
    Ok(())
}
```

expect output: PASS
actually:
```
Error: ModuleNotFound { name: "test" }

Failed with exit code: Some(1)
```
## Workaround

Avoid Rust float-math intrinsics that trigger libdevice calls. Use only basic arithmetic and manual branching in GPU kernels:

```rust
// ❌ 
let sigmoid = 1.0 / (1.0 + (-xi).exp());
let abs_xi = xi.abs();

// ✅ 
let abs_xi = if xi >= 0.0 { xi } else { -xi };
let sigmoid = xi / (1.0 + abs_xi);
```

I am relatively unfamiliar with CUDA and nvvm. Can someone tell me why I can't write it this way? Is it an issue with my environment, or am I not supposed to write it this way?

mohamedsamirx:
## Summary

Fixes #98.

Kernels that call `sin`, `exp`, `sqrt`, `atan2`, or other CUDA libdevice
functions use the NVVM IR path. Pre-Blackwell libNVVM parses LLVM 7 syntax with
typed pointers, while Blackwell and newer targets parse modern opaque-pointer
IR. cuda-oxide now emits, verifies, and loads the correct form for each target.

```text
Before: add libdevice math → pre-Blackwell libNVVM reports `parse expected type`
After:  the same kernel builds and runs through the normal `#[cuda_module]` API
```

The ordinary LLVM IR → `llc` → PTX path remains unchanged.

## What changed

- **Target-selected NVVM IR**
  - Adds one strict `CudaArch` parser for `sm_*` and `compute_*` targets,
    including suffixes such as `sm_90a`.
  - Selects LLVM 7 typed-pointer IR below compute capability 10.0, including
    `sm_90` and `sm_90a`.
  - Selects modern opaque-pointer IR for Blackwell and newer targets.
  - When an ordinary build discovers libdevice without `--arch`, uses the
    normal target order: explicit target, compatible `run` GPU hint, then the
    compiler's feature-based target. Basic headless builds select `sm_80`.
  - Uses the same resolved target for legalization, text export, libNVVM,
    nvJitLink, embedded metadata, and `.target` files.

- **Complete LLVM 7 pointer export**
  - Gives every internal pointer one neutral type per address space:
    `i8 addrspace(N)*`.
  - Converts typed results from allocas, GEPs, globals, function addresses,
    calls, and other pointer producers back to that neutral form.
  - Creates exact same-address-space typed views at loads, stores, GEPs,
    calls, and other typed uses.
  - Keeps pointer PHIs, selects, calls, returns, arrays, and structs
    type-consistent without assuming that every address has one pointee type.
  - Preserves pointer address spaces. Only an explicit source
    `addrspacecast` may cross address spaces.

- **Structured ABI boundaries**
  - Preserves pointer pointees and address spaces for device-extern
    declarations after the function body has moved to opaque pointers.
  - Adds typed adapters at supported external calls and returns.
  - Rejects signatures that cannot yet be represented exactly, including
    non-C or unwinding ABIs, variadics, unsupported small by-value values,
    scalar `f16`, and arrays passed by value.

- **Dedicated NVVM transform crate**
  - Adds `nvvm-transforms` between MIR lowering and text export:

    ```text
    mir-transforms → mir-lower → nvvm-transforms → verify → llvm-export
    ```

  - Keeps `mir-lower` focused on `dialect-mir` → LLVM conversion.
  - Rewrites `fneg` by toggling the IEEE sign bit, preserving signed zero and
    NaN payload bits.
  - Expands integer and float-to-integer saturation, modern shuffle/vote
    signatures, LLVM 7 memory intrinsic names, unsupported `nneg` flags, and
    wide bit operations.
  - Validates the module before changing it and returns an error when no
    equivalent legacy operation is available.

- **Verification and artifact handling**
  - Runs `nvvmVerifyProgram` before compilation and queries the toolkit's NVVM
    IR/LLVM versions when available.
  - Records the source target beside embedded and file-based NVVM artifacts.
  - Removes artifacts from earlier output modes before publishing new output.
  - Rebuilds file-based NVVM output instead of reusing cubins by timestamp
    alone.
  - Selects libNVVM, libdevice, and nvJitLink from the same CUDA toolkit root
    unless a component-specific path override is set.
  - Uses open integer wrappers for libNVVM and nvJitLink result codes so newer
    library error values remain valid across the FFI boundary.

- **Blackwell forward execution**
  - Native loading still requires an exact target match.
  - A standard pre-Blackwell target such as `sm_86` can be converted to
    original-target PTX and JIT-compiled by the CUDA driver on Blackwell.
  - The driver must support the PTX version emitted by the selected toolkit;
    error 222 now points users to a compatible toolkit or driver upgrade.
  - Newer output cannot run on older GPUs, and architecture-specific targets
    such as `sm_90a` are not forwarded to another compute capability.

### Capability coverage

- [x] `sm_75`, `sm_86`, `sm_90`, and `sm_90a` LLVM 7 output
- [x] `sm_100` and `sm_120` modern output
- [x] Pointer PHIs, loop backedges, calls, returns, nested structs, and pointer arrays
- [x] One address used through more than one pointee type
- [x] Structured device-extern pointer ABIs
- [x] Whole-module libNVVM verification
- [x] Automatic libdevice-triggered NVVM selection
- [x] Standard pre-Blackwell output running forward on Blackwell
- [x] Existing PTX path unchanged

## Regression examples

- **`libdevice_math`** — the issue #98/community kernel shapes, SwiGLU, and
  `sin`, `exp`, `sqrt`, `atan`, `atan2`, `acos`, and `tan`, with CPU numerical
  checks. This replaces the issue-number-based `issue98_repro` name.
- **`legacy_nvvm_pointer_shapes`** — pointers across branches, a loop,
  non-inlined calls, nested structs, an array of pointers, and `f32`/`u32`
  views of the same address.
- **`primitive_stress`** — wide bit operations, saturation, and 50 floating
  point checks through both NVVM dialects.

Compile-only smoke tests use `emit-ltoir`, so all three examples pass through
the real libNVVM verifier and compiler instead of stopping after text export.

## Testing

- [x] `cargo fmt` for the root workspace, backend, and changed examples
- [x] `cargo test --workspace --all-targets`
- [x] Strict Clippy for the changed compiler/runtime crates and backend
- [x] Warning-free workspace rustdoc and all doctests
- [x] `cargo deny check`
- [x] Strict Sphinx build with warnings treated as errors
- [x] Full LLVM 21 compile-only smoke gate: 119 / 119 examples
- [x] Zero-flag libdevice builds select `sm_80`, emit typed-pointer IR, and record `.target`
- [x] Real libNVVM compile smoke: `libdevice_math`, pointer shapes, and primitive stress
- [x] CUDA 13.0 and 13.3 LTOIR matrix: `sm_75`, `sm_86`, `sm_90`, `sm_90a`, `sm_100`, `sm_120`
- [x] Generated `sm_86` LLVM IR contains typed pointers and no opaque `ptr`
- [x] RTX 5090 automatic architecture detection: `libdevice_math`
- [x] RTX 5090 forward `sm_86` execution with CUDA 13.0: all three regression examples
- [x] RTX 5090 native `sm_120` execution: all three regression examples
- [x] Ordinary PTX execution regression

## Current limits

- This machine does not have a physical pre-Blackwell GPU or a CUDA 12
  toolkit. The legacy modules were accepted by the real LLVM 7 parser selected
  by CUDA 13 libNVVM and executed through the forward PTX path on Blackwell.
  The contributor's original CUDA 12.4 / T1200 report establishes the failure
  on `main`; native pre-Blackwell and CUDA 12 validation remain useful follow-up
  hardware checks.
- LLVM 7 NVVM output does not yet support cuda-oxide debug metadata. The
  compiler reports this instead of emitting unverified legacy debug IR.
- The legacy NVVM legalizer currently rejects LLVM atomic operations. NVVM
  supports a defined subset, but the required validation and lowering have not
  yet been implemented.
- LLVM fences require lowering to NVVM memory-barrier operations, while scalar
  `f16` is not supported by the CUDA 12 LLVM 7 dialect. Both are rejected
  instead of being emitted with incorrect semantics.
- `UnsupportedPayloadKind(272)` is a separate artifact-handling problem and is
  intentionally deferred to another PR.

## Checklist

- [x] Original contributor commit and authorship preserved
- [x] All PR commits include `Signed-off-by`
- [x] No files under `crates/cuda-bindings/` are modified
- [x] New Rust source files include SPDX headers
- [x] Documentation and example comments use plain, user-facing language


nihalpasham:
Thanks for the PR. This one actually has a larger scope.

- Typed pointers support for the older version of LLVM IR is on our to do list. 

But we'd need to land a couple of other things first, such as migrating from `dialect-llvm` to upstream `pliron-llvm` and leveraging a more structured dataflow oriented design for how pointer types are propagated across the pipeline. 

I'll keep this PR open till those land and will layer them on top of this one.

PR #102, opened 2026-06-01
mir-lower: add unit tests for memory op conversion

goog00:
## Summary

Adds end-to-end unit tests for the `dialect-mir` → `dialect-llvm` memory op conversion 

Each `convert_*` function takes a live `DialectConversionRewriter` owned by pliron's conversion driver and is not constructible standalone, so every test builds a minimal MIR module, runs the full `lower_mir_to_llvm` pass, and asserts the lowered module has the expected `dialect-llvm` shape.

## Coverage

| MIR op             | LLVM op(s)                     | Assertions                                 |
|--------------------|--------------------------------|--------------------------------------------|
| `mir.alloca`       | `llvm.alloca`                  | op count, elem type = `i32`                |
| `mir.load`         | `llvm.load`                    | op count, result width = 32                |
| `mir.store`        | `llvm.store`                   | op count, value/address operand types swapped correctly |
| `mir.ref`          | `llvm.alloca + llvm.store`     | op counts, alloca elem type, store operand wiring; both `&x` and `&mut x` |
| `mir.ptr_offset`   | `llvm.getelementptr`           | gep elem type from pointee  |
| `mir.shared_alloc` | `llvm.global + llvm.addressof` | addrspace=3, `__shared_mem_` prefix, `[64 x i32]` size, key-based dedup |
| `mir.global_alloc` | `llvm.global + llvm.addressof` | addrspace 1 vs 4, mangled name vs `__device_global_` prefix |

## Test plan

- `cargo test -p mir-lower --lib convert::ops::memory::tests` passes (9 tests)
- `cargo clippy -p mir-lower --lib --tests` clean

### DIFF
--- a/crates/mir-lower/src/convert/ops/memory.rs
+++ b/crates/mir-lower/src/convert/ops/memory.rs
@@ -686,5 +686,485 @@ fn get_or_create_extern_shared_global(
 
 #[cfg(test)]
 mod tests {
-    // TODO (npasham): Add unit tests for memory conversion
+    //! End-to-end lowering tests for `dialect-mir` memory ops.
+    //!
+    //! The `convert_*` functions in this file take a live
+    //! `DialectConversionRewriter`, which is owned by pliron's conversion
+    //! driver and not constructible standalone. So each test builds a small
+    //! MIR module, runs the full `lower_mir_to_llvm` pass on it, and asserts
+    //! the lowered module contains the expected `dialect-llvm` shape — same
+    //! pattern as `tests/lowering_test.rs`.
+
+    use super::*;
+    use dialect_llvm::op_interfaces::PointerTypeResult;
+    use dialect_llvm::ops as llvm;
+    use dialect_llvm::types::{PointerType, address_space as llvm_addr};
+    use dialect_mir::ops as mir;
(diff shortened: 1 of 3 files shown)


goog00:
The follow-up commit extracts the end-to-end test scaffolding out of `memory.rs` into a new `convert::ops::test_util` module, so the upcoming control-flow and other per-op test modules can reuse it instead of each re-copying the same
~80 lines of boilerplate.

**Moved into `test_util` (shared):** `make_ctx`, `build_kernel`, `append_block`, `append_mir_return`, `kernel_blocks`, `module_top_block`, `count_ops`, `find_first`, `find_all`. `build_kernel` / `append_mir_return` take the generic (`ret_tys` / `vals`) form; memory's void-return call sites pass empty vecs.

**Kept local to `memory.rs` (domain-specific):** `ptr_addrspace`, `append_shared_alloc`, `append_global_alloc`.

nihalpasham:
Thanks for the PR.

**Heads up:** an in-flight `dialect-llvm → llvm-export` rename will touch these imports, but we'll carry that in the migration PR, nothing for you to do on your end.

PR #103, opened 2026-06-02
Add support for cargo oxide debug and pipeline commands in non-example projects

Barrow099:
The `cargo oxide {debug, pipeline}` commands were requiring an example folder to run, thus unusable for "normal" projects. This PR aims to solve this issue by reusing the code used by the `run` and `build` commands to work in normal projects.


nihalpasham:
LGTM 👍🏾. 

Layered one polish commit on top:
- `--help` for pipeline/debug now says "required in workspace, optional for standalone projects", matching run/build
the workspace-mode usage error names the actual subcommand (Usage: cargo oxide pipeline <EXAMPLE>) instead of always saying run

PR #104, opened 2026-06-03
`type_id_u128` requires a 'static lifetime bound on T in nightly

bitemyapp:
`cargo +nightly install --git https://github.com/NVlabs/cuda-oxide.git cargo-oxide`

```
error[E0310]: the parameter type `T` may not live long enough
  --> crates/cuda-host/src/type_id.rs:48:22
   |
48 |     let id = const { core::intrinsics::type_id::<T>() };
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |                      |
   |                      the parameter type `T` must be valid for the static lifetime...
   |                      ...so that the type `T` will meet its required lifetime bounds
   |
help: consider adding an explicit lifetime bound
   |
47 | pub fn type_id_u128<T: ?Sized + 'static>() -> u128 {
   |                               +++++++++
```

This patch didn't break the build for my non-trivial and not-small cuda-oxide code. I apologize for the merge commit, I was lazy and clicked `sync branch` in GitHub and I should've realized it wouldn't rebase. I can get rid of it if you'd like.


nihalpasham:
Thanks for the report and the patch. The missing 'static bound here is intentional, and re-adding it would regress a supported use case.

The doc comment above `type_id_u128` spells out the constraint: the typed launch path has to keep accepting non-`'static` borrowing closures (e.g. a launcher that captures &[f32] from a stack frame the caller keeps alive across the launch). `core::intrinsics::type_id::<T>()` has a `T: ?Sized` bound and produces the exact same 128-bit value as the backend's `tcx.type_id_hash`, which is why we use the intrinsic rather than stable `TypeId::of`. Adding `T: 'static` would silently tighten that public API and reject those valid kernels.

I believe the actual cause of the E0310 is the toolchain: `cargo +nightly install` ... uses your default nightly, not the one cuda-oxide pins (nightly-2026-04-03 in rust-toolchain.toml). On the pinned nightly the intrinsic accepts non-'static T. Installing with cargo +nightly-2026-04-03 install --git ... cargo-oxide builds without the bound, and #73 is adding that note to the install docs.

Closing for now. If you still hit the lifetime error on the pinned nightly specifically, please reopen with your rustc --version and we'll dig in.

bitemyapp:
Fair enough! I'll try backing out the patch and see if I can work around it in my project.

Thank you very much for releasing this toolchain by the way. I'm an 8-year Rust user that has done some performance-sensitive production work in CUDA C++ and `cuda-oxide` has greatly impressed me. I found it very comfortable to use and I've gotten some awesome results out of it.

PR #111, opened 2026-06-04
mir-lower(tests): add unit tests for control flow op conversion

goog00:
## Summary

Adds end-to-end unit tests for the `dialect-mir` → `dialect-llvm` control-flow (terminator) op conversion.

Each `convert_*` function takes a live `DialectConversionRewriter` owned by pliron's conversion driver and is not constructible standalone, so every test builds a minimal MIR module, runs the full `lower_mir_to_llvm` pass, and asserts the lowered module has the expected `dialect-llvm` shape (or that malformed input is rejected with the expected error).

## Coverage

| MIR op | LLVM op(s) | Assertions |
|--------|-----------|-----------|
| `mir.return` | `llvm.return` | void → 0 operands; scalar → 1 operand; empty-tuple `()` collapses to void (ZST); multiple operands → error |
| `mir.unreachable` | `llvm.unreachable` | op count; source op erased |
| `mir.cond_branch` | `llvm.cond_br` | `[cond / true_args / false_args]` operand split; distinct args routed to each side by value identity; operand-count mismatch → error |
| `mir.assert` | `llvm.cond_br` + abort block | abort block created and terminated by `llvm.unreachable`; cond_br false side targets that abort block; missing successor → error |
| `mir.goto` | `llvm.br` | branch targets dest and forwards args; missing ZST arg padded with `llvm.undef`; missing non-ZST arg → error; too many operands → error |

## Test plan

* `cargo test -p mir-lower --lib convert::ops::control_flow::tests` passes (14 tests)
* `cargo clippy -p mir-lower --lib --tests` clean

### DIFF
--- a/crates/mir-lower/src/convert/ops/control_flow.rs
+++ b/crates/mir-lower/src/convert/ops/control_flow.rs
@@ -253,5 +253,476 @@ pub(crate) fn convert_goto(
 
 #[cfg(test)]
 mod tests {
-    // TODO: Add unit tests for control flow conversion
+    //! End-to-end lowering tests for `dialect-mir` terminator ops.
+    //!
+    //! The `convert_*` functions take a live `DialectConversionRewriter` owned
+    //! by the driver, so we can't call them directly — each test builds a
+    //! minimal MIR module, runs `lower_mir_to_llvm`, and inspects the result.
+
+    use crate::convert::ops::test_util::*;
+    use dialect_mir::ops as mir;
+    use dialect_mir::types::MirTupleType;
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the test coverage @goog00. 

Ported to main after the `pliron-llvm` migration (`dialect_llvm` -> `llvm_export`) and fixed the two goto tests that needed `find_all` + `successor` filtering to account for the prologue block's own `llvm.br`. 

All 14 tests pass, clippy clean. Merging.

PR #113, opened 2026-06-04
Pass alignment to LLVM memory ops

### LINKED ISSUE #112: Emitting vectorized memory operations
Currently, cuda-oxide does not emit memorized memory ops such as `ld.v4.f32` (https://docs.nvidia.com/cuda/parallel-thread-execution/#data-movement-and-conversion-instructions-ld) due to missing alignment information in LLVM IR. 

This is important for maximizing performance (~30% in some of my measured use cases).

paulzhng:
Fixes #112 by concretely passing the alignment to load, store & alloca ops. 

To do this correctly, I had to pull concrete alignments through the pipeline (`#[align(X)]`) - falling back to the natural alignment if no concrete alignment is given.

I added a test which verifies we emit vectorized ops and tested whether it's load bearing before and after.

Please let me know if there's anything stylistically and/or semantically wrong - happy to fix!

### DIFF
--- a/crates/dialect-mir/src/types.rs
+++ b/crates/dialect-mir/src/types.rs
@@ -291,7 +291,7 @@ impl Verify for MirDisjointSliceType {
 /// * Field types must be valid.
 #[pliron_type(
     name = "mir.struct",
-    format = "`<` $name `,` `[` vec($field_names, CharSpace(`,`)) `]` `,` `[` vec($field_types, CharSpace(`,`)) `]` `,` `[` vec($mem_to_decl, CharSpace(`,`)) `]` `,` `[` vec($field_offsets, CharSpace(`,`)) `]` `,` $total_size `>`"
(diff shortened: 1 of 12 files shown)


nihalpasham:
Thanks for the PR! This is great work -- cuda-oxide has so far produced plain vanilla LLVM IR with zero metadata attached, and alignment is likely the first piece we'll be stamping on ops. It's something we're going to be focusing on a lot more going forward as we think through the broader design for perf and debug metadata.

We ported this in #122 with a few additions on top:

- fixed a `convert_ref` gap: the synthesised `alloca+store` that materialises an SSA value back onto the stack (so we can hand out its address) wasn't picking up the pre-pass alignment, which would have been a silent miscompile for over-aligned structs taken by reference
- actually wired up `CUDA_OXIDE_NO_OPT=1`: the env var was mentioned in a comment but the check was never added; it now properly skips the `opt -O2` step
- `OP_ALIGNMENT_KEY` as a `const &str` instead of a per-call allocation, to match how `GLOBAL_ALIGNMENT_KEY` is done next to it
- extended the `natural_alignment` fallback to cover `half`, `ptr`, arrays, vectors, and structs (was only `i*`/`f32`/`f64` before)

Closing this in favour of #122 -- your authorship is preserved there via `Co-authored-by`.

PR #117, opened 2026-06-06
feat(codegen): fma contraction and an opt -O3 pass to match nvcc defaults

alejandro-soto-franco:
## Summary

cuda-oxide left two optimizations that nvcc/NVRTC enable by default off the PTX
path, so its kernels carried more instructions and registers than the equivalent
nvcc build. This PR adds both. On an identical N=256 radix-16 FFT (oxide vs the
NVRTC build of the same source) they move oxide from 1.03 to 1.13x nvcc's
instruction count to 0.94 to 0.996x, and close the gap to cuFFT to 1.02 to 1.05x
at large batch.

## 1. fma contraction

cuda-oxide emitted zero FMA: every multiply-add lowered to a separate
`mul.rn.f32` plus `add.rn.f32` (on the 256 FFT, ~37% more instructions than nvcc
and 56 vs 44 registers/thread), because nvcc sets `--fmad=true` by default and we
did not.

- `mir-lower/.../arithmetic.rs`: `add_fastmath_flags` sets only the `contract`
  fast-math flag (no reassoc/nnan/ninf/nsz) on every `fadd`/`fsub`/`fmul`.
- `mir-importer/.../pipeline.rs`: `generate_ptx` passes `-fp-contract=fast` to
  `llc` on both invocation paths, which forms the fma in the NVPTX backend.

`fma.rn.f32` 0 to 94, instructions 60.2M to 55.2M, registers 56 to 54. The only
IEEE deviation is the single (more accurate) rounding of the fused op. A
downstream FFT suite (35 numpy-tolerance tests) holds, max error 4.05e-6 to
3.30e-6 (nvcc: 3.20e-6). Applies to every fp kernel.

## 2. opt -O3 before llc

The pipeline ran only `mem2reg`, then handed the IR to `llc` (codegen -O2). The
LLVM mid-level optimizer (GVN/CSE, instcombine, reassociate, SROA,
inferaddressspace, strength reduction) never ran, leaving ~25% more dynamic
instructions and extra registers that nvcc/NVRTC (full -O3) drop.

- `mir-importer/.../pipeline.rs`: new `optimize_ll` runs `opt -O3 -S` (tries
  `$CUDA_OXIDE_OPT`, `opt-22`, `opt-21`, `opt`) on the exported `.ll` before
  `llc`, PTX mode only. On any failure it falls back to the unoptimized IR;
  `CUDA_OXIDE_NO_OPT` skips it. The NVVM-IR/libdevice path is untouched.

With both changes the FFT butterfly codegen is nvcc-competitive (numbers above).

**Caveat:** `opt -O3` defaults on, as in clang/nvcc. The arithmetic kernels are
verified; the modern-intrinsic kernels (tcgen05/TMA/WGMMA inline asm) were not
exercised here. `CUDA_OXIDE_NO_OPT` disables the pass without a rebuild if one
miscompiles.

## Rebase

Authored before the `dialect-llvm` -> `llvm-export` (pliron-llvm) migration,
reapplied onto current `main`; the only adjustment was the `FastmathFlags`
import now resolving through `llvm_export::attributes`.

## Verification

Host: sm_120, rustc nightly-2026-04-03, LLVM 22.1.6.

- fmt, clippy `--workspace -D warnings`, cargo doc `-D warnings`: pass
- `cargo test -p mir-lower -p mir-importer`: 44 passed, including two new
  end-to-end lowering tests that lock the contract-only fast-math flag on
  float `mir.mul`/`mir.add`
- `complex_mul.ptx`: `fma.rn.f32` x4 / `mul.rn.f32` x0 / `add.rn.f32` x0
  (baseline `main`: 0 / 4 / 4); `CUDA_OXIDE_NO_OPT=1` drops fma to x2
- Kernel execution not re-run here (NVML driver mismatch on this box); codegen
  is verified at the PTX level, FFT numbers measured earlier on the same host.

Both commits DCO-signed.

### DIFF
--- a/crates/cargo-oxide/README.md
+++ b/crates/cargo-oxide/README.md
@@ -36,16 +36,22 @@ cargo oxide setup                   # explicitly build the codegen backend
 
 ### Flags
 
-| Flag              | Applies to           | Description                              |
-|-------------------|----------------------|------------------------------------------|
-| `--emit-nvvm-ir`  | run, build, pipeline | Generate NVVM IR for libNVVM             |
-| `--arch <sm_XX>`  | run, build, pipeline | Target architecture override             |
-| `--features <F>`  | run, build           | Comma-separated cargo features to enable |
-| `-v, --verbose`   | run, build           | Show detailed compilation output         |
-| `--async`         | new                  | Use the async template                   |
-| `--cgdb`          | debug                | Use cgdb instead of cuda-gdb             |
-| `--tui`           | debug                | Use GDB's TUI interface                  |
-| `--check`         | fmt                  | Check formatting only                    |
+| Flag              | Applies to           | Description                                     |
+|-------------------|----------------------|-------------------------------------------------|
+| `--emit-nvvm-ir`  | run, build, pipeline | Generate NVVM IR for libNVVM                    |
+| `--arch <sm_XX>`  | run, build, pipeline | Target architecture override                    |
+| `--features <F>`  | run, build           | Comma-separated cargo features to enable        |
+| `-v, --verbose`   | run, build           | Show detailed compilation output                |
+| `--no-fmad`       | run, build, pipeline | Disable FMA contraction (keep separate mul+add) |
+| `--async`         | new                  | Use the async template                          |
+| `--cgdb`          | debug                | Use cgdb instead of cuda-gdb                    |
+| `--tui`           | debug                | Use GDB's TUI interface                         |
+| `--check`         | fmt                  | Check formatting only                           |
+
+`--no-fmad` disables FMA contraction for kernels that rely on two separate
+roundings (e.g. Dekker's algorithm, 2Sum). Equivalent to `CUDA_OXIDE_NO_FMA=1`.
+By default, FMA contraction is on (matching nvcc's `--fmad=true`): `fmul+fadd`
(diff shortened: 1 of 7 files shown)


nihalpasham:
Thanks for the PR. 

A couple of points:

**Kept:** the FMA contraction - `FastmathFlagsAttr`(`FastmathFlags::CONTRACT`) on float ops + `-fp-contract=fast` to llc. Also wired up a `--no-fmad` flag on cargo oxide (and `CUDA_OXIDE_NO_FMA=1` for the env-var path) for anyone doing exact-arithmetic work like Kahan summation or Dekker's algorithm where you need the two separate roundings.

**Dropped:** the standalone `optimize_ll` / `opt -O3` path. Main already runs `opt -O2` via LlvmToolchain (PR #153), which pairs opt and llc by LLVM major to avoid the version-mismatch crash. Redoing binary discovery (opt-22 / opt-21 / opt) would re-open that. 

**Also fixed:** the test module had use `llvm_export::op_interfaces::FastMathFlags` (the trait, capital M) but then referenced `FastmathFlags::CONTRACT` (the bitflags struct). Swapped to use `llvm_export::attributes::FastmathFlags`.

Reviewer note: All 47 unit tests + 9 integration tests green. The two new arithmetic tests lock the contract-only flag end-to-end.

and thanks for the perf analysis - the FFT numbers look good.

PR #119, opened 2026-06-06
fix(mir-importer): honour explicit repr integer widths for enums

### LINKED ISSUE #118: MIR importer lowers explicit repr enum discriminants with the wrong width
## Summary

Fieldless enums with an explicit integer representation, for example `#[repr(u32)]`, can be lowered using a discriminant width inferred from the number of variants instead of the explicit repr width.

## Impact

Pointer arithmetic over `*const Enum` can use the wrong stride. A downstream case observed a `#[repr(u32)]` enum behaving like it had an 8-byte stride rather than the expected 4-byte stride.

## Reproduction

A reduced example is prepared on branch `repr-enum-discriminant-layout`:

```bash
cargo oxide run repr_u32_enum_stride
```

Copied below for reference:
```rust
use cuda_core::{CudaContext, CudaStream, DeviceBuffer, LaunchConfig};
use cuda_host::cuda_module;
use std::sync::Arc;

/// Fieldless `#[repr(u32)]` enum used as a tag-like device-buffer element.
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
#[repr(u32)]
pub enum Tag {
    Foo = 0,
    Bar = 1,
    Baz = 2,
    Qux = 3,
}

// SAFETY: trivial repr(u32) POD, safe to copy device-host.
unsafe impl cuda_core::DeviceCopy for Tag {}

const N: usize = 4;

#[cuda_module]
mod kernels {
    use super::*;

    /// Control: read via `*const u32` with `add(i)`. Stride should be 4.
    /// Writes input[i] for i in 0..N.
    #[kernel]
    pub fn read_via_u32(input: &[u32], mut out: DisjointSlice<u32>) {
        let idx = thread::index_1d();
        let i = idx.get();
        if i >= N {
            return;
        }
        // Read through a raw u32 pointer with arithmetic.
        let base: *const u32 = input.as_ptr();
        let v = unsafe { *base.add(i) };
        if let Some(slot) = out.get_mut(idx) {
            *slot = v;
        }
    }

    /// Test: read via `*const Tag` with `add(i)`, then cast the discriminant
    /// back to u32. If stride is correctly 4 the output matches the u32
    /// control. If stride is buggy (8) the output skips slots.
    #[kernel]
    pub fn read_via_enum(input: &[u32], mut out: DisjointSlice<u32>) {
        let idx = thread::index_1d();
        let i = idx.get();
        if i >= N {
            return;
        }
        // Reinterpret the input buffer as `*const Tag`. The bytes are
        // identical (both u32-sized, repr(u32)); only pointer-arithmetic
        // stride is under test.
        let base: *const Tag = input.as_ptr() as *const Tag;
        let tag = unsafe { *base.add(i) };
        if let Some(slot) = out.get_mut(idx) {
            *slot = tag as u32;
        }
    }
}

fn run_and_report<F>(name: &str, stream: &Arc<CudaStream>, launch: F) -> Vec<u32>
where
    F: FnOnce(&Arc<CudaStream>, LaunchConfig, &DeviceBuffer<u32>, &mut DeviceBuffer<u32>),
{
    // Input designed so that any reasonable misstride produces a visible
    // wrong discriminant: 0,1,2,3 all map to distinct Tag variants.
    let input: Vec<u32> = (0..N as u32).collect();
    let dev_in = DeviceBuffer::from_host(stream, &input).unwrap();
    let mut dev_out = DeviceBuffer::<u32>::zeroed(stream, N).unwrap();

    launch(
        stream,
        LaunchConfig::for_num_elems(N as u32),
        &dev_in,
        &mut dev_out,
    );

    let host_out = dev_out.to_host_vec(stream).unwrap();
    let expected: Vec<u32> = (0..N as u32).collect();
    let verdict = if host_out == expected { "PASS" } else { "FAIL" };
    println!("  {name:<14}  {host_out:?}   {verdict}");
    host_out
}

fn main() {
    println!("=== repr(u32) enum pointer-stride miscompile minimal repro ===\n");
    println!("Input host buffer: [0, 1, 2, 3]");
    println!("Expected output (both kernels, if codegen is correct): [0, 1, 2, 3]");
    println!("Buggy expected output (enum_ptr, stride-1): slots 1..=3 read zero bytes\n");

    let ctx = CudaContext::new(0).expect("CUDA init");
    let module = kernels::load(&ctx).expect("load embedded PTX");

    let control = run_and_report("control_u32", &stream, |s, cfg, i, o| {
        module.read_via_u32(s, cfg, i, o).expect("launch")
    });
    let enum_path = run_and_report("enum_ptr", &stream, |s, cfg, i, o| {
        module.read_via_enum(s, cfg, i, o).expect("launch")
    });

    println!();
    if control == enum_path {
        println!("RESULT: control and enum-path agree — bug not reproduced (good!).");
        std::process::exit(0);
    } else {
        println!("RESULT: control and enum-path DISAGREE — bug reproduced.");
        println!("        control_u32: {control:?}");
        println!("        enum_ptr:    {enum_path:?}");
        std::process::exit(1);
    }
}
```

## Proposed fix

Consult `AdtDef::repr().int` when selecting the MIR enum discriminant type, and fall back to the existing variant-count heuristic only when no explicit integer repr is present. (See PR, about to be submitted)

niklebedenko:
Fixes #118.

Previously, cuda-oxide selected the lowered discriminant width from the number of variants. A small `#[repr(u32)]` enum therefore lowered as an `i8` discriminant, which made pointer arithmetic over `*const Enum` use the wrong stride.

The fix: consult `AdtDef::repr().int` when selecting the MIR enum discriminant type, and preserve the old variant-count heuristic only when no explicit integer repr is present.

Repro / regression test: `repr_u32_enum_stride`, a reduced example that compares pointer reads through `[u32]` and `[#[repr(u32)] enum]`.

---

Btw I have a ton of other correctness / missing feature fixes I discovered while porting a large codebase from Rust_CUDA to cuda-oxide (roughly 10 PRs worth I think) -- pls lmk if there's any style changes I should make, would love to contribute!


nihalpasham:
Thanks for the PR. 

Took it a step further on top:

- the tag width/signedness now comes from rustc's computed layout instead of the repr attribute (also covers repr(C), `default-repr` sparse/negative discriminants, and usize reprs),
- `MirEnumType` now carries total size/alignment so strides are layout-exact, with a loud error for multi-payload enums we can't faithfully lay out yet, and
- kept your example as the regression test, plus a few more enum shapes.

PR #121, opened 2026-06-06
fix(mir-importer): preserve runtime Index projections in Rvalue::Ref

### LINKED ISSUE #120: Rvalue::Ref lowering can drop runtime Index projections
`Rvalue::Ref` lowering can lose runtime `Index` projections in places shaped like `&(*ptr).field[i]`. Therefore, the generated reference points at the containing field rather than the indexed element. Kernels using closure-captured or struct-contained arrays can silently read the wrong slot.

Repro / proposed fix: See PR

niklebedenko:
Fixes #120: a MIR importer lowering bug where `Rvalue::Ref` could drop runtime `Index` projections in places shaped like `&(*ptr).field[i]`. The lowered reference pointed at the containing field instead of the indexed element, so kernels using closure-captured or struct-contained indexed data could silently read the wrong value.

The fix: 
- Teach the `Rvalue::Ref` lowering path to preserve runtime `Index` projections through the tail of the projected place.
- Add `closure_index_miscompile`, a reduced multi-kernel example that bisects the affected shape.

The example is larger than ideal because the failure depends on closure/struct projection shape, but it is self-contained and uses only generic pair-wrapper kernels.


nihalpasham:
Thanks for the PR.

Added a few things on top:

- the residual fall-through now errors loudly instead of returning a prefix address,
- routed `Rvalue::AddressOf` through the same helper (it had an identical inline loop with the same bug), and
- extended the example with write-through, `ConstantIndex`, raw-pointer and deref-tail shapes.

PR #124, opened 2026-06-06
Fix float `!=` lowering: `fcmp one` (ordered) → `une` (unordered)

### LINKED ISSUE #123: Float `!=` lowers to `fcmp one` (ordered) instead of `une` — `x != x` NaN check silently folds to `false`
**Title:** Float `!=` lowers to `fcmp one` (ordered) instead of `une` (unordered) — `x != x` NaN check silently folds to `false`

## Summary

`MirNeOp` lowers a Rust float `!=` to the **ordered** LLVM predicate `fcmp one` (ordered-not-equal). Rust's `PartialEq::ne` on floats is **unordered**: `NaN != NaN` is `true`, and the canonical NaN check `x != x` must evaluate to `true` for NaN. Because `one` is *false* when either operand is unordered (NaN), `x != x` lowers to a comparison that is always `false` for NaN, so any NaN-detection / NaN-propagation logic written in idiomatic Rust silently produces wrong results.

This is target-independent (the predicate is chosen in `mir-lower`), so it affects every architecture, not a specific GPU.

## Root cause

`crates/mir-lower/src/convert/interface_impls.rs`, in `impl MirToLlvmConversion for MirNeOp`:

```rust
super::ops::arithmetic::convert_cmp(
    ctx, rewriter, self.get_operation(), operands_info,
    ICmpPredicateAttr::NE,
    ICmpPredicateAttr::NE,
    FCmpPredicateAttr::ONE,   // <-- ordered; should be UNE (unordered) for Rust float !=
)
```

For contrast, `MirEqOp` correctly uses `FCmpPredicateAttr::OEQ` (ordered-equal) — which *is* right, because Rust `==` on floats is ordered (`NaN == NaN` is `false`). The asymmetry is the bug: `==` → `OEQ` (ordered) is correct, but `!=` → `ONE` (ordered) is wrong; `!=` must be `UNE` (unordered) so that `!(a == b)` holds for NaN.

## Minimal reproduction

```rust
#[kernel]
pub fn is_nan(x: &[f32], mut c: DisjointSlice<f32>) {
    let idx = thread::index_1d();
    let i = idx.get();
    if let Some(ce) = c.get_mut(idx) {
        let v = x[i];
        *ce = if v != v { 1.0 } else { 0.0 };  // canonical NaN check
    }
}
```

Host input `[NaN, 1.0, 0.0, -1.0, inf]`:

```
expected: [1.0, 0.0, 0.0, 0.0, 0.0]
actual  : [0.0, 0.0, 0.0, 0.0, 0.0]   <- NaN not detected
```

Generated IR (the `if v != v` comparison):

```llvm
%v22 = fcmp one float %v21, %v21    ; <- ordered; false for NaN
```

## Fix

```diff
-            FCmpPredicateAttr::ONE,
+            FCmpPredicateAttr::UNE,
```

After the fix the IR is `fcmp une float %v21, %v21` and the repro prints `[1.0, 0.0, 0.0, 0.0, 0.0]` (PASS).

## Impact

Any Rust GPU kernel that uses float `!=` where an operand may be NaN gets wrong results — including the idiomatic `x != x` NaN guard, NaN-propagating activations (e.g. a ReLU written to propagate NaN like PyTorch), and anything that relies on `!=` matching IEEE-754 / Rust unordered semantics. Found while differential-testing generated activation kernels (cuda-oxide vs nvcc vs PyTorch); a PR with the fix + a unit test locking `Ne → UNE` + a smoke example follows.

heath-hunnicutt-ruach-tov:
# Fix float `!=` lowering: `fcmp one` (ordered) → `une` (unordered)

## Summary

`MirNeOp` lowered a Rust float `!=` to the **ordered** LLVM predicate `fcmp one`. Rust's `PartialEq::ne` on floats is **unordered** — `NaN != NaN` is `true`, and the canonical NaN check `x != x` must be `true` for NaN. Since `one` (ordered-not-equal) is *false* whenever an operand is NaN, `x != x` lowered to a comparison that is always `false` for NaN, silently breaking NaN detection and NaN-propagation in idiomatic Rust kernels.

This is target-independent (the predicate is selected in `mir-lower`), so it affects every architecture.

## Root cause

`crates/mir-lower/src/convert/interface_impls.rs`, in `impl MirToLlvmConversion for MirNeOp`:

```rust
super::ops::arithmetic::convert_cmp(
    ctx, rewriter, self.get_operation(), operands_info,
    ICmpPredicateAttr::NE,
    ICmpPredicateAttr::NE,
    FCmpPredicateAttr::ONE,   // ordered; wrong for Rust float !=
)
```

`MirEqOp` correctly uses `FCmpPredicateAttr::OEQ` (Rust `==` on floats *is* ordered: `NaN == NaN` is `false`). The bug is the asymmetry: `==` → `OEQ` is correct, but `!=` → `ONE` is not — `!=` must be `UNE` so that `a != b` equals `!(a == b)`, which is `true` for NaN.

## Fix

```diff
-            FCmpPredicateAttr::ONE,
+            FCmpPredicateAttr::UNE,
```

One line in `MirNeOp::convert`.

## Reproduction (before this PR)

```rust
#[kernel]
pub fn is_nan(x: &[f32], mut c: DisjointSlice<f32>) {
    let idx = thread::index_1d();
    let i = idx.get();
    if let Some(ce) = c.get_mut(idx) {
        let v = x[i];
        *ce = if v != v { 1.0 } else { 0.0 };  // canonical NaN check
    }
}
```

Input `[NaN, 1.0, 0.0, -1.0, inf]`:

```
before: [0.0, 0.0, 0.0, 0.0, 0.0]   <- NaN not detected
after : [1.0, 0.0, 0.0, 0.0, 0.0]   <- correct
```

Generated IR for `if v != v`:

```llvm
; before
%v22 = fcmp one float %v21, %v21    ; ordered -> false for NaN
; after
%v22 = fcmp une float %v21, %v21    ; unordered -> true for NaN
```

## Test

New end-to-end smoke example `examples/fcmp_ne_nan_smoke/` computes `(x != x)` over `[NaN, 1, 0, -1, inf]`, asserts `NaN -> 1.0` and finite `-> 0.0`, and prints `SUCCESS` / `FAILURE`. Pre-fix it prints all-zeros (`FAILURE`); post-fix it prints `[1, 0, 0, 0, 0]` (`SUCCESS`).

## Verification

```
cargo oxide run fcmp_ne_nan_smoke --arch sm_61
  input : [NaN, 1.0, 0.0, -1.0, inf]
  x!=x  : [1.0, 0.0, 0.0, 0.0, 0.0]  (expect [1.0, 0.0, 0.0, 0.0, 0.0])
  SUCCESS: x != x is unordered (NaN detected)
```

Generated `.ll` shows `fcmp une float %v21, %v21`.

Device-verified on a Tesla P4 (sm_61, CUDA 12.8, driver 570.x). Found while
differential-testing generated activation kernels (cuda-oxide vs nvcc vs
PyTorch): a NaN-propagating ReLU silently flushed NaN to 0, which traced to
this predicate.

## DCO

Single change + smoke example, signed off.

Fixes #123.


nihalpasham:
Thanks for the PR. LGTM 👍🏾 

Added three things on top: 

- a predicate-lock test asserting the whole `fcmp/icmp` table against rustc's reference so a refactor can't sneak one back in, 
- a fix for the doc comment that claimed all float predicates are ordered (probably where the bug came from), and 
- some example polish, including an `#[allow(clippy::eq_op)]` since clippy flags `v != v`, which is the whole point of the test.

PR #134, opened 2026-06-07
Add scalar f16 atomic support

advpropsys:
## Summary

- Add `DeviceAtomicF16` and `BlockAtomicF16` for scalar half-precision atomic add/subtract.
- Map Rust `f16` atomics to `mir.fp16` in MIR import, and lower float `fetch_sub` as `FAdd(-x)` so the backend keeps the native add-atomic path.
- Add an `atomic_f16` example with correctness coverage, a benchmark binary, and checked-in trace artifacts.

## Notes

This adds correctness support for scalar f16 atomics. It does not claim a throughput win over f32 atomics. The `sm_103` benchmark trace shows scalar f16 global atomics are slower than f32 for the measured contention sweep.

Repeated +1 accumulation

f32 vs f16 atomic add benchmark

## Verification

- `cargo fmt --check`
- `cargo check -p cuda-device -p mir-importer`
- `cargo check --manifest-path crates/rustc-codegen-cuda/examples/atomic_f16/Cargo.toml --bins`
- `cargo oxide run atomic_f16 --arch sm_103`
- `cargo oxide run atomic_f16 --bin bench --arch sm_103`

Commit is DCO signed.


advpropsys:
I have used atomic f16 ops solely for algorithm prototyping, they were very easy to integrate and I think a lot of devs can face the same issue I had (no availability of them in oxide, but existing ptx)

nihalpasham:
The diff and the atomic_f16 suite LGTM 👍🏾

Adding the following on top:
- feat(cuda-device): added `SystemAtomicF16`. The translator already parses scope independently of the base type, so the `.sys` variant only needed the macro invocation plus example coverage.
- chore(examples): dropped the checked-in trace PNGs/CSV (~467 KB of binary history); the README keeps the sm_103 numbers in text.
- docs: removed the hardcoded `--arch sm_103` (arch is auto-detected), realigned the example-table rows, restored unrelated comment wording in the integer-atomic macro, fixed the float `fetch_sub` doc (it lowers as `fadd` of the negated value, not `fsub`), and noted the pre-sm_70 CAS-loop fallback.
- style(examples): repo SPDX header convention in the new sources + cargo fmt (the per-example fmt CI gate was failing).
- test(examples): runtime checks for f16 `swap`/`load`/`store`, which previously had only PTX-level verification.

PR #140, opened 2026-06-08
fix(mir-importer): translate `str`, fn-pointer and fn-item types

haixuanTao:
Three type-translation gaps that appear only in dead panic/formatting branches of real crates but block translation:

- `str` (unsized byte sequence) -> a `[u8]`-style slice type.
- function-pointer types (e.g. the `fmt` fn pointers in panic formatting) -> an opaque thin pointer.
- zero-sized function-item (`FnDef`) types -> an empty struct.


nihalpasham:
Thanks for this. Added 3 follow-ons

- 3e96425: on the pinned nightly, the issue's own repro still failed after import - the `fn to fn(u32) -> u32` coercion (ReifyFnPointer) reached the generic pointer cast as an empty struct and died in lowering. It now lowers to a stable per-function token (hash of the mangled name) cast `int to ptr`, so `f == f` holds and distinct `fns` compare unequal. Indirect calls still fail loudly at build time. Regression example `fn_ptr_naming` covers both directions
- 2d239d6: `&str` went through the pointer-translation path, which had no `str` case, so it became a thin pointer (8 bytes where Rust has 16). It now maps to the same fat-pointer slice type as `&[u8]`. With that, the error fixture's dead `format_args!` binding compiled and flipped the expected-failure fixture, so it now drives the real formatting engine through `core::fmt::write` (trait objects, the unsupported part).
- ebb3eda: merged main and passed the new `abi_align` argument to `get_with_full_layout`, which changed signature after this branch forked.

PR #141, opened 2026-06-08
fix(mir-lower): fall back to live operand type in arithmetic signedness check

haixuanTao:
`is_signed_int_op` errored ("expected IntegerType or MirPtrType operand") when the pre-conversion MIR type was not tracked for an operand — which happens for booleans and values produced inside short-circuit `||`/`&&`, or already-converted operands.

Fall back to the live operand type: a (signless, post-conversion) integer or a pointer is treated as unsigned, instead of failing the whole lowering.


nihalpasham:
This LGTM 👍🏾 

Added three commits layered on top:
- 4f06905: the untracked-operand fallback now consults the live `IntegerType`'s signedness instead of blanket "unsigned". Identical behavior today (the only untracked values are signless `i1` bools), but an untracked Signed operand from a future pass lowers correctly instead of silently going unsigned.
- 81a3160: lowering test locking `icmp eq i1` / `icmp ult i1` for a bool-phi comparison; reverting the fallback reproduces the original "expected IntegerType or MirPtrType operand" failure.
- 3583fce: `bool_phi_cmp` example (short-circuit `||` phi feeding `==` and `<`, all four truth combinations), picked up by examples-compile CI.

One correction folded into the first commit: "already-converted operands" in the comment was not quite right, since `replace_operation` records history for those. The untracked case is values whose type was never convertible, which today is only signless i1 bools.

PR #142, opened 2026-06-08
fix(mir-importer): lower libm float ops to libdevice intrinsics

haixuanTao:
When auto-porting existing rust-gpu / SPIR-V shader crates to cuda-oxide, their [glam](https://github.com/bitshifter/glam-rs) dependency (with `nostd-libm`) lowers float ops to `libm::math::*` (`sqrtf`, `sinf`,`cosf`, `expf`, `powf`, `floor`, ...). Intercept these at the call site and map them to the libdevice intrinsics (hardware math), and tell the collector NOT to translate the `libm` bodies (otherwise `-Zalways-encode-mir` drags in their generic software implementations and trips importer gaps).

Adds `from_libm_path` (final-segment match -> intrinsic) and an `emit_sincos` handler for `libm::sincos{f}` (returns `(sin, cos)`, so no single scalar intrinsic fits — emits a `sinf`/`cosf` pair packed into the destination tuple).


nihalpasham:
Thanks for the PR. LGTM 

Just a couple of additions:
- merged current main into the branch (one add-add conflict in the terminator dispatch; kept both arms)
- anchored the libm match to the crate root in `from_libm_path` / `libm_sincos_is_f64`: `contains("libm")` rerouted any user function whose path mentions libm (IR-proven: a user `expf` in a libm-named module compiled to `__nv_expf`); adversarial unit tests pin it
- mapped `fmaxf/fmax`, `fminf/fmin`, `rintf/rint`, `roundevenf/roundeven` onto the existing enum variants
- unmapped `libm::*` calls now fail at the call site with "libm function `X` is not yet mapped to a libdevice intrinsic" instead of the late "Symbol libm__cbrtf not found" verifier error
- promoted the repro to a `libm_math` feature example (f32/f64 lanes, `sincosf`, host comparison, plus an end-to-end guard that a user `expf` in a libm-named module stays a real call); `libm::sqrt{,f}` stays excluded for now, its inline-asm wrapper is MIR-inlined before interception
- `cbrt` intentionally left out: #148 adds the core-path cbrt, whichever lands second rebases trivially

PR #143, opened 2026-06-08
fix(mir-importer): preserve op location/span in lowering error messages

haixuanTao:
Format the dialect-conversion failure with ctx so the failing op location/span survives into the error, instead of an opaque "invalid input program" with no source pointer.


nihalpasham:
Thanks, LGTM 👍🏾

PR #144, opened 2026-06-08
fix(mir-lower): declare direct __nv_* libdevice externs at the call site

haixuanTao:
Hand-written libdevice externs (e.g. khal_std `Float::asin`/`acos` lowering to `__nv_asinf`/`__nv_acosf`) have no MIR body, so no module declaration is emitted and the verifier rejects the call ("Symbol __nv_asinf not found").

In `call::convert`, when the resolved callee name starts with `__nv_`, declare it with the call signature via the existing `ensure_intrinsic_declared` helper — mirroring the float-math intrinsic path. The symbol binds against libdevice at the cubin-link stage, exactly like `__nv_expf`/`__nv_sqrtf`.


nihalpasham:
Yeah, the guard as written never fired for the case in your commit message. So, I layered three commits on top:

- the importer named no-MIR foreign callees by FQDN (`crate____nv_asinf` after legalization), so `starts_with("__nv_")` never matched a plain `extern "C"` block; foreign items now emit under their link symbol (`Instance::mangled_name`, honours `#[link_name]`)
- pipeline step 5.5 inserted device-extern declarations unconditionally, which made the call-site declare a duplicate ("multiple definitions of Symbol __nv_asinf") for `#[device] extern` libdevice symbols; it now skips symbols already declared in the module
- added an `extern_libdevice` example covering both shapes (plain extern `__nv_asinf` + `#[device] extern __nv_acosf`), GPU-verified against host libm within 2 ULP

With those in place your call-site declare is the live mechanism: reverting it reproduces "Symbol __nv_asinf not found" on the new example.

Issue #146, opened 2026-06-08
kernel does not behave correctly with some uses of enum


mutantbob:
I had some code that was comparing some arrays to derive an Ordering.  The on-GPU version of the code was not behaving correctly.  I eventually distilled it down to this case (although it may be possible to reduce it further).

```
use crate::kernels::LoadedModule;
use cuda_core::{CudaContext, CudaStream, DeviceBuffer, LaunchConfig};
use cuda_device::cuda_module;
use std::sync::Arc;

#[cuda_module]
mod kernels {
    use cuda_device::{DisjointSlice, gpu_printf, kernel, thread};
    use std::cmp::Ordering;

    #[kernel]
    pub fn misbehave(mut dst: DisjointSlice<i32>) {
        let idx = thread::index_1d();

        if let Some(cell) = dst.get_mut(idx) {
            let lhs = [0, 0x169000, 0, 0];
            let rhs = [0, 0xbd1000, 0, 0];

            if true {
                let comparison = { helper(lhs, rhs) };

                match comparison {
                    Ordering::Less => {
                        *cell = -1;
                    }
                    Ordering::Equal => {
                        *cell = 0;
                    }
                    Ordering::Greater => {
                        *cell = 1;
                    }
                }
            } else {
                *cell = helper2(lhs, rhs) as i32;
            }
        }
    }

    fn helper(lhs: [i32; 4], rhs: [i32; 4]) -> Ordering {
        for pos in 0..4 {
            gpu_printf!("{} <=> {}\n", lhs[pos], rhs[pos]);
            if lhs[pos] < rhs[pos] {
                // *cell = -1;
                gpu_printf!("less\n");
                return Ordering::Less;
            }
            if lhs[pos] > rhs[pos] {
                // *cell = 1;
                return Ordering::Greater;
            }
        }
        Ordering::Equal
    }

    fn helper2(lhs: [i32; 4], rhs: [i32; 4]) -> i8 {
        for pos in 0..4 {
            gpu_printf!("{} <=> {}\n", lhs[pos], rhs[pos]);
            if lhs[pos] < rhs[pos] {
                gpu_printf!("LESS\n");
                return -1;
            }
            if lhs[pos] > rhs[pos] {
                return 1;
            }
        }
        0
    }
}

fn main() {
    println!("Hello, world!");
    let (stream, module) = simple_stream_and_module();

    let mut c_dev = DeviceBuffer::<i32>::zeroed(&stream, 1).unwrap();

    module
[...]

PR #147, opened 2026-06-08
fix(mir): lower BinaryOp::Cmp

### LINKED ISSUE #136: Unsupported construct: BinaryOp Cmp not yet implemented
While implementing Ord for a custom type I have encountered

```
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Translation failed: asplode: [name: "Span { id: 8, repr: "/home/thoth/.rustup/toolchains/nightly-2026-04-03-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cmp.rs:2012:21: 2012:72" }", loc: (?)] Compilation error: invalid input program.
       Unsupported construct: BinaryOp Cmp not yet implemented
```

I hope you can easily reproduce it using this kernel:

```
use cuda_device::cuda_module;

#[cuda_module]
mod kernels {
    use std::cmp::Ordering;

    #[kernel]
    pub fn asplode(mut _dst: DisjointSlice<usize>) {
        let _idx = thread::index_1d();

        let x = Foo::<u32>::default().cmp(&Foo::default());
    }

    #[derive(Eq, PartialEq, PartialOrd, Default)]
    struct Foo<T> {
        pieces: [T; 4],
    }

    impl<T> Ord for Foo<T>
    where
        T: Ord,
    {
        fn cmp(&self, other: &Self) -> Ordering {
            self.pieces[0].cmp(&other.pieces[0])
        }
    }
}
```

advpropsys:
## Summary

This adds support for MIR `BinaryOp::Cmp`, which is emitted by `Ord::cmp`, and lowers it to an enum-shaped `core::cmp::Ordering` value.

The change also preserves declared enum discriminants in `MirEnumType`, so `Ordering::Less`, `Ordering::Equal`, and `Ordering::Greater` use Rust's actual discriminant values rather than variant indices.

## Testing

- `cargo fmt --check`
- `cargo check -p dialect-mir -p mir-importer -p mir-lower`
- `cargo test -p dialect-mir`
- `cargo check --manifest-path crates/rustc-codegen-cuda/examples/ord_cmp/Cargo.toml --offline`
- `cargo oxide run ord_cmp --arch sm_103`

The `ord_cmp` example covers `u32`, `i32`, `u64`, `i64`, `usize`, and `isize` comparisons across less, equal, and greater outcomes.

Fixes #136.


nihalpasham:
Thanks for the PR. LGTM 👍🏾

Added a few things on top:

- the tag type now comes from rustc's layout instead of `discriminant_ty()` (which is `isize` for `default-repr` enums),
kept your discriminant-value threading as is; it also fixes #146, verified on hardware (`Ordering::Less` now matches as `Less`),
- hardened the tag paths (loud errors instead of silent index/value fallbacks, constant tags compared at physical tag width),
- moved the divergent-enum guard to the kernel ABI boundary so in-kernel multi-payload enums keep working,

PR #148, opened 2026-06-09
feat: Add cbrt intrinsic support for f32 and f64

goog00:
### Summary

- Registers `cbrtf32` / `cbrtf64` placeholder symbols in `dialect-mir/src/rust_intrinsics.rs`, following the same pattern as existing float math intrinsics
- Maps them to CUDA libdevice builtins `__nv_cbrtf` / `__nv_cbrt` in the MIR-lower call translator
- Hooks the new placeholders into the codegen collector
- Adds a `cbrt_smoke` example: 16 test values × 2 type variants (f32/f64), verifying GPU results are within 2 ULP of host `libm`

### Test plan

- [ ] `cargo test -p dialect-mir -p mir-lower --lib` — unit tests covering intrinsic name → enum dispatch and arg count
- [ ] `cargo oxide run cbrt_smoke` — end-to-end GPU smoke test; expected output: `SUCCESS: 16 cases × 2 variants within 2 ULP of host libm`
- Verified on RTX 5090 (sm_120a, CUDA 13.0)


nihalpasham:
Thanks for the nice addition, mirrors the atan lowering #109 . Main moved underneath the PR, so I layered fixups on top:

- #171 made bare `cuda_launch!` caller-unsafe, which broke this example's host build (E0133). Ported it to the `#[cuda_module]` typed-launch style the other math examples use; kernels, the 16-value x 2-width comparison, and the SUCCESS check are unchanged.
- Refreshed the stale lockfile (cuda-* 0.1.0 -> 0.2.1).
- rustfmt on the new importer test table.
- Module docs now lead with the path the pinned toolchain actually takes: `f{32,64}::cbrt` reaches device codegen via `core::num::imp::libm`, not `std::sys::cmath`.
- Dropped the `core::num::imp::libm::*` arms from the collector's cmath shim guard: that guard only runs in the forbidden-std branch, and those shims are extern declarations with no MIR, so the collector's no-MIR skip already excludes them. Your importer-side core arms are the load-bearing ones and stay.

`cbrt_smoke.ll` declares and calls `__nv_cbrtf` / `__nv_cbrt` on sm_90.

PR #159, opened 2026-06-11
feat: add a CUDA_OXIDE_POST_IR hook for external IR transforms

ianrgraham:
Adds `CUDA_OXIDE_POST_IR`, an opt-in hook that lets external tools rewrite the exported LLVM IR between IR export and PTX generation without forking the compiler — the IR-stage analogue of the existing `CUDA_OXIDE_LLC` override.

# Summary
- `CUDA_OXIDE_POST_IR` names one or more executables (a PATH-style list, `:`-separated on Unix) run in order on the exported `.ll`. Each is invoked as `<hook> <ll_path> <output_dir> <output_name> <target>` (`<target>` is the explicit target arch, possibly empty).
- Hooks are **transform-only** (per review): each may rewrite `<ll_path>` in place, and the next hook — ultimately the pipeline's own codegen — sees the edits. stdout is ignored; there is no artifact protocol.
- Exit 0 continues the build; a non-zero exit aborts it, surfacing the hook's stderr in the compile error (new `PipelineError::PostIr` variant).
- Follows the `BackendOptions` discipline from #314: the env read lives only in `BackendOptions::from_env` at the rustc frontend boundary; the backend receives an explicit `post_ir_hooks` list (not exposed through the experimental standalone API).
- Additive and opt-in: when unset, the pipeline is unchanged.
- Enables out-of-tree IR-stage tooling: custom LLVM passes/plugins, instrumentation, external-bitcode linking, automatic differentiation.

# Tests
- Unit tests in `cuda-oxide-codegen` drive the hook runner directly (it takes the hook list as a parameter, so tests never mutate process-global env): in-place rewrite + argv contract, multi-hook ordering, empty-list no-op, non-zero exit aborting with stderr before later hooks run, and spawn failure.
- `cargo test -p cuda-oxide-codegen -p mir-importer` passes; clippy (`-D warnings`), fmt, and rustdoc (`-D warnings`) clean.

# Examples
- **`post_ir_hook`** — minimal wiring + smoke test. The example's `.cargo/config.toml` sets the hook for the device build; `hook.sh` rewrites a marker constant in the exported IR, so the kernel computes something its Rust source does not say — the GPU result proves the PTX was generated from the edited IR. Runs in the smoketest standard category and the GPU-less examples-compile gate.
- **`enzyme_autodiff`** — the motivating use case: differentiate a `#[device]` function with Enzyme. A same-signature stub is retargeted by the hook to an `__enzyme_fwddiff` request, `opt -passes='enzyme,default<O2>'` generates the derivative, and the host validates it against the analytic one at 1024 points. Without an Enzyme plugin the hook leaves the IR untouched and the example prints `skipping:`; a set-but-broken plugin aborts the build with the hook's stderr.

Both examples verified end-to-end on a TITAN V (sm_70, `--arch` pinned). The hook is also documented in the book's lowering-pipeline page and the README examples table.


ianrgraham:
@nihalpasham mind I promote this with some added tests/examples? Or is a change like this out of scope for you now?

nihalpasham:
Sorry about the delay (just one of those weeks) 

Thanks @ianrgraham, I'm in favor of this direction and it's not out of scope. Yes, please promote it with the tests and example you offered. 

One design change before you put work in: let's make it transform-only and drop the take-over mode.

The `take-over`hook emitting a final `cubin/ptx` that we then route back through our own build ties the hook to our internal artifact handling, and that part does change over time. `Transform-only` avoids all of that, and it still fully covers your `Enzyme` case: we hand you the IR, you rewrite it, we carry on and make PTX from your edited IR.

Dropping take-over also lets us drop the stdout protocol entirely (I believe it only exists so the hook could announce an artifact). The whole contract then collapses to the same thing as `CUDA_OXIDE_LLC`:

```sh
env var names a program → we pass it the .ll path → it edits in place → exit 0 (ok) / non-zero (fail)
```
So, no stdout parsing, nothing to misformat.

Concretely, for the version you promote:

- Keep the Step 7.5 hook, transform-only. (also - you may want to remove `step` wording in the final version of the PR)
- Drop the take-over branch, the artifact-kind parsing, and the stdout result line.
- Keep "non-zero exit aborts the build with the hook's stderr" and the PipelineError::PostIr variant.
- The :-separated list of hooks is fine to keep, it just gets simpler: run each in order, each sees the previous one's edits, no "only the last may take over" rule needed.
- **Tests/example**: a smoke test where a trivial hook edits the IR and the build still produces working PTX, the "unset = pipeline unchanged" check you already have, and the Enzyme example. With take-over gone, the test matrix is just those, no artifact-kind cases to cover.

nihalpasham:
Hey @ianrgraham, just checking. Are you still working on this?

ianrgraham:
Yes! I'll be able to give it some attention and finish over this weekend

nihalpasham:
Hey @ianrgraham - just checking. Are you still working on this? As this PR seems to be behind by several commits

nihalpasham:
Hey, feel free to submit a new PR rather than rebasing this one. It would be easier. Closing this one

ianrgraham:
Hi, sorry been away. Sure thing, easy enough. Are there any further significant changes I should wait for you to land before making the new PR?

Issue #128, opened 2026-06-07
kernel silently reads the wrong struct field on the GPU (reordered/padded repr(Rust) struct)


hbtweb:
I hit this building a struct inside a kernel and reading a field back — the value came back wrong, with no error at all. On other structs the same thing fails loudly instead, in a few different shapes depending on version:

```
# LLVM module verification:
Value being inserted / extracted does not match the type of the indexed aggregate
# or (v0.2.0):
Lowering failed: IntToInt: source is not an integer
# or (v0.1.0):
Compilation error: invalid input program
```

So it's the same underlying bug showing up as either a wrong value (scary) or a confusing build error. It doesn't need generics or a custom trait — the kernel just has to build a `repr(Rust)` struct whose fields rustc reorders/pads.

### Reproduce

The trigger is a `repr(Rust)` struct whose fields rustc reorders, with a multi-word align-4 enum field that forces interior `[N x i8]` padding. To also hit the *silent* identity-fallback path, put the generic struct in a **sibling crate** so its `MirStructType` is reconstructed across the crate boundary at monomorphization.

`arena_core` (a small `no_std` crate):
```rust
#![no_std]

#[derive(Copy, Clone)]
pub enum Layout { Aos, Soa, AoSoA(u32) }   // payload variant -> align-4, multi-word enum

pub trait CellStore {
    fn bump(&self, n: u32) -> u32;
    fn store(&self, idx: u32, val: u32);
}

// declaration order: store, cap, stride, layout. rustc reorders in memory
// (fat-pointer store first, then u32s, then the enum) -> decl index != LLVM
// field index, and the lowered struct gets [N x i8] padding after the enum.
pub struct ScratchArena<S: CellStore> { store: S, cap: u32, stride: u32, layout: Layout }

impl<S: CellStore> ScratchArena<S> {
    pub fn new(store: S, cap: u32, stride: u32, layout: Layout) -> Self {
        Self { store, cap, stride, layout }
    }
    pub fn alloc(&self) -> Option<u32> {
        let s = self.store.bump(1);
        if s >= self.cap { None } else { Some(s) }
    }
    pub fn write(&self, slot: u32, field: u32, val: u32) {
        let i = match self.layout {                 // reads layout, cap, stride, store off &self
            Layout::Soa => field.wrapping_mul(self.cap).wrapping_add(slot),
            _           => slot.wrapping_mul(self.stride).wrapping_add(field),
        };
        self.store.store(i, val);
    }
}

[...]

PR #160, opened 2026-06-11
feat: field-faithful multi-payload enum layout; admit enums at the kernel ABI

nihalpasham:
## What

Multi-payload enums now lower with rustc's exact byte layout, so enum data can cross the host/device boundary in both directions. The "not yet field-faithful" divergence guard is gone; what remains is a narrower check that closes a latent unsoundness (details below).

```text
#[repr(u32)] enum E { A(u32), B(f32), C }      rustc: 8 bytes, payloads overlap at byte 4

Before:  device struct { i32, i32, i32 }  (12 bytes, payloads side by side)
         -> kernels taking E behind a ptr/slice REJECTED at the ABI boundary
After:   device struct { i32, i32 }       (8 bytes, byte-identical to rustc)
         -> &[E] / DisjointSlice<E> kernel params just work
```

## How

Four commits, each independently green:

1. **Metadata** (`dialect-mir`, `mir-importer`): `MirEnumType` records where the tag and every payload field live (byte positions from rustc's layout). The offset lookups move into a shared `translator::layout` module so type import and constant decoding cannot drift.
2. **Representation** (`mir-lower`): `build_enum_slot_map` builds the LLVM struct and the indices into it in one walk (the issue #128 `build_struct_slot_map` pattern). Fields whose bytes collide with a different-typed field of another variant keep their byte position but no struct slot; they are read/written through an alignment-stamped stack spill plus a byte-precise pointer (no `bitcast`). Identical (position, type) payloads share one slot, so common shapes like `Result<i32, u32>` stay pure SSA. All five index consumers (type conversion, construct, discriminant read, payload read, three-way compare) switched atomically. The old guard deliberately stayed active through this commit as a bisect point.
3. **Guard flip**: `enum_memory_divergence` deleted; the ABI walk narrows to `find_unmodeled_enum_in_abi`, which rejects only enums whose layout we deliberately do not record: niche-encoded shapes (`Option<&T>`) and degenerate multi-variant Singles (`Result<T, Infallible>`). **This closes a pre-existing hole**: such enums used to pass the old size-comparison guard while host and device disagreed on every byte (the host stores no tag at all; the device adds a synthetic one). New unit test pins the rejection; `kernel_param_rejects_layout_divergent_enum` flips into `kernel_param_accepts_multi_payload_enum`.
4. **Docs**: all enum-related comments rewritten in plain language with byte diagrams, readable with basic Rust knowledge.

## Evidence

- New example `enum_abi_multi_payload` proves both ABI directions on hardware: the host writes an `E { A(u32) = 7, B(f32) = 9, C = 4096 }` buffer (overlapping payloads, sparse discriminants), a kernel matches and decodes it (B's payload through the memory path), and a second kernel constructs all three variants back through `DisjointSlice<E>` for host verification. Passes on sm_120.
- Prove-it-bites: at the stage-1 commit the same example is rejected with the old "not yet field-faithful" fatal; after stage 3 it builds and passes.
- New unit tests: payload slot sharing; tag NOT at byte 0 (`{A(u64), B(u64)}` with the tag at byte 8 lowers to `{ i64, i8, [7 x i8] }` with both payloads sharing slot 0); niched-enum kernel-param rejection.
- Full GPU smoketest on sm_120: 75/76, the single failure being the documented pre-existing `cluster` Test 2 (cluster sync) baseline failure. All five enum-pinning examples (`enum_array_match`, `repr_u32_enum_stride`, `option_ref_transmute`, `option_ref_unwrap_or`, `ord_cmp`) execute correctly.
- 76/76 compile-only; all dialect-mir / mir-importer / mir-lower tests green; clippy clean.

## Notes

- The user-visible behavior change: niched enums reachable from kernel params are now rejected with an explanatory error suggesting an explicit discriminant repr. Previously they compiled and silently miscompiled.
- Known follow-ups: hoist the lowering-introduced spill allocas to the function entry block (matters only under `CUDA_OXIDE_NO_OPT=1`; `opt -O2` SROA removes them today), and modelling niched layouts themselves so `Option<&T>` can cross the boundary too.
- `SetDiscriminant` stays rejected at import; under this model it becomes a plain tag store, an enabled follow-up.


nihalpasham:
Since this whole thing is about matching device and host enum layouts, where niche-encoded enums (e.g. Option<&T>) store no tag in rustc's layout; the variant is computed from the payload (null means None).

And our construct ops only know the `load-a-tag` / `store-a-tag` shape, which is why we model these enums with an explicit variant-count tag, and `mir-lower` rebuilds the aggregate from `NicheEncodingAttr` (`emit_scalar_to_niched_enum`). 

Question: which enums does rustc apply this optimization to and why? Can we not have the same exceptions in our pipeline?

PR #176, opened 2026-06-12
Preserve device contexts when duplicate init fails

fallintoplace:
## Summary
- keep `init_device_contexts` from clearing the thread-local device map while checking for duplicate initialization
- add a regression test showing duplicate init continues to fail after the first duplicate error

## Why
`ctx.devices.replace(None).is_none()` used the duplicate-init predicate itself to replace the existing map with `None`. That made the first duplicate call return `Context already initialized`, but it also dropped the initialized state, so a later call could initialize again.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p cuda-async duplicate_init_preserves_existing_device_contexts` *(blocked locally: CUDA toolkit headers are not installed; `cuda-bindings` could not find `cuda.h` under `/usr/local/cuda`)*

### DIFF
--- a/crates/cuda-async/src/device_context.rs
+++ b/crates/cuda-async/src/device_context.rs
@@ -105,9 +105,12 @@ pub fn init_device_contexts(
     num_devices: usize,
 ) -> Result<(), DeviceError> {
     DEVICE_CONTEXTS.with(|ctx| {
+        let devices = ctx.devices.take();
+        let is_uninitialized = devices.is_none();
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM 👍🏾

- The old check wiped the live device map even on the error path; the fix restores it.

Reviewer note, verified: the added duplicate-init test passes and genuinely fails on the old code.

PR #181, opened 2026-06-12
fix(pipeline): create output_dir before emission

### LINKED ISSUE #180: Create pipeline output directories before emission
## Problem

`run_pipeline` writes generated artifacts under `PipelineConfig::output_dir`, but assumes the directory already exists. Callers that use a fresh or nested sidecar output directory fail before the first artifact is emitted with a filesystem `No such file or directory` error.

## Why this matters

The pipeline owns artifact emission. Requiring every caller to pre-create the output directory duplicates responsibility and makes otherwise valid output paths fail mechanically.

## Expected behavior

A caller should be able to pass any valid fresh `PipelineConfig::output_dir`, including a nested path. The pipeline should create it once before MIR, LLVM IR, NVVM IR, or PTX emission writes under that path.

## Acceptance criteria

- `run_pipeline` creates `PipelineConfig::output_dir` before artifact emission.
- Existing directories continue to work as a no-op.
- Directory creation errors include the failing path.
- A focused regression covers a missing nested output directory.

niklebedenko:
Closes #180.

## Problem

`run_pipeline` writes generated artifacts under `PipelineConfig::output_dir`, but previously assumed that directory already existed. Callers that choose a fresh sidecar output directory hit `Export failed: No such file or directory (os error 2)` before the first `.ll` emission.

## Fix

Create `PipelineConfig::output_dir` at the start of `run_pipeline`, before any MIR translation, LLVM export, or PTX/NVVM artifact emission work. Directory creation uses `std::fs::create_dir_all`, so existing directories remain a no-op and nested fresh paths are created in one place.

## Tests

Added a focused `mir-importer` regression that runs the pipeline with an empty function set, NVVM IR emission, and a nested missing output directory. The test avoids invoking `llc`, verifies the fresh directory is created before export, and asserts the `.ll` artifact is written under it.

## Validation

`cargo test -p mir-importer run_pipeline_creates_missing_output_dir_before_export`

### DIFF
--- a/crates/mir-importer/src/pipeline.rs
+++ b/crates/mir-importer/src/pipeline.rs
@@ -224,6 +224,8 @@ pub fn run_pipeline(
     device_externs: &[DeviceExternDecl],
     config: &PipelineConfig,
 ) -> Result<CompilationResult, PipelineError> {
+    prepare_output_dir(&config.output_dir)?;
+
     let mut ctx = Context::new();
 
     // Step 1: Register dialects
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM 👍🏾

- Layered a commit: added `debug_kind` to the new test so it compiles on current main; rebased.
- `cargo test -p mir-importer` green (34 pass).

PR #183, opened 2026-06-12
fix(mir-importer): lower array constants

### LINKED ISSUE #182: Lower bare and nested array constants in MIR import
## Problem

The MIR importer can lower pointer-to-array constants such as `&[T; N]`, but bare array value constants such as `const TABLE: [T; N]` can still fail when the whole array materializes as an operand before runtime indexing. Nested arrays such as `[[T; M]; N]` are also not handled by the flat array construction path.

## Why this matters

Static lookup tables are common in device code. The importer should handle the array constant shape generally rather than only the pointer-to-array variant.

## Expected behavior

Bare array constants and pointer-to-array constants should share the same array value construction logic. Nested arrays should be recursively decomposed when their leaves are supported primitive scalar constants.

## Acceptance criteria

- Bare `[T; N]` constants lower to `MirConstructArrayOp` values.
- Pointer-to-array constants keep working by wrapping the shared array value in `MirRefOp`.
- Nested arrays of primitive scalar leaves lower recursively.
- Existing byte-backed constants with valid padding do not regress.
- A focused example covers bare, nested, and pointer-to-array constants.

niklebedenko:
Closes #182.

## Problem

The MIR importer could lower pointer-to-array constants such as `&[T; N]`, but bare array value constants such as `const TABLE: [T; N]` could still fall through as unsupported when indexed by runtime values. Nested constants such as `[[T; M]; N]` also lacked recursive array construction, so only flat primitive arrays were effectively covered.

## Fix

Array constant lowering now has one shared value-construction path. Pointer-to-array constants decode the array value through that helper and then wrap it in `MirRefOp`; bare array constants return the constructed `MirConstructArrayOp` value directly. The helper recursively handles nested arrays whose leaves are primitive scalar elements.

Unsupported element types are rejected explicitly. The supported byte-lowered array leaves are integer, f16, f32, and f64 values, plus nested arrays of those leaves. Byte-backed constants with uninitialized bytes are now diagnosed instead of silently zero-filled.

## Diligence

Checked the sibling pointer-to-array path and routed it through the same array-value helper so bare arrays and `&[T; N]` cannot drift. Kept the existing scalar constant and aggregate paths separate, but hardened the shared `constant_bytes` helper used by byte-backed constants so unsupported uninitialized bytes are not hidden.

Struct/ADT array elements remain unsupported in this PR because they need layout-aware recursive field decoding rather than the primitive byte-slicing rule used here.

## Tests

Added `crates/rustc-codegen-cuda/examples/array_constants` covering:

- bare `[f32; 4]` constants indexed by runtime thread id,
- nested `[[u32; 3]; 2]` constants with runtime row/column selection,
- pointer-to-array `&[u32; 4]` constants to preserve existing behavior.

## Validation

Passed local targeted validation:

```bash
cargo fmt -p mir-importer -- --check
cargo fmt --manifest-path crates/rustc-codegen-cuda/examples/array_constants/Cargo.toml -- --check
scripts/smoketest.sh --compile-only --no-color --only '^array_constants$'
```


nihalpasham:
LGTM.

- layered: added the `supported-features` row
- heads up: the description says uninitialized bytes are now diagnosed - they aren't (the shared `constant_bytes` helper - still zero-fills, same as the tuple/enum paths). This I guess is harmless for the primitive leaves you accept (no padding), but dropping the claim makes more sense. 
- GPU `sm_120`: array_constants → PASS (24 threads); reverting only the importer arm reproduces Unsupported constant type

PR #185, opened 2026-06-12
fix(codegen): align export and call names

### LINKED ISSUE #184: Align definition export names with resolved call target names
## Problem

Definition-side export names and call-side names can diverge because they choose between raw FQDNs and rustc mangled names using different facts. Concrete trait impl and concrete inherent impl calls are especially sensitive: the raw `FnDef` operand can differ from the resolved monomorphic instance that the collector exports.

## Why this matters

A device call that names one symbol while the collector exports another produces unresolved calls or duplicate symbol policy bugs. Naming must be a single rule shared by collection and MIR call lowering.

## Expected behavior

Both paths should use the resolved rustc instance as the authority. Generic resolved instances should use rustc's mangled symbol name. Non-generic resolved instances should keep the raw resolved FQDN and rely on the shared legalizer to produce the final LLVM/PTX identifier.

## Acceptance criteria

- Call-side naming resolves `FnDef` operands before deciding FQDN versus mangled name.
- Definition-side naming uses the same resolved-instance rule.
- Invalid FQDN characters alone do not force mangling.
- Foreign items keep using their external link symbols.
- A focused example covers concrete trait impl and concrete inherent impl calls.

niklebedenko:
Closes #184.

## Problem

Definition-side export names and call-side names could diverge because they used different facts to decide between a raw FQDN and rustc's mangled symbol name. The collector sees a resolved rustc `Instance`, while MIR call lowering classified generic calls from raw `FnDef` substitutions and used the raw `FnDef` FQDN for non-generic calls. Concrete trait impl calls can carry operand substitutions even when they resolve to a monomorphic instance, and their trait item FQDN can differ from the resolved impl FQDN. The collector also treated invalid FQDN characters as a reason to mangle, while call lowering sends non-generic FQDNs through pliron's legalizer.

## Fix

Use resolved rustc instance data as the single rule:

- resolved instances with generic args use rustc's mangled symbol name;
- non-generic resolved instances keep the resolved `Instance::name()` FQDN and let the shared pliron legalizer produce the final identifier;
- raw-name collisions still fall back to rustc's mangled symbol name;
- foreign items keep using their link symbol because no local definition is exported.

## Diligence

Checked both duplicate naming paths:

- `rustc-codegen-cuda::DeviceCollector::compute_export_name` now documents that invalid FQDN characters are not a mangle trigger and that resolved `Instance::args` is the authority.
- `mir-importer::translator::terminator::extract_func_info` now resolves the stable MIR `FnDef` before choosing FQDN vs `Instance::mangled_name`, and uses `Instance::name()` for non-generic call symbols so concrete trait impl calls target the impl definition rather than the trait item.
- `body.rs::translate_body` already legalizes definition export names with the same `Legaliser` instance used by call lowering, so raw FQDNs match after legalization.

## Tests

Added `crates/rustc-codegen-cuda/examples/export_name_policy`, a focused regression example covering:

- concrete trait impl method naming: `<Vec2 as DeviceMeasure>::trait_impl_value`;
- sibling concrete inherent impl FQDN legalization: `Wrapper<Vec2>::inherent_concrete_value`.

## Validation

Targeted validation:

```bash
scripts/smoketest.sh --compile-only --no-color --only '^export_name_policy$'
```


nihalpasham:
Thanks for the PR. The names line up now: def side and call side both key off the resolved Instance through the same legaliser. 

- rebased onto main (was 44 behind); merges clean, nothing else needed
- GPU sm_120: export_name_policy → SUCCESS (64 lanes)
- merge first; unblocks #191 's terminator/mod.rs rebase

PR #187, opened 2026-06-12
fix(intrinsics): lower fast float ops

### LINKED ISSUE #186: Lower Rust fast floating-point intrinsics with explicit fast flags
## Problem

`core::intrinsics::{fadd_fast,fsub_fast,fmul_fast,fdiv_fast,frem_fast}` are generic compiler intrinsics with no MIR body. If they survive as placeholder calls, LLVM lowering can emit unresolved intrinsic symbols. A previous lowering attempt also used the default fast-math attribute, which is empty and does not express the Rust fast-intrinsic contract.

## Why this matters

Fast floating-point intrinsics should become LLVM floating-point binops with explicit fast-math flags. Leaving placeholders or empty flags makes the lowering either invalid or semantically weaker than the intrinsic contract.

## Expected behavior

The importer should recognize all five fast float intrinsics and emit known placeholders. MIR lowering should route those placeholders before the libdevice path and replace them with the matching LLVM float binop carrying explicit `fast` flags.

## Acceptance criteria

- `fadd_fast`, `fsub_fast`, `fmul_fast`, `fdiv_fast`, and `frem_fast` are recognized.
- Both `core::intrinsics` and `std::intrinsics` spellings are covered.
- Lowered LLVM ops use explicit `FastmathFlags::FAST`.
- No unresolved placeholder call survives MIR lowering.
- Tests cover all five operations for both `f32` and `f64`.

niklebedenko:
Closes #186.

Related: #62, #142, #148, #161.

## Problem

`core::intrinsics::{fadd_fast,fsub_fast,fmul_fast,fdiv_fast,frem_fast}` are generic compiler intrinsics with no MIR body. When the importer leaves them as ordinary calls, unresolved cuda-oxide placeholder callees can survive into LLVM lowering and fail verification or symbol resolution.

The initial lowering used `FastmathFlagsAttr::default()` while claiming to emit fast-math binops. In pliron-llvm, `FastmathFlagsAttr::default()` is `FastmathFlags::empty()`, so that did not express Rust's finite/non-NaN `f*_fast` contract.

## Fix

The importer recognizes the `core::intrinsics` and `std::intrinsics` spellings for all five `f*_fast` operations and emits the same centralized Rust-intrinsic placeholder calls as the other target-specific Rust intrinsics.

`mir-lower` now routes those placeholders before the libdevice path and replaces them with the matching LLVM float binop: `llvm.fadd`, `llvm.fsub`, `llvm.fmul`, `llvm.fdiv`, or `llvm.frem`. The fast-intrinsic path uses `FastmathFlags::FAST.into()` explicitly instead of the empty default.

## Diligence

Sibling float intrinsic paths checked:

- fixed-width `core::intrinsics` math such as `sqrtf32`, `powif64`, `fmaf32`, `round_ties_even_f64`, `maximum_number_nsz_f32`, and `cbrtf32`;
- `std::intrinsics` aliases for the same fixed-width math intrinsics;
- `libm` path matching, including re-exported names such as `libm::sqrtf` and nested names such as `libm::math::sqrt::sqrtf`;
- `std::sys::cmath` names for `atan`, `atan2`, and `cbrt`;
- ordinary MIR float arithmetic lowering, which still intentionally uses the existing empty/default fast-math attribute behavior and is not conflated with Rust's explicit `f*_fast` intrinsics;
- `fcmp` lowering, whose regression coverage depends on empty fast-math flags so NaN checks do not become poison.

The generic `f*_fast` operations are the only sibling path with no type suffix and no libdevice entry point, so they are routed to LLVM binops rather than the libdevice declaration/call path.

## Tests

Added a focused `mir-lower` regression that constructs placeholder calls for all five `f*_fast` intrinsics for both `f32` and `f64`. The test asserts:

- `FastmathFlagsAttr::default()` is distinct from explicit `FastmathFlags::FAST`;
- each placeholder lowers to the expected LLVM float binop;
- both `f32` and `f64` monomorphizations are covered for every operation;
- every lowered binop carries explicit LLVM `fast` flags;
- no unresolved `mir.call` or placeholder-named LLVM call survives.

## Validation

Targeted local validation:

```bash
cargo test -p mir-lower test_fast_float_intrinsics_lower_to_explicit_fast_binops
```

### DIFF
--- a/crates/dialect-mir/src/rust_intrinsics.rs
+++ b/crates/dialect-mir/src/rust_intrinsics.rs
@@ -152,3 +152,18 @@ pub const CALLEE_ATAN_F64: &str = placeholder!("atanf64");
 pub const CALLEE_CBRT_F32: &str = placeholder!("cbrtf32");
 /// Placeholder call used for `f64::cbrt` / `std::sys::cmath::cbrt`.
 pub const CALLEE_CBRT_F64: &str = placeholder!("cbrtf64");
+
+/// Placeholder call used for `core::intrinsics::fadd_fast` (generic over float type).
+///
+/// Lowered to `llvm.fadd` with explicit `fast` fast-math flags. The `f*_fast` intrinsics
+/// assume finite, non-NaN inputs; LLVM's fast-math flags express the same
+/// preconditions, so the binop replaces the call directly.
+pub const CALLEE_FADD_FAST: &str = placeholder!("fadd_fast");
+/// Placeholder call used for `core::intrinsics::fsub_fast` (generic over float type).
+pub const CALLEE_FSUB_FAST: &str = placeholder!("fsub_fast");
+/// Placeholder call used for `core::intrinsics::fmul_fast` (generic over float type).
+pub const CALLEE_FMUL_FAST: &str = placeholder!("fmul_fast");
+/// Placeholder call used for `core::intrinsics::fdiv_fast` (generic over float type).
+pub const CALLEE_FDIV_FAST: &str = placeholder!("fdiv_fast");
+/// Placeholder call used for `core::intrinsics::frem_fast` (generic over float type).
+pub const CALLEE_FREM_FAST: &str = placeholder!("frem_fast");
--- a/crates/llvm-export/src/export/ops.rs
+++ b/crates/llvm-export/src/export/ops.rs
@@ -25,9 +25,10 @@ use pliron::{
 
 use crate::{
     attributes::{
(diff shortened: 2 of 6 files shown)


nihalpasham:
Lowers the `f*_fast` compile gap correctly and matches rustc's flag set. 

One gap: the textual exporter was dropping the fast-math flags, so they never reached the PTX (the optimization was inert).
- Layered: taught `export_binop` to emit fast-math flags (no-op for ordinary float math) + an export regression test.
- `.ll` now shows `f.. fast float`/`double` for all five ops;

PR #189, opened 2026-06-12
fix(pipeline): preserve inline always

### LINKED ISSUE #188: Preserve Rust #[inline(always)] on device helper functions
## Problem

Rust `#[inline(always)]` intent is currently dropped before LLVM IR export for device helper functions. The emitted LLVM helper definitions therefore lack `alwaysinline`, so the existing optimizer path cannot observe the source-level inline hint.

## Why this matters

Device helper boundaries can affect generated PTX shape and optimizer behavior. cuda-oxide should preserve rustc's inline intent through the MIR importer, MIR-to-LLVM lowering, and LLVM exporter instead of silently discarding it.

## Expected behavior

Non-kernel device functions marked `#[inline(always)]` should carry an LLVM `alwaysinline` function attribute by the time LLVM IR is exported. Kernel entry points should not be marked `alwaysinline`.

## Acceptance criteria

- rustc `CodegenFnAttrs::inline` is read for collected device functions.
- The inline-always flag is threaded through `CollectedFunction`.
- MIR import attaches an `alwaysinline` attribute to non-kernel functions only.
- MIR-to-LLVM lowering preserves that attribute.
- LLVM export emits valid function-attribute syntax while preserving the existing convergent attribute group.
- Tests cover attribute propagation and export syntax.

niklebedenko:
Closes #188.

## Problem

cuda-oxide drops Rust `#[inline(always)]` before LLVM IR export, so non-kernel
device helper definitions are emitted without LLVM `alwaysinline`. That can
leave helper `.func` boundaries for the existing optimizer/PTX path and
diverges from Rust's codegen intent.

This branch claims attribute preservation only. It does not add a separate
mandatory always-inline pass; the existing matched LLVM middle-end can honor the
attribute when it runs before `llc`.

## Fix

- Read rustc `CodegenFnAttrs::inline` in `rustc-codegen-cuda`.
- Thread an `is_inline_always` flag through `CollectedFunction`.
- Attach an `alwaysinline` attribute during MIR import for non-kernel device
  functions marked `#[inline(always)]`.
- Propagate that attribute during MIR-to-LLVM lowering.
- Emit textual LLVM as `define ... ) alwaysinline #0 {`, preserving the
  existing convergent attribute group.

## Diligence

Checked the collection, MIR import, MIR-to-LLVM lowering, and LLVM export path
for the inline attribute. Kernel entry points are deliberately excluded because
they are PTX `.entry` exports rather than callees.

The pipeline already resolves a matched `opt`/`llc` pair and runs `opt -O2`
when available. This PR does not reintroduce a separate optimizer lookup or a
mandatory inliner pass, avoiding a second toolchain-selection path.

This PR draft is local-only upstreaming metadata and should stay out of the
upstream-facing commit.

## Tests

- Added `export_alwaysinline_function_attribute_uses_llvm_define_syntax` in
  `crates/llvm-export/tests/export_test.rs`.
- The test checks that exported LLVM contains `alwaysinline`.
- The test checks LLVM function-attribute placement by asserting the exact
  `define void @inline_helper() alwaysinline #0 {` line and verifying the
  convergent `#0` attribute group is still emitted.

No actual-inlining regression is included because the PR does not claim to run
a mandatory always-inline pass or guarantee helper removal after optimization.

## Validation

```bash
cargo test -p llvm-export export_alwaysinline_function_attribute_uses_llvm_define_syntax
```

### DIFF
--- a/crates/llvm-export/src/export/function.rs
+++ b/crates/llvm-export/src/export/function.rs
@@ -231,7 +231,9 @@ impl<'a> ModuleExportState<'a> {
         // Emitted as a function attribute keyword between the parameter
         // list and the body open brace.
         let alwaysinline_key: pliron::identifier::Identifier = "alwaysinline".try_into().unwrap();
-        let is_alwaysinline = attrs.0.contains_key(&alwaysinline_key);
+        let is_alwaysinline = attrs
+            .get::<pliron::builtin::attributes::StringAttr>(&alwaysinline_key)
+            .is_some();
 
         if let Some(entry_block) = entry_block_opt {
             let func_loc = func.get_operation().deref(self.ctx).loc();
--- a/crates/llvm-export/src/export/function.rs
+++ b/crates/llvm-export/src/export/function.rs
@@ -227,6 +227,12 @@ impl<'a> ModuleExportState<'a> {
             .iter(self.ctx)
             .next();
 
+        // Check for alwaysinline attribute (from #[inline(always)]).
+        // Emitted as a function attribute keyword between the parameter
+        // list and the body open brace.
+        let alwaysinline_key: pliron::identifier::Identifier = "alwaysinline".try_into().unwrap();
+        let is_alwaysinline = attrs.0.contains_key(&alwaysinline_key);
+
         if let Some(entry_block) = entry_block_opt {
             let func_loc = func.get_operation().deref(self.ctx).loc();
             let debug_scope = self.debug_subprogram_for_function(&fixed_func_name, &func_loc);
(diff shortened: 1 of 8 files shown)


nihalpasham:
LGTM

- the branch predated the `debug-info` merge, which split the define line into  `#0 { vs ) #0 !dbg !N`; your patch only touched the no-debug line, so a debug build would have dropped `alwaysinline`. Rebased into a single 4-way emission so both coexist
- layered: switched the three production sites from the raw `attributes.0` map to the wrapper `.set()` / `.get::<StringAttr>()` API (matches `gpu_kernel`); added a debug-on regression test

Reviewer note: verified, debug-off + debug-on unit tests, and a real `#[inline(always)]` helper exports as `define i64 @... alwaysinline #0`

PR #193, opened 2026-06-12
feat(cuda-core): derive DeviceCopy

### LINKED ISSUE #192: Add `#[derive(DeviceCopy)]` for CUDA transfer-safe data types
## Problem

`cuda_core::DeviceCopy` is the safety marker used for values that may be copied to and from device memory. Today users must write `unsafe impl DeviceCopy` by hand for their own plain-data structs, even when every field already implements `DeviceCopy`.

That is both noisy and easy to get wrong: a manual impl can accidentally skip the field-by-field safety check that the marker trait relies on.

## Why this matters

`DeviceCopy` is an unsafe marker trait, so the ergonomic path should make the safe case easy while still enforcing the same rule: a composite type can be `DeviceCopy` only if every stored field is also `DeviceCopy`. A derive macro can encode that rule mechanically and reduce hand-written unsafe impls.

For a comparable example, see [Rust CUDA's `#[derive(DeviceCopy)]` implementation](https://github.com/Rust-GPU/rust-cuda/blob/main/crates/cust_derive/src/lib.rs).

## Expected behavior

Users should be able to write `#[derive(DeviceCopy)]` for transfer-safe structs, enums, and unions. The derive should be available through `use cuda_core::DeviceCopy;`, matching the usual trait-plus-derive ergonomics pattern, and invalid field types should fail at compile time.

## Acceptance criteria

- `cuda-macros` provides a `DeviceCopy` derive macro.
- `cuda-core` re-exports the derive next to the `DeviceCopy` trait.
- The generated impl requires all field types to implement `DeviceCopy`.
- Generic type parameters receive the required `DeviceCopy` bounds.
- `#[derive(DeviceCopy)]` works for named structs, tuple structs, enums, unions, and generic structs whose fields are `DeviceCopy`.
- The derive is available through `use cuda_core::DeviceCopy;`, matching the common trait-plus-derive ergonomics pattern.
- Compile-pass coverage exercises representative struct, enum, union, and generic forms.
- Compile-fail coverage proves that a host reference field is rejected because it is not `DeviceCopy`.

niklebedenko:
Closes #192.

## Problem

`cuda_core::DeviceCopy` marks types that are safe to copy between host and device memory, but users currently have to write `unsafe impl DeviceCopy` manually for application data structs. That is an ergonomics gap and an API safety footgun: a manual impl can omit the field-level check that every contained type is also device-copy-safe.

## Fix

Add a `#[derive(DeviceCopy)]` proc macro in `cuda-macros` and re-export it from `cuda-core` next to the `DeviceCopy` trait. The macro generates the `unsafe impl ::cuda_core::DeviceCopy` and a hidden type-checking helper that requires each field type to implement `DeviceCopy`. Generic type parameters also receive a `DeviceCopy` bound on the generated impl.

## Diligence

This keeps the rule at the marker-trait API boundary: deriving `DeviceCopy` is only valid when every stored field is itself `DeviceCopy`. The derive handles structs, enums, and unions through the same field-check helper, so named and unnamed fields share one enforcement path. The public re-export follows the established trait-plus-derive pattern, so `use cuda_core::DeviceCopy;` brings both the trait and derive macro into scope.

## Tests

Added `cuda-core` trybuild coverage for both accepted and rejected derives:

- compile-pass coverage for named structs with scalar and array fields;
- compile-pass coverage for tuple structs containing another derived `DeviceCopy` type;
- compile-pass coverage for generic structs with the generated bound;
- compile-pass coverage for enum and union forms whose fields are `DeviceCopy`;
- compile-fail coverage proving a copyable host reference field is rejected because the field type is not `DeviceCopy`.

## Validation

- `cargo test -p cuda-core --all-targets`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -338,8 +338,10 @@ version = "0.2.1"
 dependencies = [
  "anyhow",
  "cuda-bindings",
+ "cuda-macros",
  "half",
  "oxide-artifacts",
+ "trybuild",
 ]
 
 [[package]]
--- a/crates/cuda-core/Cargo.toml
+++ b/crates/cuda-core/Cargo.toml
@@ -9,9 +9,11 @@ repository.workspace = true
 
 [dependencies]
 cuda-bindings = { path = "../cuda-bindings" }
+cuda-macros = { path = "../cuda-macros" }
 anyhow = { workspace = true }
 half = { workspace = true }
 oxide-artifacts = { workspace = true, features = ["object-read"] }
 
 [dev-dependencies]
 oxide-artifacts = { workspace = true, features = ["object-write"] }
+trybuild = "1"
--- a/crates/cuda-core/src/lib.rs
+++ b/crates/cuda-core/src/lib.rs
@@ -66,6 +66,9 @@ pub mod vmm;
 pub use context::CudaContext;
 /// Raw CUDA driver bindings re-exported for direct access when needed.
 pub use cuda_bindings as sys;
+/// `#[derive(DeviceCopy)]` macro, re-exported next to the `DeviceCopy` trait so
+/// `use cuda_core::DeviceCopy;` brings both into scope (serde trait+derive pattern).
+pub use cuda_macros::DeviceCopy;
 pub use device_buffer::{DeviceBuffer, DeviceCopy};
 pub use embedded::{EmbeddedModule, EmbeddedModuleError};
 pub use error::{DriverError, IntoResult};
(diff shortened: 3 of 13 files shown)


nihalpasham:
Struct/union derive is sound. The enum arm wasn't: `DeviceCopy` requires the all-zero pattern (`DeviceBuffer::zeroed`) to be valid, which an enum discriminant can't guarantee (`#[repr(u8)] enum E { A=1, B=2 }` → UB).
- Layered: reject enums in the derive with a diagnostic (kept struct/union), resolved the #215 conflict, made the hidden helper name collision-safe.
- trybuild green: struct/union pass; enum + reference-field fail.

PR #195, opened 2026-06-12
feat(llvm-export): AsmKind enum for inline asm optimization hints

honeyspoon:
## Summary

Replace the `convergent: bool` parameter on inline asm with an `AsmKind` enum that captures the two orthogonal GPU axes — **convergent** (thread-set dependent) × **side effects** (observable state changes):

```rust
pub enum AsmKind {
    Convergent,      // bar.sync, mma.sync, wgmma, tcgen05, cp.async
    ConvergentPure,  // shfl.sync, vote.sync, match.sync
    SideEffect,      // st.global via asm, hardware timer reads
    Pure,            // cvt.rn.f16x2.f32, cvt.rn.bf16x2.f32, prmt
}
```

Maps to LLVM IR:

| Variant | `sideeffect` | `convergent` | LLVM can… |
|---|---|---|---|
| `Convergent` | yes | yes | nothing (fully pinned) |
| `ConvergentPure` | no | yes | CSE/DCE within uniform flow, not across divergent branches |
| `SideEffect` | yes | no | not move/delete, but no thread-set constraint |
| `Pure` | no | no | CSE, DCE, LICM freely |

**Suggested labels:** `refactor`, `llvm-export`, `IR-lowering`, `codegen`

## Motivation

Inline asm ops currently only model `convergent: bool`. The `sideeffect` flag is hardcoded — every inline asm call blocks LLVM from optimizing it, even for pure data conversions like `cvt.rn.bf16x2.f32` that have no observable side effects.

Rust's own `asm!` macro (RFC 2873) separates side effects (`pure`) from memory behavior (`nomem`/`readonly`) as orthogonal axes. This PR applies the same principle to GPU inline asm, where the relevant axes are **convergence** and **side effects**.

This is complementary to #215 (`ptx_asm!` macro): `AsmKind` classifies *compiler-generated* NVVM intrinsic asm, while `ptx_asm!` handles *user-written* inline PTX. The export layer detects which path produced each `InlineAsmOp` and reads the appropriate attributes — `AsmKind` tag for NVVM ops, `sideeffect`/`convergent` bools for `ptx_asm!` ops.

## Why four variants is the complete set

The remaining LLVM inline asm axes do not apply to PTX:

- **Memory effects** (`nomem`/`readonly`): unnecessary — cuda-oxide's inline asm is either a pure register-to-register conversion or a full side-effecting op. Loads and stores go through proper LLVM ops, not inline asm.
- **`noreturn`/`may_unwind`**: PTX inline asm always returns and never unwinds.
- **`preserves_flags`/`nostack`**: CPU concepts (flags register, stack red zone) with no PTX equivalent.

## Implementation

Follows the existing `InlineAsmOpExt` trait pattern in `llvm-export`:

- **`llvm-export/src/lib.rs`**: `AsmKind` enum + `InlineAsmOpExt::build()` constructor + `asm_kind()`/`asm_kind_opt()` queries. Stores the kind as a `StringAttr` on the op (pliron's `InlineAsmOp` only models `convergent` natively).
- **`llvm-export/src/export/ops.rs`**: `asm_kind_opt()` checks for the `AsmKind` tag; if absent (i.e., a `ptx_asm!` op), falls back to reading the individual `sideeffect`/`convergent` attributes. Exhaustive matching ensures new variants are handled.
- **`mir-lower/intrinsics/`**: All call sites updated — `new_convergent(...)` → `build(..., AsmKind::Convergent)`, raw `InlineAsmOp::new(..., false)` → `build(..., AsmKind::Pure)`.

## Testing

```
cargo test -p llvm-export --all-targets  # 13 passed (4 new: AsmKind round-trip)
cargo test -p mir-lower --all-targets    # 47 passed
cargo build --workspace                  # pass
cargo fmt --check                        # pass
cargo clippy -p llvm-export -D warnings  # pass
cargo clippy -p mir-lower -D warnings    # pass
RUSTDOCFLAGS='-D warnings' cargo doc -p llvm-export -p mir-lower --no-deps  # pass
```


nihalpasham:
Thanks for this. LGTM 

I quite like the `AsmKind` abstraction that gives LLVM what it needs to CSE/DCE/LICM pure register conversions. 

**layered two small fixes:** 
- removed a duplicate docstring on `asm_kind_opt`, and migrated `convert_fma_bf16x2` (which landed after your branch was cut) to use the new API.
- Lifted `AsmKind` from a plain Rust enum serialised as `StringAttr` to a proper `#[pliron_attr]` type. I'd always recommend this as a best practice that typed attributes should be first-class IR citizens, not strings. 

Reviewer note: verified, `cvt.rn.bf16x2.f32` now emits as pure asm in the `.ll`, and all convergent ops remain correctly tagged.

PR #191, opened 2026-06-12
feat(cuda-device): add bf16x2 FMA intrinsic

niklebedenko:
Closes #190.

## Problem

cuda-oxide did not have an end-to-end representation for the PTX `fma.rn.bf16x2` instruction even though it is the packed bf16 arithmetic operation available on `sm_80+`. That left users unable to express packed bf16x2 FMA through the `cuda-device` API and meant the compiler pipeline had no focused regression covering importer recognition and MIR lowering for this instruction.

The general compiler/runtime rule is: a public CUDA device intrinsic for a concrete PTX instruction should be represented consistently through the device API, dialect op, MIR importer dispatch, MIR lowering, and a repo-native example that proves the path remains wired together.

## Fix

Add `cuda_device::bf16x2::fma_bf16x2(a, b, c) -> u32`, where each `u32` stores two bf16 lanes. The importer recognizes that intrinsic path and emits a new `nvvm.fma_bf16x2` op, and MIR lowering converts that op to inline PTX:

```ptx
fma.rn.bf16x2 $0, $1, $2, $3;
```

The new `bf16x2_fma` rustc-codegen-cuda example exercises the path from device API through codegen and checks the packed result on execution-capable systems.

## Diligence

I checked the sibling packed intrinsic path used by `tcgen05::cvt_f32x2_bf16x2`: public device declaration, dialect op registration, importer terminator dispatch, conversion interface registration, and inline-PTX emission. The new intrinsic follows that same abstraction boundary rather than adding an example-only workaround.

This PR intentionally does not add packed bf16x2 add/mul intrinsics. PTX `fma.rn.bf16x2` is supported on `sm_80+`, while `add.bf16x2` and `mul.bf16x2` require `sm_90+`; adding all three under one Ampere-facing module would imply a common architecture contract that PTX does not provide. Ampere callers can express packed add as FMA with a packed `1.0` multiplier.

All new Rust source files use the repository's NVIDIA copyright and SPDX header.

## Tests

Added `crates/rustc-codegen-cuda/examples/bf16x2_fma`, including `Cargo.lock` and a short README. The example converts `(2.0, 4.0)`, `(3.0, 5.0)`, and `(7.0, 11.0)` into packed bf16x2 operands, runs the FMA intrinsic in a kernel, and checks for packed `(13.0, 31.0)`.

## Validation

```bash
cargo fmt --all -- --check
cargo test -p dialect-nvvm --all-targets
cargo test -p mir-importer --all-targets
[...]

PR #196, opened 2026-06-12
fix(mir-importer): lower copy_nonoverlapping to memcpy

### LINKED ISSUE #194: Support device-side `copy_nonoverlapping`
## Problem

Device code that uses `core::ptr::copy_nonoverlapping` reaches MIR as `StatementKind::Intrinsic(NonDivergingIntrinsic::CopyNonOverlapping(_))`. The importer currently rejects that observable statement instead of lowering it.

## Why this matters

`copy_nonoverlapping` is Rust's standard primitive for non-overlapping raw-pointer copies. Rejecting it blocks otherwise valid device code and forces callers to rewrite low-level copy logic by hand.

## Expected behavior

The importer should preserve the copy operation explicitly, and lowering should emit an LLVM memcpy with the correct byte count for the copied pointee type.

## Acceptance criteria

- A regression using `core::ptr::copy_nonoverlapping` fails on `upstream/main` without the fix.
- The MIR operation verifies pointer/count types instead of silently accepting malformed copies.
- The LLVM lowering scales MIR element counts to memcpy byte counts.

niklebedenko:
Closes #194.

## Problem

Device code that calls `core::ptr::copy_nonoverlapping` reaches MIR as `StatementKind::Intrinsic(NonDivergingIntrinsic::CopyNonOverlapping(_))`. cuda-oxide currently rejects that observable statement instead of preserving and lowering the copy.

Related context: PR #27 made this construct a hard error because silently dropping the statement would miscompile code; this PR implements the missing lowering instead.

## Fix

Add a dedicated `mir.memcpy` operation for `CopyNonOverlapping` statements and lower it to `llvm.memcpy.p0.p0.i64`.

`mir.memcpy` keeps MIR's element-count semantics at the importer boundary. During LLVM lowering, the destination pointee type is converted and sized so the count is scaled to the byte count expected by LLVM memcpy.

The old `error_copy_nonoverlapping_unhandled` support-gap example is renamed to `copy_nonoverlapping`, converted into a positive runtime regression, and removed from the smoketest expected-failure list and `STATUS.md`.

## Diligence

The fix is routed through an explicit MIR op instead of special-casing a direct LLVM call in statement translation. That keeps verification close to the IR boundary: destination and source operands must be MIR pointers with matching pointee types, and count must be an integer.

The importer still rejects other unsupported observable statements, such as `SetDiscriminant`, rather than weakening the catch-all behavior. Store/load alignment propagation in the current lowering path is preserved.

## Tests

Updated `crates/rustc-codegen-cuda/examples/copy_nonoverlapping` to launch a kernel that copies 128 `u32` values from one device buffer to another with `core::ptr::copy_nonoverlapping`, then checks the copied host output exactly.

The regression fails on `upstream/main` without the implementation and passes on this branch.

## Validation

```bash
cargo fmt --all -- --check
cargo test -p dialect-mir -p mir-importer -p mir-lower --all-targets
cargo clippy -p dialect-mir -p mir-importer -p mir-lower --all-targets -- -D warnings
cargo oxide run copy_nonoverlapping
```


nihalpasham:
LGTM. 

I suspected the hardcoded `p0.p0` name would break a shared-memory copy; it didn't as raw pointers are `addrspacecast` to generic before the memcpy, so the operands are always p0 which I confirmed by building a shared-dest copy. 

- layered as hardening anyway: derive the overload suffix from the operand types (matches `ctpop`/`fptosi.sat`). No-op today, correct if a non-generic operand ever appears. Added `count > 1`, `u8`, and `shared-dest` cases to cover the byte-scaling branches. 


Reviewer note: GPU `sm_120`: PASS (4 copy shapes); shares interface_impls.rs with #191

Issue #58, opened 2026-05-13
Mutable array element borrow in kernel


DanilSerd:
``` rust
  #[kernel]
    pub fn life(mut out: DisjointSlice<[f32; SIZE]>) {
        let idx = thread::index_1d();
        let a = match out.get_mut(idx) {
            Some(a) => a,
            None => return,
        };
        for i in 0..SIZE {
            a.get_mut(i).map(|e| *e = 42.);
        }

        // This works as expected.
        //
        // for e in a {
        //     *e = 42.;
        // }
    }
```

Write through `get_mut()` doesn't end up being persisted in out. Out remains all zeroes.  Compiles fine.

Doctor:
```
Rust nightly toolchain... ✓ rustc 1.96.0-nightly (55e86c996 2026-04-02)
rust-toolchain.toml... ✓ present
Codegen backend... ✓ ~/.cargo/cuda-oxide/librustc_codegen_cuda.so
CUDA toolkit (nvcc)... ✓ Cuda compilation tools, release 13.1, V13.1.115
libNVVM (libnvvm.so)... ✓ libNVVM 2.0
nvJitLink (libnvJitLink.so)... ✓ nvJitLink 13.1
libdevice (libdevice.10.bc)... ✓ /usr/local/cuda/nvvm/libdevice/libdevice.10.bc
llc (LLVM)... ✓ Ubuntu LLVM version 21.1.8 (llc-21)
clang / libclang resource dir... ✓ /usr/lib/llvm-21/lib/clang/21
cuda-gdb (optional)... ✓ exec: /usr/local/cuda/bin/cuda-gdb-python3.12-tui --version

✅ Environment looks good!
```

mutantbob:
What do you see in the `out` that comes out of the kernel?  

When I try to use `get_mut(i)` I get a compile error since the API changed to take a `ThreadIndex` a little while ago.  If you do end up upgrading your cuda-oxide install, beware of https://github.com/NVlabs/cuda-oxide/issues/49 .

I don't know enough about the SIMT model to know if it is safe to access a destination cell that is not assigned to your thread.  The APi I most recently checked out does not have a `safe` route to do it.  I modified my app to include 
```
        for i in 0..5000 {
            unsafe {
                *dst.as_mut_ptr().add(i) = 77;
            }
        }
```
and it behaved as I expected.

[...]

PR #199, opened 2026-06-12
fix(mir-importer): lower nested index assignments

### LINKED ISSUE #197: Lower nested runtime-index assignments
## Problem

Assignments such as `local[i][j] = value` are represented as a two-level `Index, Index` projection. The statement translator rejects that shape in the two-projection special-case match before the generic projection-address walker can lower it.

## Why this matters

Nested fixed-size arrays are a natural way to express small device-local matrices and tiles. Runtime indexing should work consistently for reads and writes when the address walker can compute the destination address.

## Expected behavior

Two-level runtime-index assignment should be routed through the shared place-address walker and stored through the computed destination pointer.

## Acceptance criteria

- A regression assigning through `local[i][j]` fails on `upstream/main` without the fix.
- The fix reuses the existing address-walk abstraction rather than duplicating projection lowering logic.

niklebedenko:
Closes #197.

## Problem

Assignments such as `local[i][j] = value` are represented as a two-level `Index, Index` projection. The statement translator rejects that shape in the two-projection special-case match before the shared projection-address walker can lower it.

Related context: this is in the same projection-lowering family as Issue #58 (`Mutable array element borrow in kernel`) and PR #165 (`compute element addresses through fat-pointer derefs`). Those cover related borrow/address materialization paths; this PR fixes the assignment path for nested runtime indexes.

## Fix

Add a focused statement-lowering arm for two-level runtime `Index, Index` assignment projections.

The new arm routes the place through `translate_place_address`, then emits a `mir.store` through the computed destination pointer. That reuses the existing shared projection walker instead of duplicating address-lowering logic inside statement translation.

## Diligence

The branch fixes the previously rejected shape at the narrowest statement boundary while delegating correctness to the same address abstraction used by other projection materialization paths. It does not add a parallel projection walker.

The change is limited to the exact `Index, Index` assignment shape; other unsupported two-projection combinations continue to return the existing explicit unsupported diagnostic. The full compile-only smoketest also covers sibling examples such as `ref_index_projections`, `addressof_sharedarray`, `slice_get_mut`, and `array_index`.

## Tests

Added `crates/rustc-codegen-cuda/examples/nested_index_assignment`, which writes through two runtime indexes into a local `[[u32; 4]; 4]`, reads the same nested element back on device, and checks the exact host result.

The regression fails on `upstream/main` without the implementation and passes on this branch.

## Validation

```bash
cargo fmt --all -- --check
cargo oxide run nested_index_assignment
```


nihalpasham:
LGTM

- layered: replaced the hand-inlined arm with a delegation to `store_through_place_address` (the (`Deref`, `Index`) arm and the `N-projection` fallback already use it)
- added a non-square `[[u32; 3]; 5]` case so a row-stride bug can't hide behind a square array
- GPU `sm_120`: PASS for square + non-square

Reviewer note: shares statement.rs with #196 ; land this first, then #196 rebases its hunk

PR #202, opened 2026-06-12
cuda-core: add safe DeviceCopy parity impls

### LINKED ISSUE #201: Add safe standard DeviceCopy parity impls
## Problem

`cuda_core::DeviceCopy` is missing several standard-library impls that are safe under the trait's byte-copy and zero-initialization contract. In particular, `bool`, `char`, and transparent wrapper/container types such as `PhantomData<T>`, `MaybeUninit<T>`, and `Wrapping<T>` cannot currently be used in otherwise transfer-safe device buffer payloads.

## Why this matters

`DeviceCopy` is the core marker for values that may cross the host/device boundary. Missing impls force users either to avoid ordinary Rust field types or to add local unsafe wrapper impls even when the layout rule is already known and can be encoded once in `cuda-core`.

The all-zero validity requirement still matters: types such as bare `NonZero*`, blanket `Option<T>`, and blanket `Result<T, E>` should not be accepted unless their zero bit pattern is known to be valid.

## Expected behavior

`DeviceCopy` should cover standard types whose copy and zero-initialization semantics are valid and layout-stable enough for the trait:

- `bool` and `char`;
- `PhantomData<T>`;
- `MaybeUninit<T>` when `T: DeviceCopy`;
- `Wrapping<T>` when `T: DeviceCopy`.

## Acceptance criteria

- A compile regression using these standard types fails on `upstream/main` without the impls.
- The same regression passes on the patched branch.
- Device and pinned-host zeroed-buffer tests prove the newly accepted scalar values work with the existing all-zero initialization API.
- The patch does not add unsafe blanket impls for `NonZero*`, `Option<T>`, or `Result<T, E>`.

niklebedenko:
Closes #201.

## Problem

`cuda_core::DeviceCopy` did not cover several standard types whose byte-copy and zero-initialization behavior is safe for device transfer. That left users unable to put `bool`, `char`, `PhantomData<T>`, `MaybeUninit<T>`, or `Wrapping<T>` in otherwise transfer-safe payloads without writing local unsafe impls.

Related context: #192 and #193 added `#[derive(DeviceCopy)]`. This PR fills in conservative standard-library parity impls so derived application types can use these fields without downstream unsafe boilerplate.

## Fix

Add `DeviceCopy` impls at the marker-trait boundary for the safe subset:

- `bool` and `char`, whose all-zero representations are valid values;
- `PhantomData<T>`, which stores no bytes;
- `MaybeUninit<T>` when `T: DeviceCopy`, because `MaybeUninit` accepts any initializedness for its payload storage;
- `Wrapping<T>` when `T: DeviceCopy`, because it is a transparent arithmetic wrapper over `T`.

The documentation examples are adjusted so they no longer describe `bool` and `char` as rejected types. `NonZeroU32` remains the example of a type whose all-zero representation is invalid.

## Diligence

This branch deliberately keeps the trait rule narrow. I did not add bare `NonZero*` impls because `DeviceBuffer::zeroed` and `PinnedHostBuffer::zeroed` make the all-zero representation part of the public safety contract. I also did not add blanket `Option<T>` or `Result<T, E>` impls because their zero validity depends on variant layout and niche choices, not simply on `T: DeviceCopy`.

The tests cover both the type-level marker impls and the runtime zero-initialization path for the newly accepted scalar values.

## Tests

Added `crates/cuda-core/tests/device_copy_impls.rs`, which checks that the safe standard-library parity types implement `DeviceCopy`.

Extended `DeviceBuffer` and `PinnedHostBuffer` tests to zero-initialize `bool` and `char` buffers and read back `[false, false, false]` and `['\0', '\0']`.

The regression fails on `upstream/main` without the impls and passes on this branch.

## Validation

```bash
cargo fmt --all -- --check
cargo test -p cuda-core device_copy_covers_core_parity_types
cargo test -p cuda-core zeroed_supports_bool_and_char
cargo test -p cuda-core --all-targets
cargo clippy -p cuda-core --all-targets -- -D warnings
```

### DIFF
--- a/crates/cuda-core/src/device_buffer.rs
+++ b/crates/cuda-core/src/device_buffer.rs
@@ -21,6 +21,8 @@
 //! ```
 
 use std::marker::PhantomData;
+use std::mem::MaybeUninit;
+use std::num::Wrapping;
 use std::sync::Arc;
 
 use cuda_bindings::CUdeviceptr;
@@ -45,10 +47,10 @@ use crate::stream::CudaStream;
 /// `Self`, and the all-zero bit pattern must also be valid because
 /// [`DeviceBuffer::zeroed`] initializes memory with zero bytes.
 ///
-/// `Copy` alone is not enough: types such as `bool`, `char`, and
-/// `NonZeroU32` are `Copy`, but not every byte pattern is a valid value of
-/// those types. `DeviceCopy` is the stronger promise required when
-/// `DeviceBuffer` turns raw device bytes back into initialized Rust values.
+/// `Copy` alone is not enough: types such as `NonZeroU32` are `Copy`, but
+/// their all-zero bit pattern is invalid. `DeviceCopy` is the stronger promise
+/// required when `DeviceBuffer` turns raw device bytes back into initialized
+/// Rust values.
 pub unsafe trait DeviceCopy: Copy {}
 
 macro_rules! impl_device_copy {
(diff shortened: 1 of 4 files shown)


nihalpasham:
Split this one:

- kept `PhantomData`, `MaybeUninit<T>`, `Wrapping<T>` - those are genuinely sound (any/zero bit pattern is valid)
- for bool and char though, a kernel can write a non-0/1 byte (or a bad code point), and reading it back through `to_host_vec` is UB. The `DeviceCopy` docs call those two out by name, and your own derive(`DeviceCopy`) macro from #193 already says they aren't `DeviceCopy`, so letting them in here contradicts that.

Restored the safety doc + the type-level test. If we do want bool/char later, it's worth a deliberate contract change rather than a doc tweak. Everything else looks great.

PR #204, opened 2026-06-12
cuda-core: restore DeviceBuffer async methods

### LINKED ISSUE #203: Restore async DeviceBuffer compatibility methods
## Problem

`DeviceBuffer<T>` exposes stream-ordered allocation, copy, memset, and free behavior internally, but several compatibility methods that name those operations explicitly are missing from the public API. Code expecting `uninitialized_async`, `copy_from_host_async`, `copy_from_device_async`, `zero_async`, or `drop_async` cannot build even though the lower-level operations already exist.

## Why this matters

The rest of the `cuda-core` buffer API makes stream ordering explicit. Named async compatibility methods let callers express stream-ordered ownership and copy lifetimes directly without dropping down to raw pointers or duplicating allocation/copy wrappers outside the crate.

## Expected behavior

`DeviceBuffer<T>` should provide stream-ordered compatibility methods for uninitialized allocation, host-to-device copy, device-to-device copy, byte zeroing, and explicit async free. Zero-length buffers should remain valid no-ops for these methods.

## Acceptance criteria

- A compile regression using the missing async compatibility methods fails on `upstream/main`.
- Length mismatches still panic before copying.
- Zero-length buffers avoid CUDA allocation/copy/free calls and can be dropped through the async path.
- The implementation reuses the existing `cuda-core` memory helpers rather than adding a parallel driver binding path.

niklebedenko:
Closes #203.

## Problem

`DeviceBuffer<T>` already carries an explicit stream-oriented API, but it lacked several named async compatibility methods: `uninitialized_async`, `copy_from_host_async`, `copy_from_device_async`, `zero_async`, and `drop_async`.

Callers that need to express stream-ordered allocation, copies, memset, or explicit free had to either use less direct API names or drop to raw allocation/copy helpers.

## Fix

Add the missing `DeviceBuffer<T>` compatibility methods on the existing `T: DeviceCopy` impl block:

- `unsafe fn uninitialized_async(stream, len)`;
- `copy_from_host_async`;
- `copy_from_device_async`;
- `zero_async`;
- `drop_async`.

The methods delegate to the existing `cuda-core::memory` async helpers and keep the current buffer ownership model. Zero-byte buffers return the same null-pointer/zero-length representation used by the rest of `DeviceBuffer` and treat zero-byte copy, memset, and free as no-ops.

## Diligence

The branch keeps async behavior in `DeviceBuffer` instead of exposing new raw driver calls. Length checks mirror the existing synchronous copy helpers, so mismatched host/device or device/device copies fail before any CUDA call is issued.

The explicit `drop_async` consumes `self` through `into_raw_parts`, preventing the ordinary `Drop` path from freeing the same allocation a second time.

## Tests

Extended `crates/cuda-core/tests/device_buffer.rs` with `device_buffer_async_compat_methods_roundtrip`.

The test allocates uninitialized device buffers, copies host data into one buffer, copies device-to-device into a second buffer, reads exact host results back, zeroes the clone, checks the zeroed result, and then frees both buffers through `drop_async`. It also covers zero-length async allocation/free.

The regression fails on `upstream/main` without the methods and passes on this branch.

## Validation

```bash
cargo fmt --all -- --check
cargo test -p cuda-core device_buffer_async_compat_methods_roundtrip
cargo test -p cuda-core --all-targets
cargo clippy -p cuda-core --all-targets -- -D warnings
```

### DIFF
--- a/crates/cuda-core/src/device_buffer.rs
+++ b/crates/cuda-core/src/device_buffer.rs
@@ -476,4 +476,110 @@ impl<T: DeviceCopy> DeviceBuffer<T> {
             crate::memory::memcpy_htod_async(self.ptr, src.as_ptr(), num_bytes, stream.cu_stream())
         }
     }
+
+    /// Allocates `len` elements of uninitialized device memory, enqueued on
+    /// `stream`.
+    ///
+    /// Unlike [`Self::zeroed`], no `cuMemsetD8` is enqueued. The contents of
+    /// the returned buffer are undefined until the caller writes them.
+    ///
+    /// # Safety
+    ///
+    /// Reading from the returned buffer before any kernel or memcpy has
+    /// written it is undefined behavior.
+    pub unsafe fn uninitialized_async(
+        stream: &CudaStream,
+        len: usize,
+    ) -> Result<Self, DriverError> {
+        let ctx = stream.context().clone();
+        let num_bytes = len * std::mem::size_of::<T>();
+        if num_bytes == 0 {
+            // SAFETY: a null pointer with zero bytes is never dereferenced
+            // and Drop/drop_async ignore it.
+            return Ok(unsafe { Self::from_raw_parts(0, len, ctx) });
+        }
+
+        let ptr = unsafe { crate::memory::malloc_async(stream.cu_stream(), num_bytes)? };
+        Ok(Self {
+            ptr,
+            len,
+            ctx,
+            _marker: PhantomData,
+        })
+    }
+
+    /// Copies `other` into `self` device-to-device, enqueued on `stream`.
+    ///
+    /// Panics if `other.len() != self.len()`.
+    pub fn copy_from_device_async(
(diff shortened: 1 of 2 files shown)


nihalpasham:
The methods themselves are good, but the implicit `Drop` wasn't correct and compute-sanitizer caught it:

- `uninitialized_async` allocates from the async pool (`cuMemAllocAsync`), but normal drop frees with the sync `cuMemFree` - sanitizer flags free-before-alloc when a copy is still in flight
- your roundtrip test never hits it because `to_host_vec` synchronizes first

Layered a fix:
- the buffer now co-owns its stream and frees with `cuMemFreeAsync` on drop, so dropping it without an explicit `drop_async` is safe too
- added a drop-without-sync stress test (clean under compute-sanitizer, 0 errors)
- also fixed a small unrelated context refcount leak in `into_raw_parts`

Good to merge now.

PR #206, opened 2026-06-12
mir-importer: support pointer distance intrinsics

### LINKED ISSUE #205: Support raw pointer distance intrinsics
## Problem

Raw pointer distance methods such as `offset_from`, `offset_from_unsigned`, `byte_offset_from`, and `byte_offset_from_unsigned` lower through rustc pointer-distance intrinsics in MIR. cuda-oxide does not currently lower those intrinsics, so device code that computes pointer distances cannot build.

## Why this matters

Pointer distance is a standard Rust operation for low-level slice, table, and iterator-style device code. The compiler should preserve the distinction between signed and unsigned element distances and between element-based and byte-based wrappers instead of forcing users to avoid the standard methods.

## Expected behavior

The MIR importer should lower `ptr_offset_from` and `ptr_offset_from_unsigned` by exposing pointer addresses, subtracting them, and dividing by the rustc-reported pointee size. Signed intrinsics should return `isize`; unsigned intrinsics should return `usize`.

Unsupported or invalid cases, such as unknown pointer operands or zero-sized pointees, should produce explicit importer errors instead of silently defaulting to a size.

## Acceptance criteria

- A focused example using the stable pointer-distance methods fails on `upstream/main` without the lowering.
- The same example builds and passes on the patched branch.
- Both signed and unsigned distance intrinsics are covered.
- Byte-distance wrappers are covered through their rustc intrinsic lowering.
- The implementation obtains pointee size from rustc layout information and rejects zero-sized pointees explicitly.

niklebedenko:
Closes #205.

## Problem

Rust raw pointer distance methods lower to compiler intrinsics such as `ptr_offset_from` and `ptr_offset_from_unsigned`. cuda-oxide did not handle those intrinsics, so device code using `offset_from`, `offset_from_unsigned`, or their byte-oriented wrappers failed during MIR import.

The compiler rule is that pointer distance is an address difference divided by the pointee size, with signed and unsigned result types preserved.

## Fix

Add shared lowering for `core/std::intrinsics::{ptr_offset_from, ptr_offset_from_unsigned}`.

The lowering:

- exposes both pointer operands as integer addresses;
- subtracts `this - other`;
- divides by the rustc-reported pointee size;
- returns `isize` for the signed intrinsic and `usize` for the unsigned intrinsic.

The importer now errors explicitly when the pointer operand type cannot be identified, when layout lookup fails, or when the pointee is zero-sized.

## Diligence

The signed and unsigned intrinsic paths share one helper so the address-difference rule cannot drift. The helper uses rustc layout data for pointee size rather than deriving size from surface syntax or falling back to `1`.

The regression covers stable Rust pointer methods instead of calling compiler intrinsics directly, so it exercises the user-visible API shape. It includes both element-distance and byte-distance wrappers.

## Tests

Added `crates/rustc-codegen-cuda/examples/ptr_offset_from` with a standalone `Cargo.toml` and `Cargo.lock`.

The example computes four distances on device:

- positive `offset_from_unsigned`;
- negative signed `offset_from`;
- signed `byte_offset_from`;
- unsigned `byte_offset_from_unsigned`.

The host asserts exact results for a `u32` slice with `lo = 3` and `hi = 9`: `[6, -6, 24, 24]`.

The regression fails on `upstream/main` without the lowering and passes on this branch.

## Validation

```bash
cargo fmt --all -- --check
cargo check -p mir-importer
(cd crates/rustc-codegen-cuda/examples/ptr_offset_from && cargo clippy --all-targets -- -D warnings)
scripts/smoketest.sh --compile-only --only '^ptr_offset_from$' --no-color
cargo oxide run ptr_offset_from
```


nihalpasham:
Lowering looks right and it checks out on GPU ([6, -6, 24, 24], the negative slot exercises the signed path). Layered one thing on top:

- added a `supported-features.md` row for the pointer-distance intrinsics

Same types.rs insertion point as #212, so one of the two will want a quick rebase at merge. Otherwise good to go.

PR #208, opened 2026-06-12
mir-importer: preserve volatile load and store intrinsics

### LINKED ISSUE #207: Preserve volatile load and store intrinsics
## Problem

Device code that uses `core::ptr::{read_volatile, write_volatile}` lowers through rustc's volatile load/store intrinsics. cuda-oxide does not currently preserve those volatile memory semantics end to end, so volatile accesses can be rejected or represented as ordinary loads and stores.

## Why this matters

Volatile memory access is Rust's standard way to express side-effecting memory operations that must not be removed, combined, or promoted as ordinary memory traffic. Device code needs this for low-level memory-mapped, synchronization-adjacent, and externally observed access patterns.

## Expected behavior

The MIR importer should recognize volatile load/store intrinsics and create MIR load/store operations marked as volatile. MIR promotion should treat volatile accesses as non-promotable. MIR-to-LLVM lowering and LLVM text export should preserve the volatile marker and emit `load volatile` / `store volatile`.

## Acceptance criteria

- A focused example using `core::ptr::{read_volatile, write_volatile}` fails on `upstream/main` without volatile support.
- The same example builds and passes on the patched branch.
- MIR load/store operations carry explicit volatile metadata.
- Volatile load/store operations are not eligible for mem2reg promotion.
- LLVM lowering and text export preserve the volatile marker.

niklebedenko:
Closes #207.

## Problem

`core::ptr::read_volatile` and `core::ptr::write_volatile` lower through rustc volatile memory intrinsics. cuda-oxide did not carry a volatile bit through MIR import, mem2reg eligibility, MIR-to-LLVM lowering, and LLVM text export, so volatile accesses were either unsupported or could become ordinary memory operations.

The compiler rule is that volatile load/store semantics must be represented explicitly and preserved through every memory-lowering boundary.

## Fix

Add explicit volatile attributes to `MirLoadOp` and `MirStoreOp`, with helper methods to read and set them.

The importer recognizes `core/std::intrinsics::{volatile_load, volatile_store}` and emits volatile MIR load/store operations. Volatile MIR accesses return `NonPromotableUse` from the mem2reg promotion interface. MIR-to-LLVM lowering stamps a volatile marker onto the LLVM load/store op, and the LLVM exporter prints `load volatile` / `store volatile`.

## Diligence

I checked each sibling boundary that can otherwise drop the rule:

- MIR dialect representation;
- mem2reg promotion classification;
- intrinsic dispatch in the MIR importer;
- MIR-to-LLVM conversion;
- textual LLVM export.

This branch intentionally stays scoped to volatile memory intrinsics. It does not include pointer-distance intrinsic lowering or unrelated formatting churn from the reference branch.

The current `pliron-llvm` load/store wrappers in this checkout do not expose native volatile fields, so the volatile marker is carried as a cuda-oxide generic op attribute, following the existing alignment-extension pattern.

## Tests

Added focused tests in:

- `crates/dialect-mir/tests/ops_test.rs`, proving volatile MIR loads/stores are non-promotable;
- `crates/mir-lower/src/convert/ops/memory.rs`, proving volatile MIR accesses lower to volatile LLVM accesses;
- `crates/llvm-export/tests/export_test.rs`, proving textual LLVM contains `load volatile` and `store volatile`.

Added `crates/rustc-codegen-cuda/examples/volatile_memory` with a standalone `Cargo.toml` and `Cargo.lock`. The example writes through `write_volatile`, reads through `read_volatile`, writes again, reads again, and checks exact host output.

The regression fails on `upstream/main` without volatile support and passes on this branch.

## Validation

```bash
cargo fmt --all -- --check
cargo test -p dialect-mir volatile_is_not_promotable
cargo test -p llvm-export volatile
cargo test -p mir-lower preserves_volatile
cargo check -p mir-importer
cargo clippy -p mir-importer -p dialect-mir -p mir-lower -p llvm-export --tests -- -D warnings
scripts/smoketest.sh --compile-only --no-color --only '^volatile_memory$'
scripts/smoketest.sh --no-color --only '^volatile_memory$'
```


nihalpasham:
Thanks for the PR. I'm quite happy with the end-to-end plumbing of the volatile bit. Did a couple things on top:

- rebased onto current main (it had gone stale against the inline-asm + debug-info work)
- added the `supported-features.md` row
- confirmed on GPU: `volatile_memory` round-trips 128 values and the PTX has `ld.volatile` / `st.volatile`

One small thing for next time (didn't block, left as-is): the dialect-mir side stores the flag as `IntegerAttr(i1)` - the export side already uses `BoolAttr`, so a `BoolAttr` there would match the rest of the codebase.

PR #210, opened 2026-06-12
mir-importer: index slice data pointers by element

### LINKED ISSUE #209: Index fat-slice data pointers by slice element
## Problem

Mutable references and projected assignments through a fat slice can compute the wrong address when the slice element type is itself an array. For example, `rows[row][0]` on `&mut [[f32; 2]]` should step by one `[f32; 2]` row first, then select column 0. The MIR address walker currently classifies the extracted slice data pointer from its pointee type alone, so `ptr<[f32; 2]>` can be mistaken for a pointer to one array object and the row index writes inside row 0 instead.

## Why this matters

Fat slice indexing is standard Rust codegen for mutable slice borrows, `slice::get_mut`, raw address-taking, and projected assignments. Device code that works with slices of fixed-size vectors or matrix rows can silently write the wrong element instead of failing to compile.

## Expected behavior

After dereferencing a fat slice, the extracted data pointer should be indexed as a pointer to slice elements. If the slice element is `[T; N]`, the first index should move by one whole `[T; N]`; only the next index should select inside that array element.

## Acceptance criteria

- A focused regression with `&mut [[f32; 2]]` fails on `upstream/main` by writing row 0 column 1 instead of row 1 column 0.
- The same regression passes on the patched branch.
- Runtime `Index`, forward `ConstantIndex`, mutable-reference address materialization, and direct projected assignment are covered.
- Existing scalar-element fat-slice write cases continue to pass.
- The implementation distinguishes "data pointer extracted from a fat slice" from "ordinary pointer to one array object" before choosing pointer-offset versus array-element-address lowering.

niklebedenko:
Closes #209.

## Problem

The place-address walker extracts field 0 from fat slice values so it can continue indexing through the original backing storage. When the slice element type is itself an array, such as `&mut [[f32; 2]]`, that extracted data pointer has type `ptr<[f32; 2]>`.

The compiler rule is that this pointer addresses slice elements. Therefore the first `Index` or forward `ConstantIndex` after the fat-slice deref must use pointer arithmetic over `[f32; 2]` rows. It must not be reclassified as a pointer to one array object and indexed inside row 0.

## Fix

Track when the current address is the data pointer just extracted from a fat slice. For the immediately following `Index` or forward `ConstantIndex`, force the address walker to use the direct pointer-offset path. Clear the marker after consuming that indexed projection, and also clear it for ordinary sized field access.

This keeps ordinary array-object indexing unchanged: a true pointer to `[T; N]` still uses `mir.array_element_addr`; only a pointer known to originate as fat-slice data uses element-sized `mir.ptr_offset` for the first index.

## Diligence

I checked the sibling lowering paths that decide between pointer offset and array-element address:

- `translate_place_addr_from_slot`, which serves `Rvalue::Ref`, `Rvalue::AddressOf`, and projected assignment destinations;
- `emit_indexed_element_addr`, the shared dispatch point for runtime and forward constant indices;
- `MirPtrOffsetOp` lowering, which scales by the pointer pointee type;
- `MirArrayElementAddrOp` lowering, which is correct for indexing inside one array object.

## Tests

Extended `crates/rustc-codegen-cuda/examples/slice_get_mut` with three nested-array fat-slice cases:

- `write_nested_slice_row_ref`: takes `let elem = &mut rows[row][0]` and writes through the reference;
- `write_nested_slice_const_row_ref`: takes `let elem = &mut rows[1][0]` and writes through the reference;
- `write_nested_slice_row_assign`: writes through direct projected assignment `rows[row][0] = value`.

The runtime-row cases prove the `Index` path, the literal-row case proves the forward `ConstantIndex` path, and the assignment case proves the statement-level destination path that reuses the same address walker.
I also checked the raw Rust MIR dump for the literal-row kernel; the relevant projection is emitted as `(*_9)[1 of 2][0 of 1]`, so this is covering the forward constant-index arm rather than another runtime-index lowering.

All three cases initialize row 0 column 1 to a sentinel and row 1 column 0 to zero. On `upstream/main` with only the regression applied, they write the sentinel column in row 0:

```text
write_nested_slice_row_ref     FAIL    (elem[0] = [[10.0, 42.0], [0.0, 99.0]])
write_nested_slice_const_row_ref FAIL    (elem[0] = [[10.0, 62.0], [0.0, 99.0]])
write_nested_slice_row_assign  FAIL    (elem[0] = [[10.0, 52.0], [0.0, 99.0]])
```

On this branch, they write row 1 column 0:

```text
write_nested_slice_row_ref     PASS    (elem[0] = [[10.0, 77.0], [42.0, 99.0]])
write_nested_slice_const_row_ref PASS    (elem[0] = [[10.0, 77.0], [62.0, 99.0]])
write_nested_slice_row_assign  PASS    (elem[0] = [[10.0, 77.0], [52.0, 99.0]])
```

## Validation

```bash
cargo fmt --all -- --check
(cd crates/rustc-codegen-cuda/examples/slice_get_mut && cargo fmt --all -- --check)
cargo check -p mir-importer
(cd crates/rustc-codegen-cuda/examples/slice_get_mut && cargo clippy --all-targets -- -D warnings)
scripts/smoketest.sh --compile-only --no-color --only '^slice_get_mut$'
cargo oxide run slice_get_mut
scripts/smoketest.sh --no-color --only '^slice_get_mut$'
```

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -3465,6 +3465,15 @@ fn translate_place_addr_from_slot(
     let mut current_is_slice_data = false;
 
     for (proj_idx, elem) in projection.iter().enumerate() {
+        // The slice-data provenance bit only describes the pointer produced by
+        // the immediately-preceding `Deref` of a fat slice (index it by
+        // element, not as a pointer to one array object). Capture it for this
+        // iteration and clear it up front, so the invariant stays local: it is
+        // true only when the previous step handed us a slice DATA pointer, and
+        // no later projection arm can accidentally leak it forward.
+        let entered_as_slice_data = current_is_slice_data;
+        current_is_slice_data = false;
+
         match elem {
             // `*place` -- the place walked so far holds a pointer; the
             // address of the dereferenced place is that pointer VALUE, so a
@@ -3664,7 +3673,6 @@ fn translate_place_addr_from_slot(
                             // the field arm below.
                             mir::ProjectionElem::Field(..) => {
                                 current = data_ptr;
-                                current_is_slice_data = false;
                                 continue;
                             }
                             // Element access through a slice data pointer is
@@ -3724,7 +3732,6 @@ fn translate_place_addr_from_slot(
             }
 
             mir::ProjectionElem::Field(field_idx, field_ty) => {
-                current_is_slice_data = false;
                 let field_type = types::translate_type(ctx, field_ty)?;
                 let result_ptr_ty =
                     dialect_mir::types::MirPtrType::get_generic(ctx, field_type, is_mutable);
@@ -3761,7 +3768,7 @@ fn translate_place_addr_from_slot(
                     Some(kind) => kind,
                     None => return Ok(None),
                 };
-                if current_is_slice_data {
+                if entered_as_slice_data {
                     pointee_kind = PointeeKind::Direct;
                 }
 
@@ -3798,7 +3805,6 @@ fn translate_place_addr_from_slot(
                 );
                 current = next_current;
                 current_prev_op = Some(addr_op);
-                current_is_slice_data = false;
             }
 
             // Runtime `arr[i]` indexing. Without this arm, a place like
@@ -3810,7 +3816,7 @@ fn translate_place_addr_from_slot(
                     Some(kind) => kind,
                     None => return Ok(None),
                 };
-                if current_is_slice_data {
+                if entered_as_slice_data {
                     pointee_kind = PointeeKind::Direct;
                 }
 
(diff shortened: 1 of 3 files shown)


nihalpasham:
I was able to reproduce the wrong store address for `&mut [[T;N]][i][j]` on main and the fix on the branch. So this LGTM 👍🏾 i.e. GPU-verified.
- Layered: localized the slice-data flag (capture-and-clear at loop top; generated `.ll` byte-identical) + a wider-element, non-zero-column kernel (`rows[row][2]` over `&mut [[f32;3]]`).
- `cargo oxide run slice_get_mut --arch sm_90` → SUCCESS, all kernels pass.

PR #212, opened 2026-06-12
mir-importer: read closure upvars from generic suffix

### LINKED ISSUE #211: Read closure upvars from the closure substitution suffix
## Problem

Generic device code that constructs a captured closure can fail during MIR verification when the closure is passed through an out-of-line helper. rustc encodes closure substitutions as the parent generic arguments followed by the fixed closure suffix, so the tuple of captured variables is the last generic argument. The MIR importer currently reads a fixed substitution index, which can translate the closure environment as a zero-field struct even when the closure construction has captures.

## Why this matters

Captured closures are a natural way to write reusable generic device helpers. If the importer loses the closure environment fields, generic kernels can fail to compile or, in less direct shapes, reason about closure zero-sized-ness from the wrong type information.

## Expected behavior

Closure type translation should get the tupled-upvars type from rustc's closure substitution suffix. Generic parent substitutions should not change which argument describes the closure environment.

## Acceptance criteria

- A generic kernel that passes a captured closure through an out-of-line generic helper fails on `upstream/main` before the fix.
- The same kernel compiles and runs for multiple concrete type instantiations after the fix.
- Closure ZST detection and closure struct translation use the same upvar-extraction helper.
- Existing generic kernel examples continue to compile and run.

niklebedenko:
Closes #211.

## Problem

rustc suffix-encodes closure substitutions as `[parent_args..., closure_kind, closure_sig, tupled_upvars]`. The MIR importer treated `substs[2]` as the upvars tuple, which is only true when there are no parent generic arguments before the closure suffix.

In a generic kernel that constructs a captured closure and passes it through an out-of-line generic helper, the old importer translated the closure environment as a zero-field struct even though closure construction still had two capture operands. On `upstream/main` with only the regression applied:

```text
MirConstructStructOp has 2 operands but struct
'generic::kernels::closure_capture::{closure#0}' has 0 fields
```

## Fix

Add a small `closure_upvar_tys` helper that reads the tupled-upvars type from the last closure generic argument and verifies that it is a tuple. Use that helper in both places that decode closure captures:

- `is_rust_type_zst`, so captured closures in generic contexts are not mistaken for zero-sized closures;
- closure `translate_type`, so the generated MIR struct fields match rustc's closure environment.

This keeps the compiler rule in one place instead of duplicating a fixed-index assumption.

## Diligence

I checked the sibling closure paths for other fixed-index upvar decoding:

- `AggregateKind::Closure` builds the closure value from MIR operands and delegates the closure type to `types::translate_type`.
- closure function ABI handling uses `fn_sig()` and does not inspect upvar substitutions.
- closure call lowering resolves the closure body from the receiver type and substitutions, but does not decode upvars itself.
- collector closure discovery resolves closure bodies from rustc type information and does not build closure environment fields.

The initial direct-call regression was stale because rustc optimized the closure environment away before import. The committed regression keeps the closure as a real value by passing it through `#[inline(never)] apply_unary`.

## Tests

Extended `crates/rustc-codegen-cuda/examples/generic` with `closure_capture<T>`, a generic kernel that captures two generic values in a closure and passes that closure to an out-of-line generic helper:

```rust
let transform = move |x: T| (x + bias) * scale;
*out_elem = apply_unary(transform, input[idx_raw]);
```

The host launches the kernel for both `f32` and `i32`, checking that the captured arithmetic is applied correctly for each monomorphization.

## Validation

Verified the regression fails on `upstream/main` by applying only the example change and running:

```bash
CUDA_OXIDE_SHOW_RUSTC_MIR=1 cargo oxide run generic
```

Verified the patched branch with:

```bash
cargo fmt --all -- --check
cargo check -p mir-importer
(cd crates/rustc-codegen-cuda/examples/generic && cargo fmt --all -- --check)
(cd crates/rustc-codegen-cuda/examples/generic && cargo clippy --all-targets -- -D warnings)
(cd crates/rustc-codegen-cuda && cargo fmt --all -- --check)
(cd crates/rustc-codegen-cuda && cargo clippy --all-targets -- -D warnings)
scripts/smoketest.sh --compile-only --no-color --only '^generic$'
cargo oxide run generic
scripts/smoketest.sh --no-color --only '^generic$'
```

### DIFF
--- a/crates/mir-importer/src/translator/types.rs
+++ b/crates/mir-importer/src/translator/types.rs
@@ -67,6 +67,23 @@ pub fn get_usize_type(
     pliron::builtin::types::IntegerType::get(ctx, 64, pliron::builtin::types::Signedness::Unsigned)
 }
 
+/// Returns the tupled-upvars types for a `RigidTy::Closure`.
+///
+/// rustc suffix-encodes closure substitutions as
+/// `[parent_args..., closure_kind, closure_sig, tupled_upvars]`, so the upvars
+/// tuple is the last generic arg, not a fixed index.
+fn closure_upvar_tys(substs: &rustc_public::ty::GenericArgs) -> Option<Vec<rustc_public::ty::Ty>> {
+    let rustc_public::ty::GenericArgKind::Type(upvar_tuple_ty) = substs.0.last()? else {
+        return None;
+    };
+    let rustc_public::ty::TyKind::RigidTy(rustc_public::ty::RigidTy::Tuple(upvar_tys)) =
+        upvar_tuple_ty.kind()
+    else {
+        return None;
+    };
+    Some(upvar_tys)
+}
+
 /// Returns the 32-bit floating point type.
 pub fn get_f32_type(
     ctx: &mut Context,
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks for the PR. This was an interesting catch - `suffix vs fixed-index` thing. 

Merging as-is. Tiny heads up: this one and #206 both add a helper at the same spot in types.rs, so whichever lands second will need a trivial keep-both rebase. 👍🏾

PR #214, opened 2026-06-12
mir-importer: require closure receiver for rust-call lowering

### LINKED ISSUE #213: Only unpack rust-call tuples for actual closure receivers
## Problem

The MIR importer recognizes closure trait calls by looking for callee names such as `call_once`, `call_mut`, or `::call`, plus a closure appearing somewhere in the call's generic substitutions. That can misclassify ordinary methods named `call` on wrapper types that merely contain a closure in a generic parameter.

When this happens, the importer unpacks the rust-call tuple as if the receiver were a closure trait shim. The generated call can then have the wrong argument shape for the real method.

## Why this matters

Closure wrappers and adapter types commonly expose methods named `call`. Device code should be able to use those wrappers without the importer confusing normal method ABI with closure trait ABI.

## Expected behavior

The closure-call lowering path should run only when the call receiver is a closure, or a reference to a closure. Ordinary ADT methods named `call` should keep their tuple argument as one normal method argument.

## Acceptance criteria

- A device closure example with an out-of-line `ClosureWrapper<F>::call(&self, (u32,))` method fails on `upstream/main` before the fix.
- The same example compiles and runs after the fix.
- Existing genuine closure trait calls still compile and run.
- The importer predicate checks the receiver type, not just the callee name and generic substitution debug text.

niklebedenko:
Closes #213.

## Problem

`translate_call` special-cases closure trait calls because rust-call ABI passes arguments as a tuple while the closure body expects unpacked arguments. The old predicate looked for names containing `call_once`, `call_mut`, or ending in `::call`, and then checked whether any generic substitution text contained `Closure`.

That is too broad. A normal wrapper method named `call` can have a closure as a generic parameter while still expecting its tuple argument as one ordinary method argument. On `upstream/main` with only the regression applied, `ClosureWrapper<F>::call(&self, (u32,))` is misclassified and LLVM verification fails:

```text
Function call has incorrect type:
expected llvm.func <builtin.integer i32(llvm.ptr (0), llvm.struct <{ builtin.integer i32 }>) variadic = false>,
got llvm.func <builtin.integer i32(llvm.ptr (0), builtin.integer i32) variadic = false>
```

## Fix

Require the rust-call receiver to actually be a closure before using `translate_closure_call`.

The new `receiver_is_closure` helper inspects the first call operand, resolves its MIR local or constant type, unwraps references, and matches `RigidTy::Closure`. If the receiver is an ordinary ADT such as `ClosureWrapper<F>`, the call stays on the normal call-lowering path and keeps its tuple argument intact.

## Diligence

I checked the sibling closure-call paths:

- `translate_closure_call` still handles the genuine closure trait ABI path, including by-value `call_once` and by-reference `call` / `call_mut` receivers.
- `extract_closure_body_name` already requires the receiver operand to be a closure or reference-to-closure before resolving a closure body.
- the collector has a name-based closure discovery heuristic, but it only adds closure bodies to the collection when a type argument is itself a closure; it does not rewrite call operands or unpack tuples.

The initial wrapper regression used `#[inline(always)]`, which let rustc optimize the wrapper method away before import. The committed regression uses `#[inline(never)]` so the importer must classify the ordinary wrapper call.

## Tests

Extended `crates/rustc-codegen-cuda/examples/device_closures` with:

- `ClosureWrapper<F>`, where `F: Fn(u32) -> u32`;
- an out-of-line `ClosureWrapper::call(&self, (u32,)) -> u32`;
- `test_wrapper_method_named_call`, which stores a captured closure in the wrapper and calls the wrapper method from a kernel.

The existing device closure tests still exercise genuine closure calls, including closure trait calls through `FnOnce`.

## Validation

Verified the regression fails on `upstream/main` by applying only the example change and running:

```bash
CUDA_OXIDE_SHOW_RUSTC_MIR=1 cargo oxide run device_closures
```

Verified the patched branch with:

```bash
cargo fmt --all -- --check
cargo check -p mir-importer
(cd crates/rustc-codegen-cuda/examples/device_closures && cargo fmt --all -- --check)
(cd crates/rustc-codegen-cuda/examples/device_closures && cargo clippy --all-targets -- -D warnings)
(cd crates/rustc-codegen-cuda && cargo fmt --all -- --check)
(cd crates/rustc-codegen-cuda && cargo clippy --all-targets -- -D warnings)
scripts/smoketest.sh --compile-only --no-color --only '^device_closures$'
cargo oxide run device_closures
scripts/smoketest.sh --no-color --only '^device_closures$'
```

### DIFF
--- a/crates/mir-importer/src/translator/terminator/mod.rs
+++ b/crates/mir-importer/src/translator/terminator/mod.rs
@@ -1450,6 +1450,20 @@ fn translate_closure_call(
 }
 
 /// True only when the rust-call receiver is itself a closure.
+///
+/// We type the operand's base local (or constant) and ignore any
+/// `place.projection`, and we peel at most one reference. Both are safe for
+/// the rust-call receiver specifically: rustc lowers a closure-trait call
+/// (`Fn::call` / `FnMut::call_mut` / `FnOnce::call_once`) so that the receiver
+/// argument is the closure passed by value, or a single `&`/`&mut` borrow of
+/// it, materialized into its own temporary local, never an in-place projection
+/// of a larger aggregate and never behind multiple references. So a closure
+/// reached through a field (`(self.f)(x)`) still arrives here as a base local
+/// of type `&{closure}`, and one `Ref` peel plus a base-local type check covers
+/// every genuine closure-call shape. A wrapper ADT that merely carries a
+/// closure in its generic substitutions (the case this guard exists to reject)
+/// has a non-closure receiver type and is correctly left on the ordinary call
+/// path with its rust-call tuple intact.
 fn receiver_is_closure(receiver: &mir::Operand, body: &mir::Body) -> bool {
     let ty = match receiver {
         mir::Operand::Copy(place) | mir::Operand::Move(place) => {
--- a/crates/mir-importer/src/translator/terminator/mod.rs
+++ b/crates/mir-importer/src/translator/terminator/mod.rs
@@ -903,15 +903,13 @@ fn translate_call(
         }
     }
 
-    // Handle closure trait method calls (FnOnce::call_once, FnMut::call_mut, Fn::call)
-    // These calls pass arguments as a tuple, but the closure body expects unpacked args.
-    // We need to unpack the tuple before calling the closure.
-    //
-    // MIR shows: <{closure} as FnMut<(u32,)>>::call_mut(self_ref, tuple_args)
-    // But the closure body expects: fn(self_ref, unpacked_arg1, unpacked_arg2, ...)
+    // Handle genuine closure trait method calls. The receiver test matters:
+    // wrapper ADTs can carry a closure in their generic substitutions while
+    // still expecting the rust-call tuple as one argument.
     if let Some(ref name) = pattern_name
         && (name.contains("call_once") || name.contains("call_mut") || name.ends_with("::call"))
-        && substs_contains("Closure")
+        && !args.is_empty()
+        && receiver_is_closure(&args[0], body)
     {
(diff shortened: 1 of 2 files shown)


niklebedenko:
Hey @nihalpasham, sorry to land a whole batch on you at once today (#181–#214). Quick map to make triage easier:

- `mir-importer` lowering fixes: 
  - #183 
  - #196 
  - #199 
  - #206 
  - #208 
  - #210 
  - #212 
  - #214
- `cuda-core`: 
  - #193 
  - #202 
  - #204
- `pipeline`: 
  - #181 
  - #189
- misc:
  - `codegen` #185
  - `intrinsics` #187 
  - `cuda-device`, bf16x2 FMA #191

They should all be independent, merge or close in whatever order suits you, no rush on any. The `mir-importer` ones touch the same area, so if they conflict on the way in, ping me and I'll be happy to rebase (or ofc feel free to edit the branch yourself if that's easier). I'm hitting these bugs because I'm leaning on the compiler hard in real use on my own (fairly large) project, so I expect I'll keep finding fixes.

Really enjoying putting the compiler through its paces, and glad to shape how I contribute around whatever's easiest for you!

nihalpasham:
Thanks for these. Just a heads-up. I'll need some time to review these as I am working through a couple of large ticket items.

nihalpasham:
LGTM.
- Layered: a comment on why `receiver_is_closure` may ignore `place.projection` or peel one ref, plus projection-receiver and double-ref closure kernels (both still unpack the tuple).
- `cargo oxide run device_closures --arch sm_90` → all tests pass, incl. the two new ones

PR #216, opened 2026-06-13
feat: add asin/acos intrinsic support for f32 and f64

goog00:
### Summary

- Registers `asinf32` / `asinf64` / `acosf32` / `acosf64` placeholder symbols in `dialect-mir/src/rust_intrinsics.rs`, following the same pattern as the existing float math intrinsics
- Recognizes both call shapes in the mir-importer float-math dispatch:
  - the std inherent methods `f{32,64}::asin`/`acos`, which lower through the `std::sys::cmath::asinf` / `asin` / `acosf` / `acos` shims (asin/acos are std-only — there is no `core` libm form, unlike `cbrt`)
  - the `libm`-crate spellings (`libm::asinf`, `libm::math::acos::acos`, ...)  that glam's `nostd-libm` lowering emits on nvptx
- Maps them to CUDA libdevice builtins `__nv_asinf` / `__nv_asin` / `__nv_acosf` / `__nv_acos` in the MIR-lower call translator
- Hooks the new placeholders into the codegen collector's cmath-shim allowlist
- Adds an `asin_acos_smoke` example: 14 test values across the full `[-1, 1]` domain (including both endpoints and `-0.0`) × 4 variants (asin/acos × f32/f64), verifying GPU results are within 2 ULP of host `libm`

### Test plan

- `cargo test -p dialect-mir -p mir-importer -p mir-lower --lib` — unit  tests covering intrinsic name → enum dispatch, the `std::sys::cmath` +  `libm`-crate spellings, and arg count
- `cargo oxide run asin_acos_smoke` — end-to-end GPU smoke test; expected  output: `SUCCESS: 14 cases × 4 variants within 2 ULP of host libm`
- Verified on RTX 5090 (sm_120a, CUDA 13.0)


nihalpasham:
Thanks for the PR. LGTM

The libdevice routing is correct and the `is_libm_path guard` prevents false-positive matches. 

Note this slightly widens the blast radius of issue #98  (NVVM IR on pre-Blackwell targets), but that's the right layering - the intrinsics should land independently of the NVVM IR fix.

Issue #139, opened 2026-06-08
Windows MSVC support fork and signed release artifacts


ansidium:
Hi! I have been working on a Windows-support fork of cuda-oxide here:

https://github.com/ansidium/cuda-oxide-windows

The fork keeps the Linux path upstream-compatible while adding an experimental native Windows MSVC path. The first signed Windows release is published here:

https://github.com/ansidium/cuda-oxide-windows/releases/tag/windows-v0.2.0

Current Windows scope:

- Target: `x86_64-pc-windows-msvc`.
- Windows GNU is intentionally out of scope.
- Windows CI canary on `windows-latest` without a live GPU.
- Side-by-side `cargo-oxide.exe` + `rustc_codegen_cuda.dll` release layout.
- CUDA Toolkit / NVVM / nvJitLink discovery improvements for Windows paths.
- Windows smoke script and documentation.

I do not want to open one large PR that mixes fork branding, release packaging, CI, docs, and portability fixes. If you are open to upstreaming any of this, I can split the work into smaller PRs, for example:

1. Cross-platform path and dynamic-library naming helpers.
2. CUDA Toolkit / NVVM / nvJitLink discovery improvements.
3. Windows-safe `cargo oxide` command/path handling.
4. Windows CI no-GPU compile canary.
5. Windows setup docs.

Would any of these be useful upstream? If so, which order and scope would you prefer?

egorsmkv:
Thanks really needed, subscribed to your repo

nihalpasham:
Thanks for the detailed write-up, and for keeping the fork close to upstream.

**Just so that I don't set the wrong expectations:** we have not decided whether we can support Windows, and we may not be able to. The questions below are only to understand what supporting it would actually cost, before we commit either way.

1. **Ongoing upkeep:** The main codebase changes fast. It's already ~110 commits past your fork's last sync point, and every one of those changes is something a Windows build has to keep working against. How often do you currently have to fix the Windows build after pulling our changes in, and who would own that upkeep if it lived upstream?
2. **Catching Windows breakage:** GitHub's Windows runners have no GPU, and today even your fork's automatic CI doesn't rebuild the codegen backend (the part with the linker/export-limit problem) on every change. If a Windows-only break slipped into our main, what would catch it before a release?
[...]

PR #227, opened 2026-06-15
feat: Windows support (x86_64-pc-windows-msvc)

joshuapetersen:
## Summary

This PR adds Windows support to cuda-oxide. All 18 crates compile and GPU kernels execute correctly on Windows.

**Tested on:** Windows 11, RTX 4050 (SM_89, Ada Lovelace), CUDA 13.1, Rust nightly-2026-04-03, LLVM 22.1.7

### Fix 1: MSVC enum type mismatches in `cuda-core` (4 files, 10 lines)

On Linux, `bindgen` maps CUDA C enums to `u32`. On Windows MSVC, the same enums map to `i32` because MSVC defaults C enum types to `int` (signed). Added `as u32` casts at all 10 call sites in `context.rs`, `event.rs`, `lib.rs`, and `stream.rs`.

### Fix 2: PE/COFF 65535 export limit in `rustc-codegen-cuda` (3 new files)

The codegen backend DLL re-exports ~66,953 symbols from `rustc_driver`, exceeding the PE/COFF limit of 65,535. Fixed by:
- Adding a `build.rs` that passes a custom `.def` file exporting only `__rustc_codegen_backend`
- Adding `.cargo/config.toml` to use `lld-link` on Windows
- Adding `codegen_backend.def` with the single required export

### Fix 3: COFF support for PTX embedding in `oxide-artifacts` (1 file, ~30 lines)

`build_host_object_for_target()` only supported ELF object files. Added:
- Windows target detection (`BinaryFormat::Coff`) and macOS detection (`BinaryFormat::MachO`)
- COFF section flags (`IMAGE_SCN_CNT_INITIALIZED_DATA | IMAGE_SCN_MEM_READ`) alongside the existing ELF flags

### Fix 4: Windows DLL path detection in `cargo-oxide` backend discovery (1 file)

`backend.rs` had `librustc_codegen_cuda.so` hardcoded in 6 places. Added a `backend_lib_name()` helper that returns `rustc_codegen_cuda.dll` on Windows and `librustc_codegen_cuda.so` on Linux.

## Build Instructions (Windows)

```powershell
# Prerequisites: CUDA Toolkit, LLVM (for bindgen), Rust nightly
$env:CUDA_TOOLKIT_PATH = "C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.1"
$env:LIBCLANG_PATH = "C:\Program Files\LLVM\bin"

# Build workspace
cargo +nightly-2026-04-03 build

# Build and run an example
cargo +nightly-2026-04-03 oxide run vecadd
```

## Files Changed

| File | Change |
|------|--------|
| `cuda-core/src/context.rs` | 3 `as u32` casts |
| `cuda-core/src/event.rs` | 1 `as u32` cast |
| `cuda-core/src/lib.rs` | 4 `as u32` casts |
| `cuda-core/src/stream.rs` | 2 `as u32` casts |
| `oxide-artifacts/src/lib.rs` | COFF format + section flags |
| `cargo-oxide/src/backend.rs` | Platform-aware DLL/SO paths |
| `rustc-codegen-cuda/build.rs` | **[NEW]** Windows export limit fix |
| `rustc-codegen-cuda/.cargo/config.toml` | **[NEW]** lld-link config |
| `rustc-codegen-cuda/codegen_backend.def` | **[NEW]** Minimal export list |

**Total: 6 files modified, 3 files created, ~60 lines of code.**

### DIFF
--- a/WINDOWS.md
+++ b/WINDOWS.md
@@ -1,0 +1,216 @@
+# cuda-oxide Windows Port
+
+> **Fork of [NVlabs/cuda-oxide](https://github.com/NVlabs/cuda-oxide)** with Windows support added.
+
+## What is cuda-oxide?
+
+cuda-oxide is an experimental compiler project by NVIDIA Labs that lets you write GPU kernels in **pure Rust** instead of CUDA C++. You annotate functions with `#[kernel]`, and a custom rustc codegen backend compiles them directly to PTX (NVIDIA GPU assembly).
+
+```rust
+#[cuda_module]
+mod my_kernels {
+    #[kernel]
+    pub fn vector_add(a: &[f32], b: &[f32], mut out: DisjointSlice<f32>) {
+        let tid = thread::index_1d();
+        if let Some(slot) = out.get_mut(tid) {
+            *slot = a[tid.get()] + b[tid.get()];
+        }
+    }
+}
+```
+
+The compilation pipeline bypasses CUDA C++ entirely:
+
+```
+Rust source → rustc MIR → Pliron IR → LLVM IR → NVPTX → PTX
+```
+
+No `nvcc`. No `.cu` files. No C++ headers. Just Rust — with borrow checking, lifetimes, and zero-cost abstractions — compiled straight to GPU machine code.
+
+## Why This Fork Exists
+
+cuda-oxide was released as **Linux-only**. The README says so. The CI only runs on Linux. Every path in the codebase is hardcoded for ELF and `.so` shared libraries. There was no Windows support at all.
+
+We needed it on Windows. So we ported it.
+
+**This fork contains the minimum set of changes (6 fixes, ~60 lines of code) to make cuda-oxide compile and run on Windows.** Every existing example in the upstream repo works after these changes.
+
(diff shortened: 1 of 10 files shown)


nihalpasham:
Thanks for this, and appreciate the effort that went into testing it end to end on real hardware. The 4 areas you outlined (MSVC enum signedness, COFF embedding, .so vs .dll discovery, and the backend export limit) are what I'd look for in a Windows port.

We're consolidating all Windows discussion in #139, which was sbumitted a little earlier and is where we're working through whether and how we can commit to supporting Windows as maintainers. The open questions there are the ongoing ones: 
- CI without GPU runners, 
- keeping the codegen-backend build green across toolchain bumps, 
- and the sync burden as main moves. 

Rather than review two parallel Windows tracks, we'd like to keep it in one place, so I'm going to close this in favor of #139.

To be clear, this is not a "no" on the work. It's "not decided yet, and let's not split the conversation." #139 also references an existing experimental Windows fork that already carries a similar set of fixes, which is part of why we want to consolidate. If you'd like to weigh in there, please do, your results would be a useful data point. Thanks again.

PR #236, opened 2026-06-16
evaluate unevaluated array length constant for code generation

haixuanTao:
## Summary

Here's the core constraint: by the time you're in codegen, array length consts should be fully monomorphized and evaluatable. But stable_mir's TyConst::eval_target_usize() fails on TyConstKind::Unevaluated — it can't reduce those to a u64 through stable_mir's API alone.

The TyCtxt bridge is necessary because:

- TyCtxt::normalize_erasing_regions(fully_monomorphized, const) is the actual rustc mechanism to force-evaluate an unevaluated const in a monomorphized context
- That API is only available through rustc internals (rustc_middle), not through stable_mir
- The type translator lives in mir-importer, which intentionally has no direct TyCtxt dependency — hence the callback bridge rather than passing tcx directly
- The alternative would be to pass tcx directly into run_pipeline and down to the type translator, but that would couple mir-importer tightly to rustc internals, defeating its design goal of being stable_mir-only.

So the bridge is a pragmatic workaround for a gap in stable_mir's const evaluation API. If stable_mir ever exposes a way to evaluate Unevaluated consts in a fully-monomorphized context, the whole bridge could be deleted.

### DIFF
--- a/crates/mir-importer/src/lib.rs
+++ b/crates/mir-importer/src/lib.rs
@@ -87,3 +87,29 @@ pub use pipeline::{
     CollectedFunction, CompilationArtifactKind, CompilationResult, DeviceExternAttrs,
     DeviceExternDecl, PipelineConfig, PipelineError, run_pipeline,
 };
+
+// ── Array-length const evaluation bridge ─────────────────────────────────────
+// The type translator must turn `[T; N]` array lengths into concrete values,
+// but stable_mir cannot *evaluate* an `Unevaluated` length const (e.g. a `const
+// SUB_LEN: usize = 2` reference) at the type-translation layer. The codegen
+// backend (which holds `TyCtxt`) installs a lifetime-erased eval callback here
+// for the duration of one `rustc_internal::run` scope, while tcx is alive.
+type ArrayLenEval = Box<dyn Fn(&rustc_public::ty::TyConst) -> Option<u64>>;
+thread_local! {
+    static ARRAY_LEN_EVAL: std::cell::RefCell<Option<ArrayLenEval>> =
+        const { std::cell::RefCell::new(None) };
+}
+/// Install the array-length eval callback. SAFETY CONTRACT: `f` borrows the
+/// compiler `TyCtxt`; the caller MUST `clear_array_len_eval()` before that tcx
+/// is released (i.e. within the same `rustc_internal::run` scope).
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks for the PR.

**A couple of points:**

Upon reviewing this, I don't think we can take it in its current form. Wanted to explain why rather than just close it.

The change moves the compiler context (`tcx`) into `mir-importer` through a global thread-local plus an `unsafe` transmute that erases the closure's lifetime. That's a problem on two fronts:

- `mir-importer` is intentionally built only on `rustc_public` (the stable-track API). We keep compiler internals (`rustc_middle` / `rustc_internal`) out of that crate on purpose, so it stays portable as that API stabilizes — this change pulls them back in.
- The global plus a lifetime-erasing `unsafe` isn't something we can sign off on for soundness, even with the cleanup guard.

Before we land any fix, could you share a small self-contained kernel that reproduces the failure? I tried to trigger the "unevaluated array length" path from ordinary `cuda-oxide` code (const generics, const-item lengths) and it all compiles fine today, so I'd like to see exactly what you're compiling that hits it.

If it turns out to be a real gap, I'm fairly sure it lives upstream: `rustc_public`'s `eval_target_usize` calls `try_to_target_usize` without normalizing the const first, so it just gives up on an unevaluated one. The proper fix is a small change there (normalize, then read), and we'd gladly help get that into `rustc_public` rather than work around it on our side.

nihalpasham:
I'm closing this. Feel free to re-open if you have any updates.

haixuanTao:
Following up to close the loop: this is now obsolete — the problem no longer reproduces on current `main`.

I retested the original trigger (the dimforge nexus physics shaders, whose constraint banks use const-expression array lengths like `[JointConstraintElement; MAX_CONSTRAINTS_PER_JOINT]` with `const MAX_CONSTRAINTS_PER_JOINT: usize = SPATIAL_DIM * 2`) against latest `main` with our workaround fully removed: the whole crate (265 device functions) compiles to a working sm_120 cubin, and plain `TyConst::eval_target_usize()` handles every length const the build produces — nothing `Unevaluated` reaches the type translator anymore. Somewhere between the June base this PR was written against and current `main`, the gap closed.

We've dropped the tcx bridge from our fork accordingly. Thanks for the review notes — agreed in hindsight on both the crate-boundary and soundness points. No need to reopen.

PR #237, opened 2026-06-16
fix(mir-importer): materialize nested-aggregate and ZST-field constants

haixuanTao:
Three constant-translation gaps for real shader code:
- Nested aggregate constants: a const Mat3 (struct of Vec3 structs, e.g. glam Mat3::ZERO) now recursively builds each field from its byte slice instead of failing with "Struct constant field N has unsupported type".
- constant_storage_size is extended to compute the byte size of struct and array constant types (needed for the recursive field slicing above).
- ZST-struct field synthesis: a zero-sized struct still carries (zero-sized) typed fields, and MirConstructStructOp requires one operand per field; synthesize a ZST value per field type (e.g. TryFromIntError(()), hit by step_by under -Zbuild-std) instead of constructing with zero operands.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -5306,11 +5306,21 @@ fn translate_struct_constant(
                 let field_bytes = if byte_offset + byte_size <= bytes.len() {
                     &bytes[byte_offset..byte_offset + byte_size]
                 } else {
-                    return input_err!(loc, TranslationErr::unsupported(format!(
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks for the PR.

- Rebased onto main; layered a `const_aggregate` regression example (struct-of-struct, struct-of-array-of-struct, ZST-field).

Reviewer note, verified: fails on main with "Struct constant field 0 has unsupported type", passes on the branch, GPU-verified on `sm_120` (every lane = 110.0).

PR #238, opened 2026-06-16
fix(mir-lower): handle three transmute/pointer-cast edge cases

haixuanTao:
Three related cast-lowering gaps that surface when compiling real shader crates (glam/khal_std/std) to PTX:

- **array -> slice unsize** forming `&[T]` over a `#[spirv(workgroup)]` shared array (addrspace 3) now addrspacecasts the data pointer to the slice field-0 addrspace (0), so it lowers to PTX `cvta.shared` (radix-sort / LU shared-mem).
- **thin ptr -> aggregate field-0 coercion**: a `{ptr, len}` slice or empty ZST function-item source is coerced/inttoptr-d to the exact field-0 type before the `insert_value`, since rustc may hand a different addrspace/int width.
- **scalar -> aggregate Transmute** the importer did not classify as a niche-enum (e.g. `usize -> NonNull<T>` in `Arguments::from_str`) now lowers as a faithful alloca+store+load memory round-trip instead of being refused.

### DIFF
--- a/crates/mir-lower/src/convert/ops/cast.rs
+++ b/crates/mir-lower/src/convert/ops/cast.rs
@@ -493,14 +493,22 @@ fn emit_pointer_cast(
             Ok(llvm::BitcastOp::new(ctx, val, llvm_ty).get_operation())
         }
     } else if src_is_int && dst_is_struct {
-        // Scalar -> aggregate Transmute with no niche encoding: we cannot
-        // safely guess the layout. Refuse loudly rather than fall through
-        // to an invalid bitcast.
-        pliron::input_err_noloc!(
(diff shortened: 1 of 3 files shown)


nihalpasham:
I trimmed the PR to the one case that actually applies to `cuda-oxide`.

- kept the `usize -> NonNull` transmute fix. Added a test, GPU-verified.
- Dropped the other two cases: I think you mentioned they only happen with rust-gpu `#[spirv(workgroup)]` shared arrays, which `cuda-oxide` doesn't have, so that code can never run here. I'd still be open to accepting them if you have cuda-oxide specific repro's for these cases.
- cuda-oxide also blocks `std` in device code, so the `std` paths never reach this point either.

PR #239, opened 2026-06-16
fix(mir-importer): lower core::intrinsics::typed_swap_nonoverlapping

haixuanTao:
`typed_swap_nonoverlapping::<T>(x, y)` — the primitive behind `mem::swap` / `mem::replace` — has no MIR body. Lower it as a load/load/store/store crossover; the two pointers are guaranteed non-overlapping, so no temporary is needed. The element type is read off the pointer operand. Dispatched from `try_dispatch_intrinsic`.


nihalpasham:
LGTM 

- Rebased onto main, added a test that swaps two values. GPU-verified.

PR #240, opened 2026-06-16
fix(mir-importer): compute real element/field addresses for place projections

haixuanTao:
When auto porting khal_std/glam accessors like at_mut(i)/write(i,v) and raw &mut a[i].field lower to place projections (Ref, AddressOf and assignment) over slice/array references. The importer fell back to materialising the element into a fresh stack slot, so the store went to a dead copy -- writes through &mut [T] / &mut [T;N] were silently lost (e.g. gpu_add gave the unchanged output), and an atomic on a stack copy is illegal on NVPTX.

Compute the real address instead, for both the reference side and the store side:
- Rvalue::Ref / Rvalue::AddressOf of (*slice)[i] / (*slice)[i].field / (*arr)[i]: fold [Deref, Index/ConstantIndex, ...fields] to data-pointer + ptr_offset (+ field GEPs), via translate_place_addr_from_slot (now handling Deref/runtime-Index/array) and a shared ref_index_operand / emit_index_operand helper.
- Assignment to the same projections (statement.rs): the (Deref,Index)/(Deref, ConstantIndex) and general (Deref, Field-chain) store arms now emit a real element/field address store rather than writing a stack copy.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -847,6 +847,183 @@ pub fn translate_rvalue(
                 return Ok((None, result_val, last_inserted));
             }
 
+            // Case 2b: `&(*slice)[i]` — reference to a slice element through a
+            // slice reference (`&[T]` / `&mut [T]`). Compute the element address
+            // (data pointer + index) so mutation reaches the original buffer.
+            // Without this, the generic fallback (Case 5) copies the element
+            // into a fresh slot and writes are silently lost — e.g. khal_std's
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks for the PR.

But this was superseded by commits ab417b5, 1ce4e75, adaed3b, c89b480, f1c7436 that landed on main between June 11-17 via a shared `translate_place_addr_from_slot` walker. This also covers `DisjointSliceType` fat pointers, `ZST` tuple pointees, and `slice-tailed` structs.

Closing with already-fixed.

PR #249, opened 2026-06-18
fix(#231): lower projected reads through place addresses

### LINKED ISSUE #231: Projected reads from local arrays copy the whole aggregate before scalar indexing
## Problem

A projected read from an addressable local array is lowered by loading/copying the whole aggregate, then indexing into the copy.

Self-contained kernel example:
```rust

const ROWS: usize = 4;
const COLS: usize = 2;

#[kernel]
pub fn projected_read_copies_whole_array(
    row_arg: usize,
    col_arg: usize,
    mut out: DisjointSlice<f64>,
) {
    let idx = thread::index_1d();
    if idx.get() != 0 {
        return;
    }

    let mut scratch = [[0.0_f64; COLS]; ROWS];
    scratch[0] = [1.0, 2.0];
    scratch[1] = [3.0, 4.0];
    scratch[2] = [5.0, 6.0];
    scratch[3] = [7.0, 8.0];

    let row = row_arg & (ROWS - 1);
    let col = col_arg & (COLS - 1);
    let value = scratch[row][col];

    if let Some(slot) = out.get_mut(idx) {
        *slot = value;
    }
}
```
`row_arg` and `col_arg` are kernel parameters intentionally used to keep the row and column indices dynamic, so the indexing cannot be constant-folded away.

Current generated IR on `main` at `6f663d1` includes:
```llvm
%v36 = load [4 x [2 x double]], ptr %v10, align 8
%v37 = alloca [4 x [2 x double]], align 8
store [4 x [2 x double]] %v36, ptr %v37, align 8
%v38 = getelementptr inbounds [4 x [2 x double]], ptr %v37, i32 0, i64 %v32
%v39 = load [2 x double], ptr %v38, align 8
%v40 = alloca [2 x double], align 8
store [2 x double] %v39, ptr %v40, align 8
%v41 = getelementptr inbounds [2 x double], ptr %v40, i32 0, i64 %v33
%v42 = load double, ptr %v41, align 8
```
## Expected behavior

The read should stay address-based until the final scalar load, equivalent to:
```llvm
%row_ptr = getelementptr inbounds [4 x [2 x double]], ptr %scratch, i32 0, i64 %row
%val_ptr = getelementptr inbounds [2 x double], ptr %row_ptr, i32 0, i64 %col
%value = load double, ptr %val_ptr
```
No whole `[4 x [2 x double]]` load/store and no temporary aggregate copy should be needed.

## Why this matters

For small arrays this is mostly noisy IR. For larger scratch arrays it becomes a major resource issue. In a real kernel with a `[[f64; 6]; 1024]` local scratch buffer, this pattern creates a second ~49 KiB stack object and a large amount of local-memory traffic.

This looks like it may be a place/value lowering issue: ordinary place reads like `scratch[row][col]` appear to be going through a value-projection path that materializes aggregates. They should use the existing place-address machinery and emit one final scalar load.

## Repro

Add the kernel above as a small cuda-oxide example and build with NVVM IR output:

```bash
cargo oxide build <example-name> --emit-nvvm-ir --arch sm_100
```

Then inspect the generated .ll file:

```bash
grep -n "load \[4 x \[2 x double\]\]\|alloca \[4 x \[2 x double\]\]\|store \[4 x \[2 x double\]\]" <example-name>.ll
```

## Acceptance criteria

- Projected reads from addressable local arrays lower to address computation plus a final scalar load.
- The repro no longer emits a whole-array load `[4 x [2 x double]]` followed by a temporary whole-array alloca/store.
- The nested projection also avoids materializing an intermediate `[2 x double]` row copy before the final scalar load.
- Nested indexing such as `scratch[row][col]` still produces correct runtime results.

pattersonjack:
## What

Closes #231.

Projected reads like `scratch[row][col]` from an addressable local array were lowered by materializing values at each projection step. For a nested array, that meant loading/copying the whole `[4 x [2 x double]]` local, indexing into the copy, then loading/copying the `[2 x double]` row before the final scalar read.

## Fix

This PR adds a narrow, staged fix: an eligibility check for whether the projected place is representable as a real address, followed by the existing `translate_place_address` walker and one final `MirLoadOp`.

The eligibility check is separate from the address walker because the walker emits IR as it goes. Unsupported projections still fall back to the existing value-projection lowering.

This leaves the broader model refactor for a follow-up PR, where place-as-address can become the default for `Copy`/`Move`, value-projection can become the explicit fallback, and fallback cases such as unsupported enum/downcast projections, ZSTs, and computed aggregates with no addressable backing slot can be covered directly.

## Test plan

- [x] `cargo test --workspace`
- [x] `cargo oxide run projected_place_read`
- [x] `cargo oxide run ref_index_projections`
- [x] `cargo oxide build projected_place_read --emit-nvvm-ir --arch sm_100`


nihalpasham:
Thanks again for the PR. 

The design looks good i.e. dry-run eligibility check (can't call the address walker speculatively since it emits IR as it goes) along with the hard-error on checker/walker divergence. 

A few quick notes:

- Labels added on our end (bug, miscompile, codegen, IR-lowering).
- projected_place_read is auto-discovered by smoketest, no manual registration needed.
- is_mutable: false throughout the eligibility check is correct for reads; the actual walker inherits mutability from the alloca.

**Just a quick follow-up question:** if you plan to send the holistic refactor as a separate PR? The piece that makes `place-as-address` the default for all `Copy/Move` and demotes `value-projection` to the explicit fallback, with edge-case tests for enum downcast, ZST reads, and computed aggregates with no backing slot, it'd be great if you can drop a note in the discord chat, just so that folks know and don't end up working on the same thing.

PR #250, opened 2026-06-18
Preserve async free metadata across DeviceBuffer::cast_elem

fallintoplace:
## Summary

- preserve `DeviceBuffer` deallocation-stream metadata when `cast_elem` reinterprets the element type
- keep the public `into_raw_parts` API unchanged, with an internal all-parts path for ownership-preserving conversions
- add a regression test for async allocation + `cast_elem` + implicit drop while stream work is still in flight

## Why

`cast_elem` consumed a buffer with `into_raw_parts` and rebuilt it through `from_raw_parts`, which always marks the allocation as synchronous. For buffers allocated by `cuMemAllocAsync`, that stripped the stored deallocation stream and could make the reinterpreted buffer drop through `cuMemFree` instead of stream-ordered `cuMemFreeAsync`.

## Validation

- `cargo fmt --all --check`
- `git diff --check`


nihalpasham:
Thanks for the PR. LGTM 👍🏾

Reviewer note: compute-sanitizer is clean.

PR #251, opened 2026-06-18
Make borrowed host-to-device copies stream-synchronized

fallintoplace:
## Summary

- make safe borrowed host-to-device copies synchronize the stream before returning
- add explicit unsafe unchecked H2D async helpers for callers that manage source lifetimes themselves
- keep pinned H2D overlap available through the existing unsafe pinned APIs
- add pinned-source regression coverage showing safe H2D calls allow immediate source drop/reuse after return

## Why

The previous safe H2D helpers enqueued `cuMemcpyHtoDAsync` from a borrowed host slice and returned while CUDA could still be reading that slice. That made safe Rust responsible for keeping the source alive and unmodified after the borrow ended, which is not a sound safe API contract. This is especially concrete for pinned host memory, where CUDA H2D async transfers can run after the call returns.

This change keeps safe borrowed-source APIs conservative and moves true async borrowed-source H2D copies behind explicit unsafe contracts.

## Validation

- `cargo fmt --all --check`
- `git diff --check`


nihalpasham:
Thanks for the PR. LGTM 👍🏾

Reviewer note: both regression tests pass, and compute-sanitizer is clean on the drop-after-return and reuse-after-return cases.

PR #252, opened 2026-06-18
Check DeviceBuffer allocation byte sizes

fallintoplace:
## Summary

- add checked byte-size calculation for `DeviceBuffer` allocations
- store the validated byte size in `DeviceBuffer` so later transfers do not recompute a potentially overflowing length
- reject oversized `zeroed` and `uninitialized_async` allocations with `CUDA_ERROR_INVALID_VALUE`
- add regression coverage for `usize` byte-size overflow

## Why

`DeviceBuffer` previously computed byte sizes with unchecked `usize` multiplication. In release builds, an overflowing `len * size_of::<T>()` can wrap to a small allocation while the buffer still carries the huge logical element count. Later copies, memset, or kernel launches can then treat a tiny allocation as much larger.

The pinned host buffer already used checked multiplication for this case. This brings `DeviceBuffer` to the same standard and keeps the validated byte extent with the owner.

## Validation

- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/crates/cuda-core/src/device_buffer.rs
+++ b/crates/cuda-core/src/device_buffer.rs
@@ -131,6 +131,7 @@ unsafe impl DeviceCopy for half::f16 {}
 pub struct DeviceBuffer<T> {
     ptr: CUdeviceptr,
     len: usize,
+    num_bytes: usize,
     ctx: Arc<CudaContext>,
     /// When the allocation came from the stream-ordered pool
     /// (`cuMemAllocAsync`), this holds an `Arc` to the owning stream so the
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks for the PR.

A quick point: layered 1 addition as PR #250 (already merged) had restructured `from_raw_parts` to delegate to a private `from_raw_parts_with_dealloc_stream` - so the body had changed and git couldn't auto-merge.

- so, moved the `allocation_size` computation one level down into `from_raw_parts_with_dealloc_stream`, where the struct literal is actually constructed. Both PRs' intent is fully preserved 

Reviewer notes:
- #252's overflow check fires on every construction path (sync and async), and 
- #250's `cast_elem` fix still threads the `dealloc` stream through correctly.

Issue #96, opened 2026-05-27
Tracking: Tile-to-SIMT Interop (cuda-oxide as SIMT participant)


mohamedsamirx:
> **This is a tracking issue.** It coordinates work for cuda-oxide to be the **Rust SIMT participant** in CUDA's Tile-to-SIMT interop story. The original proposal from @mohamedsamirx is preserved at the bottom.

## Background

cuda-oxide is the Rust SIMT compiler. The tile programming model itself belongs to separate tile frontends (cuTile Python, CUDA Tile C++, future Rust tile DSLs). cuda-oxide's role here is to produce SIMT artifacts (PTX, LTOIR) that those frontends can launch alongside, or eventually link into, tile kernels.

Tile-to-SIMT interop comes in two flavors (per the public [cuTile Python Interoperability page](https://docs.nvidia.com/cuda/cutile-python/interoperability.html)):

- **Inter-kernel** — two separate kernels (one tile, one SIMT) launched on the same CUDA stream, sharing arrays. **Supported in the public cuTile Python release today.**
- **Intra-kernel** — a tile kernel calling a SIMT device function from inside its own body. Upstream future work; public tracker: [NVIDIA/cutile-python#69](https://github.com/NVIDIA/cutile-python/issues/69).

### Phase 0 — Inter-kernel example (today)

- [x] Minimal `cutile_inter_kernel` example combining a cuda-oxide-built SIMT kernel and a cuTile Python tile kernel sharing a CUDA stream and an array.
- [x] README walking through the build and run steps.
- [x] Smoketest coverage goes through the normal example path: `scripts/smoketest.sh -o "^cutile_inter_kernel$"` runs cargo oxide run `cutile_inter_kernel`, which builds the nested cuda-oxide SIMT PTX crate and then runs the cutile-rs host pipeline.

### Phase 1 — Tile kernels calling Rust SIMT code (gated)

Gated on [NVIDIA/cutile-python#69](https://github.com/NVIDIA/cutile-python/issues/69) exposing a public way for a tile kernel to call SIMT device functions from inside its own body. The exact shape upstream takes (a pre-compiled artifact, a plug-in library, or both) will determine the cuda-oxide side.

When the upstream surface is available:

- [ ] A one-shot `cargo oxide emit-ltoir <crate>` command for producing the SIMT artifact in a single step. Useful regardless of which mechanism lands; can ship sooner.
[...]

PR #256, opened 2026-06-19
feat(cargo-oxide): add `emit-ltoir` to build a crate's LTOIR in one step

alejandro-soto-franco:
## What

Adds `cargo oxide emit-ltoir <crate>`, which builds a crate's device code to a
binary **LTOIR** artifact in one step.

`cargo oxide build --emit-nvvm-ir` stops at NVVM IR, leaving the caller to run
libNVVM separately to get linkable LTOIR. `emit-ltoir` folds both halves into a
single command for the **Tile-to-SIMT interop** workflow (#96), where cuda-oxide
is the SIMT participant: it builds the crate in NVVM IR mode, compiles the
emitted `<crate>.ll` with libNVVM `-gen-lto`, and writes `<crate>.ltoir`.

## Details

- `--arch` is **required** (LTOIR is architecture-specific) and accepts `sm_XX`,
  `compute_XX`, or a bare `XX`. `-o/--output` overrides the default
  `<crate>.ltoir` path.
- Reuses the existing `libnvvm-sys` binding and the established NVVM IR build
  path (`codegen_build(.., emit_nvvm_ir=true, ..)`); rejects metadata-interop
  examples (point it at a single SIMT device crate instead).
- cuda-oxide exports NVVM IR 2.0 (opaque pointers), which libNVVM only accepts
  for `compute_100`+ (Blackwell). Below that, libNVVM rejects the module with a
  bare `parse expected type` and no hint at the cause (#98). On a libNVVM
  failure below `compute_100`, `emit-ltoir` appends a short explanation pointing
  at #98 and suggesting `sm_100`+. **The hint is gated on actual failure**, so it
  disappears automatically if the opaque-pointer floor ever moves (e.g. once
  typed-pointer export lands).

## Tests

- Unit tests for the arch normalization (`nvvm_compute_arch`) and the
  capability parse (`compute_capability`).
- Documents the command and the `compute_100` floor in the cargo-oxide README.

`cargo test -p cargo-oxide` (26 passed), `cargo fmt --check`, and
`cargo clippy -p cargo-oxide` are all clean.

Part of #96.

### DIFF
--- a/crates/cargo-oxide/README.md
+++ b/crates/cargo-oxide/README.md
@@ -37,19 +37,21 @@ cargo oxide setup                   # explicitly build the codegen backend
 
 ### Flags
 
-| Flag              | Applies to             | Description                                     |
-|-------------------|------------------------|-------------------------------------------------|
-| `--emit-nvvm-ir`  | run, build, pipeline   | Generate NVVM IR for libNVVM                    |
-| `--arch <sm_XX>`  | run, build, pipeline   | Target architecture override                    |
-| `--arch <sm_XX>`  | emit-ltoir             | Target architecture (required)                  |
-| `--features <F>`  | run, build, emit-ltoir | Comma-separated cargo features to enable        |
-| `-o, --output <P>`| emit-ltoir             | Output path for the `.ltoir` artifact           |
-| `-v, --verbose`   | run, build, emit-ltoir | Show detailed compilation output                |
-| `--no-fmad`       | run, build, pipeline   | Disable FMA contraction (keep separate mul+add) |
-| `--async`         | new                    | Use the async template                          |
-| `--cgdb`          | debug                  | Use cgdb instead of cuda-gdb                    |
-| `--tui`           | debug                  | Use GDB's TUI interface                         |
-| `--check`         | fmt                    | Check formatting only                           |
+| Flag               | Applies to                       | Description                                     |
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks for the PR. LGTM 👍🏾

Reviewer note: three fixes layered on our side before merge.

1. **libdevice omission (correctness)**: `compile_nvvm_to_ltoir` was adding only the kernel module to the libNVVM program. Any kernel that calls libdevice math (`expf`, `sinf`, etc.) would produce LTOIR with unresolved `__nv_*` symbols. Fixed by adding `libdevice.10.bc` first, matching the pattern in `cuda-host/src/ltoir.rs`.

2. **Double success message**: `codegen_build` printed "✓ Build succeeded" before `emit_ltoir` printed its own "✓ LTOIR written to ...". Fixed by adding a `quiet: bool` parameter to `codegen_build`; `emit_ltoir` passes `true` so only the unified LTOIR summary is shown.

3. **README flags table**: the duplicate `--arch` row was merged into one row with "Applies to: run, build, pipeline, emit-ltoir". A one-line note below the table clarifies that `--arch` is required for `emit-ltoir` only.

Partially addresses #96 (the Phase 1 emit-ltoir checklist item). Related to #98 (the helpful error hint is wired up; the typed-pointer export root cause remains a separate workstream).

PR #257, opened 2026-06-19
fix(cargo-oxide): rebuild cached backend when its source advances

alejandro-soto-franco:
## What

Out-of-tree consumers (`cargo install`ed `cargo-oxide`) keep a cached backend at
`~/.cargo/cuda-oxide/`: the built `librustc_codegen_cuda.so` plus a `src/` clone
it was built from. The staleness check only compared the cached `.so` against the
`cargo-oxide` **binary** mtime (issue #49). If a developer advanced the backend
**source** in that cached tree but left the binary unchanged, the cached `.so` was
silently reused, so the consumer kept running a stale backend.

## How

`cached_backend_status` now classifies the cache three ways:

- `Fresh` -> reuse the cached `.so`.
- `StaleVsBinary` -> the running binary postdates the cache (the existing issue #49
  signal). Recovery is unchanged: drop the cached source and re-clone fresh, since
  the cached source may no longer match the upgraded binary.
- `StaleVsSource` -> the newest backend source input (the crate's `src/**` and
  `Cargo.toml` under the cached `src/crates/rustc-codegen-cuda`) postdates the
  cached `.so`. Recovery rebuilds the `.so` from that existing cached source in
  place (the auto-fetch step already skips the clone when the source tree is
  present), so re-cloning never discards the source that triggered the rebuild.

Binary staleness takes precedence when both fire. When the source tree can't be
located, the check degrades gracefully to the existing binary-only behaviour.

## Tests

The source-mtime helper is pure and unit-tested (7 backend tests, all pass):
source newer than cache -> `StaleVsSource`, source older -> `Fresh`, absent source
tree falls back to the binary check, and binary staleness wins over source
staleness when both apply.

`cargo test -p cargo-oxide` (28 passed), `cargo fmt --check`, and
`cargo clippy -p cargo-oxide` are all clean.

### DIFF
--- a/crates/cargo-oxide/src/backend.rs
+++ b/crates/cargo-oxide/src/backend.rs
@@ -205,7 +205,7 @@ fn cached_backend_status(cached_so: &Path, source_dir: Option<&Path>) -> CacheSt
 /// `source_dir`: every file in `src/**` plus the crate `Cargo.toml`.
 ///
 /// Returns `None` when the directory cannot be located or yields no
-/// readable inputs, which lets [`cached_backend_is_stale`] degrade to the
+/// readable inputs, which lets [`cached_backend_status`] degrade to the
 /// binary-only check. The walk is best-effort: unreadable entries are
 /// skipped rather than treated as failures.
 fn newest_backend_source_mtime(source_dir: &Path) -> Option<SystemTime> {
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. LGTM 👍🏾

Reviewer note: I found one dead `rustdoc` intra-doc link (`[\`cached_backend_is_stale\`]` -> `[\`cached_backend_status\`]` in the `newest_backend_source_mtime` doc comment). All 28 cargo-oxide unit tests pass. No GPU verification needed; this is a pure build-orchestration change.

PR #264, opened 2026-06-22
feat: special registers (warpid, smid, gridid, lanemask)

### LINKED ISSUE #49: stale librustc_codegen_cuda.so may cause problems after updating cuda-oxide

honeyspoon:
## Summary

- Add 10 missing PTX special register intrinsics across all four compiler layers (cuda-device stubs, dialect-nvvm ops, mir-importer matching, mir-lower LLVM conversion)
- Warp-level: `warpid`, `nwarpid`, `lanemask_eq`, `lanemask_le`, `lanemask_lt`, `lanemask_ge`, `lanemask_gt`
- SM/grid-level: `smid`, `nsmid`, `gridid`
- All follow the existing zero-operand, single-i32-result LLVM NVVM intrinsic pattern (e.g. `llvm.nvvm.read.ptx.sreg.warpid`)

Closes #49, ref #27.

## Test plan

- [x] `cargo check -p cuda-device` passes
- [x] `cargo check -p dialect-nvvm` passes
- [x] `cargo check -p mir-importer` passes
- [x] `cargo check -p mir-lower` passes (pre-existing MmaM16N8K* errors unrelated)
- [ ] Verify each register emits correct LLVM intrinsic in a test kernel
- [ ] End-to-end test reading `smid`/`warpid` on a real GPU


honeyspoon:
Closing: PR was created against wrong repo. Recreating on honeyspoon/cuda-oxide.

PR #268, opened 2026-06-22
feat: movmatrix in-register matrix transpose

honeyspoon:
## Summary
- Adds `movmatrix.sync.aligned.m8n8.trans.b16` support across all compiler layers (cuda-device stub, dialect-nvvm op, mir-importer, mir-lower)
- New `wmma` module in each layer for warp-level matrix operations
- Each thread provides a `u32` (2 packed b16 elements); the instruction transposes the 8×8 tile distributed across the warp, avoiding a shared-memory round-trip

Closes #43
Ref #27

## Test plan
- [x] `cargo check -p cuda-device -p dialect-nvvm -p mir-importer -p mir-lower` passes
- [x] `test_movmatrix_trans_b16_lowers_to_inline_asm` verifies lowering to convergent inline PTX
- [x] All 10 existing lowering tests continue to pass

### DIFF
--- a/.github/workflows/fork-ci.yml
+++ b/.github/workflows/fork-ci.yml
@@ -1,0 +1,82 @@
+---
+name: fork-ci
+
+# Lightweight CI for our PR branches. Catches formatting, clippy, rustdoc, and
+# unit-test failures in the packages we modify (no CUDA toolkit needed).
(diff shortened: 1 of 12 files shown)


honeyspoon:
Created on wrong repo by mistake, re-creating on honeyspoon/cuda-oxide fork.

Issue #274, opened 2026-06-22
Tracking: Ampere SM80+ device intrinsics


honeyspoon:
## What problem does this solve?

cuda-oxide supports several SM 80+ intrinsics on `main` (`redux.sync`, `ldmatrix` x2/x4, `stmatrix`, `cp.async` ca/cg 16, `cvt.rn.f16x2.f32`, `fma.rn.bf16x2`, `mbarrier`, `elect.sync`, lanemask registers), but the Ampere ISA has substantially more intrinsics important for inference and training workloads: tensor core MMA, packed arithmetic, integer dot products, and additional conversion/memory variants. This tracks filling those gaps.

## Status update (July 2026)

**Upstream PR #406 merged**: All CUDA intrinsic APIs and lowering are now autogenerated from pinned LLVM 22 data. The generator admitted 781 intrinsic contracts with stable IDs. This supersedes all hand-written intrinsic PRs below — the unchecked items are now covered by the autogen system and no longer need individual PRs.

Superseded PRs: #332 (f16x2), #333 (prmt), #334 (mixed INT8 MMA), and fork PRs for dp2a.hi, cp.async pipeline, sparse MMA, stmatrix x1.

**To add a missing intrinsic**: append to `intrinsics/overlay/*.toml` + `intrinsics/abi-v1.toml` and regenerate. See #406 description for the full contributor workflow.

## Scope

Each item follows the established 4-layer pattern (`cuda-device` > `dialect-nvvm` > `mir-importer` > `mir-lower`) plus lowering tests. Inline asm ops use the `AsmKind` enum (PR #195). Multi-variant families use parameterized dispatch helpers.

### Phase 1: Self-contained extensions of existing features
- [x] Integer dot products: `dp4a.{s32,u32}`, `dp2a.{s32,u32}` (sm_61+), #275 + e2e example #282
- [x] Packed conversions: `cvt.rz.f16x2.f32`, `cvt.rn.relu.{f16x2,bf16x2}.f32`, `cvt.rz.bf16x2.f32`, #276 + e2e example #281
- [x] bf16x2 arithmetic: `add/sub/mul.rn.bf16x2`, `min/max.bf16x2`, `neg/abs.bf16x2`, `fma.rn.relu.bf16x2`, #279 (includes e2e example)
- [x] cp.async small sizes: `cp.async.ca` 4-byte and 8-byte, #280 (includes e2e example)
[...]

PR #275, opened 2026-06-22
feat: add integer dot product instructions (dp4a, dp2a) for sm_61+

honeyspoon:
Add dp4a (4-element byte dot product) and dp2a (2-element half-word dot product) intrinsics in signed and unsigned variants, lowered to pure inline PTX assembly.

## Summary

Implements the PTX `dp4a` and `dp2a` integer dot-product-with-accumulate family across the full compiler pipeline. These single-thread, non-convergent instructions perform packed-integer multiply-accumulate in a single cycle and are the building blocks for INT8/INT4 quantised inference on sm_61+ GPUs (Pascal and later).

Each operation takes three 32-bit inputs `(a, b, c)` and produces one 32-bit output `d = c + dot(a, b)`. For `dp4a`, `a` and `b` are treated as 4 packed bytes; for `dp2a`, `a` is 2 packed 16-bit values and the lower 2 bytes of `b` are used. Both come in signed (`s32`) and unsigned (`u32`) flavors. The lowering uses pure inline PTX assembly (no LLVM intrinsic exists for these), tagged with `AsmKind::Pure` (non-convergent, no side effects), matching the pattern used by `fma_bf16x2`.

| Device API | PTX Instruction | Semantics |
|---|---|---|
| `dp4a_s32(a, b, c)` | `dp4a.s32.s32 %d, %a, %b, %c;` | 4x signed byte dot product + i32 accumulate |
| `dp4a_u32(a, b, c)` | `dp4a.u32.u32 %d, %a, %b, %c;` | 4x unsigned byte dot product + u32 accumulate |
| `dp2a_s32(a, b, c)` | `dp2a.lo.s32.s32 %d, %a, %b, %c;` | 2x signed half-word × byte dot product + i32 accumulate |
| `dp2a_u32(a, b, c)` | `dp2a.lo.u32.u32 %d, %a, %b, %c;` | 2x unsigned half-word × byte dot product + u32 accumulate |

Part of #274

## Changes

- **cuda-device**: 4 new device-side stubs in `dotprod` module (`dp4a_s32`, `dp4a_u32`, `dp2a_s32`, `dp2a_u32`), each `#[inline(never)]` with `unreachable!` host-side fallback.
- **dialect-nvvm**: 4 new pliron ops (`Dp4aS32Op`, `Dp4aU32Op`, `Dp2aS32Op`, `Dp2aU32Op`), each with 3 operands (a, b, c) and 1 result, registered in the NVVM dialect.
- **mir-importer**: 4 `emit_dp*` translation functions dispatched from `try_dispatch_intrinsic` for the `cuda_device::dotprod::*` call paths.
- **mir-lower**: 4 `convert_dp*` functions lowering NVVM ops to `InlineAsmOp` with `AsmKind::Pure`; 4 `MirToLlvmConversion` trait impls wiring them up.
- **llvm-export**: No changes needed, inline asm export path already handles `InlineAsmOp`.
- **tests**: 4 new lowering tests in `mir-lower/tests/lowering_test.rs` verifying each op produces the expected PTX mnemonic in the inline asm output.

## Testing

- [x] `cargo test -p mir-lower --test lowering_test` passes (4 new + 9 existing = 13/13)
- [x] `cargo check -p dialect-nvvm -p mir-lower` passes
- [x] `cargo fmt --check` and `cargo clippy --workspace -- -D warnings` clean
- [ ] E2E example deferred (dp4a/dp2a requires sm_61+ runtime; will follow up)

## Checklist

- [x] No files under `cuda-bindings/` modified


nihalpasham:
Thanks for the PR. LGTM 👍🏾

- I'm glad you caught the result signedness: `Signed` for s32, `Unsigned` for u32 is what the `MirStoreOp` verifier needs and copying the always-unsigned bf16x2 helper would have failed the store typecheck.

Rebased onto the TypeHandle-API main and layered a small commit:
- made the four `dp4a`/`dp2a` device stubs safe `pub fn` (were `unsafe`). They're pure register ops with no memory/pointer contract, so they now match the safe-fn convention of `fma_bf16x2`/`cvt_f16x2_f32` and drop the needless `# Safety` docs + unsafe blocks at call sites.

**Reviewer note:** no GPU e2e in this env, so ptxas acceptance + numerics on sm_61+ still want a follow-up example. Ticked the dp4a/dp2a box in #274; not closing the tracking issue.

PR #280, opened 2026-06-24
feat: add cp.async 4-byte and 8-byte copy intrinsics (sm_80+)

honeyspoon:
Add cp.async.ca variants for 4-byte and 8-byte asynchronous global-to-shared memory copies, extending the existing 16-byte cp.async.ca/cg support.

## Summary

The existing `cp.async` support covers 16-byte cache-all (`ca`) and cache-global (`cg`) copies. This PR adds the smaller 4-byte and 8-byte `cp.async.ca` variants, which are needed when copying sub-cacheline data (e.g., packed scalars, individual matrix fragments) in asynchronous pipelines.

These are lowered to inline PTX assembly with `AsmKind::SideEffect` (memory write to shared memory, non-convergent). A new `inline_asm_sideeffect()` helper is added to `common.rs` alongside the existing `inline_asm_convergent()`, providing the correct non-convergent + side-effect combination for memory operations that are not warp-synchronous.

| Device API | PTX Instruction | Copy size |
|---|---|---|
| `cp_async_ca_4(dst, src)` | `cp.async.ca.shared.global [%0], [%1], 4;` | 4 bytes |
| `cp_async_ca_8(dst, src)` | `cp.async.ca.shared.global [%0], [%1], 8;` | 8 bytes |

Part of #274

## Changes

- **cuda-device**: 2 new `pub unsafe fn` stubs in `async_copy.rs` (raw pointer operations require unsafe).
- **dialect-nvvm**: 2 new NVVM ops (`CpAsyncCa4Op`, `CpAsyncCa8Op`) with 2 operands (dst shared ptr, src global ptr) and 0 results.
- **mir-importer**: 2 `emit_cp_async_ca_*` translation functions with dispatch arms.
- **mir-lower**: 2 `convert_cp_async_ca_*` lowering functions using new `inline_asm_sideeffect()` helper; 2 `MirToLlvmConversion` impls.
- **mir-lower/common.rs**: New `inline_asm_sideeffect()` helper for non-convergent side-effect inline asm (distinct from the existing `inline_asm_convergent()` which is for warp-synchronous ops).
- **tests**: 2 lowering tests verifying the PTX output.
- **example**: `cp_async_small` end-to-end GPU example testing 4-byte and 8-byte async copies via shared memory.

## Testing

- [x] `cargo test -p mir-lower --test lowering_test` passes
- [x] `cargo fmt --check` and `cargo clippy` clean
- [x] E2E example included (`cp_async_small`)

## Checklist

- [x] No files under `cuda-bindings/` modified


nihalpasham:
Thank you for adding the missing 4-byte and 8-byte async-copy forms. They are useful building blocks for kernels that cannot naturally move a full 16 bytes per lane.

I added a few safety and integration details on the branch:

- added the `sm_80` target floor and explicit shared/global address conversions;
- fixed the example's Rust 2024 static access and 8-byte shared alignment;
- completed the unsafe contract, compiler-ordered waits, exact lowering checks, and standalone lockfile.

Both copy widths now compile to the expected `sm_80` PTX and are accepted by `ptxas`.

PR #276, opened 2026-06-22
feat: add packed cvt variants (rz, relu) for f16x2 and bf16x2

honeyspoon:
Add four packed type-conversion intrinsics completing the f16x2/bf16x2 conversion family alongside the existing cvt.rn variants.

## Summary

The existing `cvt_f16x2_f32` converts two f32 values into a packed f16x2 using round-to-nearest-even (`.rn`). This PR adds the four remaining packed conversion variants that are important for inference workloads on sm_80+ hardware:

- **`cvt.rz.f16x2.f32`** and **`cvt.rz.bf16x2.f32`**: Round-toward-zero (truncation) instead of round-to-nearest-even. Truncation rounding is preferred in quantization-aware inference pipelines where deterministic floor-toward-zero behavior is needed, avoiding the slight upward bias of round-to-nearest-even on half-boundary values.

- **`cvt.rn.relu.f16x2.f32`** and **`cvt.rn.relu.bf16x2.f32`**: Fuse a ReLU clamp (negative values become zero) into the conversion in a single instruction. Fusing ReLU eliminates a separate `max(x, 0)` step, saving one instruction per element pair in activation layers.

All four instructions pack two f32 inputs into a single u32 (f16x2 or bf16x2), matching the existing `cvt_f16x2_f32` calling convention `(lo: f32, hi: f32) -> u32`.

| Device API | PTX Instruction | Description |
|---|---|---|
| `cvt_rz_f16x2_f32(lo, hi)` | `cvt.rz.f16x2.f32 d, hi, lo;` | f32 pair to packed f16x2, round-toward-zero |
| `cvt_rn_relu_f16x2_f32(lo, hi)` | `cvt.rn.relu.f16x2.f32 d, hi, lo;` | f32 pair to packed f16x2, round-to-nearest + ReLU |
| `cvt_rn_relu_bf16x2_f32(lo, hi)` | `cvt.rn.relu.bf16x2.f32 d, hi, lo;` | f32 pair to packed bf16x2, round-to-nearest + ReLU |
| `cvt_rz_bf16x2_f32(lo, hi)` | `cvt.rz.bf16x2.f32 d, hi, lo;` | f32 pair to packed bf16x2, round-toward-zero |

Part of #274

## Changes

[...]

PR #281, opened 2026-06-24
feat: add packed cvt variants end-to-end example (sm_80+)

honeyspoon:
## Summary

Follow-up to #276 -- adds the sm_80+ end-to-end GPU example that was noted as still owed.

- Adds `crates/rustc-codegen-cuda/examples/cvt_packed/` with a single kernel that exercises all five packed cvt intrinsics (`cvt_f16x2_f32`, `cvt_rz_f16x2_f32`, `cvt_rn_relu_f16x2_f32`, `cvt_rn_relu_bf16x2_f32`, `cvt_rz_bf16x2_f32`) on the same `(3.14, -2.5)` inputs
- Host unpacks each u32 result via manual f16/bf16 bit decoding and verifies:
  - **rn** (round-to-nearest): both lanes approximate the input values
  - **rz** (round-toward-zero): both lanes approximate inputs with `|result| <= |input|`
  - **relu** variants: positive lane preserved, negative lane clamped to `0.0`
- Includes sm_80+ guard with early-return on older GPUs
- Follows the same structure as the existing `cvt_f16x2` example

Part of #274

## Test plan

- [x] `cargo oxide run cvt_packed` on an sm_80+ (Ampere or later) GPU
- [x] Verify all 5 variants print correct values and the example exits with `SUCCESS`
- [x] Verify the example gracefully skips on sm_7x and below


nihalpasham:
Thank you for following through with the packed-conversion GPU example. Keeping all five variants in one kernel makes their lane behavior easy to compare.

I added a few small finishing touches on the branch:

- made the expected packed values bit-exact so nearest and toward-zero rounding cannot pass for one another;
- added the missing canonical-NaN check for the ReLU variants;
- added the standalone lockfile and cleaned up the example build.

The corrected `sm_80` output contains all five expected instructions and is accepted by `ptxas`.

PR #282, opened 2026-06-24
feat: add dp4a/dp2a end-to-end GPU example (sm_61+)

honeyspoon:
## Summary

Follow-up to #275; adds the GPU-verified example the maintainer noted as still owed.

- Single-thread kernel exercises all four dot-product intrinsics: `dp4a_s32`, `dp4a_u32`, `dp2a_s32`, `dp2a_u32`
- Host packs known byte/half-word values, launches, and asserts exact results
- Graceful skip on `sm < 61` (dp4a/dp2a require Pascal+)

## Test plan

- [x] `cargo oxide run dotprod` on sm_61+ GPU -- all four results match expected values
- [x] Verify PTX contains `dp4a.s32.s32`, `dp4a.u32.u32`, `dp2a.lo.s32.s32`, `dp2a.lo.u32.u32` instructions

Part of #274


nihalpasham:
Thank you for adding the promised end-to-end dot-product example. The operand choices and worked calculations make the signed and unsigned behavior easy to audit.

I added the standalone lockfile and made the `sm_61` target explicit in the run command. The example compiles cleanly, emits all four expected PTX instructions, and its calculations match the instruction semantics.

PR #283, opened 2026-06-24
Harden LTOIR cubin cache provenance

DaronPopov:
## Summary

Adds a persistent, content-addressed native cubin cache for file-backed NVVM IR and LTOIR. Repeated loads reuse a previously compiled GPU binary only when the complete compiler inputs and selected CUDA tools match.

Ordinary PTX, direct cubin loading, embedded artifacts, and the pre-Blackwell-to-Blackwell PTX bridge keep their existing paths.

## What changed

- **Complete cache provenance**
  - Keys include the exact NVVM IR or LTOIR bytes, target architecture, module names, ordered compiler/linker options, and libdevice bytes.
  - On Linux, the concrete libNVVM and nvJitLink files selected for the first cached build are fingerprinted from retained file descriptors; their handles are then pinned for the process lifetime.
  - Unknown or unstable CUDA-tool identity disables reuse and performs a fresh build.
  - SHA-256 is provided by the maintained RustCrypto `sha2` crate; there is no local hash or cryptography implementation.
- **Atomic, verified entries**
  - Stores entries below `.oxide-artifacts/ltoir-cubin-cache/v1` using a per-key lock.
  - Builds in a private temporary directory, flushes complete files, and publishes the directory atomically.
  - Binds the input key, cubin length/hash, and optional LTOIR length/hash in one binary manifest.
  - Rejects malformed ELF files, truncation, corruption, swapped entries, partial publication, and symlinked artifacts before reuse.
- **Artifact-pipeline integration**
  - Caches file-backed native-cubin routes. Normal artifact loading uses them only when the emitted and execution architectures match.
  - Preserves current target parsing, artifact precedence, NVVM dialect checks, whole-program verification, libdevice ordering, and Blackwell PTX bridging.
  - Loads verified owned cubin bytes instead of reopening a cache path after validation.
- **Compatibility and cleanup**
  - `build_cubin_from_ll` still writes and returns the normal sibling `<name>.cubin`, and continues to emit `<name>.ltoir`.
  - Normal compiler cleanup removes stale top-level artifacts while preserving the private cache.
  - Cache setup or publication failures fall back to the freshly built bytes; compiler or linker failures still propagate.

### Cache behavior coverage

- [x] File-backed NVVM IR to native cubin
- [x] File-backed standalone LTOIR to native cubin
- [x] Complete source, target, option, libdevice, module-name, and CUDA-tool invalidation
- [x] One build for concurrent requests for the same key
- [x] Safe repair of corrupt, truncated, swapped, partial, or symlinked entries
- [x] Ordinary PTX and cross-architecture PTX routes bypass the cache
- [x] Unknown CUDA-tool identity rebuilds without publishing a reusable entry

## Safety model

```text
Exact inputs and tools -> SHA-256 key -> per-key lock
                       -> verified atomic entry -> owned cubin bytes
```

The first cache compiler and linker handles, together with the descriptors used for their cache keys, are retained for the process lifetime. Cache reuse assumes CUDA toolkit files are not changed while the process is running; restart the process after changing or replacing a toolkit.

## Compiler and CI integration

- Adds the `sha2` dependency to `cuda-host` and updates the root plus all 124 affected standalone example lockfiles.
- Adds `libnvvm-sys` and `nvjitlink-sys` to the unit-test CI matrix.
- Adds focused cleanup tests in `cargo-oxide` and `mir-importer` so generated-file cleanup cannot erase persistent cache entries.
- Documents the cache boundary, invalidation model, location, and manual clearing behavior in `cuda-host`.

## Testing

- [x] Formatting and `git diff --check`
- [x] Clippy with `-D warnings` for every changed crate
- [x] Warning-free rustdoc for `cuda-host`, `libnvvm-sys`, and `nvjitlink-sys`
- [x] `cargo deny check`
- [x] Workspace all-target compilation with `cargo test --workspace --all-targets --no-run`
- [x] All non-GPU workspace library and binary tests
- [x] `cuda-host`: 56 passed, 1 live-toolkit test ignored by default
- [x] `libnvvm-sys`: 8 passed, 1 live-toolkit test ignored by default
- [x] `nvjitlink-sys`: 3 passed, 1 live-toolkit test ignored by default
- [x] `cuda-host` API integration tests: 4 passed
- [x] Representative standalone `--locked` checks: `vecadd`, `libdevice_math`, and `async_mlp`
- [x] GPU-less `cargo oxide build`: ordinary `sm_80` PTX and `sm_86` libdevice/NVVM IR
- [x] Live CUDA 13.0 and 13.3 build, verified hit, source-change miss, and standalone-LTOIR hit
- [ ] Execute a cached cubin on a GPU; no CUDA-capable device is available on the review machine

## Checklist

- [x] Maintainer integration commits include `Signed-off-by`
- [ ] The original contributor commit still needs the contributor's DCO sign-off
- [x] No files under `crates/cuda-bindings/` are modified
- [x] New Rust source files include SPDX headers
- [x] The contributor's original commit and authorship remain in the PR history


nihalpasham:
Thank you for starting the LTOIR cubin cache. I agree, avoiding repeated native-cubin compilation is the right direction, and the PR provided the starting point for the completed implementation.

I layered a maintainer-side rework onto the branch:
- kept ordinary PTX, embedded artifacts, and cross-architecture PTX loading on their existing paths;
- replaced the mutable cache with locked, atomic, content-addressed entries using the (well-vetted) sha2 crate;
- keyed reuse on the complete inputs and selected CUDA tools, with safe rebuilds for corruption, interruption, contention, or unknown tool identity;
- preserved the `sibling-cubin` API and added cleanup, invalidation, concurrency, and live-toolkit coverage.

The cache now builds and hits correctly with CUDA 13.0 and 13.3. 

**A note:** I am going merge this in as its been reworked but your original commit needed the contributor’s DCO sign-off.

PR #286, opened 2026-06-25
Derive DeviceCopy for zero-valid enums

### LINKED ISSUE #285: Derive DeviceCopy for enums with valid zero bit patterns
# Derive DeviceCopy for enums with valid zero bit patterns

## Problem

`#[derive(DeviceCopy)]` currently has no sound path for enum types. Some `repr`
enums and payload enums are plain device-copyable data, but deriving the trait
for enums is only sound when every field is `DeviceCopy` and the all-zero bit
pattern is valid for the concrete enum type.

## Why this matters

CUDA-facing APIs often need small tagged values to cross host/device memory
boundaries. Requiring manual unsafe impls for every valid enum makes the API
harder to use, while accepting zero-invalid enums would make `DeviceCopy`
unsound.

## Expected behavior

The derive macro should accept non-generic enums whose fields are `DeviceCopy`
and whose all-zero bit pattern is valid, and it should reject enums where
`mem::zeroed::<Enum>()` is invalid.

## Acceptance criteria

- Fieldless and payload enums with valid all-zero bit patterns can derive
  `DeviceCopy`.
- Enums with invalid all-zero bit patterns are rejected at compile time.
- Generic enums are rejected unless the implementation can prove validity for a
  concrete type.
- Trybuild coverage includes pass and fail cases for these rules.

niklebedenko:
Closes #285.

# Derive DeviceCopy for zero-valid enums

## Problem

`DeviceCopy` derive handled structs and unions, but it could not safely support
enums. The missing rule is that enum fields must be `DeviceCopy` and the enum's
all-zero bit pattern must be valid for the concrete type.

## Fix

The derive implementation now handles non-generic enums by reusing the existing
field-type checks for all variant fields and emitting a const-time
`mem::zeroed::<Enum>()` validity check. Rustc rejects expansions where the
zeroed enum value would be invalid, so the macro does not need to approximate
enum layout itself.

Generic enums are rejected for now because the macro cannot const-evaluate the
validity check for every possible instantiation.

## Diligence

The check is deliberately based on rustc's validity rules rather than variant
count or discriminant heuristics. The payload failure test uses a local
`DeviceCopy` wrapper whose zero bit pattern is invalid, so it isolates the
enum-validity rule instead of failing on a field trait bound first.

## Tests

- `cargo test -p cuda-core --test device_copy_derive -- --nocapture`
- `cargo test -p cuda-core --test device_copy_impls -- --nocapture`

Upstream red:

- The enum derive trybuild regression fails on `upstream/main`: zero-valid enum cases are rejected and the zero-invalid enum checks are absent.

### DIFF
--- a/crates/cuda-core/tests/derive_device_copy/fail_enum.stderr
+++ b/crates/cuda-core/tests/derive_device_copy/fail_enum.stderr
@@ -1,5 +1,7 @@
-error: `#[derive(DeviceCopy)]` cannot be applied to enums: `DeviceCopy` requires every bit pattern (including the all-zero pattern written by `DeviceBuffer::zeroed`) to be a valid value, but an enum's discriminant leaves most byte patterns invalid, so a zeroed or device-written buffer could materialize an out-of-range discriminant (undefined behavior). If you can guarantee every device-produced byte pattern is a valid variant, write `unsafe impl DeviceCopy for ... {}` by hand.
-  --> tests/derive_device_copy/fail_enum.rs:12:6
-   |
-12 | enum Tag {
-   |      ^^^
+error[E0080]: evaluation panicked: aborted execution: attempted to zero-initialize type `Tag`, which is invalid
+ --> tests/derive_device_copy/fail_enum.rs:8:23
+  |
+8 | #[derive(Copy, Clone, DeviceCopy)]
+  |                       ^^^^^^^^^^ evaluation of `_` failed here
+  |
+  = note: this error originates in the derive macro `DeviceCopy` (in Nightly builds, run with -Z macro-backtrace for more info)
--- a/crates/cuda-core/tests/derive_device_copy/fail_enum_generic.rs
+++ b/crates/cuda-core/tests/derive_device_copy/fail_enum_generic.rs
@@ -3,15 +3,12 @@
 
 use cuda_core::DeviceCopy;
 
-// This enum has no zero discriminant, so `DeviceBuffer::zeroed::<Tag>` would
-// create an invalid value.
 #[derive(Copy, Clone, DeviceCopy)]
-#[repr(u8)]
-enum Tag {
-    A = 1,
-    B = 2,
+enum Maybe<T> {
+    Empty,
+    Value(T),
 }
 
 fn main() {
-    let _ = Tag::A;
+    let _ = core::mem::size_of::<Maybe<u32>>();
 }
--- a/crates/cuda-core/tests/derive_device_copy/fail_enum_generic.stderr
+++ b/crates/cuda-core/tests/derive_device_copy/fail_enum_generic.stderr
@@ -1,0 +1,5 @@
+error: `#[derive(DeviceCopy)]` supports enums only when the enum is not generic, because rustc must be able to const-evaluate `mem::zeroed::<Enum>()` for the concrete enum type and reject enums whose all-zero bit pattern is invalid.
+ --> tests/derive_device_copy/fail_enum_generic.rs:7:6
+  |
+7 | enum Maybe<T> {
+  |      ^^^^^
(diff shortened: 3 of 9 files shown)


nihalpasham:
Thank you for taking on enum support and for making the zero-validity rule explicit. 

I do not think this PR can merge under the current `DeviceCopy` safety contract. `DeviceCopy` is used by safe readback APIs that turn arbitrary bytes written by the device into initialized Rust values. `mem::zeroed::<Enum>()` proves only that one bit pattern is valid. For example, 

```rust
#[repr(u8)] 
enum Tag { 
    Zero = 0, 
    One = 1 
}
```
passes this PR's check, but a kernel can write byte `2`; safe readback would then construct an invalid `Tag`, which is undefined behavior before user code can validate it. Payload enums have the same discriminant problem plus layout niches.

This is why cuda-oxide also excludes zero-valid types such as `bool` and `char`: zero validity is necessary for `DeviceBuffer::zeroed`, but every-bit validity is required for `DeviceCopy`. As its stands, Enum transfer needs a different API contract or an integer/wire representation that is validated before constructing the enum. Do you have a specific use-case in mind here?

niklebedenko:
Use case: Rust enums are cool! ;)

More specifically: I have some enum values which I'd like to pass to a kernel in a readonly way. This is to be able to select boundary conditions for a physics engine. A concrete example:

```rust
#[derive(Clone, Copy, DeviceCopy)]
enum BoundaryCondition<F>
where
    F: Fn(f32, f32, f32) -> f32,
{
    None,
    UniformDirichlet(f32),
    UniformNeumann(f32),
    UniformRobin(f32, f32),
    SpatiallyVaryingDirichlet(F), // admittedly this might be a stretch but would be super awesome
}
```

Given that the cuda-oxide kernels are themselves written in Rust, the compiler should (in the absence of any `unsafe` blocks) guarantee that invalid enum variants cannot be written, so I don't see the safety risk of passing enum values between host and device. E.g. in your example, if the kernel writes byte `2`, it should only be able to do so inside an `unsafe` block that is written incorrectly (unless I am misunderstanding whether cuda-oxide actually guarantees safety?)

One compromise option: perhaps we could have a "read-only `DeviceCopy`"? Then, the kernel will never write the value, so we don't need to worry about all-bit validity. Same thing applies to `bool` and `char` imo, I still use them in places.

I can just keep this modification on my own fork, so it's no big deal, but I thought others might find it quite helpful too.

nihalpasham:
> Given that the cuda-oxide kernels are themselves written in Rust, the compiler should (in the absence of any unsafe blocks) guarantee that invalid enum variants cannot be written, so I don't see the safety risk of passing enum values between host and device.

Agreed, safe Rust code should only create valid enum values. But the important detail here is that `DeviceCopy` is an unsafe trait. When someone writes a manual `unsafe impl DeviceCopy`, they are taking responsibility for checking that their specific use is safe. That is the existing escape hatch for cases like yours.

`#[derive(DeviceCopy)]` is different. The macro writes the unsafe implementation automatically, so cuda-oxide is taking responsibility for it. The macro therefore needs to know that the type is safe in every DeviceCopy API, not only in one read-only kernel.

Checking `mem::zeroed` only proves that zero is a valid value. It does not prove the rest of the `DeviceCopy` safety promise.

> One compromise option: perhaps we could have a "read-only DeviceCopy"? 

I assume you mean a device allocation initialized from valid host values and then exposed to kernels only as read-only input. That idea makes sense, but read-only access is a property of the allocation and how it is passed to a kernel, not a property of the Rust type itself. A type-level marker alone cannot ensure that the allocation is never zeroed, passed to a writable kernel, reinterpreted from other bytes, or otherwise used through the broader `DeviceCopy` API.

Supporting that safely would likely require a separate input-only buffer or transfer API that enforces those restrictions.
But I think for a single boundary-condition value, you may not need a buffer or `DeviceCopy` at all. Passing the enum to the kernel by value might already express what you want: a typed kernel input that the kernel only reads.

nihalpasham:
Closing, please feel free to re-open, if you think there is a discussion to be had here

PR #290, opened 2026-06-25
Select PTX ISA features by target architecture

### LINKED ISSUE #289: Select PTX ISA features from the requested target architecture
# Select PTX ISA features from the requested target architecture

## Problem

The MIR importer pipeline uses target-machine features for PTX generation. If
the selected PTX ISA feature does not match the requested target architecture,
the generated module can carry the wrong feature level.

## Why this matters

Users expect `--arch` or equivalent target selection to control the PTX ISA
level consistently. A stale or fixed feature selection can produce artifacts
that are incompatible with the requested device target.

## Expected behavior

The pipeline should derive the PTX ISA feature from the selected target
architecture and pass that feature to the backend.

## Acceptance criteria

- PTX feature selection follows the requested target architecture.
- Unsupported or unknown architectures fail clearly instead of silently using an
  unrelated feature.

niklebedenko:
Closes #289.

# Select PTX ISA features by target architecture

## Problem

PTX generation should use a feature level that matches the requested target
architecture. The importer pipeline did not centralize that mapping, making it
easy to emit with an incorrect PTX feature.

## Fix

This adds target-architecture-to-PTX-feature selection in the MIR importer
pipeline and passes the selected feature into the target machine configuration.

## Diligence

The mapping is kept at pipeline setup, where the target architecture is already
known, rather than scattering feature strings across lowering code.

## Tests

Branch green:

- `cargo test -p mir-importer test_select_ptx_mattr_tracks_arch_floor --lib -- --nocapture`
- `cargo check -p mir-importer --all-targets`

Upstream red:

- The transplanted PTX feature-selection unit test fails on `upstream/main` because `select_ptx_mattr` does not exist.

### DIFF
--- a/crates/mir-importer/src/pipeline.rs
+++ b/crates/mir-importer/src/pipeline.rs
@@ -903,6 +903,24 @@ fn select_target(features: DetectedFeatures) -> &'static str {
     }
 }
 
+/// Maps the selected GPU target to the PTX ISA feature passed to LLVM's NVPTX
+/// backend.
+///
+/// `-mcpu=sm_XX` is not enough for `llc`: without an explicit PTX ISA feature,
+/// stack intrinsics such as `llvm.stacksave` / `llvm.stackrestore` are treated
+/// as unsupported even for modern SM targets. Keep Ampere on PTX 7.6 for broad
+/// compatibility and raise the ISA floor for architectures whose selected
+/// features require newer PTX.
+fn select_ptx_mattr(target: &str) -> &'static str {
+    match arch_major(target) {
+        Some(0..=8) => "+ptx76",
+        Some(9) => "+ptx80",
+        Some(10 | 11) => "+ptx86",
+        Some(_) => "+ptx87",
+        None => "+ptx76",
+    }
+}
+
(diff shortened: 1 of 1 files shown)


nihalpasham:
About the PTX feature-selection question: the synthetic `stacksave` observation is real, but I do not think the proposed global `-mattr` table is safe to merge.

LLVM 22 already derives a target-specific PTX ISA from `-mcpu`. Direct probes produce PTX 7.8 for `sm_89`, 8.8 for `sm_103a`, and 9.0 for `sm_110`; this PR overrides those with 7.6, 8.6, and 8.6 respectively, and LLVM rejects all
three combinations. It also raises older targets unnecessarily and silently maps unknown targets to 7.6.

cuda-oxide does not currently emit the motivating stack intrinsics in the reported pipeline. If a future operation needs a PTX floor, it should be tied to that detected operation or an exact verified architecture mapping, not a coarse replacement for LLVM's `-mcpu` logic. Unknown-target validation is a real separate hardening opportunity because LLVM currently warns but exits successfully. Let me know if you have any questions.

nihalpasham:
Closing.

PR #292, opened 2026-06-25
Lower union aggregate construction

### LINKED ISSUE #291: Lower union aggregate construction in MIR import
# Lower union aggregate construction in MIR import

## Problem

MIR aggregate construction for unions needs to represent the active field while
still producing a value with the full union type. Without explicit handling,
inactive union fields can be treated like ordinary aggregate fields.

## Why this matters

Union construction has different semantics from structs and tuples. The
importer should preserve the active-field value and avoid inventing meaningful
values for inactive fields.

## Expected behavior

The importer should lower union aggregate construction by using the active
field value and representing inactive fields as undefined.

## Acceptance criteria

- Union aggregate construction lowers without treating inactive fields as
  initialized values.
- Struct, tuple, array, and enum aggregate behavior is preserved.
- Dialect verification accepts the needed zero-variant enum shape.

niklebedenko:
Closes #291.

## Summary

Models a Rust union as one shared region of bytes with several typed views.
Constructing one field no longer treats the other fields as separate values,
and the representation keeps rustc's size, alignment, stride, and pointer
requirements.

```text
Before: a union followed ordinary aggregate rules, so fields could receive
        independent storage or lose bytes when their shapes differed.
After:  every field starts at byte zero of one exact-size union value.
```

### Union type and operations

- Adds a dedicated `MirUnionType` carrying the union name, field types, rustc
  size, and rustc alignment.
- Lowers construction as `undef` plus insertion of the active field; inactive
  bytes remain undefined rather than being assigned invented values.
- Makes extraction, insertion, projected writes, and field addresses operate on
  the same byte-zero storage.
- Preserves union layout in arrays, nested structs, generic unions, returns,
  arguments, and ordinary function calls.

### Exact storage and ABI handling

- Selects a byte-faithful LLVM carrier with the exact Rust size, alignment, and
  array stride.
- Uses a pointer-preserving carrier only when it can represent every byte and
  address-space requirement; unsupported pointer layouts fail clearly.
- Avoids aggregate carriers with implicit LLVM padding that could discard union
  bytes.
- Treats non-unit zero-sized unions consistently across locals, constants,
  borrows, calls, returns, and control-flow edges while preserving over-aligned
  ZST addresses.
- Keeps ZST-returning LLVM calls as `void` for the NVPTX ABI while replacing
  live MIR value uses with a typed `undef`, so nested returns and call arguments
  remain well-typed without erasing callee side effects.

### Validation and example

- Adds dialect and lowering tests for union verification, storage selection,
  pointer provenance, exact size/alignment, and failure cases.
- Adds an end-to-end example covering cross-field reads, projected writes,
  unequal fields, arrays, nesting, generics, pointer fields, tuple padding,
  function ABI, and zero-sized/over-aligned unions.

## Safety limits

- Pointer-bearing unions are accepted only when LLVM storage can preserve the
  pointer and every union byte; otherwise compilation stops with a diagnostic.
- General allocated non-ZST union constants such as `const Bits` and
  `MaybeUninit::new(...)` remain unsupported and fail loudly. Zero-sized union
  constants are supported. This PR covers MIR union aggregate construction,
  not general union constant decoding.

## Testing

- `cargo oxide fmt --check`
- Strict mir-importer clippy
- `cargo check --workspace --all-targets`
- Full dialect-mir, mir-importer, and mir-lower tests
- Adversarial storage tests for padding, incompatible address spaces, invalid
  pointer carriers, over-alignment, ZST calls, and slot-less ZST borrows
- `cargo oxide run union_aggregate --arch sm_120`
- `CUDA_OXIDE_NO_OPT=1 cargo oxide run union_aggregate --arch sm_120`
- `RUSTFLAGS='-C opt-level=0' cargo oxide run union_aggregate --arch sm_120`
- All three RTX 5090 runs passed 21 union lanes and the by-value kernel argument

## Checklist

- [x] Struct, tuple, array, and enum behavior retained
- [x] Silent layout/provenance failures replaced with exact lowering or errors


nihalpasham:
Thanks for the PR. 

**For the union aggregate path:** The key semantic point is that union fields are typed views of one shared byte region, so we preserved your rebased commit and layered the full shared-storage model on top.

The branch now has a dedicated union type carrying 
- rustc size/alignment,
- byte-zero field extraction/insertion/addressing,
- exact LLVM storage and array
- stride, pointer-provenance checks, call/return ABI handling, and explicit
- zero-sized/over-aligned union support. Independent review caught additional
- padding, pointer-carrier, ZST-call, and slot-less-borrow cases; those now have
- regressions and fail closed where representation is not safe.

The full dialect/importer/lowering suites pass, and the example passes with normal optimization, backend optimization disabled, and rustc `-C opt-level=0` across all 21 permanent lanes plus a by-value kernel argument. General allocated non-ZST union constants remain an explicit, diagnostic limitation outside #291.

PR #294, opened 2026-06-25
Export evaluated static initializers for device globals

### LINKED ISSUE #293: Export evaluated initializers for device globals
# Export evaluated initializers for device globals

## Problem

Device global statics can be declared and referenced, but evaluated initializer
bytes are not fully carried through import, lowering, and LLVM export.

## Why this matters

Device globals need their initial values in the generated artifact. A missing
or zeroed initializer can compile successfully while producing wrong runtime
values on the GPU.

## Expected behavior

Static initializers should be evaluated, represented on the imported global,
lowered through memory conversion, and emitted by LLVM export.

## Acceptance criteria

- Static global initializer bytes are preserved through import and export.
- Pointer-to-array constants can materialize immutable globals with
  initializers.
- The focused `device_global` example observes initialized device data.

niklebedenko:
Closes #293.

## Summary

Carries evaluated Rust static-initializer bytes through MIR import and lowering
into the emitted LLVM device global. Supported globals now start with the same
observable value and bit pattern that rustc evaluated on the host side.

```text
Before: an initialized device global was rejected or could lose its initializer.
After:  representable initializers are emitted byte-for-byte; unsupported
        layouts and pointer relocations stop with a clear error.
```

### Initializer pipeline

- Records initializer bytes, alignment, storage class, and stable identity on
  imported globals and carries them through MIR-to-LLVM lowering.
- Emits aligned LLVM byte-array initializers for both modern opaque-pointer and
  legacy LLVM 7/libNVVM output.
- Preserves explicit bit patterns such as NaN payloads and normalizes rustc's
  undefined padding bytes deterministically to zero.
- Uses the declared static type when classifying global/constant address space.

### Promoted arrays

- Materializes supported promoted arrays as persistent module globals rather
  than returning pointers to temporary stack storage.
- Honors backing-allocation offsets and alignment, deduplicates by exact type
  plus bytes, and rejects nested relocations that cannot yet be represented.

### Fail-closed layout checks

- Verifies rustc size, alignment, and field offsets against the LLVM storage
  layout before emitting an initializer.
- Rejects unsupported pointer provenance, nonzero/interior pointer addends,
  mismatched pointee types, packed or overlapping layouts, niche enums,
  nonempty tuple initializers, and initialized unions instead of emitting
  plausible but incorrect data.
- Leaves shared-memory globals uninitialized as required.

## Known separate limitation

An existing path where a no-inline function returns a reference to a promoted
scalar or struct can still return callee-local storage. This behavior is also
present on `main` and is outside this PR's static/global initializer path; it
needs a separate follow-up. Supported promoted arrays in this PR use module
storage and do not have that lifetime problem.

## Testing

- `cargo oxide fmt --check`
- Full mir-importer, mir-lower, and llvm-export test suites
- Strict clippy for the root workspace, `rustc-codegen-cuda`, and every
  standalone example
- Modern and legacy LLVM/libNVVM export paths
- Positive cases for booleans, direct enums, over-aligned arrays, ZSTs,
  reordered Rust structs, array promotions, deduplication, and zero-offset
  static references
- Negative cases for packed structs, niche enums, tuples, unions, interior
  pointers, mismatched types, and unsupported provenance
- The pointer-provenance error example still fails at the intended
  one-relocation diagnostic
- RTX 5090 `sm_120` execution in optimized and no-opt modes, checking exact NaN
  payload (`0x7fc01234`), padded-struct bytes, mutable statics, and promoted data

## Checklist

- [x] Supported initializers preserve exact observable bytes
- [x] Unrepresentable layouts and relocations fail closed
- [x] Modern and legacy export paths covered


nihalpasham:
Thanks for the PR 

Evaluated static initializers: is valuable, and I preserved your rebased commit while hardening the byte/layout boundary on top.

The final branch emits 
- aligned byte-exact initializers in both modern and legacy LLVM paths, 
- preserves NaN payloads, 
- materializes promoted arrays in module storage, and 
- validates rustc size/alignment/field offsets before export. 

Packed, niche, tuple, union, interior-pointer, mismatched-pointee, and unsupported-provenance cases now stop with clear diagnostics instead of risking wrong device data.

GPU runs pass in optimized and no-opt modes, including exact NaN, padded-struct, mutable-static, and promoted-array values. 

_Reviewer note:_ the review also found a pre-existing escaping scalar/struct promotion lifetime gap; it is outside this PR and we should track it separately.

PR #296, opened 2026-06-25
Lower function-item rust-call receivers

### LINKED ISSUE #295: Lower function-item receivers in MIR call import
# Lower function-item receivers in MIR call import

## Problem

MIR call import can encounter function-item receivers in Rust-call paths. If
those receivers are treated only like closure values, function-item calls are
not lowered to the correct callee.

## Why this matters

Rust code can pass function items through call paths that look similar to
closure calls in MIR. The importer should distinguish and lower both forms
correctly instead of assuming every receiver is a closure object.

## Expected behavior

The importer should detect function-item receivers, resolve their target body,
and lower the call to the correct function.

## Acceptance criteria

- Function-item Rust-call receivers lower to the resolved function body.
- Closure receiver behavior is preserved.

niklebedenko:
Closes #295.

## Summary

Lowers Rust-call dispatch for concrete function items to the function body they
name while preserving the separate closure path.

```text
Before: passing a function item through Fn/FnMut/FnOnce could call a generated
        shim symbol that was never defined.
After:  semantic function-trait dispatch resolves the concrete item and emits
        a direct, ABI-correct call.
```

### Semantic callable detection

- Identifies the actual core `Fn`, `FnMut`, and `FnOnce` Rust-call methods by
  definition identity rather than matching strings in symbol names.
- Distinguishes function-item receivers from closure item/shim receivers and
  honors the projected operand type.
- Unpacks the rust-call tuple and emits a direct call to the concrete function
  item body; collector discovery uses the same semantic identity and receiver
  `Self` type.

### Correctness hardening

- Covers generic and const-generic items, tuples, references, unit and tuple
  returns, zero-argument calls, and `#[track_caller]`.
- Emits `mir.unreachable` after diverging function-item and closure calls, so a
  `-> !` call cannot leave an unterminated block or invent a never value.
- Rejects intrinsic, foreign, no-body, libm, and special `cuda_device` function
  items when direct dispatch would bypass required compiler handling.
- Adds a decoy-name regression so user-defined paths containing `FnOnce` cannot
  trigger callable lowering.

## Known separate limitations

- Explicitly coerced `fn` pointers still use the existing function-pointer path
  and are outside issue #295's `FnDef` receiver scope.
- The conservative `cuda_device` guard can reject ordinary MIR-backed device
  function items; relaxing it requires classifying special-dispatch functions
  rather than bypassing their handling.
- Ordinary functions with non-unit zero-sized parameters expose a pre-existing
  general ABI gap unrelated to function-item resolution.

## Testing

- `cargo oxide fmt --check`
- Strict clippy for the root workspace, `rustc-codegen-cuda`, and every
  standalone example
- 47 mir-importer tests and 5 backend collector tests
- Optimized and no-opt `sm_120` builds with exact call/definition symbol checks
- End-to-end cases for local, generic, const-generic, multi-argument,
  nested-tuple, tuple-return, reference, unit-return, zero-argument,
  `#[track_caller]`, closure, and diverging callables
- No-opt PTX retains both divergent concrete calls and non-returning loop bodies
- RTX 5090 execution passed in both optimized and no-opt modes

## Checklist

- [x] Callable traits identified semantically
- [x] Function-item and closure ABI paths kept distinct
- [x] Unsupported special dispatch fails closed


nihalpasham:
Thanks for the PR  

**For function-item Rust-call path:** I retained your rebased commit and layered a semantic implementation so the importer and collector agree on the concrete callable without relying on symbol-name substrings.

The branch now distinguishes function items from 
- closure item/shim receivers,
- unpacks the Rust-call tuple, 
- handles generic, tuple, reference, unit, zero-argument, `#[track_caller]`, and diverging cases, and 
- fails closed for functions that require intrinsic or `cuda_device` special dispatch.

Also found that diverging callable calls lacked a terminator; now both function-item and closure paths now emit `unreachable`.

**Reviewer identified follow-ups:** 
- Explicit `fn` pointers, 
- conservative rejection of ordinary `cuda_device` items, and 
- the general non-unit-ZST parameter ABI gap are separate follow-ups.

Issue #303, opened 2026-06-26
Resolve TODOs for MIR to LLVM conversion tests


uurl:
## Summary

There are existing TODOs requesting unit test coverage for MIR to LLVM conversion code paths. This issue tracks correcting those TODOs by adding focused regression tests for conversion behavior that is currently implemented but not explicitly covered.

## Scope

Add unit tests for:

* Constant conversion

  * Integer constants preserve their bit pattern.
  * Integer constants lower to signless LLVM integer types.
  * `f32` and `f64` constants preserve their floating-point attributes.
  * MIR-specific `f16` attributes are rewritten to the LLVM half attribute representation.

* Aggregate conversion

  * Slice construction lowers to the canonical `{ ptr, len }` aggregate layout.
  * Struct construction respects explicit layout slots and skips zero-sized fields.
  * ZST field extraction does not emit `llvm.extractvalue`.
  * Enum construction uses the declared discriminant value rather than the variant index.

* Cast conversion

  * Signed integer widening lowers to `llvm.sext`.
  * Unsigned integer widening lowers to `llvm.zext`.
  * Integer narrowing lowers to `llvm.trunc`.
  * Unsigned integer to float lowers to `llvm.uitofp`.
  * Float to signed integer lowers through the saturating intrinsic.
  * Pointer address exposure lowers to `llvm.ptrtoint`.

## Motivation

These tests make the lowering behavior explicit and protect several nontrivial semantic details:

* Rust/MIR signedness should not leak incorrectly into lowered LLVM integer constants.
* Explicit struct layout and padding must be preserved.
* Zero-sized types should not produce storage-based extract operations.
* Enum discriminants must use their declared bit-pattern values.
* Cast lowering must select the correct LLVM operation or intrinsic based on source and destination types.

## Acceptance criteria

* The TODOs for constant, aggregate, and cast conversion tests are resolved.
* The new tests pass with `cargo test`.
* The tests validate observable lowered LLVM operations and attributes rather than only checking that lowering succeeds.

honeyspoon:
All items in the scope section appear to be covered by tests on `main`:

| Feature | Test | Location |
|---------|------|----------|
| Integer constant bit preservation | `convert_integer_preserves_bits_and_makes_type_signless` | `constants.rs:242` |
| f16 attribute conversion | `convert_f16_constant_rewrites_mir_attr_to_builtin_half_attr` | `constants.rs:312` |
[...]

PR #302, opened 2026-06-26
test(mir-lower): Add representative unit tests for MIR cast lowering

uurl:
Closes #301

Part of #303.

## Summary

Adds six representative tests for existing MIR cast lowering. The tests build
minimal MIR functions, run the normal `lower_mir_to_llvm` pass, and inspect the
selected LLVM operations.

## What changed

- Signed integer widening lowers to `llvm.sext`.
- Unsigned integer widening lowers to `llvm.zext`.
- Integer narrowing lowers to `llvm.trunc`.
- Unsigned integer-to-float lowers to `llvm.uitofp`.
- Signed float-to-integer lowering declares and directly calls
  `llvm_fptosi_sat_i32_f32`.
- Pointer address exposure lowers to `llvm.ptrtoint`.

### Maintainer hardening

The saturation test now connects the emitted call to its declaration by
requiring a direct callee named `llvm_fptosi_sat_i32_f32`. This prevents an
unrelated call plus an unused correct declaration from passing.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p mir-lower cast::tests --lib`
- [x] `cargo test -p mir-lower`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] Adversarial wrong-callee mutation fails the strengthened test

## Checklist

- [x] Test coverage only; no compiler behavior change
- [x] Contributor authorship and DCO sign-off preserved
- [x] No `crates/cuda-bindings/` changes

PR #298, opened 2026-06-26
test(mir-lower): Add unit tests for MIR constant lowering

uurl:
Closes #300

Part of #303.

## Summary

Adds focused unit coverage for existing MIR constant lowering. The tests build
minimal MIR functions, run the normal `lower_mir_to_llvm` pass, and inspect the
resulting LLVM constant attributes.

## What changed

- Signed and unsigned 32-bit integer constants must become signless LLVM
  integers without changing their bit patterns.
- `f32` and `f64` constants must preserve their exact floating-point bits.
- MIR-specific `f16` constants must become LLVM half attributes with the same
  bits.
- A maintainer cleanup uses the current immutable context API in the test
  fixture, removing its patch-local strict-Clippy diagnostic.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p mir-lower convert::ops::constants::tests --lib`
- [x] `cargo test -p mir-lower`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] Strict test-target Clippy reports no diagnostic from `constants.rs`
  (current `main` still has unrelated diagnostics in `tests/lowering_test.rs`)

## Checklist

- [x] Test coverage only; no compiler behavior change
- [x] Contributor authorship and DCO sign-off preserved
- [x] No `crates/cuda-bindings/` changes

PR #297, opened 2026-06-26
test(mir-lower): Add aggregate lowering tests

uurl:
Closes #299

Part of #303.

## Summary

Adds focused unit coverage for existing MIR aggregate lowering. The tests build
minimal MIR functions, run the normal `lower_mir_to_llvm` pass, and inspect both
the emitted LLVM operation shape and the values flowing through it.

## What changed

- Slice construction must insert the data pointer at slot 0 and the length at
  slot 1.
- Struct construction must respect rustc layout slots, padding, and stripped
  zero-sized fields.
- Extracting a stripped zero-sized field must produce `undef`, not
  `llvm.extractvalue`.
- Enum construction must insert the declared discriminant bit pattern, not the
  variant index.

### Maintainer hardening

- Slice and struct tests now connect each inserted value to its exact slot and
  verify the insertion chain.
- The enum test follows the inserted tag to its defining constant and verifies
  the exact 8-bit value `255`.
- Patch-local strict-Clippy diagnostics in the test fixtures are removed.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p mir-lower convert::ops::aggregate::tests --lib`
- [x] `cargo test -p mir-lower`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] Adversarial pointer/length swap fails the strengthened slice test

## Checklist

- [x] Test coverage only; no compiler behavior change
- [x] Contributor authorship and DCO sign-off preserved
- [x] No `crates/cuda-bindings/` changes

PR #304, opened 2026-06-26
test(mir-lower): Add common intrinsic helper tests

uurl:
## Summary

Adds direct unit coverage for the low-level helpers in
`mir-lower/src/convert/intrinsics/common.rs`. The tests run the helpers through
a small `DialectConversionRewriter` fixture and inspect the LLVM operations and
attributes they insert.

This PR closes no open issue. In particular, #303 tracks aggregate, constant,
and cast conversion tests; those are covered separately by #297, #298, and
#302.

## What changed

- i1, i32, and i64 constant helpers must preserve exact signless types and bit
  patterns.
- Generic pointers must cast to shared address space 3 and cluster-shared
  address space 7, while pointers already in those spaces must not be recast.
- `call_intrinsic` must create the declaration and a direct call to it.
- Convergent inline assembly must preserve its template and constraints and use
  exact `AsmKind::Convergent` semantics.
- Side-effect-only inline assembly must preserve its template and constraints,
  use exact `AsmKind::SideEffect` semantics, and remain non-convergent.
- i32-to-i1 truncation must produce an i1 `llvm.trunc` result.

### Maintainer hardening

- Constant assertions now verify the values as well as widths.
- Inline-assembly assertions now pin the complete semantic kind.
- Direct coverage was added for `inline_asm_sideeffect`, which landed on
  `main` after the contributor branch was created.
- Patch-local strict-Clippy diagnostics in the test fixtures are removed.

## Testing

- [x] `cargo test -p mir-lower convert::intrinsics::common::tests --lib`
- [x] `cargo test -p mir-lower`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] Adversarial wrong-constant and wrong-`AsmKind` mutations fail the
  strengthened tests

## Checklist

- [x] Test coverage only; no compiler behavior change
- [x] Contributor authorship and DCO sign-off preserved
- [x] No `crates/cuda-bindings/` changes

### DIFF
--- a/crates/mir-lower/src/convert/intrinsics/common.rs
+++ b/crates/mir-lower/src/convert/intrinsics/common.rs
@@ -206,5 +206,369 @@ pub fn trunc_to_i1(
 
 #[cfg(test)]
 mod tests {
-    // TODO: Add tests for common intrinsic helpers
+    use super::*;
+    use dialect_mir::ops as mir;
+    use llvm_export::ops as llvm;
+    use pliron::basic_block::BasicBlock;
+    use pliron::builtin::attributes::TypeAttr;
+    use pliron::builtin::op_interfaces::{CallOpCallable, CallOpInterface, SymbolOpInterface};
+    use pliron::builtin::ops::ModuleOp;
+    use pliron::builtin::types::FunctionType;
+    use pliron::irbuild::dialect_conversion::{
+        DialectConversion, OperandsInfo, apply_dialect_conversion,
+    };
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR, 

common-helper tests are valuable i.e. the address-space cast and direct intrinsic-call checks. I corrected the PR title and description, removed the unrelated #303 closure claim, and layered stronger constant and exact `AsmKind` assertions on top. We also added direct coverage for the current side-effect helper. 

All nine focused tests and the full suite pass.

Issue #306, opened 2026-06-28
Device-side SetDiscriminant is unsupported for both direct-tag and niche-encoded enums


goog00:
## What

MIR `StatementKind::SetDiscriminant` is currently rejected on the device with:

```text
SetDiscriminant statements are not yet supported on the device; until they are lowered, enum discriminant writes would be silently dropped
```

This means any kernel code that writes an enum discriminant cannot compile. Common triggers include:

- Mutating an enum variant inside a kernel:
  ```rust
  enum DeviceState { Empty, Full(u32) }
  let mut state = DeviceState::Full(42);
  state = DeviceState::Empty;
  ```

- `Option::take(&mut opt)` or similar standard library operations that update an `Option`'s discriminant.

The diagnostic was added to replace an earlier silent miscompile where the discriminant write was dropped entirely, but the underlying lowering is still missing.

---

## Scope

This issue tracks implementing `SetDiscriminant` lowering for all enum kinds:

1. **Direct-tag enums** — enums with a separate discriminant tag in memory (e.g. `enum DeviceState { Empty, Full(u32) }`). Lowering should write the declared discriminant value to the tag slot.

2. **Niche-encoded enums** — enums whose discriminant is encoded in the payload scalar itself (e.g. `Option<&T>`, `Option<NonZeroU32>`). Lowering must rewrite the payload scalar rather than storing a separate tag.

vyncint:
This looks resolved. Both halves named in the title are implemented and covered by executing examples.

**The diagnostic is gone.** `grep` for "SetDiscriminant statements are not yet supported on the device" finds nothing on `15fa9d9d`. The importer lowers it: `crates/mir-importer/src/translator/statement.rs` builds a `MirSetDiscriminantOp` (around lines 936-1025), rejecting only the cases that genuinely cannot be encoded — an unsupported enum type, a variant index out of bounds, an uninhabited variant, an unsupported layout kind, and a place with no addressable slot.

**Three examples cover it**, and the two positive ones pass on an A10G (sm_86, driver 595.71.05):

| example | covers | result |
|---|---|---|
| `set_discriminant` | direct-tag, incl. `#[repr(i8)]` with explicit discriminants (`Empty = -5`, `Full(u32) = 7`) and a Single layout | `PASS: all 64 threads observed the SetDiscriminant write.` |
[...]

PR #307, opened 2026-06-28
feat(mir): implement device-side SetDiscriminant lowering

goog00:
## Summary
Part of #306.

Implements lowering for MIR `StatementKind::SetDiscriminant` on the device.Previously, enum discriminant writes were rejected with an "unsupported" diagnostic (after an earlier silent-miscompile regression where they were dropped entirely). This PR adds a dedicated `mir.set_discriminant` operation and lowers it through the full pipeline so that direct-tag enums can now be mutated inside `#[kernel]` functions. Niche-encoded enums are still rejected loudly until their payload-scalar encoding is handled.

## Changes

- Added `MirSetDiscriminantOp` to `dialect-mir` with verifier tests.
- Updated `mir-importer` to translate `SetDiscriminant` into `mir.constant` + `mir.set_discriminant` instead of erroring.
- Updated `mir-lower` to convert `mir.set_discriminant` to  `llvm.getelementptr` (to the enum tag slot) + `llvm.store`.
- Renamed the negative example `error_set_discriminant_unhandled` to the  positive example `set_discriminant` and updated `STATUS.md` and  `scripts/smoketest.sh` accordingly.

## Testing

- `cargo test -p dialect-mir -p mir-importer -p mir-lower --lib --tests` passes.
- `cargo clippy -p dialect-mir -p mir-importer -p mir-lower --lib -- -D warnings` passes.
- `cargo oxide run set_discriminant` passes on RTX 5090 (sm_120a): 
  PASS: all 64 threads observed the SetDiscriminant write.
- `scripts/smoketest.sh` passes.

## Follow-up

Issue #306 also covers **niche-encoded enums** such as `Option<&T>` and `Option<NonZeroU32>`, where the discriminant is encoded in the payload scalar itself rather than a separate tag. A follow-up PR will implement the payload-scalar rewrite path for those cases.

## Checklist

- [x] No files under `cuda-bindings/` modified by an external contributor (see CONTRIBUTING.md)


nihalpasham:
Thanks for the PR. It LGTM

A couple of points: 
- preserving it as a typed MIR operation and reusing the enum slot map are both in the right direction, and the it handles signed, sparse, wide, nested, indexed, global, and shared-memory cases correctly.
- I layered a couple of safety checks on top: 
  - uninhabited variants reject, 
  - direct/niche/no-tag layouts are classified explicitly, and 
  - focused compiler tests cover the edge cases. 
 
 This rounds out the inhabited direct-tag part of #306; but I think #306 should stay open for niche encoding.

Issue #107, opened 2026-06-03
CUDA Graph API


mtrsl:
Are there any plans to implement a safe interface for the CUDA Graph API? Maybe `DeviceOperation` could be used for some graph-like stuff, but it doesn't seem like a full replacement

nihalpasham:
Short answer: yes, and it arrives through the shared host crates, not oxide's own.

- oxide's `cuda-bindings` / `cuda-core` / `cuda-async` are being swapped for the crates cutile-rs publishes (#1102, draft, CI builds pass). That is why #346 and #585 were closed rather than merged: a graph layer in the old crates would be deleted with them.
- The shared `cuda-async` already has a graph module built on stream capture. Stream capture means: run your launches once while the stream is recording, get a graph back, replay the graph as one unit. The per-launch CPU cost is paid once instead of every iteration (#574 measured about 1 us saved per kernel launch).
- On `DeviceOperation`: it is not a replacement, it is the input. A `DeviceOperation` chain is what gets recorded.

```rust
// sketch of the shared cuda_async::cuda_graph API once oxide launchers plug in
let graph = CudaGraph::scope(&stream, |s| {
    s.record(kernels.gemm_async(cfg, &x, &w0, &mut h))?;   // recorded, not run
    s.record(kernels.relu_async(cfg, &mut h))
})?;
for batch in inputs {
    graph.update(memcpy_into(&mut x, &batch))?;  // new input, same buffer
    graph.launch().sync_on(&stream)?;            // replay as one launch
}
```

- What is left on my side: oxide's launchers implement a different trait (same job) from the one `record` accepts today, so the shared crate needs a small bridge for them, plus an oxide example with buffers allocated outside the capture. That pair is what this issue waits on.

Which workload do you want to replay (decode-step loop, iterative solver, something else), so the example matches it?

mtrsl:
Thanks for the update. A common base for `cuda-oxide` and `cutile-rs` sounds great.

I am more likely to use explicitly defined graphs than captured graphs. No particular workload in mind currently, but I've previously encoded the dependencies between compartmental model components in a CUDA graph allowing for parallel model updates

midagedev:
We are running the dense block of a DeepSeek-V2-Lite decode step (33 launches, one token) as one captured graph on an RTX 3090 with cuda-oxide launchers, MoE blocks next, so this thread is the one we depend on. Two things, one small.

[...]

PR #308, opened 2026-06-28
Add experimental CUDA program graph macro

DaronPopov:
## Summary

This PR introduces an experimental `#[cuda_program]` macro that provides a semantic program-graph layer on top of the existing `#[cuda_module]` system.

A `#[cuda_program]` function describes an ordered sequence of kernel launches. The generated code keeps module loading, stream management, memory transfers, and launch submission explicit while producing graph metadata and a typed `BoundProgram` that can be launched directly or submitted through CUDA Oxide's existing execution infrastructure.

## Features

* Introduces the experimental `#[cuda_program]` macro.
* Generates semantic graph metadata, including:

  * operations
  * resource roles
  * argument roles
  * dependency edges
* Adds `cuda_host::program` runtime types.
* Introduces `BoundProgram` with direct launch support.
* Integrates `BoundProgram` with `cuda_async::DeviceOperation` behind the `cuda-host/async` feature.
* Generates graph builder functions such as `forward_graph(...)`.
* Provides compile-time diagnostics for invalid program definitions and invocations.
* Adds the `cuda_program_chain` example demonstrating a five-kernel execution pipeline with host-side reference validation.

## Design

The goal of this abstraction is to provide a higher-level way to structure CUDA programs without introducing a new scheduler or compiler IR.

`#[cuda_program]` builds on the existing `#[cuda_module]` infrastructure and preserves CUDA Oxide's current compilation pipeline and runtime execution model while exposing program structure as typed graph metadata.

## Validation

Validated with:

* `cargo fmt --check`
* `cargo clippy -p cuda-device -p cuda-host -p cuda-macros -- -D warnings`
* `cargo test -p cuda-host --features async`
* `cargo test -p cuda-macros`
* `cargo oxide run cuda_program_chain`
* `BENCH_ITERS=10000 cargo oxide run cuda_program_chain`


nihalpasham:
Thanks for the PR. Just a quick note, I'd recommend starting a discord discussion, next time for topics such as this one, where the community can quickly chime in on this.

As I understand this PR, the goal is 
- describe several kernel launches once, 
- collect metadata about them, and 
- execute them as one program.

After reviewing the design and implementation, this ends up re-inventing part of CUDA Graphs in host code. `cuda-oxide` already has two established solutions for the main use cases:

```text
Compose launches and use Rust async:
    with DeviceOperation

OR

Build and repeatedly execute a GPU graph:
    CUDA Graphs (planned, tracked in #107)

This PR provides:
    static metadata + one-shot sequential launches
```

`BoundProgram` contains a `Box<dyn FnOnce>`, and `launch(self)` consumes it. The only lowering is `SequentialLaunches`, so every execution rebuilds the wrapper and still submits each kernel separately. There is no CUDA Graph creation, capture, instantiation, update, or replay.

The other thing is the safety boundary. An unsafe kernel may have requirements that the caller must prove, but the macro hides the unsafe call behind a safe API:

```text
unsafe kernel requirement
        ↓
generated unsafe { kernel(...) }
        ↓
safe graph.bind(...).launch(...)
```

This allows safe host code to launch an unsafe kernel without proving its preconditions.

The generated dependency graph is also not reliable enough for future scheduling. For example:

```text
operation 0: read(x)
operation 1: write(x)

required edge: 0 → 1
generated edge: none
```

The implementation tracks only the last writer, so it misses this read-before-write dependency. Expressions such as `&*x`, indexing, and field access can also lose their resource identity entirely. Sequential execution hides these mistakes today, but a future scheduler could reorder operations incorrectly.

I think the cleaner direction is:

1. Use or extend `DeviceOperation` for launch composition and async execution.
2. Use safe native CUDA Graph support (when #107 lands) when replay and reduced launch overhead are required.
3. Add optional syntax sugar only after that execution primitive exists.

If a separate semantic graph layer is still desired, it first needs a concrete consumer and an explicit, sound resource-effect model rather than dependency inference from variable names.

nihalpasham:
Closing this one. Please feel free to reopen if you believe there's more to discuss.

PR #309, opened 2026-06-29
feat: add missing special registers (warpid, smid, gridid, smem_size)

honeyspoon:
## Summary

Add 7 PTX special register intrinsics across all four compiler layers
(cuda-device, dialect-nvvm, mir-importer, mir-lower):

- `warpid` / `nwarpid`: hardware warp identification within the SM
- `smid` / `nsmid`: streaming multiprocessor identification
- `gridid`: grid launch identification
- `dynamic_smem_size` / `total_smem_size`: shared memory size queries

All follow the established zero-operand, single-i32-result LLVM NVVM
intrinsic pattern (same as `lane_id`, `lanemask_*`).

Ref #274 (Phase 4: Miscellaneous)

## Test plan

- [x] 4 new lowering tests verifying correct LLVM intrinsic names
- [x] All existing tests pass (31 integration + 80 unit in mir-lower)


nihalpasham:
Thanks for the PR. LGTM 

I layered a couple of changes on top
- The branch resolves both public paths, reads shared-memory sizes through portable inline PTX, preserves repeated `warpid`/`smid` samples, and returns the full 64-bit `gridid`.

Also, verified the generated PTX with LLVM 21 and 22 and CUDA 13.3 `ptxas`.

PR #310, opened 2026-06-29
feat: add movmatrix in-register matrix transpose

honeyspoon:
## Summary

Add `movmatrix.sync.aligned.m8n8.trans.b16` across all four compiler
layers (cuda-device, dialect-nvvm, mir-importer, mir-lower).

The instruction transposes an 8x8 matrix of b16 elements distributed
across a warp's registers without a shared-memory round-trip. Each
thread provides one u32 (2 packed b16 elements); the hardware
collectively transposes the tile and returns the result to each lane.

This is useful for switching fragment layouts between chained MMA
operations (e.g., row-major to column-major) without going through
shared memory.

Requires sm_90+ (Hopper).

Ref #274 (Phase 2: Matrix operations)

## Test plan

- [x] 1 new lowering test verifying convergent inline PTX output
- [x] All existing tests pass


nihalpasham:
Thanks for the PR. LGTM. 

I kept the hardware floor at `sm_75`, added exact packed-`i32` verification, and covered mixed target requirements. Runtime lane values were not tested on a GPU. 

**Reviewer note:** This work partially addresses #274, which remains open.

PR #311, opened 2026-06-29
feat: add ldmatrix warp-level shared memory matrix loads

honeyspoon:
## Summary

Add all 6 `ldmatrix` variants across all four compiler layers
(cuda-device, dialect-nvvm, mir-importer, mir-lower):

- `ldmatrix_x1` / `ldmatrix_x1_trans`: load 1 tile, returns `u32`
- `ldmatrix_x2` / `ldmatrix_x2_trans`: load 2 tiles, returns `[u32; 2]`
- `ldmatrix_x4` / `ldmatrix_x4_trans`: load 4 tiles, returns `[u32; 4]`

These are the warp-cooperative building blocks for loading matrix
fragments from shared memory in the layout expected by `mma.sync`
tensor core instructions.

**Design notes:**

- x1 variants use the standard scalar-return pattern (1 operand, 1 result)
- x2/x4 variants use the alloca-slot pattern (same as `stmatrix`): the
  importer passes the destination slot pointer as an operand, and the
  lowered inline PTX writes results directly to the slot via `st.b32`
- All variants emit convergent inline assembly with `cvta.to.shared`
  pointer conversion (shared memory 32-bit addressing)

Ref #274 (Phase 2: Matrix operations)

## Test plan

- [x] 5 new lowering tests covering x1, x1_trans, x2, x4, x4_trans
- [x] All existing tests pass (32 integration + 80 unit in mir-lower)


nihalpasham:
Thanks for the PR. LGTM. 

I confirmed that all six `ldmatrix` forms now assemble at the true `sm_75` / PTX 6.5 floor.

A couple of points: 
- I changed `x2`/`x4` to return register tuples directly, modeled every load as a convergent shared-memory read, and completed the target and lane-address rules. 
- Since we are here, also fixed `stmatrix` x4-transpose, its packed-register contract, its `sm_90` / PTX 7.8 target selection, and related target and cluster-register detection. 

**Reviewer note:** This work partially addresses #274, which remains open. Runtime fragment values and performance were not tested on a GPU.

PR #319, opened 2026-07-01
feat: add proof-carrying device views

nihalpasham:
## Summary

- Adds proof-carrying 32-bit element and fixed-tile views for safe GPU writes.
- Uses the clearer public API:
  - `#[kernel(launch_context = launch_context)]`
  - `thread::index_1d_u32(launch_context)`
  - `thread::coord_2d_u32(launch_context)`
- Keeps the launch context local to the device entry; it is not a runtime kernel argument.
- Checks a complete tile once, then allows compile-time positions through `at_const` without another runtime bounds check.
- Documents that `ROW_STRIDE` is the caller-declared logical row pitch.
- Updates rustdocs, examples, diagnostics, and the cuda-oxide book.

## Small example

```rust
#[kernel(launch_context = launch_context)]
#[launch_contract(domain = 1, coordinates = u32, block = (64, 1, 1))]
fn update(mut values: DisjointSlice<u32, LinearTiles<4>>) {
    let thread_index = thread::index_1d_u32(launch_context);

    if let Some(mut tile) = values.tile_thread32(thread_index) {
        let value = tile.at_const::<0>().read();
        tile.at_const::<0>().write(value + 1);
    }
}
```

## Validation

- `cargo test -p cuda-device --all-targets`
- `cargo doc --no-deps --workspace`
- Sphinx book build with warnings as errors
- `proof_carrying_views --verify-ptx`


alejandro-soto-franco (commented):
Reviewed the soundness core, since that is what the whole design rests on: `at()` skips the bounds branch and does an unchecked `ptr.add`, so the tile constructor has to carry the entire proof. It does, and the algebra holds:

- `scaled_tile_axis_fits(origin, width)` gives `origin*width + (width-1) <= u32::MAX`, so `start_row`/`last_row` and `start_col`/`last_col` cannot overflow.
- `checked_linear_2d` then bounds `last_row*stride + last_col` inside `u32`, and `checked_row_major_tile_start` rejects unless `last < len`.
- The largest interior offset `at()` can form is `(ROWS-1)*ROW_STRIDE + (COLS-1)`, which is exactly `last - start`. So the `wrapping_mul`/`wrapping_add` cannot wrap, and `ptr.add(offset)` stays within the rectangle that was checked once. The comment claiming this is accurate, and the `u32`-boundary unit tests pin it.

`last_col < stride` keeping each tile inside one logical row is also what makes adjacent X threads disjoint, so the disjointness claim in the `tile_2d32` SAFETY comment follows from the same check rather than from convention. Good.

Three things:

**1. The base branch has merged.** This targets `feat/typed-launch-contracts`, which landed on main as #318 on 4 July (29396b7). GitHub reports MERGEABLE/CLEAN, but that is against an already-merged branch, so the green state is misleading. This needs the retarget to `main` and the rebase the description already flags, and CI has not run on it since.

**2. `LocalIndex32::constant` cannot panic, and its docs say it can.** The asserts sit in a `const { }` block, so an out-of-range `I` is a compile-time error, not a runtime panic. That is the right choice for device code, where a runtime panic is far worse than a failed build, but the `# Panics` section says the opposite of what the code does. Worth saying "fails to compile" instead, on an API whose whole job is to be trusted.

**3. `ROW_STRIDE` is not checked against the parent's real row width.** `valid_row_major_shape` requires `ROW_STRIDE >= COLS`, and the rectangle is bounds-checked against `len`, so nothing here is unsound. But if a kernel declares `RowMajorTiles<1, 2, 64>` over a buffer whose rows are actually 32 wide, every access stays in bounds and silently addresses the wrong elements. Since `len` is known, is it worth rejecting `len % ROW_STRIDE != 0`, or at least documenting that the declared stride is taken on trust?

nihalpasham:
@alejandro-soto-franco - Thanks for the review. Appreciate it. 

- rebased and retargeted this to main, 
- corrected the `LocalIndex32::constant` docs to say invalid values fail at compile time, and 
- clearly documented that `ROW_STRIDE` is supplied by the caller and must match the buffer layout.

PR #320, opened 2026-07-01
feat: add compile-time kernel policy primitives

nihalpasham:
## Summary

- Lets one generic kernel produce several compile-time-tuned specializations.
- Supports policy constants in `launch_bounds`, such as 64 threads with unroll 2 and 256 threads with unroll 4.
- Carries each policy's `MAX_THREADS` into its typed host launch contract, so an oversized block is rejected before the CUDA driver call.
- Keeps `MIN_BLOCKS` as a compiler occupancy hint; it is not launch geometry.
- Keeps policy selection out of the runtime kernel arguments.
- Adds metadata-only names for shapes, layouts, memory spaces, tiles, scopes, and operations. These descriptions do not allocate or access memory.
- Updates rustdocs, READMEs, the policy example, and the cuda-oxide book.

## Small example

```rust
#[kernel]
#[launch_bounds(P::MAX_THREADS, P::MIN_BLOCKS)]
#[launch_contract(domain = 1)]
fn transform<P: TransformPolicy>(mut output: DisjointSlice<u32>) {
    let index = thread::index_1d();
    if let Some(value) = output.get_mut(index) {
        #[unroll(P::UNROLL)]
        for _ in 0..8 {
            *value = value.rotate_left(1);
        }
    }
}

let small = module.prepare_transform::<Small>(LaunchConfig1D::new(1, 64, 0))?;
let wide = module.prepare_transform::<Wide>(LaunchConfig1D::new(1, 256, 0))?;
```

`MAX_THREADS` is a maximum, not an exact block size. Use an explicit `block = (x, y, z)` contract when the full block shape must match.

## Validation

- `cargo test -p cuda-device --all-targets`
- `cargo doc --no-deps --workspace`
- Sphinx book build with warnings as errors
- `policy_config --verify-ptx`


alejandro-soto-franco (commented):
The const-block validation (`validate_launch_bounds`) plus the `compile_fail` coverage for zero threads, invalid unroll, and unresolved policies is the right shape: an out-of-range policy fails the build rather than the launch, which is what you want on the device side.

Two things, one procedural and one about how this composes with #319.

**1. The base branch has merged.** This targets `feat/typed-launch-contracts`, which landed on main as #318 on 4 July (29396b7). GitHub shows MERGEABLE/CLEAN, but that is against an already-merged branch, so the state is misleading and CI has not run since. Same for #319. Both need a retarget to `main`.

**2. A policy kernel cannot use a launch contract, and that is where the two PRs pull apart.** The macro rejects the combination:

> launch_bounds const expressions are supported for device compilation, but a kernel with launch_contract currently requires a literal maximum-thread bound so its host-side contract can be checked

So `#[launch_bounds(P::MAX_THREADS, P::MIN_BLOCKS)]` opts a kernel out of the contract path. That path is where `__prepare` calls `validate_live_shape(C::SPEC, raw, limits, function_max_threads)` against the function's compiled `maxntid`. A policy kernel loses that pre-flight check and instead finds out at `cuLaunchKernel`, as a driver error rather than a typed one. Not unsound, since the driver does reject an oversized block, but it means the kernels most likely to want a policy (the ones with a tuned block size, which is the motivating example) are exactly the ones that fall back to `unsafe fn` and a raw launch.

The restriction looks like an implementation limit rather than a fundamental one: at the launch site `P` is monomorphised, so `P::MAX_THREADS` is as much a constant there as a literal would be. Is there a reason the contract cannot read the bound through the same associated const it already reads on the device side? Landing #319 and #320 together with a gap between the safe-launch path and the policy path seems a shame when the two are so close to composing.

nihalpasham:
Thanks, agreed on the launch-contract gap. 

- rebased and retargeted this to `main`, and policy values such as `P::MAX_THREADS` now flow into each typed host launch contract. 
- Oversized blocks are therefore rejected before calling the CUDA driver. 
- Tests, examples, and book documentation were added too.

PR #322, opened 2026-07-01
Add packed atomic add intrinsics (f16x2, bf16x2)

### LINKED ISSUE #289: Select PTX ISA features from the requested target architecture

honeyspoon:
## Summary

Adds low-level packed FP16x2 and BF16x2 global atomic-add intrinsics to
cuda-device, with exact PTX semantics and target requirements.

Each 16-bit lane is atomic independently. The returned lanes are the previous
lane values, but they are not guaranteed to form one coherent old 32-bit
snapshot.

Partially addresses #274. It does not fix #289's general requested-target/PTX
mapping problem; both issues remain open.

### Device API and memory contract

- Adds unsafe `atom_add_f16x2(*mut u32, u32) -> u32` and
  `atom_add_bf16x2(*mut u32, u32) -> u32` operations.
- Documents relaxed ordering, device scope, global memory, independent lane
  atomicity, unspecified lane order, alignment, and overlapping-access rules.
- Keeps the API raw and register-packed; no typed or emulated operation is
  implied.

### Compiler pipeline

- Imports each public stub through a shared generic operation helper.
- Verifies a mutable generic/global pointer to unsigned `u32`, an unsigned
  `u32` addend, and an unsigned `u32` result.
- Lowers to side-effecting inline PTX with a memory clobber.

### Target selection

- FP16x2 selects cuda-oxide's `sm_70` floor and PTX 6.2.
- Native BF16x2 selects `sm_90` and PTX 7.8.
- BF16x2 is deliberately native-only; this PR does not claim CUDA C++'s older
  hardware emulation.
- Explicit unsupported targets and mnemonic near misses are rejected.

### Example and tests

- Adds a `packed_atomic_add` example with a compile-only path and a runtime
  lane-sum smoke test for `sm_90+` hardware.
- Adds dialect, importer, lowering, target-detection, PTX, and assembler
  regressions.

## Testing

- [x] Root and example formatting checks
- [x] `cargo test -p dialect-nvvm -p mir-lower -p mir-importer --all-targets`
- [x] `cargo check --workspace --all-targets`
- [x] Workspace and example clippy with warnings denied
- [x] Combined example emits PTX 7.8 / `sm_90` and both exact atomics
- [x] CUDA 13.3 `ptxas -arch=sm_90` accepts the combined PTX
- [x] Explicit `sm_89` and `sm_80` reject native BF16x2 before PTX generation
- [x] FP16-only build emits PTX 6.2 / `sm_70`; the instruction assembles
  forward for `sm_75` on CUDA 13.3
- [x] Runtime lane-sum smoke test passes on `sm_120`
- [x] Compute Sanitizer memcheck reports zero errors on the runtime example

## Checklist

- [x] Contributor authorship preserved
- [x] Every PR commit has DCO sign-off
- [x] Public API and safety documentation updated
- [x] No `crates/cuda-bindings/` files changed


nihalpasham:
Thanks for the PR.

The key semantic detail is that the two packed lanes are atomic independently:

```text
u32 word:  [ high 16-bit lane | low 16-bit lane ]
atomicity: [ atomic separately | atomic separately ]
           not one atomic 32-bit snapshot
```

The final hardware floors are:

- `f16x2`: PTX 6.2, available on cuda-oxide's `sm_70` minimum.
- Native `bf16x2`: PTX 7.8 and `sm_90+`, not `sm_80`. Older CUDA C++ support uses emulation, while this API exposes the native instruction.

Layered the following on top: 
- added complete pointer/type verification, 
- precise unsafe API documentation,
- side-effecting inline PTX with a memory clobber, 
- exact feature detection, 
- architecture rejection tests, and an 
- end-to-end example.

This completes the packed-atomic part of #274, but #274 remains open for its other tracked work. It does not fix #289, which covers broader PTX ISA selection.

PR #324, opened 2026-07-01
feat(cuda-macros): collect #[cuda_module] kernels from nested modules

bitemyapp:
## Summary

`#[cuda_module]` can now organize kernels in nested inline Rust modules without moving their generated launchers into another namespace.

The namespace matters because Rust resolves an unqualified type such as `Params` where generated code is emitted:

```text
Before: child::kernel(child::Params) -> parent launcher could accept parent::Params
After:  child::kernel(child::Params) -> child launcher accepts only child::Params
```

Each participating inline namespace receives its own `LoadedModule` view. A child view is constructed from its immediate parent and shares the same loaded CUDA module and generic-function cache.

### Namespace-preserving launchers

- Root kernels remain methods on `kernels::LoadedModule`.
- Child kernels become methods on `kernels::child::LoadedModule`.
- Deeper namespaces chain `LoadedModule::from_parent` through their immediate parent.
- Parameter types, private generic helpers, `pub(super)` visibility, raw module identifiers, and conditional compilation remain in the kernel's defining Rust scope.

```rust
let root = kernels::load(&ctx)?;
let child = kernels::child::LoadedModule::from_parent(&root)?;
child.run(&stream, config, kernels::child::Params { /* ... */ })?;
```

### Safe module boundary

- Inline module bodies are traversed recursively.
- File-backed modules, name-value `#[path]` modules, and `include!` invocations are preserved but not inspected.
- Kernels hidden behind those boundaries do not receive generated launchers.
- The procedural-macro filesystem crawler and partial rustc module-resolution emulation were removed.

### PTX identity restriction

Launcher methods are namespace-qualified, but PTX entry names are still bare function names. Kernel names must therefore remain unique throughout one `#[cuda_module]` inline tree. Duplicate sibling names, including cfg-gated alternatives, produce a clear diagnostic instead of risking lookup of the wrong entry.

Raw module identifiers are supported. Raw kernel-function identifiers remain an older unsupported PTX-naming edge and are not presented as part of this feature.

### Artifact and cfg handling

- Root `load()` pins every active non-generic artifact in the complete inline tree.
- Anchor references carry effective ancestor-plus-local `cfg` and `cfg_attr` conditions.
- A root module containing only nested kernels remains independently loadable from a library archive.
- Nested views share the parent's module and generic-function cache.
- `LoadedModule` is explicitly reserved in participating inline namespaces. Generated method names are also guarded: `as_cuda_module` in every namespace and `from_parent` in child namespaces.

### Examples and tests

- Reworked `cuda_module_nested` around namespace-local views and inline-only nesting.
- Extended `cuda_module_in_lib` so a nested-only library artifact must survive `.rlib` extraction.
- Added compile pass/fail coverage for shadowed 4-byte/16-byte types, cfg ancestry, direct and nested `cfg_attr`, private generics, restricted visibility, raw module names, raw/plain collision guards, reserved generated names, duplicate PTX names, deep bridge modules, and file/`#[path]`/include boundaries.

## Testing

- [x] Clippy for `cuda-macros`, all targets/features, with warnings denied
- [x] `cargo test -p cuda-macros`
- [x] `cargo test -p cuda-macros --features async`
- [x] `cargo test -p cuda-host`
- [x] `cargo test -p cuda-host --features async`
- [x] `cargo test --workspace`
- [x] `cargo oxide build cuda_module_nested --arch sm_80`
- [x] Expanded Rust and generated PTX inspected; all four nested entries are present
- [x] `cargo oxide build cuda_module_in_lib --arch sm_80`
- [x] Nested-only library and binary bundles verified in the final executable and `.oxart` section
- [ ] GPU execution after the maintainer repair; no working NVIDIA driver was available

## Checklist

- [x] Contributor commit and authorship preserved
- [x] Contributor and maintainer commits have DCO `Signed-off-by` trailers
- [x] No `crates/cuda-bindings/` changes
- [x] SPDX headers included on new Rust files


nihalpasham:
Thanks for the PR. This (i.e. nested kernel organization) is useful. 

But it needed a few additions that keeps each launcher in its defining inline namespace and removes the procedural-macro filesystem crawler. 
- The four-byte parent versus 16-byte child type case (in the PR description) now fails at compile time; 
- `cfg`/`cfg_attr`, private generics, restricted visibility, raw module names, nested-only, library artifacts, and file/include boundaries have regression coverage.
- All compiler, async, lint, PTX-generation, and archive checks pass.

PR #327, opened 2026-07-02
Add mma.sync.aligned.m16n8k16 f16 tensor core intrinsic

honeyspoon:
## Summary

Adds the raw Ampere F16 warp-level matrix multiply-accumulate operation:

```ptx
mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32
    {%d0, %d1, %d2, %d3},
    {%a0, %a1, %a2, %a3},
    {%b0, %b1},
    {%c0, %c1, %c2, %c3};
```

This is the low-level, unsafe register atom used to build typed matrix and tile APIs. It partially addresses tracking issue #274; it does not close the issue.

## What changed

- Added the `cuda-device` stub `mma_m16n8k16_f32_f16`.
- Added the dialect-nvvm operation, MIR importer handling, register-only convergent lowering, and PTX 7.0 / `sm_80` target detection.
- Documented the exact per-lane A/B/C/D coordinates, Rust-array-to-PTX register order, packed F16 order, and warp participation contract.
- Made the shared MMA fragment diagnostic operation-neutral.
- Added positive and negative operation-verifier tests.
- Added a checked-in device-stub -> MIR importer -> lowering -> PTX regression.

## Safety and register contract

```text
each lane: C[f32 x4] + A[b32 x4] * B[b32 x2]
                         |
                         v
                     D[f32 x4]
```

All 32 lanes must execute the same instruction with the same qualifiers, with no lane exited. A and B are raw `.b32` carriers containing two F16 values each in low-to-high order. Passing another lane layout computes a different matrix operation. The instruction requires PTX ISA 7.0 and `sm_80` or newer.

## Testing

- [x] `cargo test -p cuda-device --lib`
- [x] `cargo test -p dialect-nvvm --all-targets` (17 passed)
- [x] `cargo test -p mir-importer --lib` (66 passed)
- [x] `cargo test -p mir-lower --all-targets` (85 unit + 40 integration passed)
- [x] Focused strict clippy, formatting, and diff checks
- [x] `cargo oxide run standalone_device_fn --arch sm_80` (10/10)
- [x] Generated exact PTX 7.0 / `sm_80`; CUDA 13.3 `ptxas` accepted it
- [x] Explicit `sm_75` compilation rejected the operation with the required `sm_80` floor
- [ ] Full-warp numerical comparison on a GPU, including Compute Sanitizer

## Checklist

- [ ] Original contributor commit has the required DCO `Signed-off-by` trailer
- [x] Maintainer follow-up commit is signed off
[...]

PR #328, opened 2026-07-02
Add mma.sync.aligned.m16n8k8 tf32 tensor core intrinsic

honeyspoon:
## Summary

Adds the raw Ampere TF32 warp-level matrix multiply-accumulate operation:

```ptx
mma.sync.aligned.m16n8k8.row.col.f32.tf32.tf32.f32
    {%d0, %d1, %d2, %d3},
    {%a0, %a1, %a2, %a3},
    {%b0, %b1},
    {%c0, %c1, %c2, %c3};
```

This is the low-level, unsafe register atom used to build typed matrix and tile APIs. It partially addresses tracking issue #274; it does not close the issue.

## What changed

- Added the `cuda-device` stub `mma_m16n8k8_f32_tf32`.
- Added the dialect-nvvm operation, MIR importer handling, register-only convergent lowering, and PTX 7.0 / `sm_80` target detection.
- Documented the exact per-lane A/B/C/D coordinates, Rust-array-to-PTX register order, and warp participation contract.
- Documented that A and B must already contain valid TF32 register values. Ordinary `f32::to_bits()` is not a TF32 conversion; use `cvt.rna.tf32.f32` or an equivalent TF32-producing operation.
- Made the shared MMA fragment diagnostic operation-neutral and added verifier boundary tests.
- Added checked-in raw and f32-to-TF32 device paths through the complete compiler pipeline.
- Integrated current `main` after #327 while retaining the complete F16 and TF32 APIs, detectors, dispatch arms, lowerings, tests, and PTX checks.

## Safety and TF32 contract

```text
six lane-local f32 values
          |
          | cvt.rna.tf32.f32 x6
          v
 A[tf32 x4], B[tf32 x2] -> mma.sync -> D[f32 x4]
```

All 32 lanes must execute the same instruction with the same qualifiers, with no lane exited. A and B use raw `.b32` TF32 register carriers in the documented lane layout. The instruction requires PTX ISA 7.0 and `sm_80` or newer.

## Testing

- [x] `cargo test -p cuda-device --lib`
- [x] `cargo test -p dialect-nvvm -p mir-importer -p mir-lower` (18 dialect verifier, 67 importer, 85 lowering unit, and 41 lowering integration tests passed)
- [x] Strict workspace and standalone-example clippy with warnings denied
- [x] Root, rustc-codegen-cuda, and complete per-example formatting checks
- [x] `cargo oxide run standalone_device_fn --arch sm_80` (14/14)
- [x] Generated one exact F16 MMA, two exact TF32 MMAs, and six `cvt.rna.tf32.f32` instructions
- [x] Generated PTX 7.0 / `sm_80`; CUDA 13.3 `ptxas` accepted it
- [x] Explicit `sm_75` compilation rejected the TF32 operation with the required `sm_80` floor
- [ ] Full-warp numerical comparison on a GPU, including Compute Sanitizer

## Checklist

- [ ] Original contributor commit has the required DCO `Signed-off-by` trailer
- [x] Maintainer follow-up and merge commits are signed off
- [x] No `crates/cuda-bindings/` changes in the PR delta


nihalpasham:
Thanks for the PR 

The raw `TF32 MMA` atom and register order are correct, and I added a follow-up that makes the `TF32` producer rule and exact lane map explicit. The checked-in path emits six
- `cvt.rna.tf32.f32` conversions feeding the MMA, passes 12/12, and assembles  for `sm_80`. 

**Reviewer note:** A typed TF32 wrapper can be layered above this raw primitive. This partially addresses #274.

PR #329, opened 2026-07-02
Add mma.sync.aligned.m16n8k32 s8 tensor core intrinsic

honeyspoon:
## Summary

Adds the raw Ampere signed-INT8 warp-level matrix multiply-accumulate operation:

```ptx
mma.sync.aligned.m16n8k32.row.col.s32.s8.s8.s32
    {%d0, %d1, %d2, %d3},
    {%a0, %a1, %a2, %a3},
    {%b0, %b1},
    {%c0, %c1, %c2, %c3};
```

This is the low-level, unsafe register atom used to build typed quantized matrix and tile APIs. It partially addresses tracking issue #274; it does not close the issue.

## What changed

- Added the `cuda-device` stub `mma_m16n8k32_s32_s8`.
- Added the dialect-nvvm operation, MIR importer handling, and register-only convergent lowering.
- Documented the exact per-lane A/B/C/D coordinates and low-byte-to-high-byte packing of four signed 8-bit values in each `u32` carrier.
- Documented that this non-`.satfinite` form wraps signed accumulator overflow.
- Extended target detection to recognize both legal plain and `.satfinite` spellings while rejecting near misses.
- Made the shared MMA fragment diagnostic operation-neutral and added verifier boundary tests.
- Added a checked-in device-stub -> MIR importer -> lowering -> PTX regression.
- Integrated current `main` while keeping the BF16, F16, TF32, INT8, and FP64 paths separate.

## Safety and integer contract

```text
each lane: C[s32 x4] + A[b32 x4] * B[b32 x2]
             wrapping    four s8 values per b32
                              |
                              v
                          D[s32 x4]
```

All 32 lanes must execute the same instruction with the same qualifiers, with no lane exited. A and B must use the documented per-lane layout and byte order. The instruction requires PTX ISA 7.0 and `sm_80` or newer.

## Testing

- [x] `cargo test -p cuda-device -p dialect-nvvm -p mir-importer -p mir-lower`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] Root, `rustc-codegen-cuda`, and every example crate pass `cargo fmt --all -- --check`
- [x] `cargo oxide run standalone_device_fn --arch sm_80` passes 18/18
- [x] Generated PTX is version 7.0 targeting `sm_80`, with F16 MMA = 1, TF32 MMA = 2, INT8 MMA = 1, TF32 conversions = 6, and INT8 `.satfinite` = 0
- [x] CUDA 13.3 `ptxas` accepts the integrated `sm_80` PTX
- [x] `ptxas` rejects the same INT8 instruction at `sm_75` with the required `sm_80` floor
- [x] Plain and `.satfinite` detector boundary tests
- [ ] Full-warp numerical comparison on a GPU, including overflow cases and Compute Sanitizer

## Checklist

- [ ] Original contributor commit has the required DCO `Signed-off-by` trailer
- [x] Maintainer follow-up and integration commits are signed off
- [x] No `crates/cuda-bindings/` changes relative to `main`


nihalpasham:
Thanks for the PR 

The signed `INT8` operation survives the full compiler path with the correct `D4`/`A4`/`B2`/`C4` register order. I added a follow-up 
- with the exact lane/byte map, wrapping and exited-lane contract, official `.satfinite` detection, neutral diagnostics, verifier negatives, and checked-in PTX coverage. 

This partially addresses #274.

PR #330, opened 2026-07-02
Fix cargo oxide debug executable discovery

mohamedsamirx:
## Summary

Make `cargo oxide debug` launch the executable Cargo actually built instead of guessing a path under `target/release`. The same artifact-selection helper is also used by the run and sanitizer paths.

```text
Before: Cargo builds one package/target -> cargo-oxide may launch another path
After:  Cargo metadata selects packages -> Cargo JSON selects the exact executable
```

This also adds `debug --bin` / `--features` support and resolves `cuda-gdb` through the configured CUDA toolkit.

### Executable selection

- Read Cargo metadata to determine the current package or workspace default members.
- Keep each emitted executable's Cargo `package_id` and target name.
- Honor `package.default-run` and an explicit `--bin` only when Cargo emitted that exact artifact.
- Allow a unique `--bin` to disambiguate multiple default members.
- Ignore library-only members and executable artifacts from dependencies or other packages.
- Treat zero or multiple matching artifacts as an error instead of silently choosing another program.
- Choose from Cargo's emitted artifacts rather than metadata target declarations, so bins disabled by `required-features` are handled correctly.

### Backend build isolation

- Build `rustc_codegen_cuda` explicitly for Cargo's host tuple in its own absolute target directory.
- Remove inherited application `CARGO_BUILD_TARGET` settings from the backend build.
- Read the exact backend dylib path from Cargo's JSON output.
- Reject missing or ambiguous backend artifacts; there is no fallback to a stale guessed `target/debug` path.
- Make `cargo oxide doctor` report the corresponding host-triple path.

### CLI and documentation

- Add `cargo oxide debug --bin <name>` and `--features <features>`.
- Use the configured CUDA toolkit when locating `cuda-gdb`.
- Document the new debug behavior in the cargo-oxide README and book.

## Safety and failure behavior

The selection rule is deliberately strict:

```text
selected Cargo packages
        -> emitted bin artifacts
        -> --bin / default-run filter
        -> exactly one executable, or a clear error
```

This prevents a successful build from causing cuda-oxide to debug a stale binary, a build-tool artifact, or a different workspace member.

## Testing

- [x] `cargo test -p cargo-oxide` (72 tests)
- [x] `cargo clippy -p cargo-oxide --all-targets -- -D warnings`
- [x] `cargo check --workspace --locked`
- [x] `git diff --check`
- [x] Real backend `setup` with an invalid inherited `CARGO_BUILD_TARGET` and a separate application `CARGO_TARGET_DIR`
- [x] `cargo oxide doctor` resolves the host-triple backend path
- [ ] GPU execution (not required for this host-tooling change)

Regression coverage includes configured target directories, custom bin names, virtual and non-virtual workspace defaults, multiple default members, `default-run`, library-only members, dependency artifacts, duplicate bin names, missing requested targets, `required-features`, and stale backend-path rejection.

### DIFF
--- a/crates/cargo-oxide/README.md
+++ b/crates/cargo-oxide/README.md
@@ -45,7 +45,8 @@ cargo oxide setup                   # explicitly build the codegen backend
 |------------------------------|----------------------------------|-------------------------------------------------|
 | `--emit-nvvm-ir`             | run, build, pipeline             | Generate NVVM IR for libNVVM                    |
 | `--arch <sm_XX>`             | run, sanitize, build, test, pipeline, emit-ltoir | Target architecture override |
-| `--features <F>`             | run, sanitize, build, build passthrough, emit-ltoir | Comma-separated cargo features to enable |
+| `--features <F>`             | run, sanitize, debug, build, build passthrough, emit-ltoir | Comma-separated cargo features to enable |
+| `--bin <NAME>`               | run, sanitize, debug                | Specific binary target to build/run      |
 | `--tool <T>`                 | sanitize                         | Compute Sanitizer tool: `memcheck`, `racecheck`, `initcheck`, or `synccheck` |
 | `-o, --output <P>`           | emit-ltoir                       | Output path for the `.ltoir` artifact           |
 | `--cargo-target-dir <PATH>`  | build/test passthrough           | Cargo target directory                          |
@@ -216,7 +217,13 @@ cargo oxide pipeline device_ffi_test --emit-nvvm-ir --arch sm_120
 
 ### `cargo oxide debug <example>`
 
-Builds with debug info (`-C debuginfo=2`) and launches cuda-gdb. Supports `--tui` for GDB's TUI mode and `--cgdb` for the cgdb frontend.
+Builds with debug info (`-C debuginfo=2`) and launches cuda-gdb. Supports
+`--tui` for GDB's TUI mode and `--cgdb` for the cgdb frontend.
+
+`debug` uses Cargo's reported executable artifact, so custom binary names,
+`package.default-run`, configured target directories, host target triples, and
+virtual workspaces do not require path guessing. Use `--bin <name>` when a
+package has multiple binaries and no unambiguous default.
 
 ### `cargo oxide new <name> [--async]`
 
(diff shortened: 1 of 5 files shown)


nihalpasham:
Thanks @mohamedsamirx - using cargo’s reported executable path makes sense, and the normal single-package cases work. 

A couple of things to fix before this can merge:
- Keep each artifact’s `package_id` and select only the executable belonging to cargo’s chosen package and target. If `default-run` or `--bin` names a target cargo did not build, return an error instead of silently debugging another binary.
- Use cargo metadata to honor a virtual workspace’s default member and that package’s `default-run`.
- Ignore executable artifacts from binary dependencies. They are build tools, not the program the user requested.
- Isolate the rustc-codegen backend build from the application’s `CARGO_TARGET_DIR`, or obtain the backend path from cargo’s artifact output. A fresh custom target directory currently makes cargo-oxide look in the wrong place.

Please add one regression test for each case. The existing artifact-path and toolkit-lookup work can stay as it is.

mohamedsamirx:
Thanks @nihalpasham, addressed in the latest push.

What changed:

- Kept `package_id` with each Cargo executable artifact and now select only the artifact matching Cargo’s selected package and target.
- Use `cargo metadata` to resolve the package being debugged, including virtual workspace `default-members` and the selected package’s `default-run`.
- Return an error when `--bin` or `default-run` names a target that Cargo did not build, instead of falling back to another executable.
- Ignore executable artifacts from other packages, including binary/build-tool dependencies.
- Isolated the rustc-codegen backend build from the application target directory by forcing the backend build to its own local `target` dir.

Added regression coverage for each requested case:

- virtual workspace default member + package `default-run`
- requested `--bin` target not built
- `default-run` target not built
- executable artifacts from another package
- backend build isolation from `CARGO_TARGET_DIR`

Validation:

```bash
cargo fmt --check
cargo test -p cargo-oxide
cargo clippy -p cargo-oxide -- -D warnings
git diff --check origin/main..HEAD

nihalpasham:
The update version looks good. I added a quick follow-up for a few workspace and host-target edge cases we found during adversarial testing. The final branch now follows cargo's 

- selected packages and emitted artifacts, rejects missing or ambiguous targets, and obtains the backend from Cargo's exact host build output.

PR #336, opened 2026-07-03
fix: avoid auto-NVVM when module uses both f16 types and libdevice

honeyspoon:
## Summary

- When a kernel uses both f16 types (MMA/WMMA intrinsics) and libdevice calls (`__nv_*` math functions), the pipeline previously auto-forced NVVM IR mode
- On pre-Blackwell targets (sm<100), this selects the legacy LLVM 7 dialect which rejects scalar f16 — creating an impossible constraint
- Now detects f16 types in the module before choosing the compilation path and stays on PTX when both are present, with a warning that `__nv_*` calls must be linked externally

## Details

**Root cause**: `pipeline.rs:462` auto-forces NVVM IR when `__nv_*` libdevice symbols are detected. On sm<100, NVVM IR uses the legacy LLVM 7 dialect (`legalize.rs:172-235` rejects f16). Code using both libdevice functions AND f16 types (e.g., MMA/WMMA + `sqrt()`) hit an impossible constraint.

**Fix**:
- Added `module_contains_f16()` / `op_contains_f16()` functions that recursively scan the module for f16 types in operands, results, and function signatures
- Modified the auto-NVVM logic: when both f16 types and libdevice calls are present AND the user did not explicitly request `--emit-nvvm-ir`, stay on the PTX path
- Emit a warning explaining the conflict and that `__nv_*` calls must be linked externally (via `llvm-link` or the LTOIR pipeline)

**Behavior changes**:
- Modules with only libdevice calls (no f16): unchanged, still auto-NVVM
- Modules with only f16 types (no libdevice): unchanged, stays on PTX
- Modules with both: **new** — stays on PTX with warning (previously failed with f16 rejection error)
- Explicit `--emit-nvvm-ir`: unchanged, always forces NVVM IR regardless

## Test plan

- [x] 4 new unit tests for `module_contains_f16()`:
  - `test_module_contains_f16_detects_f16_return_type`
  - `test_module_contains_f16_detects_f16_arg_type`
  - `test_module_contains_f16_negative_for_f32_only`
  - `test_module_contains_f16_negative_for_i32_only`
- [x] All 72 `mir-importer` tests pass
- [x] `cargo fmt` clean
- [x] `cargo clippy` clean

## Related

- `crates/nvvm-transforms/src/legalize.rs:172-235` — f16 rejection check in legacy NVVM
- `crates/libnvvm-sys/src/lib.rs:81-84` — `uses_legacy_llvm()` returns true for sm<100
- See analysis: `analysis/upstream-bug-const-generic-naming.md` (secondary issue section)

### DIFF
--- a/crates/mir-importer/src/pipeline.rs
+++ b/crates/mir-importer/src/pipeline.rs
@@ -460,8 +460,23 @@ pub fn run_pipeline(
     // The example is then expected to feed the `.ll` through the LTOIR
     // pipeline (compile_ltoir + link_ltoir) and load the resulting cubin.
     let needs_libdevice = module_uses_libdevice(&ctx, module_op_ptr);
-    let emit_nvvm_ir = config.emit_nvvm_ir || needs_libdevice;
-    if needs_libdevice && !config.emit_nvvm_ir && config.verbose {
+    // When a pre-Blackwell module uses BOTH libdevice (__nv_* calls) AND f16
+    // types (MMA/WMMA intrinsics), auto-forcing NVVM IR mode would fail
+    // because the legacy LLVM 7 dialect rejects scalar f16. In that case,
+    // stay on the PTX path and let the user link libdevice externally (via
+    // llvm-link or the LTOIR pipeline). This matches how build_and_run.sh
+    // already handles the combination.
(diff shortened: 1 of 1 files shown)


honeyspoon:
Opened on wrong repo — moving to fork.

PR #340, opened 2026-07-04
feat(codegen): opt-in build-time cubin materialization for embedded artifacts

bitemyapp:
## What this PR does

Some cuda-oxide deployments finish compiling embedded NVVM IR or LTOIR when
the program first loads. That adds startup work and can require libNVVM and
nvJitLink on the deployment machine.

This PR adds an explicit opt-in that finishes that compilation during the host
build and embeds native GPU code instead:

    cargo oxide build vecadd --materialize-cubin --arch sm_86
    NVVM IR + libdevice -> LTOIR -> nvJitLink -> embedded sm_86 cubin

Default artifact behavior is preserved: without `--materialize-cubin`, builds
continue to embed PTX.

## Why this is useful

For a deployment with a known GPU architecture, the program can load the cubin
directly through the CUDA driver. It does not need to run libNVVM or nvJitLink
at startup.

The trade-off is intentional: a cubin targets one real GPU architecture. It
does not provide the forward compatibility or multiple architectures of PTX
or a CUDA fatbinary.

## How it works

- Build-time and load-time CUDA finalization share one driver-independent
  implementation.
- The selected architecture, FMA policy (including `--no-fmad`), and debug
  policy are preserved through libNVVM and nvJitLink.
- Cache keys include the exact source artifact, ordered options, compiler and
  linker identities, toolkit inputs, and `libdevice`.
- The codegen backend still has no direct dependency on the CUDA driver or
  toolkit libraries.
- Missing architecture, incompatible output modes, multiple final artifacts,
  and unresolved late-link inputs fail with clear errors instead of producing
  an incomplete cubin.

## Cargo cache correctness and CI performance

The previous head was not a performance repair. It made target, output, and
materializer fingerprints global `CARGO_ENCODED_RUSTFLAGS`, so every mode or
architecture change invalidated Cargo's shared dependency cache.

Hosted comparison before this repair:

- `main` (`52accfa4`): compile-examples step 10m25s.
- Previous PR head (`7616282c`): 22m52s.
- `generated_intrinsics`: 3s -> 127s.
- `generated_intrinsics_blackwell`: 2s -> 122s.
- `tcgen05`: 4s -> 124s.

This repair splits the cache boundary according to ownership:

- The exact backend `.so` SHA-256 remains global because that backend compiles
  every crate.
- Architecture, output mode, FMA/debug policy, owner filter, and exact
  materializer provenance are a full SHA-256 environment dependency recorded
  by CUDA procedural macros only in crates that can own or instantiate device
  artifacts.
- Exact materializer provenance is tracked separately and is still rediscovered
  and byte-checked by the backend before cubin generation.

The workflow now checks Cargo compilation events rather than imposing a timing
threshold. PTX -> cubin -> PTX may rebuild `vecadd`, and an `sm_86` -> `sm_90`
cross-crate switch may rebuild `kernel-lib` plus its consumer, but either
transition fails CI if an unrelated shared dependency recompiles.

Local validation of the repaired candidate compiled all 145 examples in 181s.
The formerly regressed examples completed in 2s, 2s, and 3s respectively. A
focused retained-log run confirmed that each compiled only its owning example
crate. These local timings are supporting evidence; the authoritative hosted
comparison will be recorded after the repaired head runs in CI.

## Validation

- 86 `cargo-oxide` tests, including cold/warm/architecture/output/provenance
  Cargo JSON `fresh` assertions for an owner, downstream consumer, and plain
  shared dependency.
- 69 CUDA macro unit tests and the existing macro compile-test suite.
- 17 codegen-backend tests.
- Strict clippy for the affected root and backend crates.
- Full GPU-less compile sweep: 145/145 examples passed in 181s.
- Real PTX -> cubin -> PTX and `sm_86` -> `sm_90` cross-crate cache transitions.
- Extracted cubin inspection: CUDA ELF `sm=86` with exported `vecadd` entry.
- Current macros compiled successfully with the exact previous backend head,
  which recognized `vecadd` and embedded a non-empty PTX artifact.

No CUDA device was available, so this validation does not make a GPU-execution
claim.

## Issue relationship

Issues #378, #388, and #390 are related, but this PR does not close them.

## History

The contributor-authored, DCO-compliant commit is preserved. Separate signed
maintainer commits rebuild the feature around the shared finalizer and repair
the Cargo cache boundary.

### DIFF
--- a/.github/workflows/examples-compile.yml
+++ b/.github/workflows/examples-compile.yml
@@ -22,7 +22,9 @@ jobs:
     # backend build + shared dep tree + per-example
     # device codegen. Typically well under an hour with the shared
     # CARGO_TARGET_DIR; the timeout leaves headroom for cold runners.
-    timeout-minutes: 60
+    # Leave job-level headroom around the 50-minute full compile and the
+    # focused 10-minute materialization gate, plus toolkit/LLVM setup.
+    timeout-minutes: 75
     steps:
       - name: Clone repository
         uses: actions/checkout@v4
@@ -36,7 +38,11 @@ jobs:
           cuda: "13.1.1"
           method: network
           sub-packages: '["nvcc", "cudart-dev"]'
-          non-cuda-sub-packages: '["libcurand-dev"]'
+          # nvJitLink is packaged separately from cuda-nvcc on Linux. Keep it
+          # in this already-toolkit-backed job so the materialization gate
+          # below exercises the real public command without adding work to
+          # the latency-sensitive fmt job.
+          non-cuda-sub-packages: '["libcurand-dev", "libnvjitlink-dev"]'
 
       # The device pipeline shells out to LLVM's llc/opt for PTX
       # generation. LLVM 21 is the documented floor for the intrinsics we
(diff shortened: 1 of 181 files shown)


nihalpasham:
Thanks for the PR. I pushed a maintainer follow-up to fix the Cargo cache regression introduced while completing the cubin materialization path.

- `--materialize-cubin` still finishes compilation during the host build and embeds an architecture-specific cubin.
- Architecture and PTX/cubin changes now rebuild only the CUDA crate and its consumers, rather than unrelated shared dependencies.
- Backend and materializer inputs remain fingerprinted, and stale output from older macros is rejected.

For example:

```text
Before: sm_86 -> sm_90 -> shared dependencies + kernel + application rebuild
After:  sm_86 -> sm_90 -> kernel + application rebuild
```

**Reviewer note:** All 145 compile-only examples passed locally in 181s; the three previously regressed examples completed in 2s, 2s, and 3s. Package and backend tests, strict clippy, formatting, compatibility checks, and independent post-repair review passed. No CUDA device was available, so this is compile and artifact validation only. Hosted CI is pending and remains the final merge gate.

#378, #388, and #390 are related; this PR does not close them.

PR #342, opened 2026-07-04
fix(cargo-oxide): preserve Cargo test profile flags

### LINKED ISSUE #341: `cargo oxide` compiles tests without `debug_assertions`
Thanks for making this tool. I found #288 particularly useful, as I could now _test_ my code, as well as just running it. However, I noticed that the tests are run with heavy optimizations and without `debug_assertions`. 

Here is an example piece of code that fails to compile. My main reason for using this is that I can do an expensive check while testing, but not include it in the release build.

```rust
#[cfg(debug_assertions)]
fn test_sum() {
    // Usually an expensive check.
    let x = 1;
    debug_assert_eq!(x + 1, 2);
}

#[test]
fn test_test_sum() {
    test_sum()
}
```

George-Ogden:
Fixes #341

## Summary

`cargo oxide test` now leaves optimization, debug assertions, and debug-info
settings to the Cargo profile selected by the user. Plain tests therefore use
Cargo's test/dev defaults, while `--release` and `--profile <name>` retain their
own policies.

`cargo oxide build` keeps cuda-oxide's release-like defaults, and
`cargo oxide debug` keeps release-like optimization plus debug information.
Backend selection, the JumpThreading restriction, symbol mangling, device
cfgs, and the codegen fingerprint remain enforced in every mode.

## Implementation

- Uses a small typed passthrough subcommand and explicit profile policy.
- Adds no `strum` dependency and no unrelated `Context: Clone`.
- Tests default test mode, `--release`, custom `--profile`, build mode, debug
  mode, and compiler invariants.

## Maintainer verification

- `cargo check -p cargo-oxide`
- `cargo test -p cargo-oxide` — 73 passed
- `cargo clippy -p cargo-oxide -- -D warnings`
- Independent post-repair audit — passed

The original contributor commit had no DCO trailer. The final branch was
independently reimplemented from current `main` as a maintainer-authored,
signed-off commit; no sign-off was added on the contributor's behalf.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -248,7 +248,7 @@ pub fn codegen_run(
     }
 
     apply_common_codegen_env(&mut cmd, ctx, verbose, no_fmad);
-    apply_codegen_rustflags(&mut cmd, ctx, false, &[]);
+    apply_codegen_rustflags(&mut cmd, ctx, CodegenProfilePolicy::ReleaseLike, &[]);
     apply_output_mode(&mut cmd, emit_nvvm_ir, target_arch);
     apply_device_arch_hint(&mut cmd, target_arch, detected_device_arch.as_deref());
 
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks for the PR and for identifying the underlying test-profile problem.

A couple of points. Added fixes for how `cargo oxide test` handles Cargo profiles. Tests now respect the profile Cargo selected instead of silently receiving cuda-oxide’s release-style optimization and assertion settings.

```sh
    Before:
      cargo oxide test
      -> release-like device flags could remove debug assertions

    After:
      cargo oxide test
      -> Cargo controls optimization, assertions, and debug information
```
The original contribution identified the correct problem, but it also forced debug information, added unnecessary machinery, and lacked DCO. The branch was independently reimplemented from current `main` as a signed, DCO-compliant maintainer commit. Required backend, mangling, MIR, cfg, and fingerprinting flags remain enforced.

Verification passed:

- Formatting
- `cargo check -p cargo-oxide`
- All 73 `cargo-oxide` tests
- Strict package clippy
- Independent post-repair review

This closes #341. The repaired version has now been merged.

Issue #337, opened 2026-07-03
Tracking: SM89 (Ada Lovelace) device intrinsics — FP8 types, conversions, and MMA


honeyspoon:
## What problem does this solve?

Ada Lovelace (SM89) introduced FP8 data types (E4M3 and E5M2) and FP8 tensor core operations. This tracks adding FP8 conversions, FP8 MMA, and FP8 sparse MMA for SM89+.

Split from #335 for per-architecture tracking. See also: SM80 (#274), SM90 (#338), SM100 (#339).

## Status (2026-07-27)

PR #406 (pinned-LLVM intrinsics autogen) landed most of this issue:

- **Phase 2 (FP8 MMA): complete.** All eight checklist shapes are generated in `crates/cuda-device/src/generated/register_mma.rs` as the `mma_m16n8k16_fp8_*` and `mma_m16n8k32_fp8_*` families, each in both f32- and f16-accumulator forms (16 functions), lowering to e.g. `mma.sync.aligned.m16n8k16.row.col.f32.e4m3.e4m3.f32`. Blackwell-class `kind::f8f6f4` m16n8k32 variants (FP8 mixed with FP6/FP4 inputs) were generated alongside.
- **Phase 1 (conversions): partial (4 of 10).** The four f32-source packs (`cvt_rn_satfinite[_relu]_{e4m3x2,e5m2x2}_f32`) are generated in `crates/cuda-device/src/generated/convert.rs`. The f16x2-source packs, the unpacks back to f16x2, and the scalar `cvt.f32.e4m3` / `cvt.f32.e5m2` are still missing.
- **Phase 3 (sparse MMA): not covered for sm_89.** What exists is the `sp::ordered_metadata` + `kind::f8f6f4` form (32 functions in `crates/cuda-device/src/generated/sparse_mma.rs`, f16 and f32 accumulators), which targets newer architectures. The plain sm_89 `mma.sp.sync.aligned.m16n8k64` FP8 forms tracked below are still unimplemented, so those boxes stay unchecked.

Remaining items go through the #406 workflow: append to `intrinsics/overlay/*.toml` + `intrinsics/abi-v1.toml` and regenerate.

## Scope

Each item follows the established 4-layer pattern (`cuda-device` > `dialect-nvvm` > `mir-importer` > `mir-lower`) plus lowering tests.

---

### Phase 1: FP8 conversions (sm_89+, PTX ISA 8.1+)

[...]

PR #344, opened 2026-07-04
Add cvt.rn.satfinite.e4m3x2.f32 packed FP8 conversion

obeis:
## Summary

Adds the first FP8 conversion from #337 (Phase 1): `cvt.rn.satfinite.e4m3x2.f32`, which packs two f32 values into an e4m3x2 pair (u16) with round-to-nearest-even and saturate-to-finite. This is the foundation for FP8 compute on sm_89+ (Ada Lovelace). Wired through all four layers following the established pattern, with a lowering test and a hardware-verified example. Unlike the existing f16x2/bf16x2 packs, the result is a 16-bit register (`=h` constraint), since the two FP8 values occupy 8 bits each.

## Changes

- `cuda-device`: new `convert::cvt_rn_satfinite_e4m3x2_f32(lo, hi) -> u16` intrinsic (sm_89+, PTX ISA 8.1+)
- `dialect-nvvm`: new `CvtRnSatfiniteE4m3x2F32Op` (2 f32 operands, 1 result) + registration
- `mir-importer`: intrinsic dispatch arm and emitter producing the op with a u16 result
- `mir-lower`: inline-asm lowering to `cvt.rn.satfinite.e4m3x2.f32 $0, $2, $1;` with `"=h,f,f"` + `MirToLlvmConversion` impl
- `mir-lower` tests: `test_cvt_rn_satfinite_e4m3x2_f32_lowers_to_inline_asm` (suite: 37/37 pass)
- New `cvt_e4m3x2` example: verifies 256 packed values bit-exactly against a host-side nearest-even e4m3 reference, pins lane order with constant operands (`0xC13C`), checks satfinite clamping to ±448 and the NaN lane pattern; skips on pre-sm_89 GPUs (verified passing on sm_120)


nihalpasham:
Thanks for the PR.

Packed E4M3 FP8 conversion is useful. But, current `main` implements this conversion through the generated intrinsic framework. (merged in with PR #406 )

```
    low  = 1.5   -> 0x3c
    high = -2.25 -> 0xc1

    packed result = 0xc13c
```

- The public API
- A generated typed operation and verifier
- Typed LLVM NVPTX lowering where available
- A reviewed inline-PTX fallback for libNVVM
- Target-version checks and tests

Note: the 256-case GPU numerical example is still valuable. I would welcome that as a fresh tests-only PR using the generated API, with explicit infinity handling in the scalar reference.

PR #345, opened 2026-07-06
Fix collector ICE on closures from non-local crates

mbrobbel:
## Summary

Fixes a compiler ICE when device code reaches a closure defined in a dependency
crate.

`should_collect_from_crate` used `tcx.item_name` while checking whether a
cross-crate item was a kernel entry point. Closures have no item name, so that
query ICEs. The fix uses `opt_item_name`; unnamed items skip the kernel-name
exception and continue through the existing crate policy.

The regression example keeps its closures in an external helper crate and
routes them through `FnOnce` receivers. `gated_load` uses an `if` plus a closure
passed to `apply`; `double_even` uses `bool::then`; neither uses `Option::map`.

## Maintainer verification

- Root, backend, and example formatting checks passed.
- Example and backend all-target checks passed with locked dependencies.
- `scripts/smoketest.sh --compile-only -o '^extern_crate_closure$'` — 1/1
  compiled.
- Independent post-repair audit — passed.

The compiler fix remains the contributor's signed commit. The documentation
correction is a separate signed maintainer commit.


nihalpasham:
Thanks for the PR. LGTM.

The compiler now uses `opt_item_name` when inspecting cross-crate items, so unnamed definitions such as closures follow the normal collection policy instead of causing an ICE.

I pushed a small maintainer follow-up to correct the example documentation; the compiler fix and contributor history were preserved.

**Reviewer note:** Formatting, relevant checks, and the focused external-crate closure compile test passed. There is no matching open issue to close.

PR #347, opened 2026-07-07
Extend cast lowering coverage

uurl:
Follow-up coverage related to #303; this PR does not close that issue because
its listed acceptance cases already landed independently on `main`.

## Summary

Adds focused unit coverage for existing MIR cast lowering paths:

- same-width integer casts to `llvm.bitcast`;
- signed integer-to-float casts to `llvm.sitofp`;
- unsigned float-to-integer casts through `llvm_fptoui_sat_*`;
- float widening and narrowing through `llvm.fpext` / `llvm.fptrunc`;
- pointer-with-exposed-provenance casts to `llvm.inttoptr`;
- cross-address-space pointer casts to `llvm.addrspacecast` with the destination
  address space checked.

Every route asserts exactly one expected operation and no residual `mir.cast`.
The saturating paths also check the exact direct callee and declaration.

## Maintainer verification

- Focused cast tests — 13 passed
- Full `mir-lower` suite — 135 passed; 7 doctests ignored
- Library clippy with warnings denied — passed
- All-target clippy still reports ten pre-existing warnings in unchanged
  `tests/lowering_test.rs`; this PR introduces none.

Test coverage only; no production lowering behavior changes.

### DIFF
--- a/crates/mir-lower/src/convert/ops/cast.rs
+++ b/crates/mir-lower/src/convert/ops/cast.rs
@@ -1015,12 +1015,12 @@ mod tests {
     use dialect_mir::types::MirPtrType;
     use llvm_export::ops as llvm;
     use pliron::builtin::op_interfaces::{CallOpCallable, CallOpInterface, SymbolOpInterface};
-    use pliron::builtin::types::{FP32Type, IntegerType, Signedness};
+    use pliron::builtin::types::{FP32Type, FP64Type, IntegerType, Signedness};
     use pliron::context::{Context, Ptr};
     use pliron::linked_list::ContainsLinkedList;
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. LGTM.

I kept the change as submitted. The tests add direct coverage for seven MIR cast routes and verify both the expected LLVM operation and the removal of the original `mir.cast`.

For example, an `i32 -> f32` cast must produce `llvm.sitofp`, not leave an unresolved MIR cast behind.

**Reviewer note:** The focused tests and complete `mir-lower` suite passed. #303 is related, but its acceptance cases were already implemented by earlier PRs, so this PR does not close it.

PR #349, opened 2026-07-08
feat(mir): support SetDiscriminant lowering for niche-encoded enums (incl. multi-field payloads)

goog00:
## What this PR fixes

cuda-oxide previously represented some Rust enums differently on the GPU by adding a synthetic discriminant. That could disagree with the memory layout chosen by rustc, especially for niche-encoded enums such as `Option<NonZeroU32>` and `Option<&T>`.

This PR makes device enums use rustc's real physical layout. The same checked layout model is used for construction, matching, `SetDiscriminant`, payload access, constants, and transmute lowering.

For example:

    Option<NonZeroU32>::Some(7) -> physical word 7
    SetDiscriminant(None)       -> physical word 0
                                   no extra tag is added

## Why this matters

Host and device code must agree on the bytes used to represent an enum. Using rustc's layout avoids inventing a second GPU-only representation and fixes the original niche-`SetDiscriminant` support gap without creating ABI mismatches elsewhere.

## How it works

One rule drives everything: **the device uses exactly the bytes rustc chose on the host.** Anything the compiler cannot reproduce byte-for-byte is a clear compile error, never an approximation.

- **The dialect records rustc's real layout.** For each enum the importer stores what rustc decided: the layout kind (direct tag, niche, single, empty), total size and alignment, where the tag or niche carrier lives, and the byte offset of every payload field. The encoding API is typed (real enums for the kinds, one `u128` niche start, named constructor fields), so a malformed encoding is a compile error or verifier failure, not a silently transposed integer.

- **Direct tags store the declared value.**

      #[repr(i8)] enum E { Below = -1, Zero = 0 }
      // tag byte is 0xFF for Below; reading sign-extends it back to -1

- **Niche enums hide the discriminant in invalid payload values**, using rustc's exact wrapping arithmetic:

      Option<NonZeroU32>:  None = 0,  Some(n) = n      (no extra tag byte)
      Option<&T>:          None = null pointer
      // SetDiscriminant to the Some/payload variant never touches
      // the payload bytes it shares storage with

- **Tuples know their real field offsets.** rustc reorders tuple fields exactly like struct fields, and the importer now records where they landed:

      Option<(u32, u32, &T)>          // rustc puts the pointer FIRST
      bytes 0..8   = &T               // also the niche: null here = None
      bytes 8..12  = u32
      bytes 12..16 = u32

  Structs and tuples flow through the same lowering. When a payload shares bytes with a pointer carrier, lowering accepts it only if the pointer sits at the same offset, with the same width and address space, on both sides. A pointer is never rebuilt from integer bits.

- **Bools are one canonical byte, even nested inside a payload.**

      Option<struct { pad: u32, flag: bool }>
      // flag byte at offset 4:  0 = false,  1 = true,  2 = None

  LLVM's `i1` leaves seven undefined bits in its byte, so enum storage never contains a raw `i1`: writes zero-extend each bool into its byte, reads load it back as a bool.

- **Alignment always matches rustc.** If the payload bytes alone would under-align the LLVM struct (say an `i8` carrier inside a 4-aligned enum, or `repr(align(8))`), a zero-sized anchor field raises it, the same trick union storage already uses.

## Examples

- The former `error_set_discriminant_niche` support-gap fixture is now the positive `set_discriminant_niche` ABI regression, covering seven niche shapes with byte-level host/device verification, including a device-verified `Option<(u32, u32, &u32)>` tuple pointer niche and an `Option<struct { u32, bool }>` nested-bool niche.
- Its later-field case replaces the separate `set_discriminant_niche_multifield` example.
- `error_set_discriminant_uninhabited` remains a required negative test.
- The former `error_enum_nested_bool_niche` fixture is retired: its shape is now supported and runtime-verified.
- Three negative fixtures cover the remaining honest boundaries: enum-constant pointer provenance, pointer/integer overlapping payloads, and target-mode-dependent shared-memory pointers.

## Remaining boundaries (fail closed with diagnostics)

- Enum constants containing pointer relocations (needs device-global materialization).
- Pointer payloads overlapping integer storage (`repr(C)` unions of pointer and bits).
- Shared-memory (AS3) pointers in enum layouts, until the exporter's pointer-width mode is threaded into lowering.
- Pointer-bearing arrays/vectors in overlapping payload positions, and `Option<(&T, &U)>`-style second pointer words without their own pointer slot.
- Bools hidden behind union byte blobs.

This partially addresses #306, which should remain open for the boundaries above. Issues #382 and #393 are related but are not fixed by this PR.

## Validation

- Full `dialect-mir`, `mir-importer` (802), and `mir-lower` (131 unit + 73 integration) suites.
- Workspace Clippy with warnings denied, rustdoc with warnings denied.
- All 148 examples in the GPU-less compile gate.
- `set_discriminant_niche` and `option_ref_transmute` executed on real hardware (sm_120) with byte-level host/device agreement checks.

## History

The contributor-authored, DCO-compliant commit is preserved, followed by signed maintainer commits: the rustc-layout rebuild, Clippy/rustdoc cleanup, the nested-pointer provenance repair, tuple field-offset support (removing the interim two-word shape restriction), and the typed encoding API plus nested-bool canonicalization.


nihalpasham:
Thanks for the PR. The original `SetDiscriminant` support gap is fixed. I pushed maintainer follow-ups that rebuilt the enum path on rustc's real layouts so the fix holds for every enum shape the compiler accepts, not just the initial niche case.

- Device enums now use rustc's exact physical layout for construction, matching, `SetDiscriminant`, payload access, constants, and transmutes. No synthetic device-only tag is introduced.
- Niche encodings write the discriminant into the carrier rustc chose, and selecting the untagged variant never touches the payload bytes it shares storage with.
- Tuples carry rustc's field offsets like structs do, so reordered multi-field payloads such as `Option<(u32, u32, &T)>` lower byte-identically to the host.
- Bools use canonical byte storage even nested inside payloads (`Option<struct { u32, bool }>` works), and enum storage alignment always matches rustc.
- Layouts that cannot be reproduced byte-for-byte fail with a clear diagnostic; the remaining boundaries are listed in the description and covered by negative fixtures.

For example:

```
Option<NonZeroU32>::Some(7) -> physical word 7
SetDiscriminant(None)       -> physical word 0   (no extra tag)

Option<(u32, u32, &T)>      -> pointer first in memory, per rustc;
                               null in that slot means None
```

**Reviewer note:** 1037 unit and integration tests across dialect-mir, mir-importer, and mir-lower passed, plus all 148 examples in the GPU-less compile gate. Workspace clippy with warnings denied, rustdoc, and formatting are clean. set_discriminant_niche (64 byte-verified host/device roundtrips across seven niche shapes) and option_ref_transmute executed successfully on an RTX 5090 (sm_120). The branch is rebased onto current main; hosted CI on the final head is the remaining merge gate.

This partially addresses #306, which stays open for the documented unsupported layouts. #382 and #393 are related but not fixed here.

PR #350, opened 2026-07-08
fix(mir-lower): coerce unsize data pointer to the slice's address space

haixuanTao:
## Summary

Fixes shared-memory array unsizing. A shared `&[T; N]` carries an address-space-3
data pointer, while cuda-oxide's canonical slice stores a generic
address-space-0 pointer. Inserting the shared pointer directly made the LLVM
aggregate type-invalid.

`emit_unsize_cast` now reads both real pointer address spaces and emits one
`addrspacecast` only when they differ before constructing the ordinary slice
fat pointer. This matches the NVPTX shared-to-generic boundary (`cvta.shared`).

```rust
let row: &[f32; N] = unsafe { &TILE[0] };
let total = sum_slice(row); // shared array reference unsizes to a slice
```

The PR adds a shared-memory regression, a generic-pointer no-cast control, and
a standalone compile/run example.

## Maintainer verification

- Rebased onto `main` at `6a92f548`; the sole conflict retained all cast
  tests from #347 and both unsize tests from this PR.
- All 15 cast-lowering tests passed.
- Full `mir-lower` all-target suite — 168 passed (95 library and 73
  integration tests).
- Root workspace, `rustc-codegen-cuda`, and example clippy with warnings
  denied — passed.
- Repository-wide `cargo oxide fmt --check`, including every standalone
  example — passed.
- Compile-only `shared_slice_unsize` device-pipeline smoke test — 1/1 passed.
- Raw LLVM contains the shared-to-generic `addrspacecast`; LLVM 21 at `-O0`
  lowers it to `cvta.shared.u64`.
- Independent RTX 5090 run — `PASS shared_slice_unsize: sum=8128 (expected
  8128)`.
- Compute Sanitizer — memcheck: 0 errors; racecheck: 0 hazards; synccheck:
  0 errors.
- All hosted workflows passed: CI, all-examples device compilation,
  docs/doctests, CodeQL, naming guard, and status guard.

The broader question of preserving Rust reference kind and exclusivity through
MIR import/lowering is tracked separately in #413. It is not caused by this
address-space fix, and no current miscompile from the erased exclusivity
information has been demonstrated.


haixuanTao:
Added a minimal standalone repro that doesn't depend on our (admittedly custom) rust-gpu-style shader stack: `examples/shared_slice_unsize`, built only from cuda-oxide constructs.

The key insight while minimizing: the address space has to survive into the importer's pointer typing for the bug to trigger. A raw-pointer round-trip (`SharedArray::as_mut_ptr()` cast back to `*mut [T; N]`) erases it and compiles fine. What does trigger it is indexing — `SharedArray<[f32; N], 1>` indexed as `&TILE[0]` yields a genuine addrspace(3) `&[f32; N]`, and passing that to a `#[device] fn(&[f32])` is the unsize:

```rust
static mut TILE: SharedArray<[f32; N], 1> = SharedArray::UNINIT;
// ...threads stage input into TILE[0]...
let row: &[f32; N] = unsafe { &TILE[0] };
let total = sum_slice(row);   // fn sum_slice(values: &[f32]) -> f32
```

- On `main` without this fix: `cargo oxide build shared_slice_unsize` fails device codegen with *"Value being inserted / extracted does not match the type of the indexed aggregate"*.
- With the fix: builds, and `cargo oxide run shared_slice_unsize` on an RTX 5080 (sm_120) prints `PASS shared_slice_unsize: sum=8128 (expected 8128)`.

So this reproduces with plain `SharedArray` indexing — no external framework needed. This is also the same shape a nested shared tile (`[[u32; 256]; 4]`-style, e.g. radix-sort LDS rows) produces whenever a row is borrowed and passed to slice-taking code.

nihalpasham:
Thanks for the PR. LGTM.

- The slice builder now converts shared-memory pointers from address space 3 to the generic address space expected by Rust slices.
- The cast is only added when the source and destination address spaces differ.
- A maintainer follow-up fixed two clippy warnings in the new tests.
- The branch was rebased onto current `main`, preserving both #347’s cast tests and this PR’s unsize tests.

For example:
```sh
    shared pointer (AS3) -> addrspacecast -> slice pointer (AS0)
```

**Reviewer note:** Compiler-side lowering was independently verified through `addrspacecast` and LLVM 21 `cvta.shared.u64`. The example passed on an RTX 5090 with the expected sum of 8128. Compute Sanitizer reported 0 memcheck errors, 0 racecheck hazards, and 0 synccheck errors. All hosted CI checks passed.

A broader reference-exclusivity question is tracked separately in #413 and is not caused by this fix.

PR #354, opened 2026-07-08
Document oxide-artifacts compression policy and v2 header

### LINKED ISSUE #353: Document compression policy for embedded oxide artifacts
The `oxide-artifacts` README currently states that compression is not part of the wire format, but the TODO section still asks whether compression is useful or necessary for embedded payloads, especially large PTX bundles, and whether it belongs in this crate or in a higher-level packaging layer.

This should be resolved as a documentation-only clarification.

Suggested direction:

* Keep the base artifact wire format uncompressed.
* Treat `oxide-artifacts` as a neutral container for device-code payloads and entry metadata.
* Leave compression to higher-level packaging, distribution, or loader layers unless measurements show that per-payload compression is required.
* If per-payload compression is added later, introduce it as an explicit versioned extension instead of overloading the current payload flags.

Rationale:

`oxide-artifacts` currently stores payloads such as PTX, NVVM IR, LTOIR, and cubin bytes. Runtime crates can then decide how to consume those payloads. Keeping compression outside the base format preserves that separation and avoids adding codec, allocation, and compatibility concerns to this low-level container.

Expected change:

* Replace the compression TODO with a short documented policy.
* Do not change parsing, serialization, object emission, payload records, or runtime behavior.

### LINKED ISSUE #355: Document artifact format v2 header field
`crates/oxide-artifacts/README.md` still documents artifact format version `1` as the current wire format.

That is now outdated. The current format version is `2`, and the final 32-bit header field is used for compile option bits in version-2 bundles. Version-1 bundles treated this field as reserved zero and should continue to be described as the legacy default-policy format.

### Proposed change

Update `crates/oxide-artifacts/README.md` only:

* Document version `2` as the current artifact format version.
* Update the artifact header diagram so the final field is described as `Compile Options (v2) / Reserved zero (v1)`.
* Add a short compatibility note explaining that version-1 blobs map to the historical default compile policy.
* Keep the documented header size at 32 bytes.
* Do not change code.
* Do not make unrelated compression-policy or TODO changes in this issue.

### Acceptance criteria

* The README wire-format description matches the current version-2 header layout.
* Version-1 compatibility behavior is documented.
* The change is documentation-only.
* No compression-related text is added, removed, or modified unless it is already part of a separate accepted change.

uurl:
Closes #353
Closes #355

## Summary

Documents two previously unresolved parts of the `oxide-artifacts` wire-format
contract.

### Compression policy

- Payload bytes remain uncompressed in the base artifact format.
- Optional compression belongs in a higher-level packaging, distribution, or
  loader layer unless measurements justify a format extension.
- Future per-payload compression requires an explicit versioned extension with
  codec and decompressed-size metadata; current zero flags must not be
  reinterpreted.

This deliberately chooses an accelerator-neutral outer-envelope model. NVIDIA
nvFatbin supports in-container compression by default, while Clang's offload
bundler demonstrates a versioned outer compression envelope.

### Version-2 header

- Version 2 is the latest supported format.
- The final header field is a 64-bit compile-options bitset at bytes 24–31 (the
  half-open range `[24, 32)`), not a 32-bit field.
- Bit 0 disables FMA contraction; unknown bits are rejected.
- Version-1 bundles define that field as reserved zero and map to the historical
  default policy.
- For compatibility, the writer still emits v1 when the options bitset is zero
  and v2 only for non-default policy.

## Maintainer verification

- `cargo test -p oxide-artifacts` — 12 passed
- Header prose checked against `write_u64`/`read_u64`, version selection, and
  parser tests
- Independent post-repair audit — passed

Documentation only; serialization, parsing, object emission, and runtime
behavior are unchanged.

### DIFF
--- a/crates/oxide-artifacts/README.md
+++ b/crates/oxide-artifacts/README.md
@@ -163,9 +163,20 @@ cubin from embedded NVVM IR/LTOIR before loading.
   section.
 - `object`: enables both read and write support.
 
+## Compression Policy
+
+Compression is intentionally not part of the base artifact wire format.
+
+`oxide-artifacts` stores device-code payload bytes as provided by the compiler
+or packaging layer. This keeps the crate as a neutral container for PTX, NVVM IR,
+LTOIR, cubin, and entry metadata, while allowing runtime or loader layers to
+decide how those payloads should be consumed.
+
+If embedded payload size becomes a measured problem, compression should first be
+handled by a higher-level packaging, distribution, or loader layer. Per-payload
+compression should only be added to this format as an explicit versioned
+extension with codec and decompressed-size metadata.
+
 ## TODO
 
-- Investigate whether compression is useful or necessary for embedded payloads,
-  especially for large PTX bundles, and whether it belongs in this crate
-  or in a higher-level packaging layer.
 - Consider Windows support later.
--- a/crates/oxide-artifacts/README.md
+++ b/crates/oxide-artifacts/README.md
@@ -56,7 +56,8 @@ Artifact Blob
  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  |       Payload Count           |        Entry Count            |
  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
- |                         Reserved (zero)                       |
+ |        Compile Options (v2) / Reserved Zero (v1)              |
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. LGTM.

- The documented policy keeps the base `.oxart` payload uncompressed while allowing compression around the complete artifact.
- Any future internal payload compression must use a new, explicitly versioned format.
- A maintainer follow-up incorporated #355 and corrected the v2 header description: compile options occupy bytes 24-31 as a `u64`.

**Reviewer note:** This changes documentation only; it does not alter the artifact format, serializer, parser, or runtime behavior. It closes #353 and #355.

PR #314, opened 2026-06-30
cuda-oxide-codegen: rustc-independent dialect-mir to PTX backend

alejandro-soto-franco:
## Summary

This PR adds an experimental, rustc-independent PTX backend for frontends that already produce cuda-oxide's `dialect-mir` representation.

Previously, another frontend had to depend on rustc-private code or reproduce part of the compiler pipeline. The new `cuda-oxide-codegen` crate keeps frontend translation separate and owns one shared post-translation path used by both `mir-importer` and standalone frontends.

The API is intentionally experimental. Consumers must pin cuda-oxide, Pliron, `dialect-mir`, and `dialect-nvvm` to the same exact revision.

### One rustc-free backend pipeline

- Adds `cuda-oxide-codegen` without `rustc_private` or `rustc_driver`.
- Moves verification, mem2reg, annotated unrolling, device-extern insertion, MIR-to-LLVM lowering, libdevice/NVVM selection, final verification, LLVM export, and PTX generation behind one orchestrator.
- Keeps Rust MIR collection and translation in `mir-importer`.
- Makes the existing Rust path call that same orchestrator while retaining its NVVM/libdevice and artifact-sidecar workflow.

### Experimental public API

The supported standalone surface is `cuda_oxide_codegen::experimental`:

- `CodegenModule` owns its Pliron context and root module together.
- `Compiler` reuses an explicit `Toolchain`.
- `Target` and `CompileOptions` validate compilation inputs.
- `Compilation`, `Diagnostic`, and `CompileError` report structured results and failures.
- Kernel entries are marked by symbol through the owned module.

Implementation modules are private. The doc-hidden `__private` bridge exists only for the matching `mir-importer` revision and is not a supported frontend contract.

### Safety and failure behavior

- Compilation clones the input before destructive passes, so caller IR remains available after success or failure.
- Input and lowered LLVM IR are both verified.
- The standalone v1 rejects libdevice and other unresolved external functions instead of returning PTX that cannot be loaded.
- Requested standalone optimization fails clearly when `opt` is missing or fails.
- The legacy Rust path keeps its warning-and-unoptimized fallback.
- LLVM tool discovery is explicit and reusable; library code does not print unless the caller supplies a trace sink.
- Scratch artifacts use unique, atomically created RAII directories.

## Experimental v1 contract

```text
Rust MIR frontend -+
Other frontend ----+-> owned dialect-mir module
                       -> one shared backend
                       +-> standalone: self-contained PTX
[...]

PR #357, opened 2026-07-08
codegen: harden the standalone experimental backend (follow-up to #314)

alejandro-soto-franco:
Follow-up to #314 (the `cuda-oxide-codegen` rustc-independent MIR→PTX extraction, now merged). This gathers the fixes from a full review of that merged PR. No behaviour change for the existing env-driven rustc pipeline; the work hardens the standalone experimental `Compiler`/`CompileOptions` API and its diagnostics.

## Correctness

- **No IR leak on panic.** `Compiler::compile` clones the caller's module and runs the destructive pipeline on the clone. The previous "compile, then erase" sequence skipped the erase if `compile_clone` panicked, permanently leaking the clone in the shared `Context` arena. The erase now runs from an RAII `EraseGuard`, so it fires on every exit path including an unwinding panic. Covered by a new unit test.
- **`tempfile` floor.** `ScratchDirectory` now uses `tempfile::Builder` with a `0o700` mode via `Builder::permissions`, which first shipped in tempfile 3.9. The workspace floor is raised `3.3 → 3.9` accordingly; a downstream consumer resolving 3.3–3.8 would otherwise fail to compile.

## Diagnostics and error provenance

- **Standalone API no longer blames a legacy env var.** Target/feature-floor validation failures from the standalone API used to reference `CUDA_OXIDE_TARGET`, which those callers never touch, and were classified as `CompilationStage::Codegen`. They now carry the caller's own provenance and are classified as input errors. `BackendOptions` gains a `target_arch_source` label so the env-driven rustc pipeline and the standalone API each describe where their target came from; a new `PipelineError::TargetSelection` distinguishes "bad requested target" from "`llc` failed".
- **Explicit `Toolchain` path keeps its diagnostics.** The explicit-toolchain compile path discarded pipeline diagnostics, and `CompileOptions` had no way to request verbose output through the public surface. `CompileOptions` gains `with_verbose`/`verbose`, and `Toolchain::from_paths` now reports the tools it selected instead of always returning empty diagnostics.

## API and cleanup

- **`Compiler::compile_owned`.** Single-use callers can now consume the module and skip the defensive clone that `compile` always pays for, running passes directly on the owned IR. Useful for large, one-shot modules.
- **Deduplicated module-liveness check** shared by `mark_kernel_entry` and `Compiler::compile` into one helper.
- **Documented the double LLVM-IR render** on the libdevice path: the preview (pre-legalization) and final (post-legalization, different export config) renders are of genuinely different module states, so caching one for the other would be unsound. A real fix needs feature detection to move off rendered text entirely; this documents the constraint rather than attempting an unsound shortcut.

## Testing

- `cargo test -p cuda-oxide-codegen` — 75 passed across 5 suites, including new tests for the panic-safe erase guard and the source-label threading in `resolve_ptx_target`.
- `cargo clippy -p cuda-oxide-codegen --all-targets` — clean.


alejandro-soto-franco:
Closing in favour of three smaller PRs. This one bundled eight unrelated fixes across 12 files, which put the small ones behind a dependency discussion and made the whole thing awkward to review in a sitting.

* #415 clone, liveness and diagnostic paths: the `EraseGuard` leak fix, `compile_owned`, the deduplicated liveness check, and the verbose/diagnostic plumbing.
* #416 target selection provenance: `target_arch_source`, the `TargetSelection` variant, and the reclassification from `Codegen` to `Input`.
* #417 scratch directory on tempfile: the one that adds a dependency, kept on its own so a no there does not block the other two.

The three are independent. Each applies cleanly to `main` on its own and they merge cleanly in any order.

The split also picked up review fixes that were not in this PR: the reclassification here routed feature-floor rejections through `CompileError::InvalidTarget`, which called a valid `sm_75` invalid and printed the architecture twice, so #416 adds a distinct variant. The env-var mis-attribution this PR set out to remove was still present on the `mir-importer` path, where `PipelineConfig::target_arch` overrode the target and left the `CUDA_OXIDE_TARGET` label behind; #416 fixes that too. The 0o700 scratch-dir mode, which is the sole justification for the tempfile floor, now has a test in #417.

PR #358, opened 2026-07-08
fix(collector): don't root items nested inside kernel bodies as kernels

haixuanTao:
## Summary

`is_kernel_function` matches the kernel marker by substring, so a named `fn` defined *inside* a `#[kernel]` body also matches — its def path carries the kernel's name as a path prefix (`...::cuda_oxide_kernel_<hash>_k::helper`). The collector's root scan then treats the nested fn as a kernel entry point: a generic monomorphization gets exported under a `<kernel>__<fn>_TID_<hash>` name (and tagged `gpu_kernel`) while its call sites reference the canonical mangled symbol, so the module fails verification with *"Symbol ... not found"*.

Repro shape (verified against current `main`):

```rust
#[kernel]
pub fn nested_helper(a: &[f32], b: &[u32], mut out: DisjointSlice<f32>) {
    #[inline(never)]
    fn pick<T: Copy>(xs: &[T], i: usize) -> T { xs[i % xs.len()] }
    // ...
    *slot = pick::<f32>(a, i) + pick::<u32>(b, i) as f32;  // Symbol ..pick.. not found
}
```

The existing closure skip already handled the anonymous flavor of exactly this problem; this PR folds both into one shared predicate.

## Changes

- New `is_kernel_entry_def_path` predicate: a def path is a kernel entry point only when its **final** segment carries the kernel marker; closures and named fns nested inside kernel bodies are rejected and left to the call-graph walk, which collects them under their mangled symbols as before.
- The root scan's inline closure check is replaced by the predicate (one skip path instead of two).
- Four unit tests in `collector.rs` covering: bare/qualified kernel paths, nested named fns (single and deep nesting), closures, and non-kernel paths.

## Testing

- `cargo test` in `crates/rustc-codegen-cuda` — 12 passed (4 new).
- Repro kernel above: fails device codegen on `main` with `Symbol ..._pick... not found` (the def exported as `nested_helper__pick_TID_<hash>`); with the fix it builds and runs correctly on an RTX 5080 (sm_120).
- Downstream validation: a shader stack whose reduction kernels define generic tree-reduce helpers inside kernel bodies compiles and runs bit-identically with this fix.

- [x] `cargo test --workspace` passes (crate-local suite; see above)
- [ ] New example added — not needed: covered by unit tests on the predicate (repro kernel included above for reference)

## Checklist

---

**Maintainer note (2026-07-22):** Follow-up commits extended this PR beyond the original description: `is_kernel_function` now delegates to the exact final-segment classifier directly (the root scan's manual skip path is gone entirely), near-miss negative tests were added, the `helper_fn` example gained a nested generic helper with two monomorphizations, and the smoketest gained a compile-only PTX assertion pinning the expected shape, so the "new example" item above is now covered after all. The branch was also rebased across #340's scoped-cache protocol: the classifier accepts the current or legacy kernel prefix on the final def-path segment (`is_current_kernel_symbol` / `is_legacy_kernel_symbol`), keeping legacy roots on the explicit protocol diagnostic path instead of silently dropping them.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/helper_fn/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/helper_fn/src/main.rs
@@ -70,6 +70,22 @@ mod kernels {
     /// an owned value.
     #[kernel]
     pub fn vecadd_with_helper(a: &[f32], b: &[f32], c: DisjointSlice<f32>) {
+        // A named generic helper nested inside a kernel carries the kernel's
+        // generated symbol in an earlier def-path segment. It must remain a
+        // normal device callee under its canonical mangled symbol, not become
+        // a second `_TID_` kernel entry.
+        #[inline(never)]
+        fn nested_identity<T: Copy>(value: T) -> T {
+            value
+        }
+
+        // Keep two concrete monomorphizations reachable without changing the
+        // example's result. `inline(never)` makes them visible to the device
+        // collector and this guard is always false at runtime.
+        if nested_identity::<f32>(0.0) != 0.0 || nested_identity::<u32>(0) != 0 {
+            return;
+        }
+
         // Call the device function by its original name
         vecadd_device(a, b, c);
     }
--- a/crates/rustc-codegen-cuda/src/collector.rs
+++ b/crates/rustc-codegen-cuda/src/collector.rs
@@ -396,6 +396,24 @@ pub fn is_kernel_function(tcx: TyCtxt<'_>, def_id: DefId) -> bool {
     is_kernel_symbol(&tcx.def_path_str(def_id))
 }
 
+/// Returns `true` when `def_path` names a kernel entry point *itself*, as
+/// opposed to an item nested inside a kernel body.
+///
+/// [`is_kernel_function`] matches by substring, so a closure or a named `fn`
+/// defined inside a `#[kernel]` also matches — its def path has the kernel's
+/// name as a path *prefix* (`...::cuda_oxide_kernel_<hash>_k::helper`). Only
+/// a def path whose *final* segment carries the kernel marker is the kernel:
+/// nested items are plain device functions, exported under their mangled
+/// symbol by the call-graph walk. Rooting them as kernels would give generic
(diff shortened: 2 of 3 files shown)


nihalpasham:
Thanks for the PR. LGTM. The diagnosis looks right: substring matching let items nested inside a `#[kernel]` body be rooted as kernels themselves.

- The collector now classifies kernel entry points by the *final* def-path segment, so named fns and closures nested inside a kernel body are no longer rooted; they are collected transitively through the call-graph walk under their canonical mangled symbols.
- This fixes the failure where rooting a nested generic fn minted a `_TID_<hash>` export name that no call site references, failing module verification with "Symbol ... not found".
- I pushed maintainer follow-ups that unify `is_kernel_function` on the exact classifier (making the manual root-scan skip block redundant), add near-miss negative tests, give the `helper_fn` example a nested generic helper with two monomorphizations, and add a compile-only smoketest assertion pinning the expected PTX shape.
- The branch was rebased onto current `main` across #340's scoped-cache protocol change: the classifier accepts the current or legacy prefix on the final segment (via `is_current_kernel_symbol` / `is_legacy_kernel_symbol`), so legacy roots still flow to the explicit protocol diagnostic instead of silently disappearing, and the tests derive their names from the prefix constants.

For example:

```text
my_crate::<KERNEL_PREFIX>softmax                 -> kernel entry point (rooted)
my_crate::<KERNEL_PREFIX>softmax::helper         -> device helper (call-graph walk)
my_crate::<KERNEL_PREFIX>softmax::{closure#0}    -> device helper (call-graph walk)
my_crate::not_<KERNEL_PREFIX>softmax             -> ordinary host function
```

**Reviewer note:** All 21 backend unit tests passed, strict clippy and formatting are clean, and the compile-only smoketest passed with the new `helper_fn` PTX assertions (exactly one `.entry`, both mangled nested-helper `.func` definitions present, no stray `_TID_` exports). Three independent adversarial reviews of the rebase reconciliation found no defects; the legacy-root diagnostic path is byte-identical to `main`. All hosted CI checks passed on the final head. There is no matching open issue to close.

PR #359, opened 2026-07-08
feat: device-only crate builds (auto-detected) + llvm.* link-name intrinsic dispatch

haixuanTao:
## Summary

Two coupled features that together let cuda-oxide compile *device-only* Rust crates — GPU libraries written once against `cfg(target_arch)` for multiple device targets — wholesale to PTX:

1. **Device-only crate builds**: compile the crate front-end for the real `nvptx64-nvidia-cuda` target (`-Z build-std=core`). `cfg(target_arch = "nvptx64")` is then *true*, so shader-style crates pick their GPU code paths (real barriers, scalar math) with no host-target workarounds; the emitted standalone `<crate>.ptx` is the deliverable. Device-only crates are auto-detected (a library with no binary target and no cuda-host dependency cannot be a unified build), with `--device` as an explicit override. Three small pieces: the cargo-oxide build-mode selection, gating the host-side `CudaKernel` marker out of `#[kernel]` expansion on nvptx64, and skipping the host-object embed step for an nvptx `llvm_target`.

2. **Dispatch `#[link_name = "llvm.*"]` externs as intrinsics**: foreign items declared via the `link_llvm_intrinsics` pattern (sreg reads, `llvm.nvvm.barrier0`) dispatch on their link symbol instead of the wrapper's arbitrary Rust path, and the sreg/barrier match arms gain their `llvm.nvvm.*` aliases. Previously such calls fell through to a plain call against a symbol no definition provides, failing module verification with *"Symbol llvm_nvvm_read_ptx_sreg_tid_x not found"*.

They ship together because (2) is only reachable through (1): on the host target these externs cannot exist (x86 ISel rejects nvvm intrinsics — a unified crate carrying them dies with `Cannot select: intrinsic %llvm.nvvm.barrier...`), and (1) without (2) fails on the first sreg read any real shader crate performs.

Motivation: with these two changes (plus #350/#358) we compile an existing multi-crate physics/RL shader codebase (dimforge nexus + vortx) to sm_120 cubins *verbatim* and train on it — simulation results match the WebGPU reference backend on the scenes we test.

## Changes

- `cargo-oxide`: `build` auto-detects device-only crates (a library with no binary target and no cuda-host/cuda-core dependency cannot be a unified build) and selects the nvptx64 path; `--device` remains as an explicit override (adds `--target nvptx64-nvidia-cuda -Z build-std=core`, points `CUDA_OXIDE_PTX_DIR` at the crate dir).
- `cuda-macros`: `#[cfg(not(target_arch = "nvptx64"))]` on the host-side `CudaKernel` marker so the same `#[kernel]` source compiles as unified *and* as pure device crate.
- `rustc-codegen-cuda`: skip host-object embed for nvptx `llvm_target`.
- `mir-importer`: `translate_call` dispatches `llvm.*` callees on the link symbol; 12 sreg arms + the barrier arm gain `llvm.nvvm.*` aliases.
- New example `examples/llvmlink_device`: minimal device-only crate using the extern pattern.

## Testing

- `cargo oxide build llvmlink_device` (device mode auto-detected; same with explicit `--device`) emits PTX with real `%tid.x`/`%ntid.x`/`%ctaid.x` reads and `bar.sync`; assembles clean with `ptxas -arch=sm_120`. Without the dispatch commit it fails verification with `Symbol llvm_nvvm_read_ptx_sreg_ctaid_x not found`.
- Unified path regression: `cargo oxide build vecadd` builds and runs correctly on an RTX 5080 (all 1024 elements verified).
- `cargo test -p mir-importer` and the `rustc-codegen-cuda` suite — all green.
- Downstream validation: the dimforge nexus (265 device fns) + vortx shader crates build through `--device`-style invocations and run a full physics + PPO training loop on sm_120.

- [x] `cargo oxide run vecadd` passes (unified regression)
- [x] `cargo test` passes on touched crates
- [x] New example added (`llvmlink_device`)

## Checklist

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -1213,11 +1213,29 @@ fn run_compute_sanitizer(
 /// Same as [`codegen_run`] but uses `cargo build --release` instead of
 /// `cargo run`. Useful for cross-compilation or when the target hardware
 /// (e.g., Blackwell tensor cores) isn't available on the build machine.
+/// Detect a device-only crate: a library with no binary target and no
+/// dependency on the host-side crates (`cuda-host` / `cuda-core`).
+///
+/// Such a crate cannot be a unified host+device build — there is nothing to
+/// produce for the host — so `build` selects the `--device` path (real
+/// `nvptx64-nvidia-cuda` target) automatically. The explicit `--device` flag
+/// remains as an override for crates the heuristic cannot see through
+/// (e.g. a lib that optionally depends on host crates behind a feature).
+fn is_device_only_crate(crate_dir: &Path) -> bool {
+    let Ok(manifest) = std::fs::read_to_string(crate_dir.join("Cargo.toml")) else {
+        return false;
+    };
+    let has_bin_target = crate_dir.join("src/main.rs").is_file() || manifest.contains("[[bin]]");
+    let depends_on_host = manifest.contains("cuda-host") || manifest.contains("cuda-core");
+    !has_bin_target && !depends_on_host
+}
+
 pub fn codegen_build(
     ctx: &Context,
     example: &str,
     verbose: bool,
     emit_nvvm_ir: bool,
+    device: bool,
     arch: Option<&str>,
     features: Option<&str>,
     no_fmad: bool,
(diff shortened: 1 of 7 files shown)


haixuanTao:
Withdrawing this one. After more thought we reworked our downstream stack to fit cuda-oxide's unified compilation model instead: the shader libraries now call the `cuda_device` API surface (which the importer recognizes by name on any target) rather than `#[link_name = "llvm.*"]` externs, and compile as ordinary host-target crates through the standard interception path — no device-target mode needed.

That approach works end-to-end for us (multi-crate physics/RL codebase, results identical to our previous device-target builds), and it respects the single-compilation-model design here rather than adding a second one. Closing; the branch stays available should real-nvptx64 device builds ever become interesting upstream.

One small genuine issue surfaced while converting, filed separately: the intrinsic dispatch has `cuda_device::thread::*` aliases for the sreg accessors but not for `sync_threads`, so a call resolved via the canonical `cuda_device::thread::sync_threads` path fails with "Symbol ... not found".

PR #360, opened 2026-07-08
fix(mir-importer): dispatch sync_threads via its canonical thread:: path

haixuanTao:
## Summary

Every sreg accessor arm in `try_dispatch_intrinsic` matches both the crate-root re-export path and the canonical module path (`"cuda_device::threadIdx_x" | "cuda_device::thread::threadIdx_x"`), but the barrier arm only matches `"cuda_device::sync_threads"`. A call that resolves through the canonical `cuda_device::thread::sync_threads` path — e.g. a dependency crate calling the API by full path — misses intrinsic dispatch and falls through to a plain function call against a symbol no definition provides, failing module verification with *"Symbol cuda_device__thread__sync_threads not found"*. (Found while porting a multi-crate shader codebase whose support crate calls `cuda_device::thread::sync_threads()` directly; mentioned in #359.)

## Changes

- Add the missing `"cuda_device::thread::sync_threads"` alias to the barrier dispatch arm, matching the sreg arms' pattern.
- Swept the other dispatch arms against the `cuda_device` module tree: `sync_threads` was the only function with a root-path arm but no canonical-path alias.

## Testing

- `cargo test -p mir-importer` — all green.
- Downstream validation: a support crate calling `cuda_device::thread::sync_threads()` by full path fails module verification without this change and compiles + runs correctly (barrier lowers to `bar.sync`) with it.

- [x] `cargo test` passes
- [ ] New example — not applicable (one-line dispatch alias)

## Checklist

### DIFF
--- a/crates/mir-importer/src/translator/terminator/mod.rs
+++ b/crates/mir-importer/src/translator/terminator/mod.rs
@@ -2644,9 +2644,14 @@ fn try_dispatch_intrinsic(
         // =================================================================
         // Synchronization (from intrinsics::sync)
         // =================================================================
-        "cuda_device::sync_threads" => Ok(Some(intrinsics::sync::emit_sync_threads(
-            ctx, target, block_ptr, prev_op, block_map, loc,
-        )?)),
+        // Match the canonical `thread::` path too, like every sreg accessor
+        // above: `sync_threads` lives in `cuda_device::thread`, and a call
+        // resolved via that path (rather than the crate-root re-export)
+        // otherwise falls through to a plain call against a symbol no
+        // definition provides.
+        "cuda_device::sync_threads" | "cuda_device::thread::sync_threads" => Ok(Some(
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. The diagnosis was correct: `sync_threads` had no dispatch entry for its canonical `cuda_device::thread::sync_threads` path. But current `main` resolves this through the generated intrinsic framework (merged in with PR #406).

- Dispatch for both `cuda_device::sync_threads` and `cuda_device::thread::sync_threads` def-paths
- Generated def-path table entries, so aliases stay in sync with the intrinsic definitions instead of a hand-written match arm
- A regression test (`sync_threads_uses_only_canonical_or_compatibility_defpaths`) locking both paths in place

The hand-written dispatch arm this PR patches no longer exists on `main`. Closing as already fixed.

PR #334, opened 2026-07-03
Add mixed-signedness INT8 mma.sync variants (s8*u8, u8*s8)

honeyspoon:
## Summary

Adds four warp-level MMA intrinsics for mixed-signedness INT8 matrix multiply-accumulate, extending the same-signedness s8 variant merged in #329. Mixed-signedness is common in quantized inference where weights and activations use different signedness.

```ptx
mma.sync.aligned.m16n8k32.row.col.s32.s8.u8.s32
    {%d0, %d1, %d2, %d3},
    {%a0, %a1, %a2, %a3},
    {%b0, %b1},
    {%c0, %c1, %c2, %c3};
```

New intrinsics:
- `mma_m16n8k32_s32_s8_u8` — signed A × unsigned B, m16n8k32 shape
- `mma_m16n8k32_s32_u8_s8` — unsigned A × signed B, m16n8k32 shape
- `mma_m16n8k16_s32_s8_u8` — signed A × unsigned B, m16n8k16 shape
- `mma_m16n8k16_s32_u8_s8` — unsigned A × signed B, m16n8k16 shape

All variants require sm_80+ / PTX ISA 7.0 and use convergent inline assembly.

Part of tracking issue #274 (Phase 3).

## What changed

- `cuda-device`: Added 4 user-facing `#[inline(never)]` stubs in `wmma.rs` with full per-lane fragment documentation.
- `dialect-nvvm`: Added 4 operations with 10/7 operands and 4 results respectively; satfinite attribute support; verifier tests for all variants including negative cases.
- `mir-importer`: Added dispatch table entries and MMA fragment emitter with mixed-type PTX mnemonic construction.
- `mir-lower`: Added lowering conversions using convergent inline asm. Mixed-signedness is encoded in the PTX type specifiers (`.s8.u8` vs `.u8.s8`).
- Pipeline detection for mixed-signedness INT8 MMA instructions.
- Lowering tests for all 4 variants.

## Safety and register contract

```text
m16n8k32 (10-operand): C[s32 x4] + A[b32 x4] * B[b32 x2]  →  D[s32 x4]
m16n8k16  (7-operand): C[s32 x4] + A[b32 x2] * B[b32 x1]  →  D[s32 x4]
```

All 32 lanes must execute the same instruction with the same qualifiers, with no lane exited.

## Testing

- [x] `cargo fmt --check` clean
- [x] `cargo test -p dialect-nvvm --all-targets` — verifier tests pass (including satfinite variants)
[...]

PR #323, opened 2026-07-01
Add mma.sync.aligned.m8n8k4 f64 double-precision tensor core intrinsic

honeyspoon:
## Summary

Adds the Ampere FP64 `mma.sync.aligned.m8n8k4` tensor-core operation as an
unsafe, register-valued cuda-device intrinsic.

The operation consumes the calling lane's two C registers plus one A and one B
register, and returns two D registers. It generates no hidden pointer traffic
or fixed PTX temporary-register declarations.

Partially addresses #274. Related to #289; this PR does not close either issue.

### Device API and safety

- Adds `mma_m8n8k4_f64([f64; 2], f64, f64) -> [f64; 2]`.
- Documents the exact lane-to-matrix mapping and full-warp participation rules.
- Documents round-to-nearest-even behavior and the PTX 7.0 / `sm_80+`
  requirements.

### Compiler pipeline

- Represents the operation as four scalar f64 SSA operands and two scalar f64
  SSA results.
- Verifies the exact operand/result count and types.
- Extracts and reconstructs the Rust accumulator array in the MIR importer.
- Lowers to one convergent inline PTX instruction with direct `f64` register
  constraints and no memory clobber.

### Target selection and regressions

- Detects the exact FP64 MMA spelling as PTX 7.0 / `sm_80`.
- Rejects explicit lower targets before invalid PTX can be emitted.
- Adds verifier, importer, lowering, target-detection, one-call, and repeated-
  call coverage.
- Preserves the strongest architecture and PTX floor when FP64 MMA is combined
  with BF16 MMA and the packed atomics already on `main`.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo check --workspace --all-targets`
- [x] Focused dialect, importer, and lowering test suites (203 passed)
- [x] Strict workspace clippy with warnings denied
- [x] One-call and two-call Rust-to-PTX builds select PTX 7.0 / `sm_80`
- [x] CUDA 13.3 `ptxas -arch=sm_80` accepts both kernels
- [x] SASS contains one/two `DMMA.884` instructions with `STACK:0 LOCAL:0`
- [x] Explicit `sm_75` is rejected before PTX generation
- [x] A post-#322 all-four kernel selects PTX 7.8 / `sm_90`, emits both MMA
  forms and both packed atomics, and assembles with zero stack or spills
- [ ] Numerical execution on a physical `sm_80+` GPU

## Checklist

- [x] Contributor authorship preserved
- [x] Every PR commit has DCO sign-off
[...]

PR #321, opened 2026-07-01
Add mma.sync.aligned.m16n8k16 bf16 tensor core intrinsic

honeyspoon:
## Summary

Adds the Ampere BF16 `mma.sync.aligned.m16n8k16` tensor-core operation as an
unsafe, register-valued cuda-device intrinsic.

The API consumes the calling lane's C, A, and B fragments as scalar register
arrays and returns the four D registers. It does not hide pointer loads,
stores, a local stack frame, or memory ordering inside the MMA operation.

Partially addresses #274. Related to #289; this PR does not close either issue.

### Device API and safety

- Adds `mma_m16n8k16_f32_bf16([f32; 4], [u32; 4], [u32; 2]) -> [f32; 4]`.
- Documents the exact per-lane A, B, C, and D fragment mapping.
- Documents the full-warp convergence, exited-lane, PTX 7.0, and `sm_80+`
  requirements.
- Marks the operation `unsafe` and `#[must_use]`.

### Compiler pipeline

- Represents the operation as ten scalar SSA operands and four scalar SSA
  results in the NVVM dialect.
- Verifies the exact operand count, result count, and register types.
- Extracts and reconstructs Rust arrays in the MIR importer without pointer
  transport.
- Lowers to one convergent inline PTX instruction using direct register
  constraints and no memory clobber.

### Target selection

- Detects this exact MMA spelling as requiring PTX 7.0 and `sm_80`.
- Rejects explicit lower targets before emitting invalid PTX.
- Includes whitespace/escaped-whitespace and near-miss detection tests.

### Regression coverage

- Adds dialect verifier, MIR importer, lowering, and target-selection tests.
- Covers one and two MMA calls so repeated use cannot redeclare fixed PTX
  temporary names.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo check -p cuda-device -p dialect-nvvm -p mir-importer -p mir-lower --all-targets`
- [x] `cargo test -p dialect-nvvm -p mir-importer -p mir-lower`
- [x] Strict affected-crate clippy with warnings denied
- [x] One-call and two-call Rust-to-PTX builds select PTX 7.0 / `sm_80`
- [x] CUDA 13.3 `ptxas` accepts both kernels
- [x] SASS contains one/two `HMMA.16816.F32.BF16` instructions with zero stack
  bytes and no local loads/stores
- [x] Explicit `sm_75` is rejected before PTX generation
- [ ] Numerical execution on a physical `sm_80+` GPU

## Checklist

[...]

PR #363, opened 2026-07-09
Add mma.sync.aligned.m16n8k32 u8 tensor core intrinsic

aymuos15:
> I am still getting used to the repo. So I purposely followed the structure and style  #329, for the description. Felt that this also allows the two `m16n8k32` INT8 intrinsics to read the same way and hence, are easy to compare side by side.

## Summary

Adds the raw Ampere unsigned-INT8 warp-level matrix multiply-accumulate operation:

```ptx
mma.sync.aligned.m16n8k32.row.col.s32.u8.u8.s32
    {%d0, %d1, %d2, %d3},
    {%a0, %a1, %a2, %a3},
    {%b0, %b1},
    {%c0, %c1, %c2, %c3};
```

It is the unsigned twin of the signed `s8` form (#329). The per-lane register layout is identical, and the `sm_80` / PTX 7.0 floors are the same. Only the `.u8.u8` operand types differ. This is the low-level, unsafe register atom used to build typed quantized matrix and tile APIs. It partially addresses tracking issue #274 (item "INT8 u8 m16n8k32"). It does not close the issue.

## What changed

- Added the `cuda-device` stub `mma_m16n8k32_s32_u8`.
- Added the dialect-nvvm operation, MIR importer handling, and register-only convergent lowering.
- Documented the exact per-lane A/B/C/D coordinates and the low-byte-to-high-byte packing of four unsigned 8-bit values in each `u32` carrier.
- Documented that this non-`.satfinite` form wraps signed accumulator overflow.
- Extended target detection to recognize both the plain and `.satfinite` spellings. It rejects near-misses, and it also rejects the signed and mixed-signedness (`s8.u8` / `u8.s8`) forms, so neither detector cross-matches the other.
- Extended the shared MMA fragment diagnostic guard and added verifier boundary tests.
- Added a checked-in device-stub to MIR importer to lowering to PTX regression.
- Kept the BF16, F16, TF32, INT8 (signed and unsigned), and FP64 paths separate.

## Safety and integer contract

```text
each lane: C[s32 x4] + A[b32 x4] * B[b32 x2]
             wrapping    four u8 values per b32
                              |
                              v
                          D[s32 x4]
```

All 32 lanes must execute the same instruction with the same qualifiers, and no lane may have exited. `A` and `B` must use the documented per-lane layout and byte order. The instruction requires PTX ISA 7.0 and `sm_80` or newer. cuda-oxide selects both floors automatically and rejects a lower explicit target.

## Testing

- [x] `cargo test -p dialect-nvvm -p cuda-oxide-codegen -p mir-lower -p mir-importer`
- [x] `cargo clippy` on the changed crates, clean
- [x] Root and example workspace pass `cargo fmt -- --check`
- [x] `cargo oxide run standalone_device_fn --arch sm_86` passes 20/20
- [x] The unsigned INT8 MMA selects the PTX 7.0 and `sm_80` floors. `sm_75` is rejected with the required `sm_80` floor (target-detection tests).
- [x] Plain and `.satfinite` detector boundary tests, plus cross-guards that reject the signed and mixed-signedness forms.
- [x] **Full-warp numerical comparison on an `sm_86` GPU against a CPU-reference matmul.** One warp computed `D = A x B` for `A` (16x32) and `B` (32x8) into `s32`. All 16x8 = 128 outputs matched the reference exactly. Sample values: `A[0][0..4] = [11, 48, 85, 122]`, `B[0..4][0] = [7, 175, 87, 255]`, `D[0][0] = 527216` (gpu == ref). The inputs deliberately include bytes above 127 (`175`, `255`). Under a signed lowering those decode as `-81` and `-1`, which would change the result. The exact match is therefore positive evidence of unsigned operand lowering. This was verified locally. The numerical example is not shipped, matching the MMA-family PTX-check convention set by #327, #321, #328, #323, and #329.

## Notes

This may need a small rebase if #334 (mixed-signedness INT8) lands first. That PR touches the same `m16n8k32` detection and emit paths. The overlap is additive.

## Checklist

- [x] All commits carry the DCO `Signed-off-by` trailer
- [x] No `crates/cuda-bindings/` changes relative to `main`

### DIFF
--- a/crates/cuda-device/src/wmma.rs
+++ b/crates/cuda-device/src/wmma.rs
@@ -443,3 +443,59 @@ pub unsafe fn mma_m16n8k32_s32_s8(c: [i32; 4], a: [u32; 4], b: [u32; 2]) -> [i32
     let _ = (c, a, b);
     unreachable!("mma_m16n8k32_s32_s8 called outside CUDA kernel context")
 }
+
+/// Multiply one warp-distributed unsigned INT8 tile and add an i32 accumulator.
+///
+/// Together, the 32 lanes compute `D = A × B + C` for row-major `A` with
+/// shape 16×32, column-major `B` with shape 32×8, and `C`/`D` with shape
+/// 16×8. Each lane supplies its fragments in registers and receives four i32
+/// result registers. The call itself does not access memory or act as a fence.
+///
+/// Each `a` and `b` register packs four unsigned `u8` values from low byte to
+/// high byte. In other words, logical element `i` is stored in bits
+/// `(i % 4) * 8..(i % 4 + 1) * 8` of register `[i / 4]`.
+///
+/// For lane `lane`, let `group = lane / 4` and `thread = lane % 4`:
+///
+/// ```text
+/// A element i=0,...,15:
+///   row = group       for 0 <= i < 4 or 8 <= i < 12; otherwise group + 8
+///   col = thread*4 + (i&3) + (if i >= 8 { 16 } else { 0 })
+///
+/// B element i=0,...,7:
+///   row = thread*4 + (i&3) + (if i >= 4 { 16 } else { 0 })
+///   col = group
+///
+/// C/D register i=0,...,3:
+///   row = group + (if i >= 2 { 8 } else { 0 })
+///   col = thread*2 + (i&1)
+/// ```
+///
+/// This is the non-`.satfinite` form. The i32 accumulator is signed, so
+/// accumulator overflow wraps rather than clamping to `i32::MIN` or `i32::MAX`.
+///
+/// # PTX
+///
+/// ```ptx
+/// mma.sync.aligned.m16n8k32.row.col.s32.u8.u8.s32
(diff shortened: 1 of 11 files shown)


aymuos15:
Havent checked but since this got the fixed tag please feel free to close this. Thanks.

nihalpasham:
Thanks for the PR. The `u8 m16n8k32` shape was a real gap in the hand-written set. But current `main` implements this intrinsic through the generated intrinsic framework (merged in with PR #406).

- The public API: `mma_m16n8k32_s32_u8(c: [i32; 4], a: [u32; 4], b: [u32; 2]) -> [i32; 4]`, identical in name and signature to this PR's stub
- Typed importer dispatch and codegen emitting `mma.sync.aligned.m16n8k32.row.col.s32.u8.u8.s32` with the same sm_80 / PTX 7.0 floors
- The `satfinite` and mixed-signedness (`u8_s8`, `s8_u8`) variants as well

Closing as already fixed, per your comment above. Thanks again.

PR #366, opened 2026-07-10
Optimize common equal-results path in difftest

### LINKED ISSUE #365: Optimize difftest equivalence-class construction for equal backend results
## Context

`difftest/src/lib.rs` has a TODO in `ExecResults::from_exec_results` to optimize the common case where all backend execution results are equivalent.

Current behavior builds equivalence classes incrementally for every result. This works, but it does unnecessary class scanning in the expected common case where all backends produce the same output.

## Problem

Most differential-test executions should produce equivalent backend results. The current implementation still follows the general equivalence-class construction path from the first result onward.

There is also a subtle constraint: successful executions are not compared with full `ExecResult` equality. The current logic treats two successful results as equivalent when their `stdout` values match, while non-success cases use structural equality. Any optimization must preserve that behavior.

## Proposed change

Add a fast path in `ExecResults::from_exec_results`:

1. Take the first `(backend_name, ExecResult)` as the reference result.
2. Compare subsequent results directly against that first result using the existing equivalence semantics.
3. If all results are equivalent, return a single equivalence class.
4. If a mismatch is found, fall back to the normal equivalence-class construction path for the remaining results.

This should preserve current semantics while avoiding unnecessary class scanning in the dominant all-equal case.

## Acceptance criteria

* Preserve the existing equivalence semantics:

  * `Ok(_)` results are equivalent when `stdout` matches.
  * error results use structural equality.
* Preserve behavior for empty input.
* Preserve behavior when multiple equivalence classes are required.
* Avoid introducing unnecessary cloning of `ExecResult`.
* Add a small helper for equivalence comparison if it improves readability.
* `cargo fmt` passes.
* `cargo test -p difftest` passes.
* `cargo clippy -p difftest --all-targets -- -D warnings` passes.

uurl:
Closes #365 

This optimizes `ExecResults::from_exec_results` for the common case where all backends produce equivalent outputs.

The implementation first compares all results against the first result directly and only falls back to equivalence-class construction when a mismatch is found. This preserves the existing equivalence semantics: successful executions are compared by stdout, while errors are compared structurally.

It also removes an unnecessary clone when inserting a new equivalence class.

### DIFF
--- a/crates/fuzzer/rustlantis/difftest/src/lib.rs
+++ b/crates/fuzzer/rustlantis/difftest/src/lib.rs
@@ -207,3 +207,139 @@ pub fn run_diff_test<'a>(
 
     ExecResults::from_exec_results(exec_results.into_iter())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{ffi::OsString, process::ExitStatus};
+
+    fn exit_status(code: u32) -> ExitStatus {
+        #[cfg(unix)]
+        {
+            use std::os::unix::process::ExitStatusExt;
+            ExitStatus::from_raw((code as i32) << 8)
+        }
+        #[cfg(windows)]
+        {
+            use std::os::windows::process::ExitStatusExt;
+            ExitStatus::from_raw(code)
+        }
+        #[cfg(not(any(unix, windows)))]
+        {
+            let _ = code;
+            unimplemented!("test exit statuses are only defined for Unix and Windows")
+        }
+    }
+
+    fn process_output(status: u32, stdout: &str, stderr: &str) -> backends::ProcessOutput {
+        backends::ProcessOutput {
+            status: exit_status(status),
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. LGTM.

- `ExecResults::from_exec_results` now takes a fast path for the common case: every result is compared against the first, and equivalence-class construction only runs when a mismatch appears.
- The original equivalence semantics are preserved exactly: successful executions compare by stdout, errors compare structurally.
- Maintainer follow-ups restacked the branch on #368 (they share the difftest clippy base) and added four deterministic unit tests covering empty input, Ok stdout equivalence, Err structural equivalence, and the mismatch fallback, matching #365's acceptance criteria.

**Reviewer note:** `cargo test -p difftest --lib` passes 4/4, strict clippy and formatting are clean, and all hosted CI checks passed. Merged after #368, which it was stacked on. This closes #365.

PR #368, opened 2026-07-10
Fix difftest Clippy warnings under -D warnings

### LINKED ISSUE #367: Fix difftest Clippy warnings under -D warnings
## Context

Running Clippy on `crates/fuzzer/rustlantis/difftest` with warnings denied currently fails because of existing warnings in `difftest/src/backends.rs`.

These warnings are unrelated to the `ExecResults` optimization work, so they should be handled in a separate cleanup PR.

## Problem

The following command fails:

```bash
```

The failures are caused by existing Clippy diagnostics in `difftest/src/backends.rs`:

* `clippy::let_and_return` in `run_compile_command`
* `clippy::upper_case_acronyms` for `LLVM`
* `clippy::needless_borrow` when invoking `rustup run ... miri`
* `clippy::upper_case_acronyms` for `GCC`

## Proposed change

Apply the minimal cleanup needed to make strict Clippy pass:

* Return the `compiler.wait_with_output().expect(...)` expression directly.
* Rename private struct `LLVM` to `Llvm`.
* Rename private struct `GCC` to `Gcc`.
* Remove the unnecessary borrow in the Miri `rustup run` command.

The renamed structs are private implementation details, so this should not affect the public API.

## Acceptance criteria

* `cargo fmt` passes.
* `RUST_TEST_THREADS=1 cargo test` passes.
* `cargo clippy --all-targets -- -D warnings` passes.
* No behavioral changes are introduced.
* The PR only touches Clippy cleanup in `difftest/src/backends.rs`.

uurl:
Closes #367 

This fixes existing Clippy warnings in `crates/fuzzer/rustlantis/difftest/src/backends.rs` so the crate can pass strict Clippy validation with warnings denied.

Changes:

* Return the `wait_with_output().expect(...)` expression directly instead of binding and returning it.
* Rename private implementation structs `LLVM` and `GCC` to `Llvm` and `Gcc`.
* Remove an unnecessary borrow in the Miri `rustup run ... miri` command.

These are mechanical cleanup changes only. The affected structs are private, so this should not change the public API or runtime behavior.

Validation:

```bash
cargo fmt
RUST_TEST_THREADS=1 cargo test
```

### DIFF
--- a/crates/fuzzer/rustlantis/difftest/src/backends.rs
+++ b/crates/fuzzer/rustlantis/difftest/src/backends.rs
@@ -80,7 +80,7 @@ pub fn from_config(config: Config) -> HashMap<String, Box<dyn Backend>> {
             BackendConfig::MiriRepo { path, flags } => {
                 Box::new(Miri::from_repo(path, flags).unwrap())
             }
-            BackendConfig::LLVM { toolchain, flags } => Box::new(LLVM::new(toolchain, flags)),
+            BackendConfig::LLVM { toolchain, flags } => Box::new(Llvm::new(toolchain, flags)),
             BackendConfig::Cranelift { toolchain, flags } => {
                 Box::new(Cranelift::from_rustup(toolchain, flags))
             }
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks for the PR. LGTM.

- The difftest backends now pass strict clippy with warnings denied: `LLVM`/`GCC` renamed to `Llvm`/`Gcc`, an unnecessary borrow removed, and the `wait_with_output` expression returned directly.
- I pushed a small maintainer follow-up removing an unused `<'a>` lifetime in `ExecResults::from_exec_results`, which was the last remaining `-D warnings` failure (`clippy::extra_unused_lifetimes`).

**Reviewer note:** `cargo clippy -p difftest --all-targets -- -D warnings` passes at the merged head and failed before the follow-up, confirming the acceptance criterion is now met. All hosted CI checks passed. This closes #367.

PR #370, opened 2026-07-10
Use configured array length for usize literal generation

### LINKED ISSUE #369: Use configured array length when generating usize literals
`generate/src/literal.rs` uses the hard-coded range `0..8` when generating small `usize` literals:

```rust
0 => rng.random_range(0..8), // FIXME: ARRAY_MAX_LEN
```

The maximum array length is already configurable through `TyConfig::array_max_len` and available from `TyCtxt`.

The literal generator should use the configured value instead of duplicating the default value `8`.

### Proposed change

* Parameterize the `usize` literal distribution with the configured array-length bound.
* Use `tcx.config.array_max_len` when generating small `usize` values.
* Remove the existing `FIXME`.
* Keep `isize` generation independent from the `usize`-specific configuration.

### Acceptance criteria

* No hard-coded array maximum remains in `generate/src/literal.rs`.
* Small `usize` literals are generated in `0..tcx.config.array_max_len`.
* Existing `isize` literal generation behavior remains unchanged.
* Formatting, Clippy, and workspace tests pass.

uurl:
Closes #369 

## Summary

Use `TyConfig::array_max_len` when generating small `usize` literals instead of relying on the hard-coded value `8`.

## Changes

* Split the existing `Sombrero` distribution into `usize`-specific and `isize`-specific distributions.
* Parameterize the `usize` distribution with the configured upper bound.
* Pass `tcx.config.array_max_len` from `GenLiteral::gen_literal`.
* Remove the `FIXME` associated with the hard-coded array-length value.

## Rationale

Array lengths are configurable, so duplicating the default value in the literal generator can produce inconsistent behavior when `array-max-len` is changed.

Using the value from `TyCtxt` keeps literal generation aligned with type generation.

## Validation

* `cargo test`
* `cargo clippy --all-targets`
* `cargo fmt --all --check`
* `cd crates/fuzzer/rustlantis && cargo test -p generate usize_sombrero_uses_configured_small_values_upper_bound`
* `cd crates/fuzzer/rustlantis && cargo test -p generate`
* `cd crates/fuzzer/rustlantis && cargo clippy -p generate --all-targets`
* `cd crates/fuzzer/rustlantis && cargo fmt --all -- --check`

### DIFF
--- a/crates/fuzzer/rustlantis/generate/src/literal.rs
+++ b/crates/fuzzer/rustlantis/generate/src/literal.rs
@@ -5,19 +5,29 @@ use mir::{
 use rand::{Rng, RngCore, seq::IndexedRandom};
 use rand_distr::Distribution;
 
-struct Sombrero;
+struct UsizeSombrero {
+    small_values_upper_bound: usize,
+}
 
-impl Distribution<usize> for Sombrero {
+impl UsizeSombrero {
+    fn small_values_range(&self) -> std::ops::Range<usize> {
+        0..self.small_values_upper_bound
+    }
+}
+
+impl Distribution<usize> for UsizeSombrero {
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> usize {
         match rng.random_range(0..=1) {
-            0 => rng.random_range(0..8), // FIXME: ARRAY_MAX_LEN
+            0 => rng.random_range(self.small_values_range()),
             1 => rng.random_range(usize::MIN..=usize::MAX),
             _ => unreachable!(),
         }
     }
 }
 
-impl Distribution<isize> for Sombrero {
+struct IsizeSombrero;
+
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. LGTM.

- Small `usize` literal generation now draws from the configured `TyConfig::array_max_len` instead of the hard-coded `0..8` range, so generated indices track the configured array sizes.
- I pushed a maintainer follow-up: `array_max_len` comes from user TOML with no lower-bound validation, so a zero bound now falls back to full-width sampling instead of panicking in `random_range(0..0)`; the range helper was inlined; and the original helper-only test was replaced with three deterministic behavioral tests driven by a sequence RNG.

**Reviewer note:** All 17 `generate`-crate tests passed at the head commit, formatting is clean, and clippy reports no warnings in the touched file. All hosted CI checks passed. This closes #369.

PR #372, opened 2026-07-11
Document descriptor offset output

### LINKED ISSUE #371: Replace placeholder descriptor offsets in README output
## Problem

The WGMMA example README still shows placeholder values in the expected Hopper output:

```text
  Leading dimension offset: xxx (raw bits)
  Stride offset: xxx (raw bits)

uurl:
Closes #371 

## Summary

- Replace placeholder WGMMA descriptor offset values in the README expected output.
- Document the current raw leading-dimension and stride offsets as `8`.

## Details

The WGMMA lowering builds the descriptor with `0xC000000800080000`, and the example prints:

- `(desc >> 16) & 0x3FFF`
- `(desc >> 32) & 0x3FFF`

Both fields resolve to `8`, so the README output should not keep `xxx` placeholders.

## Validation

Docs-only change. No runtime validation required.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/wgmma/README.md
+++ b/crates/rustc-codegen-cuda/examples/wgmma/README.md
@@ -78,8 +78,8 @@ Loading PTX from: wgmma.ptx
 Launching wgmma_sync_test kernel...
 SMEM descriptor: 0xC000000000000xxx
 ✓ Swizzle mode correct (128B)
-  Leading dimension offset: xxx (raw bits)
-  Stride offset: xxx (raw bits)
+  Leading dimension offset: 8 (raw bits)
+  Stride offset: 8 (raw bits)
 
 === WGMMA Test Complete ===
 ```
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. LGTM.

- The README's expected output now shows the real descriptor offset fields instead of placeholders; both the raw leading-dimension and stride offsets decode to 8.
- A maintainer follow-up fixed the remaining inconsistency (the `SMEM descriptor:` line still showed zeroed offset fields) and added a short paragraph explaining that the low bits are the variable 14-bit shared-memory address field while the `0008` fields are the fixed offsets.

**Reviewer note:** All documented values were verified against the actual descriptor construction (`0xC000000800080000` combined with the encoded address in cuda-device's wgmma builder) and the example's extraction code (`(desc >> 16) & 0x3FFF` and `(desc >> 32) & 0x3FFF` both resolve to 8). This is a documentation-only change; it does not alter compiler or runtime behavior. All hosted CI checks passed. It closes #371.

PR #373, opened 2026-07-11
Translate constant assert conditions

uurl:
## Summary

Fixes a MIR importer feature gap where `translate_assert` rejected non-place assert conditions, including constant boolean operands.

The assert translator now delegates condition lowering to `rvalue::translate_operand`, which already supports `Copy`, `Move`, `Constant`, and `RuntimeChecks` operands. This preserves the existing place path while enabling constant assert conditions.

## Testing

- cargo check -p mir-importer
- cargo clippy -p mir-importer -- -D warnings
- scripts/smoketest.sh --compile-only primitive_stress

### DIFF
--- a/crates/mir-importer/src/translator/terminator/mod.rs
+++ b/crates/mir-importer/src/translator/terminator/mod.rs
@@ -361,13 +361,29 @@ fn translate_assert(
 
     // Translate the condition operand.
     //
-    // MIR assert conditions are operands, not necessarily places. In particular,
-    // rustc may leave constant boolean conditions as `Operand::Constant`, and
-    // runtime-check operands also lower to a boolean value through the shared
-    // operand translator. `translate_operand` already preserves the old
-    // Copy/Move path by delegating to `translate_place`.
-    let (cond_value, mut last_inserted) =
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks for the PR. LGTM.

- `translate_assert` no longer bails on constant assert conditions: `Copy`, `Move`, and `Constant` operands all delegate to the shared `rvalue::translate_operand`, which materializes constants through the dialect's `MirConstantOp`. The existing place path is preserved.
- I pushed a maintainer follow-up that keeps `RuntimeChecks` conditions fail-closed behind an explicit diagnostic instead of routing them through the same delegation; session-dependent runtime-check flags must not silently lower to constant `false` in assert conditions.
- The follow-up also added two compile-only regression kernels covering both assert polarities (an always-false condition via a division and an always-true condition via an out-of-bounds index on an empty array).

For example:

```text
assert with constant condition -> materialized MirConstantOp, normal assert lowering
assert with RuntimeChecks flag -> explicit diagnostic, not constant-folded
```

**Reviewer note:** The match is exhaustive over the pinned toolchain's operand variants, and the negation ordering for expected-false asserts was verified. `cargo check -p mir-importer`, strict package clippy, and the compile-only smoke gate passed; all hosted CI checks passed on the final head. There is no matching open issue to close.

PR #377, opened 2026-07-12
fix(mir): accept empty enum types

### LINKED ISSUE #376: MIR enum verification rejects valid zero-variant types
**Description**

`MirEnumType` requires at least one variant, although Rust permits zero-variant enums and the MIR importer already has an explicit `VariantsShape::Empty` translation path. The dialect invariant should permit the uninhabited type while continuing to reject every operation that would manufacture or inspect a value of it.

**Minimal reproducer**

Construct a `MirEnumType` with empty variant names, discriminants, and field metadata, then call its verifier:

```rust
let empty = MirEnumType::get(
    &mut ctx,
    "Infallible".to_string(),
    discr.into(),
    vec![],
    vec![],
);
empty.deref(&ctx).verify(&ctx)?;
```

**Expected behavior**

The type verifies as a valid uninhabited enum. Construct, payload, get-discriminant, and set-discriminant operations on it fail with explicit diagnostics.

**Actual behavior**

Type verification fails with:

```text
MirEnumType must have at least one variant
```

**Additional context**

This is a dialect-invariant correction, not a claim of a currently observed end-to-end compiler failure. It is separate from niche and single-live-variant enum layout support tracked by #306.

niklebedenko:
## Summary

Closes #376.

Permit zero-variant Rust enum types in the MIR dialect while keeping every enum-specific value and discriminant operation fail-closed.

```text
Before: a zero-variant MirEnumType fails type verification
After:  the type verifies; construct, payload, get-discriminant, and
        set-discriminant operations reject it explicitly
```

### Empty enum type verification

- Treat zero variants as a valid uninhabited Rust type.
- Retain the existing consistency checks for variant metadata and recorded layout.
- Update the type documentation to state the zero-variant contract.

### Fail-closed operation verification

- Keep construction and payload extraction rejected by the shared variant bounds.
- Explicitly reject discriminant reads and writes because an empty enum has no valid discriminant.
- Assert the precise diagnostics for all four enum-specific operation kinds.

## Known separate limitations

- This does not change niche-encoded or single-live-variant enum ABI modeling.
- It does not make values of a zero-variant enum constructible.
- No end-to-end compiler failure caused by the old verifier has been identified; the regression is the dialect invariant itself.

## Testing

- `cargo oxide fmt --check`
- `cargo test -p dialect-mir`: 5 const-fold tests and 17 operation tests passed
- `cargo clippy -p dialect-mir --all-targets -- -D warnings`

## Checklist

- [x] Branch based directly on current `upstream/main`
- [x] Correctness claim covered by a focused regression
- [x] Unsupported operations fail clearly
- [x] Diff contains no fork-only scaffolding

### DIFF
--- a/crates/dialect-mir/src/ops/enum_ops.rs
+++ b/crates/dialect-mir/src/ops/enum_ops.rs
@@ -211,6 +211,13 @@ impl Verify for MirGetDiscriminantOp {
             }
         };
 
+        if enum_ty.variant_count() == 0 {
+            return verify_err!(
+                op.loc(),
+                "MirGetDiscriminantOp cannot inspect an enum with no variants"
+            );
+        }
+
         // Result must match the enum's discriminant type
         let result = op.get_result(0);
         let result_ty = result.get_type(ctx);
@@ -289,6 +296,12 @@ impl Verify for MirSetDiscriminantOp {
                 let pointee = ptr_type.pointee.deref(ctx);
                 match pointee.downcast_ref::<MirEnumType>() {
                     Some(enum_ty) => {
+                        if enum_ty.variant_count() == 0 {
+                            return verify_err!(
+                                op.loc(),
+                                "MirSetDiscriminantOp cannot mutate an enum with no variants"
+                            );
+                        }
                         let expected_discr_ty = enum_ty.discriminant_type();
                         if discr_ty != expected_discr_ty {
                             return verify_err!(
--- a/crates/dialect-mir/src/types.rs
+++ b/crates/dialect-mir/src/types.rs
@@ -735,8 +735,8 @@ impl EnumVariant {
 /// `variant_field_counts` to split the `all_*` vectors per variant.
 ///
 /// # Verification
-/// * Must have at least one variant.
-/// * Discriminant type must be an integer type.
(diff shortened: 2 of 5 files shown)


nihalpasham:
Thanks for the PR. Accepting zero-variant enums here is correct. But #349 has since rebuilt the enum path on rustc's real layouts, and current `main` covers everything here in stronger form:

- Zero-variant enums verify when `layout_kind == EnumLayoutKind::Empty`, rather than unconditionally
- `MirGetDiscriminantOp` rejects any enum with no inhabited variant, not just zero-variant ones, and `MirSetDiscriminantOp` rejects uninhabited or out-of-bounds variants
- The metadata consistency checks from the follow-up commit on this branch (checked field-count fold, flattened-count validation) also exist on `main`
- `VariantsShape::Empty` is mapped end to end through the importer and lowering, with unit tests and an `error_set_discriminant_uninhabited` fixture

The branch now hard-conflicts with the new enum model, and resolving in this PR's favor would drop the Empty-layout gate, so I'm closing this as already fixed rather than rebasing.

Note: the `empty_enum_argument` example (a kernel taking a zero-variant enum with `match value {}`) has no direct example-level equivalent on `main`. I would welcome that as a fresh tests-only PR against the current enum model.

PR #383, opened 2026-07-12
fix(mir-importer): lower tuple array constants with rustc layout

### LINKED ISSUE #382: Array value constants with tuple elements fail MIR import
## Bug

Bare array constants support primitive elements and nested arrays, but an array whose element is a tuple is rejected during MIR import.

```rust
const TABLE: [(u8, u32); 2] = [(1, 0x1122_3344), (2, 0x5566_7788)];

#[inline(never)]
fn lookup(i: usize) -> u32 {
    let (tag, value) = TABLE[i & 1];
    value + tag as u32
}
```

```text
Actual:   byte lowering rejects the tuple element type
Expected: tuple elements are decoded using rustc's array stride and field offsets
```

The example intentionally places padding between `u8` and `u32`. Treating tuple fields as tightly packed reads the second field from padding even if tuple construction itself succeeds.

## Correctness boundary

Tuple decoding must use rustc layout. Constants containing pointer provenance must not reinterpret relocation placeholder bytes as exposed-provenance pointers; they should fail clearly until relocation-aware aggregate materialization exists.

## Acceptance criteria

- Arrays of tuples with supported scalar, enum, nested-tuple, and ZST fields lower correctly.
- Internal field padding and trailing element-stride padding are preserved.
- Direct padded tuple constants use the same layout-aware decoder.
- Pointer-bearing tuple arrays fail with a stable, actionable diagnostic.
- Existing primitive, nested-array, and pointer-to-array constant behavior remains unchanged.
- Supported-feature documentation records the new support and its limitations.

niklebedenko:
Closes #382.

## Summary

MIR import now materializes byte-backed array constants whose elements are supported tuples. Array stride and tuple field positions come from rustc layout, while pointer-bearing allocations remain explicitly unsupported rather than being reconstructed from relocation placeholder bytes.

```text
Before: tuple array elements were rejected, and packed tuple slicing could ignore padding.
After:  supported tuples use rustc offsets and stride; pointer provenance fails closed.
```

### Layout-aware aggregate decoding

- Carries the Rust array type through recursive byte-backed array construction and uses rustc's element size as the stride.
- Routes tuple elements through the shared typed constant decoder.
- Uses rustc field offsets and field sizes for array elements and standalone tuple constants.
- Preserves primitive and recursively nested array handling.

### Correctness and safety hardening

- Validates field counts, layout shape, byte ranges, and checked offset arithmetic before constructing MIR values.
- Rejects array constants containing pointer provenance before relocation bytes can be mistaken for exposed-provenance pointers.
- Updates feature documentation and registers the negative example with the error-smoke suite.

## Known separate limitations

- Struct and initialized-union array elements remain unsupported.
- Pointer-bearing tuple arrays remain unsupported until aggregate relocations can be represented faithfully.

## Testing

- `cargo oxide fmt --check`
- Strict `-D warnings` clippy for `mir-importer`, `array_constants`, and `error_tuple_array_provenance`.
- 17 `mir-importer` tests and error-example status synchronization.
- Optimized and no-opt LLVM assertions for padded direct, nested-ZST, and enum tuple constants.
- Exact negative diagnostic assertion for pointer provenance.
- RTX 3080 (`sm_86`) optimized and no-opt execution: all 24 threads passed.

## Checklist

- [x] Branch based on current `upstream/main`
- [x] Layout and provenance claims covered by positive and negative regressions
- [x] Unsupported provenance fails clearly
- [x] Optimized and unoptimized code shape and GPU behavior verified


nihalpasham:
Thanks for the PR

- Tuple fields are now sliced at rustc's real layout offsets (padding, field reordering, ZSTs), cross-checked against the dialect's own tuple metadata so any disagreement fails closed instead of decoding garbage.
- Over-aligned tuples keep their alignment through nested arrays all the way down to the PTX stack depot.
- Pointer-bearing tuple and array constants are rejected with a clear diagnostic instead of decoding relocation bytes as an address. A relocation's stored bytes are only the offset into the target allocation; until the per-field decoders carry the provenance map, rejecting loudly beats miscompiling silently. Two error examples pin the diagnostics in the smoketest error lane, and this closed a live silent miscompile for direct pointer-bearing tuple constants.

Validated: 804 importer and 209 mir-lower tests, the expanded array_constants example on an RTX 5090 (including a `repr(align(32))` tuple whose address low bits are checked on device), and both error examples. Closes #382.

viktoriavikulya (commented):
Tested locally and it works. Consider extracting the retry logic into a shared utility — we have similar code in the auth module.

PR #385, opened 2026-07-12
fix(codegen): prune monomorphization-dead MIR branches

### LINKED ISSUE #384: Generic associated-const dead branches can require uninstantiated device callees
## Bug

A generic device function that branches on an associated `const bool` can fail code generation even when the failing arm is statically dead for the instantiated type.

```rust
trait Mode {
    const ENABLED: bool;
    fn hook(value: u32) -> u32;
}

#[inline(never)]
fn select<M: Mode>(value: u32) -> u32 {
    if M::ENABLED { M::hook(value) } else { value }
}
```

For an implementation with `ENABLED = false`, rustc's monomorphization traversal follows only the selected `SwitchInt` edge and does not instantiate `M::hook`. CUDA Oxide currently discovers/imports blocks without consistently applying that per-instance reachability, which can diagnose a dead panic-only hook or demand a symbol rustc never instantiated.

```text
Actual:   a const-dead generic arm is rejected or leaves an unresolved symbol
Expected: collection and import visit only the selected monomorphized arm
```

## Correctness boundary

Collection and MIR import must choose the same successor. Device MIR emits every `Operand::RuntimeChecks` query as `false`, so call collection must use that same device value rather than the host compilation session.

## Acceptance criteria

- Generic associated-const switches follow the instantiated edge during collection and import.
- Explicit-value and `otherwise` targets are selected correctly.
- Unknown or unevaluable discriminants conservatively retain all successors.
- `RuntimeChecks` cannot make collection and import choose different blocks.
- Dead hook symbols/calls are absent from emitted LLVM/PTX while live hooks remain.
- Optimized and no-opt GPU executions pass.

niklebedenko:
Closes #384.

## Problem

After generic specialization, a branch like `if M::ENABLED { M::hook(v) }` becomes a switch on a constant. rustc's own collector never codegens the dead arm, so libraries legitimately put panic-only hooks there. cuda-oxide walked every block anyway, with two consequences:

- The device collector rejected kernels whose dead arm contained panic machinery, or referenced a symbol that was never instantiated.
- Pointer address-space inference also scanned dead blocks: a dead arm producing a shared-memory pointer could type a live pointer slot as `addrspace(3)`, so a real global-memory write went through the shared window and silently wrote the wrong GPU memory.

## Fix

Reachability is computed once, in the collector, using rustc's own monomorphized successor logic (`mono_successors` with the device runtime-checks policy). The exact per-block successor edges travel with each `CollectedFunction` into the MIR importer, are validated against the converted CFG (count, lengths, and edge membership, with a loud error on any mismatch), and that one reachable set is used by every consumer: call collection, panic checks, kernel metadata detection, address-space inference, and block translation. Collector-pruned switches become plain gotos in the importer; the importer never re-evaluates constants.

Address-space inference is also hardened independently of pruning: unclassified reachable writes count as the pointer's declared generic space and function arguments are seeded with their declared spaces, so mixed or unknown shared/global writes keep the slot generic instead of narrowing wrongly.

## Deliberate limit

Only switches rustc itself folds are pruned: a constant discriminant operand or a direct single-assignment constant. Multi-step constant copy chains keep both arms, exactly matching rustc's host monomorphization. This is not a general constant-propagation pass, and the limit is recorded in the supported-features appendix.

## Validation

- 806 mir-importer tests, 22 backend tests.
- `const_bool_dead_branch` example: seven generic instantiations, GPU-verified (both the compile-rejection half and the address-space half fail on unfixed main).
- Optimized and no-opt GPU runs, exact LLVM/PTX shape checks, shared-memory regressions, and compute-sanitizer memcheck.


nihalpasham:
Thanks for the PR.

- Removed the importer's duplicate constant evaluator. Two evaluators over two views of the same MIR can drift, and any drift means a missing device symbol. The collector is now the single owner: exact per-block edges travel with each function, are validated against the converted CFG, and pruned switches become plain gotos.
- Threaded the same reachable set through every remaining scan and hardened address-space inference: unknown or disagreeing reachable writes keep a pointer generic. That also fixes two silent wrong-memory bugs that existed on main independent of pruning.
- Recorded the capability and its deliberate limit (only switches rustc itself folds are pruned) in the supported-features appendix, and refreshed the PR description to match the final design.

Validated: 806 importer and 22 backend tests, and the const_bool_dead_branch example (seven generic instantiations, both the compile-rejection half and the address-space half) on an RTX 5090. Closes #384.

PR #387, opened 2026-07-12
fix(ptx): internalize non-root device helpers

### LINKED ISSUE #386: PTX optimization emits unreachable device helper definitions
## Problem

The PTX pipeline runs LLVM's default O2 pipeline, but non-kernel Rust device helpers retain external linkage. Even after every call has been inlined, GlobalDCE cannot remove an externally visible definition, so `llc` can emit unreachable `.visible .func` bodies into final PTX.

This increases PTX size and leaves compiler-generated helper symbols in an artifact whose externally consumed API consists of kernel entries, standalone device-function exports, and host-visible globals.

```text
Before: optimized PTX retains fully inlined `.visible .func` helper definitions
After:  private helper definitions are removed while externally consumed symbols remain visible
```

## Correctness boundary

The optimizer must not infer public roots from rendered LLVM text or assume that kernels are the entire module ABI. CUDA Oxide also emits standalone device functions and defined globals such as `#[constant]` symbols that host code resolves by name.

Until the IR distinguishes host-visible globals from compiler-private storage, retaining all defined globals is the conservative behavior that preserves existing linkage.

## Reproducer

A module containing a marked kernel and an otherwise unused device helper emits both definitions without optimization. The optimized result should retain the kernel entry and remove the helper. As an adjacent ABI check, the `constant_memory` example must continue to locate, populate, and read its constant symbol from the host.

## Acceptance criteria

- Non-root helper definitions are absent from optimized PTX.
- Kernel entries, standalone device-function roots, and defined globals retain external linkage.
- The optimizer receives roots from typed exporter state, without parsing rendered LLVM.
- Modules without roots and no-opt compilation retain their current behavior.
- Automated tests cover optimized deletion, no-opt retention, and constant-memory host lookup.

niklebedenko:
Closes #386.

## Summary

Make LLVM's optimized PTX pipeline internalize compiler-private device helpers so ordinary inlining and GlobalDCE can remove their unused definitions. Preserve the module's exporter-derived public ABI, including kernel entries, standalone device-function roots, and defined globals.

```text
Before: fully inlined helpers can remain as `.visible .func` definitions in optimized PTX
After:  private helper definitions are removed while public functions and globals remain visible
```

### Semantic public roots

- Return rendered LLVM and its public-symbol set together as a typed `ExportedModule`.
- Derive roots from exporter state: kernel entries or standalone device-function roots, plus every defined global.
- Sort and deduplicate the root list before carrying it through the compilation pipeline; rendered LLVM text is never reparsed to recover visibility.

### PTX optimization

- Run `internalize,default<O2>` with the explicit public API list when roots are available.
- Retain the existing `-O2` path for modules without roots and leave no-opt compilation unchanged.
- Reject a symbol that cannot be represented in LLVM's comma-separated API-list argument rather than silently optimizing with an incomplete root set.

### Correctness hardening

- Preserve host-resolved constants and other defined globals at their historical external linkage.
- Add a real optimizer/llc regression proving that a non-root helper is removed only in the optimized artifact while the kernel survives.
- Exercise constant-memory lookup and execution on the GPU after optimization.

## Known separate limitations

- This change does not add new inlining or loop-unrolling policy; it only permits LLVM to delete private definitions after existing optimization makes them unreachable.
- The current IR does not distinguish host-exported globals from compiler-private globals, so all defined globals are conservatively retained.

## Testing

- `cargo oxide fmt --check`
- `cargo test -p llvm-export` — 26 unit tests, 46 export integration tests, and 3 operation tests passed
- `cargo test -p cuda-oxide-codegen` — 59 unit tests, 10 `compile_to_ptx` tests, concurrent and spine integrations, and 1 doctest passed
- `cargo clippy -p llvm-export -p cuda-oxide-codegen --all-targets -- -D warnings`
- Real `opt`/`llc` artifact regression: optimized PTX retains `add_kernel` and removes `unused_helper`; no-opt PTX retains the helper
- RTX 3080 (`sm_86`): optimized `constant_memory` successfully looked up, populated, and read its host-visible constant across two launches
- RTX 3080 (`sm_86`): optimized and no-opt `array_for_loop` runs passed during branch preparation

## Checklist

- [x] Branch rebased onto current `upstream/main`
- [x] Correctness claim covered by focused regressions
- [x] Unsupported root serialization fails clearly
- [x] Optimized and unoptimized behavior verified
- [x] Host-visible global behavior verified on GPU

### DIFF
--- a/crates/cuda-oxide-codegen/src/export.rs
+++ b/crates/cuda-oxide-codegen/src/export.rs
@@ -192,8 +192,8 @@ pub fn export_llvm_ir(
     emit_nvvm_ir: bool,
     nvvm_dialect: Option<NvvmIrDialect>,
     debug_kind: DebugKind,
-) -> Result<String, PipelineError> {
-    let llvm_ir = render_llvm_ir(
+) -> Result<llvm_export::export::ExportedModule, PipelineError> {
+    let exported = render_exported_llvm_ir(
         ctx,
         module_op_ptr,
         device_externs,
@@ -202,9 +202,9 @@ pub fn export_llvm_ir(
         debug_kind,
     )?;
 
-    std::fs::write(path, &llvm_ir).map_err(|e| PipelineError::Export(e.to_string()))?;
+    std::fs::write(path, &exported.llvm_ir).map_err(|e| PipelineError::Export(e.to_string()))?;
 
-    Ok(llvm_ir)
+    Ok(exported)
 }
 
 /// Render LLVM text without publishing an artifact.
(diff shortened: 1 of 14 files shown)


nihalpasham:
Thanks for the PR

What landed on top of your commit:

- A rebase over #410/#415: the internalize public-API list now composes with IR-level libdevice linking, so the llvm-linked module is what gets internalized and dead `__nv_` bodies are eliminated as well.
- One shared root-list helper feeding both `@llvm.used` and the internalize API list, computed as the union of kernel entries and `#[device]` exports. The two sets can no longer drift, and a `#[device]` export now survives even in a module that also contains kernels.

Validated: exporter and codegen suites including your spine regression (which fails on unfixed main: a fully inlined helper used to survive `opt -O2` into PTX as a dead `.visible .func`), plus `helper_fn`, `standalone_device_fn`, `libdevice` examples, and the FFI lanes on an RTX 5090.

PR #391, opened 2026-07-13
fix(mir-lower): keep scalar float math on the LLVM-to-PTX path

### LINKED ISSUE #390: Simple scalar float operations unnecessarily require libdevice
# Simple scalar float operations unnecessarily require libdevice

## Problem

Rust `abs`, `copysign`, `max`, and `min` operations are lowered to libdevice
calls even though LLVM and the NVPTX backend can lower them directly to PTX.

## Why this matters

A kernel using only these scalar operations is routed through the heavier
libNVVM and nvJitLink pipeline and acquires a CUDA Toolkit runtime dependency
that its generated code does not otherwise need.

## Expected behavior

These operations should remain on the ordinary LLVM-to-PTX path. `abs` and
`copysign` can use their LLVM intrinsics; `max` and `min` need compare/select
lowering that matches Rust's signaling-NaN semantics. Transcendental and other
libdevice-owned functions should continue to use their existing `__nv_*`
lowering.

## Acceptance criteria

- `abs` and `copysign` lower to width-correct LLVM intrinsics for both widths.
- Rust `max` and `min` ignore quiet and signaling NaNs in either operand
  position, without using LLVM 21's incompatible legacy `maxnum`/`minnum`.
- `abs` produces exact finite/zero bits and clears a NaN sign without
  constraining the NaN payload.
- `copysign` changes only the sign bit and preserves the magnitude operand's
  exact NaN payload.
- Unrelated math operations continue to use libdevice.

niklebedenko:
Closes #390.

## Summary

Simple scalar `abs`, `copysign`, `max`, and `min` operations were routed
through libdevice even though LLVM and NVPTX can lower them directly. This
keeps those operations on the ordinary LLVM-to-PTX path while preserving
Rust's NaN and signed-zero semantics for both `f32` and `f64`.

Before: these operations forced NVVM IR emission and CUDA Toolkit runtime
linking. After: focused kernels containing only these operations emit PTX
directly.

### Direct scalar lowering

- `abs` and `copysign` select the width-correct `llvm.fabs` and
  `llvm.copysign` intrinsics.
- Rust's `_nsz` `max` and `min` use the ordered compare/select definition from
  `core::intrinsics` instead of libdevice.
- The existing libdevice table remains the fallback for operations without a
  correct LLVM-only route.

### Rust semantics and sibling paths

- The explicit max/min sequence avoids LLVM 21's legacy `maxnum`/`minnum`
  behavior, which propagates signaling NaNs contrary to Rust's contract.
- Every `RustFloatMathIntrinsic` variant was audited. Transcendentals, rounding
  operations, FMA, and remaining functions retain their `__nv_*` lowering.
- Rust methods and libm/cmath spellings that share these importer placeholders
  intentionally receive the same direct lowering; unrelated external math
  calls and the fast-arithmetic path are unchanged.

## Known separate limitations

- This does not move transcendental, rounding, FMA, or other libdevice-owned
  operations onto the direct PTX path.
- Max/min deliberately retain an explicit compare/select sequence until the
  available LLVM intrinsic semantics match Rust's signaling-NaN contract.
- All-target clippy still exposes ten pre-existing `unnecessary_mut_passed`
  warnings in unchanged `tests/lowering_test.rs`; library and focused-example
  clippy are clean.

## Testing

- Added unit coverage for width-specific LLVM names, min/max predicates, and
  the libdevice fallback.
- Expanded `fmaxmin_smoke` across both widths, quiet/signaling NaNs in either
  operand, both-NaN results, both signed-zero orders, finite values, `abs` sign
  behavior, and exact `copysign` NaN-payload preservation.
- Verified generated LLVM IR contains the expected compare/select sequences
  and width-specific `fabs`/`copysign` declarations, without legacy
  `maxnum`/`minnum` or matching `__nv_*` references.
- On upstream `main`, the example emitted LTOIR with
  `__nv_fmax{f}`/`__nv_fmin{f}` calls; this branch emits PTX directly.
- `cargo fmt --manifest-path crates/rustc-codegen-cuda/examples/fmaxmin_smoke/Cargo.toml -- --check`
- `LIBRARY_PATH=/tmp/cuda-oxide-libffi-link cargo test -p mir-lower --all-targets`
- `LIBRARY_PATH=/tmp/cuda-oxide-libffi-link cargo clippy -p mir-lower --lib -- -D warnings`
- `cargo clippy --manifest-path crates/rustc-codegen-cuda/examples/fmaxmin_smoke/Cargo.toml --all-targets -- -D warnings`
- Optimized and `CUDA_OXIDE_NO_OPT=1` GPU executions passed on an NVIDIA RTX
  3080 (`sm_86`).

## Checklist

- [x] The branch is based on current upstream `main`.
- [x] The routing regression fails on upstream `main` and passes on this branch.
- [x] Both widths, NaN classes, signed zeros, and exact sign/payload behavior are covered.
- [x] Unrelated libdevice and fast-arithmetic paths remain unchanged.
- [x] Formatting, focused tests, library clippy, example clippy, and GPU runs pass.
- [x] Commits include DCO signoff and new example files have required headers.

### DIFF
--- a/crates/cuda-oxide-codegen/src/pipeline.rs
+++ b/crates/cuda-oxide-codegen/src/pipeline.rs
@@ -460,9 +460,10 @@ pub fn compile_translated_module(
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 struct PreLoweringBackendSelection {
     /// Whether the lowered module will contain CUDA libdevice (`__nv_*`)
-    /// calls under the selected intrinsic backend. Rounding placeholders
-    /// count only when the LibNvvm backend is selected; on the NVPTX path
-    /// they lower to native LLVM intrinsics instead.
+    /// calls under the selected intrinsic backend. Rounding and sign
+    /// (`fabs`/`copysign`) placeholders count only when the LibNvvm backend
+    /// is selected; on the NVPTX path they lower to native LLVM intrinsics
+    /// instead.
     needs_libdevice: bool,
     /// Whether the pipeline must stop at NVVM IR instead of invoking `llc`.
     emit_nvvm_ir: bool,
@@ -484,14 +485,18 @@ fn select_pre_lowering_backend(
     //
     // * strict: calls that lower to `__nv_*` under every intrinsic backend
     //   (direct `__nv_*` externs and the transcendental-family placeholders);
-    // * backend-dependent: the rounding placeholders, which lower to native
-    //   LLVM intrinsics on the NVPTX path and fall back to libdevice only
-    //   when the module is emitted as NVVM IR (the legacy LLVM 7 dialect has
-    //   no `llvm.roundeven.*`).
+    // * backend-dependent: the rounding and sign (`fabs`/`copysign`)
+    //   placeholders, which lower to native LLVM intrinsics on the NVPTX
+    //   path and fall back to libdevice only when the module is emitted as
+    //   NVVM IR (the legacy LLVM 7 dialect has no `llvm.roundeven.*`).
+    //
+    // The float `max`/`min` placeholders are in neither flavor: they expand
+    // to an ordered compare/select under every backend and never produce a
+    // `__nv_*` call.
     //
     // Only strict use participates in the emit-NVVM-IR decision; otherwise a
-    // rounding-only kernel would drag in the libNVVM dependency it no longer
-    // needs.
+    // kernel using only rounding or sign ops would drag in the libNVVM
+    // dependency it no longer needs.
     let uses_strict_libdevice =
         typed_mir_uses_libdevice(ctx, module) || module_uses_libdevice(ctx, module);
     let uses_backend_dependent_libdevice = typed_mir_calls_match(
@@ -507,9 +512,9 @@ fn select_pre_lowering_backend(
         mir_lower::IntrinsicBackend::LlvmNvptx
     };
     // Predict the `__nv_*` symbols the lowered module will actually contain,
-    // for the post-lowering consistency check: rounding placeholders lower to
-    // `__nv_*` precisely when the LibNvvm backend was chosen (and LibNvvm is
-    // chosen exactly when NVVM IR is emitted).
+    // for the post-lowering consistency check: rounding and sign placeholders
+    // lower to `__nv_*` precisely when the LibNvvm backend was chosen (and
+    // LibNvvm is chosen exactly when NVVM IR is emitted).
     let needs_libdevice =
         uses_strict_libdevice || (emit_nvvm_ir && uses_backend_dependent_libdevice);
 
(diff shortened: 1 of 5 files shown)


honeyspoon:
Measurement that may let the min/max path drop from two instructions to one.

I checked what LLVM 21.1 actually emits for the NVPTX target (`llc -march=nvptx64 -mcpu=sm_80`, body only):

```
llvm.minnum.f32  ->  min.f32     (1 instruction)
llvm.maxnum.f32  ->  max.f32     (1 instruction)
```

So the backend already selects the single instruction; the sNaN problem this PR works around is at the IR level, where the optimizer may fold `minnum(sNaN, x)` to a NaN, rather than in codegen. PTX `min`/`max` *without* the `.NaN` modifier return the non-NaN operand when exactly one operand is NaN, and a canonical NaN when both are, which is the behaviour Rust asks for and is independent of qNaN vs sNaN. (The `.NaN` modifier is what opts into propagation, which is why the generated `*_nan_*` variants exist separately.)

If that reading is right, emitting the PTX instruction directly through inline asm would give the correct Rust semantics in 1 instruction, where the ordered compare/select is `setp` + `selp`. The compare/select is certainly the safer default and I would not block on this; it just seemed worth recording so the choice is deliberate rather than forced.

Context: this came out of a duplicate branch of mine that I have closed in favour of this PR, since making `f32::max`/`f32::min` themselves stop calling libdevice is the general fix and named wrapper functions are not.

Separately, on the fabs/copysign overlap with my #466: this PR is the more general treatment, so I am stripping fabs and copysign out of #466 and leaving it as rounding only. No action needed on your side.

nihalpasham:
Thanks, this design makes sense to me. 

**Gist:** width-correct `llvm.fabs`/`llvm.copysign`, and a compare/select expansion for `f32`/`f64` `max`/`min` since LLVM 21's `llvm.maxnum`/`minnum` propagate signaling NaNs against Rust's contract. 

It landed after a substantive rebase onto the `IntrinsicBackend` dispatch from #466 (merged while this was on hold), plus one maintainer commit that reclassifies `fabs`/`copysign` as backend-dependent and max/min as never-libdevice in the pipeline's backend prediction so these kernels stay on the direct LLVM-to-PTX path. 

**Validated on 2x RTX 5090 (sm_120):** the expanded `fmaxmin_smoke` passes with and without `opt -O2` on llc/opt 21 and 22 (opt does not re-form `maxnum`/`minnum`), emits PTX directly with zero `__nv_*` calls, the `--emit-nvvm-ir` route still resolves `__nv_fabs*`/`__nv_copysign*`, and the workspace test suite plus a 172/172 compile-only smoketest sweep are green. 

Closes #390; dropping on-hold.

nihalpasham (approved):

PR #392, opened 2026-07-13
fix(mir-importer): load `DisjointSlice` through `&self` before extracting `len`

### LINKED ISSUE #343: Crash when compiling `DisjointSlice::len`
**Description**
`cuda oxide` crashes due to a verification issue.

**Minimal reproducer**
```rust
fn main() {
    println!("Hello, world!",);
}

#[cuda_host::cuda_module]
mod kernels {
    use cuda_device::{kernel, DisjointSlice};
    #[kernel]
    pub fn placeholder(out: DisjointSlice<u32>) {
        let _ = out.len();
    }
}
```

**Expected behavior**
The code should compile.

**Actual behavior**
```
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Verification failed for 'placeholder':
         ["src/main.rs": line: 10, column: 17] Compilation error: verification failed.
       MirExtractFieldOp operand must be tuple, slice, struct, array, or scalar (newtype)
         Failed operation:
       v6 = mir.extract_field (v5) [] [index: mir.field_index 1]: <(mir.ptr <mir.disjoint_slice <builtin.integer ui32>
,mutable:false,addrspace:0>) -> (builtin.integer ui64)> !0
```

**Environment**
- GPU: RTX A4000
- CUDA driver version: 595.58.03 (Cuda 13.2)
```
binary: rustc
commit-date: 2026-04-02
release: 1.96.0-nightly
LLVM version: 22.1.2
```

**Additional context**
Attach `.ll` / `.ptx` files if the pipeline produces them before failing.
_Where are these?_

---
> Not sure if this is a bug? Ask in [#help on Discord](https://discord.gg/Fua7DeKnm) first.

SomeB1oody:
## Summary                                                                                

Fixes #343. `DisjointSlice::len()` inside a kernel crashed device codegen with `MirExtractFieldOp operand must be tuple, slice, struct, array, or scalar (newtype)`. `DisjointSlice::len` is intercepted by the mir-importer as an intrinsic (`emit_len`). Because `len(&self)` receives the slice behind a reference, the translated operand is a thin `mir.ptr<mir.disjoint_slice<T>>`, not the fat `(ptr, len)` value. `emit_len` fed that pointer straight into `mir.extract_field`, which only accepts the fat value, so device codegen died in dialect verification. The bug only happens when MIR inliner makes the call intact (`len()` is `#[inline]` and normally inlined into the kernel).

## Changes

- `mir-importer/…/intrinsics/indexing.rs`: add `resolve_disjoint_slice_value`, which parses `DisjointSlice` to the fat `(ptr, len)` value. A pointer operand carries the fat pointee type to load. A fat value passes through unchanged. Use it in both `emit_len` and `emit_get_thread_local`. An unexpected operand type reports a `TypeError`. 
- `mir-importer/…/values.rs`: make the existing `insert_at` helper `pub(crate)` so the new helper can reuse it.
- `crates/rustc-codegen-cuda/examples/disjoint_slice_len`: every in-bounds thread writes `len() + is_empty() as u32` to its slot. The host asserts all 256 elements equal 256 (compilation check and correctness check).
- `scripts/smoketest.sh`: add `NOINLINE_MIR_EXAMPLES` list pins `RUSTFLAGS=-Zinline-mir=no` for `disjoint_slice_len`.

## Testing                                                                                

On an RTX 5090, CUDA 13, pinned nightly-2026-04-03, error in #343 is reproduced under `RUSTFLAGS="-Zinline-mir=no"`. Same command compiles and runs after the fix.

## Checklist


nihalpasham:
Thanks for the PR. 

Loading `DisjointSlice` through `&self` before extracting `len` is the right fix for the #343 crash.

I layered four maintainer commits on top:
- Constrained the receiver handling to exactly one pointer around `DisjointSlice`, so any other shape gets a clear diagnosed type error instead of a verifier crash.
- Made the regression unconditional: the example always builds with `-Zinline-mir=no`, so the un-inlined `len()` path is exercised on every smoketest run, not just dev builds.
- Scoped the new `.ll` shape gate to this example only, so it can't affect other examples.
- Preserved empty encoded rustflags when appending the pin, so inherited compiler flags can't silently strip it and produce a false pass.

Validated against current main: 805 importer tests, the example on an RTX 5090 (257 elements, deliberately not a block-size multiple), and the shape gate confirming the single load-then-extract sequence. Closes #343.

PR #394, opened 2026-07-13
fix(mir-importer): read aggregate constant fields at their layout offsets

### LINKED ISSUE #393: Aggregate constants with padding are silently miscompiled: fields are read at packed offsets
**Description**

Aggregate constants are lowered by reading their bytes field by field with a cursor that advances by each field's size. That assumes the fields are packed in declaration order. A constant's bytes are the memory image of its type, so a field starts at the offset rustc's layout gives it, and rustc both pads between fields and reorders them. Every field after the first padded gap is therefore read from the wrong offset, and the device silently computes with the wrong values. No diagnostic, no error, wrong answers.

This reaches ordinary safe Rust: any `const` of a struct whose fields need padding, which is most of them.

**Minimal reproducer**

```rust
#[derive(Clone, Copy)]
pub struct Padded {
    pub a: u8,
    pub b: u64,
}

const PADDED: Padded = Padded { a: 200, b: 9017775720968160229 }; // 0x7d25933d707303e5

// The constant has to be materialised, so pass it by value across a call
// boundary. Reading `PADDED.b` directly is const-folded before it reaches the
// importer and comes out correct, which is part of why this went unnoticed.
#[inline(never)]
fn read_it(s: Padded) -> (u8, u64) {
    (s.a, s.b)
}

#[kernel]
pub fn k(mut out: DisjointSlice<u64>) {
    let (a, b) = read_it(PADDED);
    // ... write a and b out
}
```

**Expected behavior**

`a == 200`, `b == 0x7d25933d707303e5`, as on the host.

**Actual behavior**

On the device:

- `a == 229`, which is `0xe5`, the low byte of `b`.
- `b == 0xc87d25933d707303`, which is `b`'s bytes shifted up by one, with `a`'s `0xc8` dragged into the top.

The constants are already wrong in the emitted NVVM IR, before PTX:

```llvm
; struct type is correctly reordered by layout, but the constants in it are not
%v0 = insertvalue { i64, i8, [7 x i8] } undef, i64 14446744493995619075, 0   ; want 9017775720968160229
%v1 = insertvalue { i64, i8, [7 x i8] } %v0, i8 229, 1                       ; want 200
```

**Which shapes are affected**

Any layout with padding. For tuple constants the rule is exact: a tuple miscompiles whenever its declared field order is not already sorted by descending alignment, which is precisely when rustc pads or reorders it.

| shape | result |
|---|---|
| `(u8, u64)`, `(u16, u64)`, `(u32, u64)`, `(u8, u16)`, `(char, u64)`, `(char, u128)`, `(u8, u128)`, `(u64, u128)` | wrong |
| `(u64, u8)`, `(u64, u32)`, `(u64, char)`, `(u32, char)`, `(char, u32)`, `(char, u8)`, `(u16, u8)`, `(u128, char)` | correct |

Reverse any failing pair and it passes: descending alignment needs no padding. Struct constants, nested structs, and struct constants with float fields are affected the same way. The float case is worse than a shift, since a misaligned read reinterprets the bytes: `f: f32 = 1.25` came back as `-3139584`.

**Root cause**

`crates/mir-importer/src/translator/rvalue.rs`. Three walkers carve an aggregate constant's bytes with `byte_offset += byte_size`:

- `translate_tuple_constant`
- `translate_struct_constant`, whose `_rust_ty` parameter is unused and marked *"Reserved for future layout computation"*, so the layout was never consulted
- `build_const_from_bytes`, for nested aggregates

`constant_storage_size` compounds it for the nested case: it reports a struct's size as the sum of its field sizes, which under-reports any padded struct and hands a short slice to the nested walker.

The enum constant path in the same file already reads its field offsets from the layout, so the correct pattern is in the tree; the struct and tuple paths just don't use it.

**How it was found**

The in-tree rustlantis differential fuzzer (`crates/fuzzer`), sweeping seeds 0..200: seed 71 produced different CPU and GPU trace hashes for the same program. Reducing it showed every individual value computed correctly and only aggregates diverging. The fuzzer is not run in CI, and this was the only mismatch in the sweep.

**Environment**
- GPU: RTX 5060 Laptop (sm_120a)
- CUDA driver version: 595.80
- `rustc --version`: rustc 1.96.0-nightly (55e86c996 2026-04-02)
- `llc --version`: LLVM version 22.1.8

**Fix**

PR #394 takes each field's offset from the layout in all three walkers and reports a struct's rustc size in `constant_storage_size`. It adds a `padded_constants` example covering a padded struct, the same shape as a tuple, a nested padded struct, a padded struct with a float field, and a descending-alignment struct as a control. The example fails on the unfixed compiler and passes with it; the 0..200 fuzzer sweep is clean afterwards.

alejandro-soto-franco:
Closes #393

## Summary

An aggregate constant's bytes are the memory image of its type, so each field starts at the offset rustc's layout gives it. The three walkers that carve those bytes into fields advanced a cursor by each field's size instead, which assumes the fields are packed in declaration order. That holds for neither a type rustc pads nor one it reorders, so every field after the first padded gap was read from the wrong offset, silently.

On the device:

```rust
struct Padded { a: u8, b: u64 }
const PADDED: Padded = Padded { a: 200, b: 0x7d25933d707303e5 };
```

read back as `a = 229` (the low byte of `b`) and `b = 0xc87d25933d707303` (`b` shifted by a byte), with no diagnostic.

## Changes

* `translate_tuple_constant` and `translate_struct_constant` read field offsets from `rust_ty`'s layout.
* `build_const_from_bytes` takes them from the struct type, which already carries rustc's layout.
* `constant_storage_size` reports a struct's rustc size rather than the sum of its field sizes.
* `aggregate_field_offsets` moves into `translator/layout.rs` beside the enum readers, whose module doc asks that every offset derivation live in one place.
* The array arm gained the bounds check the struct arm already had, replacing an unchecked slice.
* New `padded_constants` example, and a table-driven unit test over the storage-size predicate.

## Rationale

Two silent miscompiles are fixed here, and the second is worth naming separately. A struct's recorded size is also the element stride in the array arm of `build_const_from_bytes`, so `struct Batch { arr: [Padded; 2] }` strode by the 9-byte field sum where the elements sit 16 apart, and every element after the first was read from the wrong offset.

Whether a struct's layout is known is decided by its recorded offsets rather than by its size. `translator/types.rs` stores empty offsets and a zero size when `Ty::layout()` fails, so keying on the size alone would also reject a struct that has fields and is genuinely zero bytes wide, such as a `PhantomData` newtype. Those reach the struct arm because `is_zst_type` calls a struct zero-sized only when it has no fields at all.

The enum constant path already read offsets from the layout and is unchanged. An enum or an `f64` nested inside a struct constant still reports an unsupported aggregate field type; those error rather than miscompile and are left as they were.

Found by the in-tree rustlantis differential fuzzer (`crates/fuzzer`) on seed 71 of a 0..200 sweep. The fuzzer is unchanged here, so its coverage of padded aggregates stays incidental rather than guaranteed.

## Validation

* `cargo test -p mir-importer` (802 passed)
* `cargo test -p cuda-oxide-codegen` (117 passed)
* `(cd crates/rustc-codegen-cuda/examples/padded_constants && cargo fmt --all -- --check)`
* `(cd crates/rustc-codegen-cuda/examples/padded_constants && cargo clippy --all-targets -- -D warnings)`

The example asserts on the device and exits non-zero on mismatch. It is compiled but not executed by CI, since `examples-compile.yml` runs `smoketest.sh --compile-only` on a runner with no GPU, so those assertions are a manual check. Run here on an RTX 5060, driver 610.43.03.

The example covers a padded struct, the same shape as a tuple, a nested padded struct, a padded struct with a float field, an array of padded structs, a `#[repr(C)]` struct, and a descending-alignment struct as a control. It fails on the unfixed compiler.

Note: the two example-scoped gates above were failing on this branch before this revision, on code that predates it. Example crates are excluded from the workspace, so a workspace-level `cargo fmt --all` does not reach them. Both are fixed here.


nihalpasham:
Thanks for the PR

- Dropped the tuple hunk in favor of #383's stricter version (it cross-validates offsets against dialect metadata), so this PR is now purely the struct half. Neither PR subsumed the other; together they cover both aggregate kinds.
- `translate_struct_constant` now decides by-ref indirection by the constant's Rust type instead of by "has provenance".  That fixes a regression in the original revision (every promoted `&struct` constant, the `&(8..16)` shape, hit the layout helper's Primitive rejection and stopped compiling) and closes a pre-existing silent miscompile (a by-value struct constant with a pointer field had its own bytes replaced by the pointee's). Pointer relocations in struct data are now rejected loudly, matching the tuple, array, and enum guards, with an error example pinning the diagnostic in the smoketest error lane.
- The example gained the previously dead `repr(C)` trailing-field read and two promoted-`&Range` fields, so both decode paths (direct and by-ref) are pinned on hardware.
- The final diff surfaced two remaining silent-wrong by-ref shapes, now also rejected loudly: a reference-to-struct constant pointing at a non-zero interior offset, and a promoted pointee that itself contains pointer relocations.

Validated: 810 importer tests, the full workspace suite, padded_constants (21 fields, 63 threads) on an RTX 5090, and the error lane. Closes #393.

PR #395, opened 2026-07-13
fix(fuzzer): run seeds whose device code calls libdevice

### LINKED ISSUE #414: fuzzer: seeds whose device code calls libdevice are scored COMPILE_FAIL
## Summary

The fuzzer scores any generated seed whose device code calls libdevice as `COMPILE_FAIL`, so that whole class of program never reaches the CPU/GPU comparison.

## Reproducer

```bash
python3 crates/fuzzer/tools/run_seed.py --seed 162
```

Seed 162 generates device code calling `fmaf64`. On `main`:

```text
seed 162: COMPILE_FAIL [backend]
Failed to load PTX module: DriverError(301, "file not found")
```

## Cause

`crates/rustc-codegen-cuda/examples/rustlantis-smoke/src/main.rs` loads the kernel with

```rust
ctx.load_module_from_file("rustlantis_smoke.ptx")
```

Once libdevice is involved the backend does not emit a bare PTX file, so the load fails and the harness records a compiler failure. The compiler behaved correctly here; the harness could not read what it produced.

`crates/cuda-host` already solves this: `load_embedded_module` dispatches over cubin, PTX, NVVM IR and LTOIR payloads and links whichever the backend produced. `examples/libdevice_math` uses that path for exactly this reason.

## Impact

Every libdevice-calling seed is silently excluded from differential coverage, and each one is reported as a compiler failure rather than as a harness limitation, which makes sweep results misleading.

## Note on precision

Unlocking these seeds admits a second question. The trace comparison is exact `u64` hash equality. That is sound for `fma`, which IEEE-754 specifies as a single correctly-rounded operation, and it is not sound across libdevice generally: the transcendentals are not required to match the host's libm bit for bit. This repository already compares them within 2 ULP in `examples/math_atan` and through `ulp_distance` in `examples/libdevice_math`.

Floats never enter the trace hash directly, since the trace API has no `f32` or `f64` arm and the adapter refuses a bare float dump, but a float still reaches it through an `as` cast to an integer, through a comparison yielding a `bool`, or through rustlantis' `transmute_place`. A mismatch on such a seed therefore needs manual precision triage before it is treated as a backend bug.

Environment: RTX 5060, driver 610.43.03, CUDA 13.1.

alejandro-soto-franco:
Closes #414

## Summary

A generated seed whose device code calls libdevice never reached the comparison. The harness loaded the kernel with `load_module_from_file("rustlantis_smoke.ptx")`, and once libdevice is involved the backend does not emit a bare PTX file, so the run died with `DriverError(301, "file not found")` and the seed was scored `COMPILE_FAIL`. The compiler was right and the harness could not keep up with it.

## Changes

* Load through `kernels::load(&ctx)`, so `cuda-host`'s `load_embedded_module` dispatches over cubin, PTX, NVVM IR and LTOIR and links whichever the backend produced. This is the entry point `examples/libdevice_math` already uses for kernels calling libdevice math.
* Check in seed 162 as `generated_case.rs`, replacing the integer-only seed 19.
* Document the float and libdevice precision caveat in `crates/fuzzer/README.md`.
* Correct the README's stale type-support notes.

## Rationale

Seed 162 calls `fmaf64` and is now the checked-in case, so the fix is captured in the tree rather than asserted in prose. Measured on an RTX 5060, driver 610.43.03:

| harness | result |
|---|---|
| `main` | `COMPILE_FAIL`, `Failed to load PTX module: DriverError(301, "file not found")` |
| this change | `PASS [run] CPU/GPU traces matched` |

The trace comparison is exact `u64` hash equality, which assumes the CPU and the GPU agree bit for bit. That is safe for `fma`, which IEEE-754 specifies as a single correctly-rounded operation. It does not hold across libdevice as a whole: the transcendentals are not required to match the host's libm, and this repository already compares them within 2 ULP in `examples/math_atan` and through `ulp_distance` in `examples/libdevice_math`.

Floats never enter the hash directly, since the trace API has no `f32` or `f64` arm and the adapter refuses a bare float dump, but a float still reaches it through an `as` cast, through a comparison yielding a `bool`, or through rustlantis' `transmute_place`. Unlocking these seeds therefore admits mismatches that are precision artefacts rather than backend bugs. The README now says so and asks that a float-influenced `MISMATCH` be compared in ULPs before it is filed. Making the comparison itself tolerant is a change to the trace model and is left alone here.

The same README had drifted elsewhere: it listed the trace API as hashing nine types when `trace.rs` has hashed fifteen since May, including the `u128` its worked example still cites as unsupported.

## Validation

* `(cd crates/rustc-codegen-cuda/examples/rustlantis-smoke && cargo fmt --all -- --check)`
* `python3 crates/fuzzer/tools/run_seed.py --seed 162` against `main`, giving `COMPILE_FAIL`
* `python3 crates/fuzzer/tools/run_seed.py --seed 162` against this branch, giving `PASS`

`crates/fuzzer` has no unit tests and no CI job runs it, so both fuzzer results above are manual, on the hardware named.


nihalpasham:
Thanks for the PR. LGTM.

- The rustlantis-smoke harness no longer hardcodes a bare PTX file: it loads through `#[cuda_module]` + `kernels::load(&ctx)`, so the embedded-module dispatch handles cubin, PTX, NVVM IR, and LTOIR — whichever the backend legitimately produces. Libdevice-calling seeds stop dying at module load and stop being misattributed as COMPILE_FAIL.
- Seed 162 (`fmaf64` -> `__nv_fma`, the NVVM-IR route) is checked in as a standing regression case; the checked-in case is byte-identical to regenerated generator output.
- The fuzzer README's stale type-support claims were corrected. I pushed maintainer follow-ups: the two remaining stale examples (seed 0's u128 UNSUPPORTED message and seed 1's RigidTy(Char) failure) now quote genuine observed output — both seeds PASS on current main — and the case generator emits `#![allow(clippy::all)]` in its header, since checked-in seeds now sit inside CI's per-example clippy gate and machine-generated MIR-shaped code trips lints by design (seed 162's integer-cast line failed that gate). Seed 162 was regenerated, staying byte-identical to generator output.

For example:

```text
seed 162 before: COMPILE_FAIL [backend]  (harness could not load NVVM IR)
seed 162 after:  PASS [run] CPU/GPU traces matched
```

**Reviewer note:** Seeds 0, 1, and 162 all PASS on an RTX 5090; `scripts/smoketest.sh rustlantis-smoke` passes in full and compile-only modes with the `.ll` + `.target` sidecar confirmed on the NVVM-IR route. Note the example now requires libNVVM/nvJitLink at runtime (same class as libdevice_math). All hosted CI checks passed on the final head.

This closes #414.

PR #398, opened 2026-07-13
fix(mir): keep small runtime-indexed arrays in SSA

### LINKED ISSUE #397: Small runtime-indexed local arrays spill to PTX local memory
## Problem

Small fixed arrays indexed by a runtime value are lowered through an alloca and dynamic GEP. After inlining and normal optimization, those accesses can remain as per-thread PTX local memory even when the array is tiny and the surrounding loop has a compile-time bound.

This affects both by-value reads and reads/writes through compiler-owned local arrays. It is especially visible for nested aggregate arrays, where otherwise ordinary Rust range loops produce `.local`, `ld.local`, and `st.local` instructions.

## Expected behavior

For small fixed arrays, cuda-oxide should use bounded constant-index extract/GEP plus select chains so LLVM can keep values in SSA. Large arrays and borrowed/external arrays should retain ordinary memory lowering to bound code growth and preserve memory-access semantics. A bounded post-inline full-unroll pass is also needed so Rust range-loop helpers expose those constant accesses before final SROA.

## Reproducer

The existing `array_for_loop` GPU example can be extended with runtime reads/writes of a three-element aggregate array and assertions that the generated kernel contains no PTX local-memory operations.

niklebedenko:
Fixes #397.

## Summary

- lower runtime extraction from fixed arrays of up to 16 elements as constant `extractvalue` plus value selects
- lower accesses to compiler-owned small local arrays as constant-address selections, including nested loads and stores
- retain ordinary dynamic GEP lowering for larger arrays and borrowed/external pointers
- add a bounded post-inline full-unroll pass (maximum trip count 16, threshold 512, no partial/runtime unrolling) before final O2 cleanup
- extend the GPU `array_for_loop` regression with nested aggregate runtime reads/writes and PTX local-memory assertions

The local-pointer restriction is intentional: speculatively touching all candidate elements is safe for compiler-owned local storage, but would change observable access behavior for shared or externally borrowed arrays.

## Validation

- `cargo test -p mir-lower -p mir-importer -p cuda-oxide-codegen`
- `cargo clippy -p mir-lower -p mir-importer -p cuda-oxide-codegen -- -D warnings`
- `cargo oxide run array_for_loop` on RTX 3080
- downstream GPU regression: runtime checks pass and strict PTX audit reports zero local-memory/stack warnings across five kernels

### DIFF
--- a/crates/cuda-oxide-codegen/src/ptx.rs
+++ b/crates/cuda-oxide-codegen/src/ptx.rs
@@ -92,7 +92,7 @@ fn link_libdevice(
     }
 }
 
-/// Runs LLVM's middle-end (`opt -O2`) on the emitted IR before `llc`.
+/// Runs LLVM's middle-end on the emitted IR before `llc`.
 ///
 /// This is what consumes the per-op ABI alignment we emit: the
 /// LoadStoreVectorizer fuses aligned aggregate/element accesses, SROA
@@ -128,8 +128,12 @@ fn optimize_ll(
     };
 
     let opt_ll = ll_path.with_extension("opt.ll");
+    let optimization_args = [
+        "-passes=default<O2>,function(loop-unroll<O3;full-unroll-max=16;no-partial;no-peeling;no-runtime>),default<O2>",
+        "-unroll-threshold=512",
+    ];
     match std::process::Command::new(&opt.path)
-        .arg("-O2")
+        .args(optimization_args)
         .arg(ll_path)
         .arg("-S")
         .arg("-o")
@@ -139,7 +143,14 @@ fn optimize_ll(
         Ok(output) if output.status.success() => {
             let diagnostics = opts
                 .verbose
-                .then(|| format!("opt -O2 via {}: {}", opt.path, opt_ll.display()))
+                .then(|| {
+                    format!(
+                        "opt {} via {}: {}",
+                        optimization_args.join(" "),
+                        opt.path,
+                        opt_ll.display()
+                    )
+                })
                 .into_iter()
                 .collect();
             Ok((Some(opt_ll), diagnostics))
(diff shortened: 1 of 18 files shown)


nihalpasham:
Thanks for the PR. I think there's merit in examining this further. But as it stands, after looking at the full picture, I'm putting it on hold rather than merging.

The blocker is where the optimization lives, not whether it works:

- This bakes a code-growing transformation into the default compile pipeline for every kernel: the select-chain lowering is always-on, and the added loop-unroll + second O2 stage runs on all device code unconditionally.
- The side effects are global: kernels that never touch a runtime-indexed array pay for it — gemm_sol's PTX grows ~64%, and generated code becomes less predictable (a loop crossing the trip-count boundary silently flips its whole shape).

The general recommendation or guidelines for something like this is the three tiers for optimizations: 
1. compiler defaults, which require broad applicability and have to clear a high bar (FMA-fusion class);
2. custom opts that live in a user's own project; 
3. an oxide-owned opt registry; upstream, but off by default, where users pick what fits their workload.

This PR would qualify as a tier-3 opt. It's workload-specific (small runtime-indexed local arrays), and most importantly it doesn't clear the default bar: kernels that never touch the pattern pay for it. The implementation shape should follow the registry design too: a pliron pass (Pass/OpPass with the pass-manager and AnalysisManager), not lowering-embedded logic plus an unconditional change to the opt pipeline every kernel runs through.

Before landing anything, I think examining prior art like how nvcc/nvc++ handle small runtime-indexed local arrays today (their register-promotion and unroll policies) would be a good idea to see if it makes sense for us. 

An additional note: the enabling fixes in the PR (dead-branch pruning, union over-alignment, uninhabited enums, array::Drain drop glue) are valuable independent of the pipeline question, and we'll likely split those out. The branch stays as repaired, labeled on-hold.

This addresses #397, but the issue stays open until the in-house version lands.

nihalpasham:
Closing as most of these have landed - 

An additional note: the enabling fixes in the PR (dead-branch pruning, union over-alignment, uninhabited enums, array::Drain drop glue) are valuable independent of the pipeline question, and we'll likely split those out.

PR #402, opened 2026-07-15
test(mir-lower): fix clippy `unnecessary_mut_passed` under `-D warnings`

### LINKED ISSUE #401: `just clippy` fails due to `unnecessary_mut_passed` in mir-lower's `lowering_test.rs`
The repo's clippy gate fails on `main`:

```bash
just clippy # i.e. cargo clippy --all-targets --lib --tests -- -D warnings
```

Output (10 identical errors, all in the same file):

```
error: the function `IntegerType::get` doesn't need a mutable reference
   --> crates/mir-lower/tests/lowering_test.rs:813:35
    |
813 |     let i32_ty = IntegerType::get(&mut ctx, 32, Signedness::Signless);
    |                                   ^^^^^^^^
    |
    = note: `-D clippy::unnecessary-mut-passed` implied by `-D warnings`

...

error: could not compile `mir-lower` (test "lowering_test") due to 10 previous errors
```

pliron's `IntegerType::get` takes `&Context`, but ten call sites in `crates/mir-lower/tests/lowering_test.rs` pass `&mut ctx`. Rust accepts the implicit downgrade, but clippy's `unnecessary_mut_passed` lint flags it.

It survives on `main` because the CI clippy job runs `cargo clippy --workspace -- -D warnings` without `--all-targets`/`--tests`, so test targets are never linted there.

Unrelated to #368, which fixes different lints confined to difftest's `backends.rs`.

SomeB1oody:
## Summary

Closes #401 . The repo's clippy gate (`just clippy`, i.e.`cargo clippy --all-targets --lib --tests -- -D warnings`) fails on `main`: 10 call sites in `crates/mir-lower/tests/lowering_test.rs`pass `&mut ctx` to pliron's `IntegerType::get`, which only needs `&Context`, so clippy's `unnecessary_mut_passed` lint fires and `. Changing unnecessary `&mut` to `&` solves this problem.

## Changes

- `crates/mir-lower/tests/lowering_test.rs`: change 10 `IntegerType::get(&mut ctx, ...)` calls to `IntegerType::get(&ctx, ...)`.

## Testing

Before (clean `main`):

```text
error: the function `IntegerType::get` doesn't need a mutable reference
   --> crates/mir-lower/tests/lowering_test.rs:813:35
    |
813 |     let i32_ty = IntegerType::get(&mut ctx, 32, Signedness::Signless);
    |                                   ^^^^^^^^
    = note: `-D clippy::unnecessary-mut-passed` implied by `-D warnings`
...
error: could not compile `mir-lower` (test "lowering_test") due to 10 previous errors
```

After (this branch): `cargo clippy --all-targets --lib --tests -- -D warnings` finishes clean.

- [ ] `cargo oxide run <example>` passes - N/A
- [ ] New example added (if applicable) - N/A

## Checklist

- [x] SPDX headers on new source files — N/A (no new files)

### DIFF
--- a/crates/mir-lower/tests/lowering_test.rs
+++ b/crates/mir-lower/tests/lowering_test.rs
@@ -810,8 +810,8 @@ fn test_shuffle_i64_lowers_to_inline_asm() -> Result<(), anyhow::Error> {
     use pliron::builtin::types::{IntegerType, Signedness};
 
     let mut ctx = make_test_ctx();
-    let i32_ty = IntegerType::get(&mut ctx, 32, Signedness::Signless);
-    let i64_ty = IntegerType::get(&mut ctx, 64, Signedness::Signless);
+    let i32_ty = IntegerType::get(&ctx, 32, Signedness::Signless);
+    let i64_ty = IntegerType::get(&ctx, 64, Signedness::Signless);
     // Kernel args: [mask (i32), value (i64), lane/delta (i32)].
     let (module_ptr, entry) =
         build_test_kernel(&mut ctx, vec![i32_ty.into(), i64_ty.into(), i32_ty.into()]);
@@ -1887,7 +1887,7 @@ fn test_cvt_rz_f16x2_f32_lowers_to_inline_asm() -> Result<(), anyhow::Error> {
 
     let mut ctx = make_test_ctx();
     let f32_ty = FP32Type::get(&ctx);
-    let i32_ty = IntegerType::get(&mut ctx, 32, Signedness::Signless);
+    let i32_ty = IntegerType::get(&ctx, 32, Signedness::Signless);
     let (module_ptr, entry) = build_test_kernel(&mut ctx, vec![f32_ty.into(), f32_ty.into()]);
 
     let lo_val = entry.deref(&ctx).get_argument(0);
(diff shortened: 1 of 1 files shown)


SomeB1oody:
Without CI coverage this is likely to recur. Current clippy CI runs `cargo clippy --workspace -- -D warnings`, while the 2 sibling steps in the same file already use `--all-targets`. I suggest adding `--all-targets` for clippy CI. Happy to do it in this PR or a separate PR if you'd like.

nihalpasham:
Thanks for the PR. The fix was correct when filed: passing `&ctx` where the parameter is `&Context` is exactly what `unnecessary_mut_passed` asks for, and the change is behavior-neutral (the same reborrow happens either way).

But `main` has since moved under this branch: PR #406's generated-intrinsics rewrite deleted the tests containing 8 of these 10 hunks, so the branch no longer applies, and the two surviving call sites are covered by #442 — which also fixes the newer `unnecessary_mut_passed` and `unnecessary_lazy_evaluations` sites introduced after #406 that this branch predates. Merging #402 alone would still leave `just clippy` red.

One piece of this PR outlives it: the suggestion to run clippy with `--all-targets` in CI so test-only lint breakage (#401's class) stops recurring. That is worth landing separately.

Closing as already fixed in favor of #442.

PR #403, opened 2026-07-16
examples: barrier_div — convergent-barrier + JumpThreading regression guard

haixuanTao:
## What

A small runnable example guarding a correctness invariant that currently lives only in a comment: `cargo-oxide` disables rustc's `JumpThreading` MIR pass (`-Z mir-enable-passes=-JumpThreading`) because it can duplicate a convergent `bar.sync` across the arms of a divergent branch.

The kernel has two sequential lane-0-only sections (through a `#[inline(never)]` helper, mirroring the shape of a real physics kernel where we first hit this), each followed by `thread::sync_threads()`, with a parallel section in between. With JumpThreading enabled, the first barrier gets duplicated into the `tid == 0` arm → asymmetric per-lane `bar.sync` → CTA deadlock. With the flag (i.e. any normal `cargo oxide run`), it passes and self-checks its output.

## Why an example

If the `build_rustflags` line is ever refactored and the flag dropped, nothing in the test suite would catch it — the failure mode is a hang on real hardware, not a compile error. This example makes the invariant executable:

```
$ cargo oxide run barrier_div
out[0..8]=[0, 10, 20, 30, 40, 50, 60, 70] (expect i*7+i*3=i*10); PASS=true
```

Verified against current `main` (updated for the typed-launch-contract API from #318).

https://claude.ai/code/session_01U2n9RqmxTJb8UG5d1Sjw4W


haixuanTao:
Closing after a second look: the deadlock this guards against only manifests at runtime, and CI here compiles examples but can't execute them (no GPU runners) — so as a regression tripwire it can't actually trip. The repro stays available on my fork (haixuanTao/cuda-oxide, branch test/barrier-div-guard) if GPU CI ever lands or anyone wants to poke at the JumpThreading/convergent-barrier interaction.

PR #405, opened 2026-07-17
fix(mir-lower): trap on assert failure so opt cannot delete bounds checks

### LINKED ISSUE #396: kernel reads out of bounds: checked &[T] indexing (src[j]) no longer emits a bounds check (regressed since v0.2.0)
## Summary
 
A `#[kernel]` that indexes a device slice with a data-dependent index using **safe, checked Rust indexing** (`src[j]`) no longer emits a bounds check. On an out-of-range index the kernel performs an **out-of-bounds global read** (CUDA 700) instead of terminating the thread.
 
I bisected this to the codegen backend (`librustc_codegen_cuda`):
- **v0.2.0 (`faea395`, 2026-06-05):** the bounds check is present; the access is safe.
- **`a9748f4` (2026-06-11):** already gone.
- **current `main` (`29396b7`, 2026-07-04):** still gone.
So the change landed within about a week of the v0.2.0 release.
 
This looks distinct from #132 (`Option::unwrap_or(&literal)` miscompile) and #392 (a `DisjointSlice::len()` codegen crash): this is plain `&[T]` checked indexing losing its bounds check at runtime.
 
Suggested labels: `safety`, `miscompile`, `codegen`.
 
## Kernel
 
```rust
#[kernel]
pub fn gather(src: &[f32], indices: &[u32], mut out: DisjointSlice<f32>) {
    let idx = thread::index_1d();
    let i = idx.get();
    if let Some(out_elem) = out.get_mut(idx) {
        let j = indices[i] as usize;   // checked index load
        *out_elem = src[j];            // checked index; j is runtime data, not statically provable
    }
}
```
 
`j` is loaded from device memory, so the compiler cannot prove `j < src.len()`. Under Rust semantics `src[j]` must bounds-check.
 
## Reproduction
 
Set one index out of range and run the kernel:
 
```rust
let mut idx_host: Vec<u32> = (0..1024).map(|i| (1023 - i) as u32).collect();
idx_host[0] = 1_000_000;   // src has 1024 elements
```
 
```
compute-sanitizer --tool memcheck cargo oxide run --arch sm_89
```
 
## Observed behavior
 
**Current `main` (`29396b7`, 2026-07-04):**
 
```
========= Invalid __global__ read of size 4 bytes
=========     at gather+0x110
=========     by thread (0,0,0) in block (0,0,0)
=========     Access to 0x... is out of bounds
=========     and is 3987713 bytes after the nearest allocation at 0x... of size 4096 bytes
========= ERROR SUMMARY: 6 errors
```
 
(1024 f32 = 4096 bytes; index 1_000_000 x 4 = ~4 MB past the start — matching the reported offset.)
 
**v0.2.0 (`faea395`, 2026-06-05), identical program:**
 
```
Kernel finished without trapping.
out[0] (from OOB index 1_000_000) = 0
========= ERROR SUMMARY: 0 errors
```
 
The v0.2.0 PTX contains the guard: a `setp.ge.u64` on `indices[i] < src.len()` that branches to a clean `exit` when out of range, so the out-of-bounds load never executes. The current-`main` PTX has no such guard and issues the `ld.global` unconditionally.
 
## Expected behavior
 
Out-of-bounds checked indexing should not perform the read. Either the documented panic-to-`trap` behavior, or the v0.2.0 behavior (guard the access and skip it) — not a silent out-of-bounds global read.
 
## PTX difference (gather kernel)
 
- v0.2.0: two bounds checks present (`setp.ge.u64` → `exit` for `indices[i] < indices.len()` and for `indices[i] < src.len()`); generic loads.
- current `main`: only the thread guard remains; no bounds checks; `ld.global` on the indexed access.
## Environment
 
- Backend commits tested: `faea395` (v0.2.0, safe), `a9748f4` (2026-06-11, unsafe), `29396b7` (2026-07-04 `main`, unsafe).
- Frontend: cargo-oxide v0.2.1. rustc nightly-2026-04-03. LLVM 21.1.8. CUDA 12.8. GPU: NVIDIA L4, driver 580.159.03, `sm_89`.
I have the full repro (both kernels, before/after PTX, and both compute-sanitizer logs) and can share the repository link if useful.

SomeB1oody:
## Summary

Fix #396. Checked slice indexing inside kernels (`src[j]`) lost its bounds check. An out-of-bounds Rust index triggers a MIR Assert terminator. `mir-lower` incorrectly translated the assert's failure branch into a block containing only LLVM `unreachable` (`unreachable` means "control flow can never reach this point", not "stop when control flow reaches here"). Based on that semantics, `opt -O2` (SimplifyCFG) folds the entire check into `llvm.assume(j < len)`. The branch disappears and the optimizer treats the condition as always true. So an out-of-bounds index still executes, which is the `Invalid __global__ read` reported by compute-sanitizer.

## Changes

- `crates/mir-lower/src/convert/ops/control_flow.rs`: `convert_assert` now declares `@llvm.trap` and inserts a direct call to it at the head of the abort block, ahead of `llvm.unreachable`.
- `crates/dialect-mir/src/ops/control_flow.rs`: `MirAssertOp` docs updated: assert failure aborts the kernel via trap. The failure path must stay observable or the optimizer deletes the check.
- Unit test `convert_assert_creates_abort_block_with_trap` asserts the abort block is exactly `[call @llvm_trap, llvm.unreachable]` and that the trap intrinsic is declared in the module.
- New example crate `slice_index_bounds_check`: Regression test. A gather kernel indexes with runtime data. The in-range pass must produce correct values. An out-of-range index must trap (or skip the access) without the load reaching memory (`CUDA_ERROR_ILLEGAL_ADDRESS` fails the test).

## Testing

`cargo oxide run slice_index_bounds_check` (sm_120a):

```
=== Checked Slice Indexing Bounds Check Test (issue #396) ===
in_bounds_gather: PASS
out_of_bounds_gather: PASS (kernel trapped: DriverError(719, "unspecified launch failure"))
SUCCESS
```

- PTX inspection: the gather kernel regains both `setp` bounds guard.  The abort path is a PTX `trap`.
- `compute-sanitizer --tool memcheck` on the issue reproducer: 0 memory violation (previously 6 invalid `__global__` reads of size 4).

## Checklist


midiareshadi:
Verified on **sm_89 (NVIDIA L4, driver 580.159.03, CUDA 12.8)** using the original repro from #396 (`indices[0] = 1_000_000` into a 1024-element `src`).

**The fix works.** Backend `5fbf122`, clean rebuild:

PTX (gather kernel):

| | before (`29396b7`) | after (`5fbf122`) |
|---|---|---|
| `trap` | 0 | 2 |
| `setp` | 1 | 3 |
| total instr | 37 | 45 |

Both bounds guards are back, and each abort path carries a `trap`.

Runtime under `compute-sanitizer --tool memcheck`:
- **No `Invalid __global__ read`** — the out-of-bounds read is gone (previously 6 of them).
- The kernel traps as intended: `Trace/breakpoint trap at gather+0x1d0, by thread (0,0,0) in block (0,0,0)`.

One small note on the "0 errors" claim: on sm_89 the sanitizer still prints `ERROR SUMMARY: 6 errors`, but they are not memory violations — 1 is the `Trace/breakpoint trap` itself, and 5 are `CUDA_ERROR_LAUNCH_FAILED (719)` on post-trap cleanup calls (`cuMemcpyDtoHAsync_v2`, `cuModuleUnload`, 3x `cuMemFree_v2`), since the context is poisoned after the trap. So the memory-safety result is clean; the summary count just reflects the cascade. Might differ from sm_120a depending on how the host tears down after the trap.

SomeB1oody:
Thanks for the verification. Your reading of the error summary is exactly right.

I can reproduce the same breakdown on sm_120a:
- 0 `Invalid __global__ read`
- 1 `Trace/breakpoint trap` entry
- a cascade of sticky-error entries on the post-trap cleanup calls (`cuMemcpyDtoHAsync_v2`, `cuMemFree_v2` from Rust `Drop`,  `cuModuleUnload`)
- `ERROR SUMMARY: 10 errors`: the count differing from your 6 only because the two host programs issue a different number of CUDA calls after the trap, not because of the architecture.

That cascade is inherent to trap semantics, not bug. A device-side trap poisons the context and the error is sticky, so every subsequent driver call on that context returns 719.

I will change my PR's claim from "0 error" to "0 memory violations".

nihalpasham:
Thanks for the PR. LGTM.

- `convert_assert` now emits `call void @llvm.trap()` before `unreachable` on the assert-failure edge, so `opt -O2` can no longer delete bounds checks — previously a bare `unreachable` told LLVM the failure edge was impossible and the whole check folded away, letting out-of-bounds indexing silently touch wild memory.
- Kernels are panic=abort, so a device trap (sticky error 719) is the correct device analog of the panic, matching nvcc's device assert.
- The new `slice_index_bounds_check` example pins the behavior end to end. I pushed a trivial maintainer follow-up fixing a test comment typo.

For example:

```text
let j = idx[i] as usize;
*out_elem = src[j];        // j >= src.len() must panic

before: opt deletes the check; OOB reads wild memory
after:  PTX keeps the check; OOB traps, host observes error 719
```

**Reviewer note:** mir-lower tests pass; the example exits 0 with SUCCESS on an RTX 5090 after the deliberate trap; the NVVM-IR path accepts `llvm.trap`; the PTX failure edge contains `trap`. Kernels that previously "passed" only because a failing assert was deleted will now visibly trap — correct, and may surface latent bugs. Merge together with companion #408, which closes the same bug class at the mir-importer site. All hosted CI checks passed on the final head.

This closes #396.

PR #408, opened 2026-07-17
fix(mir-importer): trap on dropped diverging calls so opt cannot delete panic paths

### LINKED ISSUE #407: `opt -O2` deletes the panic path and produces silent wrong results
**Description**

When a kernel hits a diverging panic call (`Option::unwrap` on a runtime `None`, explicit `panic!`, `unwrap_failed`, etc.), `mir-importer` drops the call entirely and emits a bare `mir.unreachable` (`crates/mir-importer/src/translator/terminator/mod.rs`, line 1212–1219). 

The comment there says: "*we emit an unreachable terminator since panics can't actually execute*". LLVM's  `unreachable` tells the middle end that reaching the panic is UB, so the `opt -O2` stage  folds the `None` branch away, not preserving it. 

This is the same problem as #396 (fixed by #405), but at a different lowering site. #405 fixed `TerminatorKind::Assert` in mir-lower's `convert_assert`. Diverging `TerminatorKind::Call` still produces the deletable pattern. 

**Minimal reproducer**

```rust
use cuda_host::cuda_module;

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    pub fn unwrap_get(src: &[f32], j: usize, mut out: DisjointSlice<f32>) {
        if let Some(out_elem) = out.get_mut(thread::index_1d()) {
            *out_elem = src.get(j).copied().unwrap(); // None when j >= src.len() -> must panic
        }
    }
}

fn main() {
    let ctx = CudaContext::new(0).expect("context");
    let module = kernels::load(&ctx).expect("load module");

    let src = DeviceBuffer::from_host(&stream, &[1.0f32, 2.0, 3.0, 4.0]).expect("src");
    let mut out = DeviceBuffer::from_host(&stream, &[-1.0f32]).expect("out");

    // SAFETY: one thread, one output element
    let result = unsafe {
        module.unwrap_get(
            &stream,
            LaunchConfig::for_num_elems(1),
            &src,
            1_000_000usize, // src.get(1_000_000) is None -> unwrap() must panic
            &mut out,
        )
    }
    .and_then(|()| out.to_host_vec(&stream));

    match result {
        Err(e) => println!("PASS (kernel trapped: {})", e),
        Ok(out) => {
            println!("FAIL (unwrap(None) did not panic, out[0] = {})", out[0]);
            std::process::exit(1);
        }
    }
}
```

**Expected behavior**

`src.get(1_000_000)` returns `None` and `.unwrap()` panics. According to `rustc-codegen-cuda` README (*"If a panic condition is hit at runtime, the GPU traps"*), the kernel must trap and the host must see the launch fail. At minimum, the `None` arm must never be compiled into code that produces a value.

**Actual behavior**

The launch **succeeds** and `out[0]` is overwritten with garbage (`0.0` here, whatever an uninitialized register holds). No trap, no error code, and compute-sanitizer is completely silent:

```
FAIL (unwrap(None) did not panic, out[0] = 0)
```

```
========= COMPUTE-SANITIZER
FAIL (unwrap(None) did not panic, out[0] = 0)
========= Target application returned an error
========= ERROR SUMMARY: 0 errors
```

**Environment**

- cuda-oxide: `main`@29396b7f
- sm_120a
- CUDA driver version: 595.71.05 (CUDA toolkit 13.3, nvcc V13.3.73)
- `rustc --version --verbose` (repo-pinned `nightly-2026-04-03`):
  ```
  binary: rustc
  commit-date: 2026-04-02
  release: 1.96.0-nightly
  LLVM version: 22.1.2
  ```
- `llc-21 --version`: Ubuntu LLVM version 21.1.8 (`opt-21` from the same toolchain)

SomeB1oody:
## Summary

Fixes #407. When a kernel hits a diverging panic call (`Option::unwrap` on a runtime `None`, `panic!`, `unwrap_failed`, etc.), `mir-importer` drops the call (the panic machinery cannot run on the GPU) and terminated the block with only a bare `mir.unreachable`. That marks the runtime-reachable panic path as undefined behavior, so the `opt -O2` stage folds it into the surrounding code: `unwrap(None)` compiled into storing an uninitialized register into the output (silent wrong data, invisible to compute-sanitizer). This PR emits `nvvm.trap` ahead of the `mir.unreachable` at that site, so the panic path stays observable through `opt` and the GPU traps as the documented panic contract requires.

## Changes

- `crates/mir-importer/src/translator/terminator/mod.rs`: the diverging-call site now terminates the block using `emit_trap_unreachable_after` (`nvvm.trap` + `mir.unreachable`).
- Add doc comments on `emit_unreachable_after`: it is only for paths that can never execute (rustc-proven unreachability) or code after a call that is emitted into the module and genuinely never returns. Those sites are deliberately unchanged, so proven-unreachable paths still fold as before.
- New example crate `option_unwrap_trap`: one thread runs `src.get(j).copied().unwrap()` with an out-of-range `j`. The launch must fail with the kernel trapping.

## Testing

sm_120a, CUDA 13.3, driver 595.71.05:

Before the fix, the example's launch succeeds with garbage output:

```
FAIL (unwrap(None) did not panic, out[0] = 0)
```

After the fix:

```
PASS (kernel trapped: DriverError(719, "unspecified launch failure"))
```

- `cargo oxide run option_unwrap_trap` passes (also with `CUDA_OXIDE_NO_OPT=1`).
- Proven-unreachable paths are unaffected.
- `compute-sanitizer --tool memcheck` reports no memory violations on the example.
- `scripts/smoketest.sh`: 134/134 PASS, including the new example.
- `cargo oxide fmt --check` clean; `cargo clippy -p mir-importer --all-targets -- -D warnings` clean.

## Checklist


SomeB1oody:
I pushed a second commit extending the fix to the `assert_inhabited` intrinsic handler.

`assert_inhabited::<T>` with an uninhabited `T` is not a proven-unreachable marker. The commit lowers the uninhabited case through the same `emit_trap_unreachable_after` helper. It also adds the `assume_init_uninhabited_trap` regression example. 

before the fix it failed with the launch succeeding. After, the launch aborts with the kernel trapping.

midiareshadi:
Verified #408 on **sm_89 (NVIDIA L4, driver 580.159.03, CUDA 12.8)** with your `unwrap_get` reproducer (`src.get(1_000_000).unwrap()`).

**Before** (`29396b7`): reproduces the silent failure exactly — the launch succeeds and `out[0]` gets an uninitialized register. The program prints `FAIL (unwrap(None) did not panic, out[0] = 0)` and compute-sanitizer reports `ERROR SUMMARY: 0 errors` — completely silent, as you noted.

**After** (`6a58392`, both commits): the panic path is preserved and the kernel traps.

PTX (panic path): `trap` 0 → 1, total 38 → 44 — the `None` arm is no longer folded away.

Runtime: the program prints `PASS (kernel trapped: DriverError(719, "unspecified launch failure"))`, and the sanitizer shows a `Trace/breakpoint trap`. No `Invalid` reads; the remaining `ERROR SUMMARY: 5 errors` are all the post-trap 719 cascade on cleanup (`cuMemcpyDtoHAsync_v2`, 2x `cuMemFree_v2`, `cuModuleUnload`), consistent with the sticky-context behavior from #405 — no memory violations.

So on sm_89 the fix behaves exactly as on sm_120a: silent wrong data becomes a clean trap.

nihalpasham:
Thanks for the PR. LGTM.

- The mir-importer now emits a side-effecting trap instead of bare `unreachable` when a diverging call is dropped (`unwrap`/`panic!` shims, and `assert_inhabited` for uninhabited types in the second commit), so `opt` cannot fold panic paths away. This closes the importer half of the bug class #405 fixes in mir-lower.
- The `option_unwrap_trap` and `assume_init_uninhabited_trap` examples pin both shapes with host-verified failed launches.
- I pushed maintainer follow-ups: a code comment recording the call-drop semantic (a resolvable `-> !` device callee is dropped and trapped rather than emitted — pre-existing, now safe), LICENSE files matching the example convention, and the rebase + ABI-marker fix described in the note below.

For example:

```text
let v = src.get(j).copied().unwrap();   // None when j >= len

before: opt deletes the panic path; execution continues into UB
after:  PTX contains trap; the launch fails and the host observes it
```

**Reviewer note:**  
- since #406, `nvvm.trap` is a generated intrinsic and module verification rejects it without its exact ABI marker — as submitted, every trap-reaching example (including pre-existing `function_item_call`, `slice_reslice`, `step_by`) failed to compile against current main. 
- A maintainer commit constructs the trap through the typed `TrapOp::build` and attaches marker `v1:i0295`, exactly as the generated `debug::trap` dispatch does. 
- A second follow-up scoped `#[allow(invalid_value, clippy::uninit_assumed_init)]` to the one deliberately-invalid statement in `assume_init_uninhabited_trap` — the construct is the example's entire point, and CI's per-example clippy gate rightly flags it in ordinary code. 
- After the fixes: mir-importer 803 tests pass, all five trap-path examples PASS on an RTX 5090, and both new example dirs pass `cargo clippy --all-targets -- -D warnings`. Merge together with companion #405. All hosted CI checks passed on the final head.

This closes #407.

PR #409, opened 2026-07-18
feat(mir): add Drop glue translation for device code

honeyspoon:
## Summary

Enable types with effectful `Drop` implementations on device by emitting actual `drop_in_place::<T>` calls when the no-op proof fails. Preserves the existing no-op fast path as an optimization.

## Motivation

`TerminatorKind::Drop` is currently a support gap (tracked as `error_drop_glue` in `STATUS.md`). Destructors that do observable work (logging, releasing resources, resetting state) fail compilation. This PR closes that gap by falling back to real drop glue when the no-op proof fails.

## Implementation

- **mir-importer** (`drop_glue.rs`): add `emit_drop_glue()` and `compute_drop_place_address()` -- when `drop_glue_is_noop()` fails, emit a `MirCallOp` to the monomorphized `drop_in_place::<T>`
- **mir-importer** (`terminator/mod.rs`): change `translate_drop()` from `if noop → branch; else → error` to `if noop → branch; else → emit_drop_glue()`
- **rustc-codegen-cuda** (`collector.rs`): discover and collect `drop_in_place::<T>` functions via `TerminatorKind::Drop`, allow `InstanceKind::DropGlue` through the instance filter
- Convert `error_drop_glue` from expected-failure to positive test (`drop_glue` example with end-to-end verification)
- Update `STATUS.md` and `smoketest.sh` to remove `error_drop_glue` from error examples

## Testing

```
cargo test -p mir-importer         # 805 passed
cargo fmt --all -- --check         # clean
cargo clippy -p mir-importer -p cuda-oxide-codegen --tests -- -D warnings  # clean
scripts/smoketest.sh               # 133/141 passed (8 failures are sm_90+ features)
```


nihalpasham:
Thanks for the PR. 

The makes sense for device RAII, device-side calls to the monomorphized `drop_in_place::<T>` shim, discovered by the collector like any other device function, mirroring how rustc's own mono-item collector handles drop glue. 

- As submitted, though, the widened type-level "no-op" fallback silently skipped exactly the destructors the PR enables — including this PR's own `drop_glue` example, which failed on GPU at the submitted head (all 256 outputs 0x00000000, expected 0xDEADBEEF). A silent miscompile replacing main's loud error.
- I pushed a maintainer commit removing the fallback in all three coupled places and unifying them onto one shared predicate (`drop_instance_is_noop`, body-level proof only) consulted by the importer's emit decision, the collector, and device_codegen's filter; it also reverts the Assert-edge widening in `body_is_noop` and makes the collector skip provably-no-op shims at discovery, so dead `Drop::drop` bodies are never collected.
- A second commit scopes STATUS.md to actual support. Net: destructors are never silently skipped — provably no-op drops lower to a branch, everything else emits the real call or fails loudly.

For example:

```text
struct DropMarker { target: *mut u32 }
impl Drop for DropMarker { fn drop(&mut self) { unsafe { *self.target = 0xDEADBEEF; } } }

// 256 threads; each writes 0 to its own slot, then builds a
// DropMarker aimed at it. The destructor fires at end of scope:
#[kernel] fn drop_glue_kernel(mut out: DisjointSlice<u32>) {
    let slot = ...;              // this thread's output slot
    *slot = 0;
    let _m = DropMarker { target: slot };
}   // <- _m drops here; slot must become 0xDEADBEEF

main:        loud compile error
PR as filed: compiles, destructor silently skipped - all 256 slots read 0
repaired:    compiles, destructor runs - all 256 slots read 0xDEADBEEF
```

**Reviewer note:** mir-importer 803 and backend 17 tests pass; `drop_glue` PASSES on an RTX 5090 with the sentinel host-verified; `array_for_loop` passes (IntoIter shims prove no-op at collection time); drop/array compile-only subset 7/7; ltoir check passes. Land before #398; the drop-glue predicate reconciliation for #398's `array::Drain` reduction is spelled out in the review record. All hosted CI checks passed on the final head.

#437 is related but stays open.

PR #410, opened 2026-07-18
feat: link libdevice.10.bc at IR level via llvm-link on the PTX path

honeyspoon:
## Summary

When a kernel uses both f16 types (MMA/WMMA) and libdevice calls (`__nv_*` math functions), the pipeline auto-forces NVVM IR mode. On pre-Blackwell targets (sm<100), this selects the legacy LLVM 7 dialect which rejects f16 -- making f16 and libdevice mutually exclusive. Fix: link `libdevice.10.bc` at the LLVM IR level via `llvm-link` before `opt` and `llc`, keeping everything on the PTX path which has no f16 restrictions.

## Motivation

The PTX path (`llc`) handles f16 natively. The NVVM IR path uses `legalize_for_nvvm()` which rewrites the module to LLVM 7 syntax and rejects f16 on pre-Blackwell. Libdevice calls (`__nv_sinf`, `__nv_sqrtf`, etc.) currently force the NVVM IR path because libNVVM resolves them -- but this creates an impossible constraint for kernels that also use f16.

The fix matches what `build_and_run.sh` already does (line 91): `llvm-link kernel.ll libdevice.10.bc -o linked.bc`, then `opt -O2`, then `llc`. The `__nv_*` calls get resolved and inlined during `opt`, producing self-contained PTX.

## Implementation

- **cuda-oxide-codegen** (`llvm_tools.rs`): add `llvm_link: Option<OptTool>` to `LlvmToolchain`, resolved with the same co-located/sysroot/PATH discovery as `opt`. Add generic `resolve_sibling_tool()` and `sibling_tool_candidates()` helpers.
- **cuda-oxide-codegen** (`ptx.rs`): add `link_libdevice()` function that runs `llvm-link -S`. Insert linking step before `opt -O2` in `generate_ptx_impl()`. Pass `libdevice_path: Option<&Path>` through `generate_ptx()` and `generate_ptx_with_toolchain()`.
- **cuda-oxide-codegen** (`pipeline.rs`): modify `should_emit_nvvm_ir()` so `needs_libdevice` only forces NVVM IR when `libdevice.10.bc` is not found. Discover `libdevice.10.bc` early via `libnvvm_sys::find_libdevice()` so the backend decision accounts for IR-level linking availability.
- **cuda-oxide-codegen** (`api.rs`): resolve `llvm-link` as sibling tool in `Toolchain::from_paths()`.

Pipeline with libdevice: `.ll → llvm-link libdevice.10.bc → opt -O2 → llc → .ptx`

Fallback (no `libdevice.10.bc` or no `llvm-link`): existing NVVM IR path unchanged.

## Testing

```
cargo test -p cuda-oxide-codegen   # 120 passed (107 unit + 12 integration + 1 doc)
cargo test -p mir-importer         # 805 passed
cargo fmt --all -- --check         # clean
cargo clippy -p cuda-oxide-codegen --tests -- -D warnings  # clean
scripts/smoketest.sh               # 133/141 passed (8 failures are sm_90+ features)
```

## Related

- Complementary to #390/#391 (reducing unnecessary libdevice deps for scalar float ops) -- this PR handles the case where libdevice IS needed but shouldn't force NVVM IR
- `build_and_run.sh:91` -- reference implementation of the `llvm-link` pipeline
- `crates/libnvvm-sys/src/lib.rs:780` -- `find_libdevice()` discovery

### DIFF
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -393,9 +393,11 @@ pub struct Toolchain {
 impl Toolchain {
     /// Discover LLVM 21+ tools from the Rust sysroot and `PATH`.
     ///
-    /// Discovery is explicit and does not read cuda-oxide environment knobs.
-    /// It may return a toolchain without `opt`; that toolchain can compile only
-    /// with [`Optimization::None`].
+    /// Discovery is explicit and does not read cuda-oxide environment knobs,
+    /// with one exception: `llvm-link` (used for IR-level libdevice linking)
+    /// honors `CUDA_OXIDE_LLVM_LINK` and otherwise must share the chosen
+    /// `llc`'s LLVM major. It may return a toolchain without `opt`; that
+    /// toolchain can compile only with [`Optimization::None`].
     pub fn discover() -> Result<Self, CompileError> {
         let opts = BackendOptions::default();
         let inner = LlvmToolchain::resolve(&opts).ok_or_else(|| CompileError::Toolchain {
@@ -413,6 +415,12 @@ impl Toolchain {
     }
 
     /// Use explicit LLVM tools, verifying that they run and share one major.
+    ///
+    /// `llvm-link` is not an explicit parameter: it is taken from
+    /// `CUDA_OXIDE_LLVM_LINK` when set, otherwise discovered next to `llc`,
+    /// in the Rust sysroot, or on `PATH`, requiring the same LLVM major as
+    /// `llc`. It stays unset (disabling IR-level libdevice linking) when
+    /// nothing resolves.
     pub fn from_paths(llc: impl Into<PathBuf>, opt: Option<PathBuf>) -> Result<Self, CompileError> {
         let llc = llc.into();
         let llc_text = llc.to_string_lossy().into_owned();
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -4,7 +4,7 @@
  */
 
 use crate::error::PipelineError;
-use crate::llvm_tools::{LlvmToolchain, OptTool, probe_runnable};
+use crate::llvm_tools::{LlvmToolchain, OptTool, probe_runnable, resolve_sibling_tool};
 use crate::options::BackendOptions;
 use crate::pipeline::{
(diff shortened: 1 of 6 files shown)


nihalpasham:
Thanks for the PR. LGTM 

- I pushed maintainer commits completing it:
  - the `llvm-link` invocation now passes `--internalize --only-needed` (clang's LinkOnlyNeeded | InternalizeLinkedSymbols); without them every libdevice body has external linkage, survives GlobalDCE, and a one-call kernel's PTX ballooned from ~132 to ~21,800 lines with 349 exported `__nv_*` bodies. A regression test links a real kernel and asserts no unreferenced `__nv_*` definition survives;
  - the PTX-vs-NVVM decision now requires a resolvable same-major `llvm-link` at decision time (not just a libdevice file on disk) and falls back to the NVVM-IR path exactly as before the PR; once the PTX path is chosen, a link failure is a hard error instead of warn-and-continue emitting PTX with unresolved `.extern .func __nv_*`;
  - plus cleanups: `linked.ll`/`linked.opt.ll` in stale-artifact cleanup, toolchain doc contracts reconciled with `CUDA_OXIDE_LLVM_LINK`, sibling-tool discovery deduplicated, and the regression detector hardened (value-returning `.visible .func (.param ...) __nv_*` forms and identifier-boundary reference matching — proven against the 349-definition bloat).

For example:

```text
#[kernel] fn k(out: &mut [f32], h: f16, x: f32) {
    out[0] = f32::from(h) + x.sin();     // f16 AND __nv_sinf together
}

before: __nv_sinf forces the NVVM-IR path; legacy dialect rejects f16
after:  llvm-link --internalize --only-needed -> 527-line self-contained PTX via llc
```

**Reviewer note:** cuda-oxide-codegen tests pass; `libdevice_math` and `math_atan` PASS on an RTX 5090 via the new path with zero `__nv_*` residue in PTX; the fallback path (llvm-link unresolvable) still runs via NVVM-IR; an sm_90 build assembles under `ptxas -arch=sm_90`; fuzzer seed 162 passes with the new PTX artifact kind. Land after #415 (both rework cuda-oxide-codegen; the reconciliation is spelled out in the review record). Related: #414 is closed by #395; #390/#380/#386 stay open. All hosted CI checks passed on the final head.

PR #411, opened 2026-07-18
feat(cuda-macros): add multi-output support for ptx_asm! macro

honeyspoon:
## Summary

Extend `ptx_asm!` to support up to 8 output operands (previously limited to 1), enabling MMA intrinsics, vectorized loads, and other multi-register PTX directly from user code without waiting for per-intrinsic compiler support.

## Motivation

Many PTX instructions produce multiple register results (MMA, ldmatrix, vectorized loads). Currently users cannot express these via `ptx_asm!` because only a single output operand is allowed. This forces workarounds (pointer-based output, splitting into multiple asm blocks) or waiting for each instruction to get dedicated compiler support.

```rust
// With this PR:
let (d0, d1, d2, d3): (f32, f32, f32, f32);
ptx_asm!(
    "mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32
     {%0, %1, %2, %3}, {%4, %5, %6, %7}, {%8, %9}, {%10, %11, %12, %13};",
    out("=f") d0, out("=f") d1, out("=f") d2, out("=f") d3,
    in("r") a0, in("r") a1, in("r") a2, in("r") a3,
    in("r") b0, in("r") b1,
    in("f") c0, in("f") c1, in("f") c2, in("f") c3,
);
```

## Implementation

- **cuda-macros** (`ptx_asm.rs`): parse multiple `out(...)` operands, generate tuple return type for multi-output, maintain backward compatibility for single-output
- **dialect-nvvm** (`ops/asm.rs`): remove the max-1-result verifier check on `InlinePtxOp`
- **mir-importer** (`intrinsics/asm.rs`): add `count_output_constraints()` to parse constraint strings, decompose tuple return types into per-element results
- **mir-lower** (`intrinsics/asm.rs`): build LLVM struct return type for multi-output asm, emit `extractvalue` for each result element
- Trybuild pass tests for 2, 4, 8 outputs and mixed types
- Compile-fail test for >8 outputs

## Testing

```
cargo test -p cuda-macros -p mir-importer -p mir-lower  # 1026 passed
cargo fmt --all -- --check                               # clean
cargo clippy -p cuda-macros -p dialect-nvvm -p mir-importer --tests -- -D warnings  # clean
```


nihalpasham:
Thanks for the PR. LGTM.

- `ptx_asm!` now supports up to eight output operands, threaded correctly through all four layers: the proc macro destructures a tuple return, `nvvm.inline_ptx` carries N results, the importer decomposes the tuple destination and repacks with `MirConstructTupleOp`, and mir-lower emits a struct-returning LLVM inline asm plus one extractvalue per output. This unblocks mma.sync-, ldmatrix-, and vectorized-load-style PTX without waiting for per-instruction intrinsics.
- I pushed maintainer follow-ups: the three stale doc sites that still said "zero or one out" now state the real contract; the inline_ptx example gained a host-verified multi-output kernel; a mir-lower lowering test pins the struct-return conversion; an `InlinePtxOp` verifier enforces result count == `=`-constraint count; and the counting logic is deduplicated onto `InlinePtxOp::count_output_constraints`, used by both the verifier and the importer.

For example:

```text
// one PTX block, two outputs (%0 and %1):
let sum: u32;
let prod: u32;
unsafe {
    ptx_asm!(
        "add.u32 %0, %2, %3; mul.lo.u32 %1, %2, %3;",
        out("=r") sum, out("=r") prod,
        in("r") x, in("r") y,
        options(register_only),
    );
}

Before: compile error - ptx_asm! allowed at most one out()
After:  x = 6, y = 7  ->  sum = 13, prod = 42, host-verified on GPU
        (unequal results prove the two outputs are not swapped)
```

**Reviewer note:** dialect-nvvm, mir-importer, and mir-lower suites pass; `cargo oxide run inline_ptx` PASSES on an RTX 5090 including the multi-output kernel; the emitted `.ll` and PTX were inspected for correct constraint strings and register binding. All hosted CI checks passed on the final head.

nihalpasham:
Superseded: #411 merged at the stale-clone head before this could be posted. The maintainer follow-ups now live in PR #443; 

A force-push earlier today replaced the review follow-ups that were on this branch (verifier, multi-output E2E test, doc updates); they are restored in #443.

PR #412, opened 2026-07-18
feat(llvm-export): support i8/u8/i16/u16/f16/bool in device extern signatures

honeyspoon:
## Summary

Remove the hard errors that reject small types (`i8`, `u8`, `i16`, `u16`, `f16`, `bool`) in device extern function signatures. Emit proper LLVM `signext`/`zeroext` ABI attributes for sub-32-bit integer types, enabling clean FFI boundaries for INT8 and FP16 kernels without pointer indirection.

## Motivation

Device extern functions currently reject any parameter or return type smaller than 32 bits. This forces users to widen all arguments to `i32`/`u32` on both sides of the FFI boundary, adding unnecessary casting and losing type safety. INT8 quantized inference and FP16 kernels are common use cases that need clean small-type signatures.

```llvm
; Before: rejected with error
; After:
declare void @my_kernel(i32 signext, i32 zeroext, half)
```

## Implementation

- **rustc-codegen-cuda** (`device_codegen.rs`): replace rejection errors with proper type mapping -- signed types (`i8`, `i16`) map to `SignExtInteger(32)`, unsigned types (`u8`, `u16`) map to `ZeroExtInteger(32)`, `f16` maps to `Float16`, `bool` maps to `ZeroExtInteger(32)`
- **llvm-export** (`externs.rs`): add `SignExtInteger(u32)` and `ZeroExtInteger(u32)` variants to `DeviceExternType`, add `write_llvm_with_attr()` for attribute emission
- **llvm-export** (`module.rs`, `ops.rs`): emit `signext`/`zeroext` attributes on both declarations and call sites
- **cuda-oxide-codegen** (`lower.rs`): handle new variants in `device_extern_type_to_pliron()`

## Testing

```
cargo test -p llvm-export -p cuda-oxide-codegen -p mir-importer  # 1002 passed
cargo fmt --all -- --check                                        # clean
cargo clippy -p llvm-export -p cuda-oxide-codegen --tests -- -D warnings  # clean
```


nihalpasham:
Thanks for the PR. This is the right direction, the old rejection message itself invited this ("rejected until their signext or zeroext attributes can be emitted"). The ABI shape needed a redo, which I pushed as maintainer commits on top.

- Widening the declared IR type to i32 mismatched both the clang/nvcc-compiled definitions being linked against and cuda-oxide's own narrow i8/i1 SSA values (no promotion was inserted, so by-value integer/bool cases still failed, just later); and the return-position attribute emission produced IR llvm-as cannot parse (`declare i8 signext @f()`).
- The redo keeps the NARROW type with the attribute — the clang NVPTXABIInfo / rustc nvptx64 shape: variants store the original width, `declare signext i8 @f(i8 signext)` with the attribute before the type in return position, zero inserted conversions. The legacy libNVVM path cleanly rejects sub-32-bit by-value externs, mirroring its existing f16 rejection.
- Tests now assert the canonical strings plus an llvm-as parse gate, and device_ffi_test gained eight small-scalar extern functions (nvcc -dc -dlto LTOIR) with GPU-verified round-trips.

For example:

```text
// CUDA C++, linked as LTOIR:  __device__ int8_t scale(int8_t x);
#[device] extern "C" { fn scale(x: i8) -> i8; }

before PR:   error: i8 rejected in device extern signatures
PR as filed: declares i32 -> mismatched definition, unparseable return attrs
after redo:  declare signext i8 @scale(i8 signext) -> links and runs
```

**Reviewer note:** llvm-export, mir-lower, and backend tests pass; the emitted module parses under llvm-as-21; `device_ffi_test --emit-nvvm-ir` passes 5/5 on an RTX 5090 including negative-i8 sign extension, u16 zero extension, bool, and f16 round-trips; compile-only smoketest passes. All hosted CI checks passed on the final head.

PR #415, opened 2026-07-20
codegen: tighten the standalone compiler's clone, liveness and diagnostic paths

alejandro-soto-franco:
One of three parts of #357, split so each can be reviewed on its own. #357 bundled eight unrelated fixes across 12 files, which put the small ones behind a dependency discussion. The three parts are independent: each applies cleanly to `main` alone, and they merge cleanly in any order. #357 will be closed once all three are open.

Sibling parts: #416 (target selection provenance), #417 (scratch directory on tempfile).

## Summary

Follow-up to #314's `cuda-oxide-codegen` extraction, covering the review findings that stand on their own: a leak on the panic path, an unavoidable module clone, a duplicated liveness check, and diagnostics that the standalone API could neither request nor receive.

## Changes

* `EraseGuard` owns the erase of the cloned module, so a panic inside `compile_clone` no longer leaks it into the shared `Context` arena.
* `Compiler::compile_owned` consumes the module and runs the pipeline on the caller's IR, making the clone opt-in.
* `validate_live_module_op` replaces the liveness check duplicated between `mark_kernel_entry` and the compile entry points.
* `CompileOptions::with_verbose`/`verbose`, so the standalone API can ask for the diagnostics the pipeline has always produced behind `BackendOptions::verbose`.
* `generate_ptx_with_toolchain` returns `GeneratedPtx`, matching the discovery path, and `compile_clone` forwards the diagnostics it receives.
* `Toolchain::discover` and `Toolchain::from_paths` both record which `llc` and `opt` they selected.
* The double LLVM-IR render on the libdevice path is documented in place.

## Rationale

Two paths dropped diagnostics even when asked for them: `generate_ptx_with_toolchain` discarded everything but the resolved target string, and `compile_clone` never copied the pipeline's diagnostics into the `Compilation`. Reporting nothing about toolchain selection left a caller unable to tell which of several installed LLVM versions produced an artifact.

`Operation::erase` still runs from `Drop`, so a panic inside the erase during an unwind aborts. That is left alone and documented: an erase fails only once the arena is already inconsistent, and swallowing the second panic would trade a loud abort for a silent leak.

The double render is documented rather than removed. The preview renders pre-legalisation and the export renders post-legalisation through a different export config, so the two texts describe different module states and caching one for the other would be unsound. Removing the second render needs feature detection that works off the IR rather than off rendered text.

## Validation

* `cargo test -p cuda-oxide-codegen` (123 passed)
* `cargo test -p mir-importer` (801 passed)

### DIFF
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -107,6 +107,7 @@ pub struct CompileOptions {
     optimization: Optimization,
     fma_contraction: bool,
     debug_info: DebugInfo,
+    verbose: bool,
 }
 
 impl CompileOptions {
@@ -117,6 +118,7 @@ impl CompileOptions {
             optimization: Optimization::O2,
             fma_contraction: true,
             debug_info: DebugInfo::None,
+            verbose: false,
         }
     }
 
@@ -140,6 +142,17 @@ impl CompileOptions {
         self
     }
 
+    /// Request progress and tool-selection diagnostics.
+    ///
+    /// Without this, [`Compilation::diagnostics`] still reports the
+    /// toolchain's own selection diagnostics and a final success note, but
+    /// omits per-compilation detail such as which target-selection source won
+    /// or why `opt -O2` was skipped.
+    pub fn with_verbose(mut self, verbose: bool) -> Self {
+        self.verbose = verbose;
+        self
+    }
+
     /// Requested CUDA target.
     pub fn target(&self) -> &Target {
         &self.target
@@ -158,6 +171,11 @@ impl CompileOptions {
     /// Requested device debug-information tier.
     pub fn debug_info(&self) -> DebugInfo {
         self.debug_info
+    }
+
+    /// Whether progress and tool-selection diagnostics were requested.
+    pub fn verbose(&self) -> bool {
+        self.verbose
     }
 }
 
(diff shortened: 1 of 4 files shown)


nihalpasham:
Thanks for the PR. LGTM as submitted. 

- The RAII `EraseGuard` fixes is good: a panic in the destructive pipeline left the cloned module in the shared pliron Context arena; it now erases on every exit path, unit-tested.
- Explicit-toolchain callers get their diagnostics back: `CompileOptions::with_verbose` plus the `GeneratedPtx` return plumb through which llc/opt produced the artifact and why opt was skipped, previously discarded.
- `compile_owned` consumes the module and skips the defensive clone; the module-liveness check is deduplicated into `validate_live_module_op`.

For example:

```text
let compiler = Compiler::new(Toolchain::from_paths(llc, Some(opt))?);
// a panic mid-pipeline no longer leaks the cloned module;
// compilation.diagnostics() reports the exact toolchain selection
```

**Reviewer note:** all 123 cuda-oxide-codegen tests pass with LLVM 21/22 on PATH (the discover-diagnostic test genuinely executed rather than skipping); crate clippy and fmt are clean. Merge-order caution: siblings #416/#417 touch the same files and will need rebasing after this lands — #416 also rewords the target-selection diagnostic this PR's test greps for. Land this before #410. All hosted CI checks passed on the final head.

PR #416, opened 2026-07-20
codegen: attribute target selection to whoever chose the target

alejandro-soto-franco:
One of three parts of #357, split so each can be reviewed on its own. #357 bundled eight unrelated fixes across 12 files, which put the small ones behind a dependency discussion. The three parts are independent: each applies cleanly to `main` alone, and they merge cleanly in any order. #357 will be closed once all three are open.

Sibling parts: #415 (clone, liveness and diagnostic paths), #417 (scratch directory on tempfile).

## Summary

A rejected target produced an error naming `CUDA_OXIDE_TARGET`, and classified the failure as a codegen stage failure. Both are wrong for every caller that reaches target selection by some other route.

## Changes

* `BackendOptions` carries `target_arch_source` alongside `target_arch`, and `resolve_ptx_target` reports the label it was given.
* `PipelineConfig` carries the same label, defaulting to `PipelineConfig::target_arch` and set to `CUDA_OXIDE_TARGET` by `rustc-codegen-cuda`, which reads that variable itself.
* The environment and config merge moves into `backend_options_for`, so the precedence is testable without running a compilation.
* Target and feature-floor rejections travel as `PipelineError::TargetSelection` and land in `CompilationStage::Input`.
* `CompileError` gains a matching `TargetSelection` variant.
* New `tests/target_provenance.rs`, plus three `mir-importer` tests over the label precedence.

## Rationale

`resolve_ptx_target` hardcoded `"CUDA_OXIDE_TARGET"` as the provenance of any explicit override, because the env-driven rustc pipeline was once the only caller. The standalone experimental API passes a `Target` the caller built in Rust and never touches the environment, so a caller who requested `sm_75` for a module needing `sm_80` was told to go and look at an environment variable they had never set.

The same mis-attribution existed one layer up: `PipelineConfig::target_arch` replaced `target_arch` and left `from_env`'s label behind, so a caller setting the config field directly was still blamed on the environment. That is fixed here too, which is the only in-tree site that needed the new mechanism.

Target and feature-floor rejections are decided before `llc` is reached and describe the caller's request, so `Codegen` was the wrong stage for them.

`CompileError::TargetSelection` is a new variant rather than a reuse of `InvalidTarget`. `InvalidTarget` renders as ``invalid CUDA target `{target}`: {reason}``, which suits a parse failure from `Target::parse` and misdescribes a feature-floor rejection: `sm_75` is a valid architecture, and the reason text already names it, so routing feature-floor failures through `InvalidTarget` called a valid target invalid and printed the architecture twice. `TargetSelection` renders its reason alone, which is already phrased for a user. `InvalidTarget` keeps its parse-failure meaning and its rustdoc now points at the new variant.

`target_arch_source` stays `&'static str`. Every label in the tree is a fixed string. Widening it to `Cow<'static, str>`, which would let a caller build a label carrying the target text itself, can wait until a caller needs one.

## Validation

* `cargo test -p cuda-oxide-codegen` (123 passed)
* `cargo test -p mir-importer` (804 passed)
* `(cd crates/rustc-codegen-cuda && cargo check)`, since this adds a field to a `PipelineConfig` literal in `device_codegen.rs` and that crate is excluded from the workspace

### DIFF
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -532,6 +532,7 @@ impl Compiler {
         let ptx_path = scratch.path().join("module.ptx");
         let backend_options = BackendOptions {
             target_arch: Some(options.target.sm()),
+            target_arch_source: "the requested Target",
             device_arch_hint: None,
             no_opt: options.optimization == Optimization::None,
             no_fma: !options.fma_contraction,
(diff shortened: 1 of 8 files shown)


nihalpasham:
LGTM, the `Input-stage` classification and the `InvalidTarget`/`TargetSelection` split match how nvrtc separates `INVALID_OPTION` from `COMPILATION`.

I pushed one follow-up: the provenance label was threaded through but only printed on the verbose success path, so env and `--arch` users lost the pointer to the knob that chose the target. All three rejection reasons now carry it, e.g. `invalid CUDA target `sm_9x`: compute capability must contain at least two digits (target from CUDA_OXIDE_TARGET)`, and the parse error is no longer double-wrapped in a second "invalid CUDA target" prefix. Unit test pins both. 

One follow-up to track separately: the NVVM-IR path's resolver still reports target rejections as `Export-stage`. Merging.

nihalpasham (approved):
Validated; see merge comment.

PR #417, opened 2026-07-20
codegen: back the standalone compiler's scratch directory with tempfile

alejandro-soto-franco:
One of three parts of #357, split so each can be reviewed on its own. This part adds a dependency, which is the discussion the other two were sitting behind. The three parts are independent: each applies cleanly to `main` alone, and they merge cleanly in any order. #357 will be closed once all three are open.

Sibling parts: #415 (clone, liveness and diagnostic paths), #416 (target selection provenance).

## Summary

`ScratchDirectory` was a hand-rolled RAII temp directory: a process-id and nanosecond-timestamp name, a compilation counter to break ties, a 32-attempt retry loop around `DirBuilder`, and a `Drop` that removes the tree. That is a reimplementation of `tempfile`, carrying the collision and cleanup edge cases a widely used crate has already worked through.

## Changes

* Replace it with `tempfile::Builder`. The 0o700 mode the previous code set through `DirBuilderExt` survives as `Builder::permissions`, and `TempDir`'s own `Drop` keeps the cleanup behaviour.
* Add a workspace `tempfile` entry pinned at 3.9.
* Record the new transitive dependencies in `dependency-licenses.csv`.
* Add a unix test asserting the scratch directory's mode is 0o700.

## Rationale

3.9 is the first release exposing `Builder::permissions`; 3.3 through 3.8 would fail to compile against this code. The workspace had no `tempfile` entry before, so this adds one. The 3.3.0 in `crates/fuzzer/rustlantis/difftest` belongs to a separate vendored workspace that is excluded from this one and is unaffected.

The scratch directory holds the caller's LLVM IR and the generated PTX in a shared temp root, so its mode is a security property. It had no test, and it is also the sole reason for the 3.9 floor, so it has one now.

`tempfile` brings six transitive crates: `rustix`, `linux-raw-sys`, `fastrand`, `getrandom`, `errno`, `r-efi`. All are permissively licensed and recorded in `dependency-licenses.csv`. `errno` and `r-efi` are platform-conditional and are not compiled on `x86_64-unknown-linux-gnu`. Lockfiles were regenerated rather than transcribed, so `fastrand` resolves to 2.5.0 here.

If the dependency is unwanted, the alternative is to keep the hand-rolled implementation and add the 0o700 test to it. The other two parts of #357 do not depend on this one either way.

## Validation

* `cargo test -p cuda-oxide-codegen` (118 passed)
* `cargo test -p mir-importer` (801 passed)
* `cargo deny check`, giving `advisories ok, bans ok, licenses ok, sources ok`


nihalpasham:
Taking the dependency: `tempfile` is the ecosystem answer here, and the hand-rolled naming/retry/Drop machinery it deletes was quite a bit of maintenance surface. The `0o700` test is a nice touch.

The branch had gone conflicting, so I pushed a merge with main: both lockfiles regenerated, licenses CSV rows confirmed against the resolved versions. cargo deny (licenses + bans) green, `0o700` test green, device pipeline unaffected. Merging.

nihalpasham (approved):
Taking the dependency; see merge comment.

nihalpasham (approved):
Taking the dependency; validated on the merged head. See merge comment.

PR #423, opened 2026-07-20
Reuse primitive integer type IDs in generation

### LINKED ISSUE #422: Deduplicate primitive integer type IDs in generation
## Problem

`generate/src/generation/mod.rs` manually enumerates all signed and unsigned primitive integer `TyId`s when selecting the right-hand operand of shift operations.

The same list is also duplicated when selecting candidates for `SwitchInt` generation.

There are already `TyKind::INTS` and `TyKind::UINTS` constants, but those contain `TyKind` values, while the generation selectors operate on `TyId`s.

## Proposed change

Add a reusable `TyCtxt::INTEGER_TYPES` associated constant containing the twelve primitive signed and unsigned integer type IDs.

Use this constant for:

* `Shl` and `Shr` right-hand operand selection.
* `SwitchInt` candidate selection.

Remove the existing TODO and duplicated inline arrays.

## Acceptance criteria

* Shift operands can still use all primitive signed and unsigned integer types.
* `SwitchInt` considers the same set of integer types as before.
* No generation behavior changes beyond removing the duplicated lists.
* Formatting and workspace tests pass.

uurl:
Closes #422 

## Summary

* Add `TyCtxt::INTEGER_TYPES` containing all primitive signed and unsigned integer type IDs.
* Use the shared constant when selecting shift right-hand operands.
* Reuse the same constant when selecting `SwitchInt` candidates.
* Remove the obsolete compile-time concatenation TODO and duplicated arrays.

## Rationale

The generation code previously maintained multiple copies of the same twelve `TyId`s.

The existing `TyKind::INTS` and `TyKind::UINTS` constants cannot be used directly because generation selectors consume `TyId` slices rather than `TyKind` values. Keeping the shared list on `TyCtxt` places it alongside the primitive `TyId` definitions and makes it reusable by generator components.

This is a refactoring only and does not change the set or ordering of eligible integer types.

## Testing

* `cargo test --workspace`

### DIFF
--- a/crates/fuzzer/rustlantis/generate/src/generation/mod.rs
+++ b/crates/fuzzer/rustlantis/generate/src/generation/mod.rs
@@ -189,24 +189,7 @@ impl GenerationCtx {
                 Shl | Shr => {
                     // left operand same type as lhs, right can be uint or int
                     let l = self.choose_operand(&[lhs_ty], lhs)?;
-                    // TODO: use a compile time concat
-                    let r = self.choose_operand(
-                        &[
-                            TyCtxt::ISIZE,
-                            TyCtxt::I8,
-                            TyCtxt::I16,
-                            TyCtxt::I32,
-                            TyCtxt::I64,
-                            TyCtxt::I128,
-                            TyCtxt::USIZE,
-                            TyCtxt::U8,
-                            TyCtxt::U16,
-                            TyCtxt::U32,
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks for the PR. This makes sense. We only need one list to maintain instead of two, same types and order, and the stale TODO is gone. Thanks!

PR #425, opened 2026-07-20
Use HashSet membership in overlap checks

### LINKED ISSUE #424: Replace nested overlap scan with HashSet membership
## Context

`PlaceGraph::overlap` collects the transitive subfields of two places and checks whether both collections contain a common `PlaceIndex`.

The current implementation performs this intersection check using nested loops, resulting in `O(n * m)` membership comparisons. There is also an existing TODO asking whether a hash-based lookup should be used.

## Proposed change

Collect one of the subfield collections into a `HashSet<PlaceIndex>` and use `Iterator::any` with `HashSet::contains` to check for a shared node.

The existing fast paths for identical places and places from different allocations should remain unchanged.

## Expected impact

* Replace the quadratic intersection scan with expected linear-time membership checking.
* Make the set-intersection intent more explicit.
* Remove the existing TODO.
* Preserve the current behavior of `PlaceGraph::overlap`.

This is an asymptotic improvement. It does not claim a measured runtime improvement for small place graphs, where hashing and allocation overhead may dominate.

## Acceptance criteria

* The nested loops in `PlaceGraph::overlap` are replaced with `HashSet` membership checking.
* The existing `overlap_check` test passes.
* The complete test suite passes.
* No unrelated code is modified.

uurl:
Closes #424 

## Summary

Replace the nested subfield comparison in `PlaceGraph::overlap` with explicit `HashSet` membership checking.

The change:

* Collects one transitive-subfield collection into a `HashSet<PlaceIndex>`.
* Uses `Iterator::any` and `HashSet::contains` to detect a shared node.
* Removes the existing hash-map TODO.
* Preserves the early returns for identical places and different allocation IDs.

The intersection phase changes from `O(n * m)` comparisons to expected `O(n + m)` work. No measured performance improvement is claimed for small collections.

## Testing

```text
cargo test -p generate overlap_check
RUST_LOG=debug cargo test
```

The existing `overlap_check` unit test covers positive, symmetric, and disjoint overlap cases.

## Scope

Only `generate/src/pgraph.rs` is modified. No behavioral changes are intended.

### DIFF
--- a/crates/fuzzer/rustlantis/generate/src/pgraph.rs
+++ b/crates/fuzzer/rustlantis/generate/src/pgraph.rs
@@ -360,7 +360,7 @@ impl PlaceGraph {
         // Copy ret
         self.copy_place(old_frame.return_destination, callee_ret);
 
-        // TODO: the following to loops can probably be merged
+        // TODO: the following two loops can probably be merged
         // Remove ref edges into about to be deallocated places (necessary to prevent dangling references)
         let mut ref_edges = vec![];
         for pidx in old_frame.locals.right_values() {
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR. We get the same answers with way less work per check, nice speedup for the generator.

PR #426, opened 2026-07-20
Remove stale Literal::ty TODO

uurl:
## Summary

Remove the obsolete TODO comment above `Literal::ty()`.

The method already determines its `TyId` without requiring a `TyCtxt`, so the comment no longer describes pending work.

## Changes

* Remove the stale `TODO: this doesn't need tcx` comment from `mir/src/syntax.rs`.
* Preserve the existing `Literal::ty()` API and implementation.

## Validation

* [ ] `cargo fmt --check`
* [ ] `cargo test --workspace`

### DIFF
--- a/crates/fuzzer/rustlantis/mir/src/syntax.rs
+++ b/crates/fuzzer/rustlantis/mir/src/syntax.rs
@@ -422,7 +422,7 @@ impl TyId {
         }
     }
 
-    // If doesn't contain printer
+    // If it doesn't contain printer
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks, the TODO is indeed stale, but this file is a byte-for-byte copy of upstream rustlantis and we want to keep it that way for clean re-syncs. 

Please send the same fix to cbeuw/rustlantis and we'll pick it up on the next sync (and while you're in there, the other comment should say "pointer" rather than "printer").

PR #430, opened 2026-07-21
Fix Miri handling of nonzero program exit codes

### LINKED ISSUE #429: Distinguish Miri diagnostics from nonzero program exit codes
The Miri backend currently treats every nonzero process exit status as a backend error:

```rust
if !miri_out.status.success() {
    return Err(CompExecError(miri_out.into()));
}
```

This assumes that tested programs always exit with status code 0 and that any nonzero status originates from Miri itself, such as an undefined behavior or type-checking diagnostic.

That assumption is incorrect for valid programs that intentionally return a nonzero exit code. In that case, compiled backends return `Ok(ProcessOutput)` and preserve the program status, while the Miri backend returns `Err(CompExecError)`. This creates inconsistent backend semantics and can produce false differential-test failures.

Miri and rustc diagnostics can be emitted as structured JSON through `--error-format=json`. The backend should use those diagnostics to distinguish Miri errors from the interpreted program's own exit status.

Expected behavior:

* Miri diagnostics with error severity are returned as `CompExecError`.
* Termination by signal is returned as `CompExecError`.
* A valid program that exits with a nonzero status is returned as `Ok(ProcessOutput)`.
* Program-written stderr is preserved.
* Human-readable Miri diagnostics remain available for existing reporting and UB detection.

uurl:
Closes #429 

## Summary

Fix the Miri backend so that a valid program returning a nonzero exit code is not classified as a backend error.

## Problem

The backend previously treated every unsuccessful Miri process status as `CompExecError`. This conflated two distinct cases:

* Miri or rustc reported a diagnostic error.
* The interpreted program completed normally with a nonzero exit code.

Compiled backends already preserve the second case as `Ok(ProcessOutput)`, so Miri behaved inconsistently.

## Changes

* Run Miri with `--error-format=json`.
* Parse structured rustc diagnostics from Miri stderr.
* Classify error-level diagnostics as `CompExecError`.
* Treat termination without an exit code as a backend error.
* Preserve stderr produced directly by the interpreted program.
* Convert structured diagnostics back to their rendered human-readable form.
* Add `serde_json` to the `difftest` dependencies.
* Add a regression test covering a valid program that:

  * writes to stderr;
  * exits with status code 1;
  * remains an `Ok` result for both Miri and LLVM.

## Validation

Executed from `crates/fuzzer/rustlantis`:

```text
cargo fmt --all --check
cargo check -p difftest
cargo test -p difftest nonzero_exit_is_not_a_backend_error -- --nocapture
cargo test -p difftest
```

All four `difftest` integration tests pass, including the existing invalid MIR and undefined-behavior cases.

### DIFF
--- a/crates/fuzzer/rustlantis/Cargo.lock
+++ b/crates/fuzzer/rustlantis/Cargo.lock
@@ -182,6 +182,7 @@ dependencies = [
  "log",
  "rayon",
  "serde",
+ "serde_json",
  "tempfile",
 ]
 
@@ -317,6 +318,12 @@ name = "is_terminal_polyfill"
 version = "1.70.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7943c866cc5cd64cbc25b2e01621d07fa8eb2a1a23160ee81ce38704e97b8ecf"
+
+[[package]]
+name = "itoa"
+version = "1.0.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
 
 [[package]]
 name = "lazy_static"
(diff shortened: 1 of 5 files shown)


nihalpasham:
Thanks for the PR. Nice catch, a program choosing to exit nonzero is not a Miri failure, and the JSON check keeps real errors loud. Thanks!

PR #434, opened 2026-07-21
Clarify unsupported backend compilation

### LINKED ISSUE #433: Clarify unsupported Backend::compile behavior
## Problem

The default implementation of `Backend::compile` currently uses:

```rust
panic!("not implemented")
```

Some backends, such as Miri, override `execute` and intentionally do not support a separate compilation phase. The current panic message makes this look like an accidentally missing implementation rather than an unsupported operation.

## Proposed change

Replace the generic `panic!` with an explicit `unimplemented!` invocation:

```rust
unimplemented!("this backend does not support compilation")
```

This keeps the existing behavior while documenting the intended contract more clearly.

## Acceptance criteria

* `Backend::compile` uses `unimplemented!` in its default implementation.
* The panic message clearly states that compilation is unsupported by the backend.
* Existing backend implementations remain unchanged.
* The workspace builds and all tests pass.

uurl:
Closes #433 

## Summary

* Replace the generic `panic!("not implemented")` in the default `Backend::compile` implementation with `unimplemented!`.
* Clarify that some backends intentionally do not support a separate compilation phase.

## Rationale

Miri overrides `Backend::execute` and executes source code directly, so it does not implement `compile`. Using `unimplemented!` makes this intentional limitation explicit without changing the trait contract or runtime behavior.

## Testing

```text
cargo build
RUST_LOG=debug cargo test
```

No functional behavior change is intended.

### DIFF
--- a/crates/fuzzer/rustlantis/difftest/src/backends.rs
+++ b/crates/fuzzer/rustlantis/difftest/src/backends.rs
@@ -109,7 +109,7 @@ pub struct BackendInitError(pub String);
 
 pub trait Backend: Send + Sync {
     fn compile(&self, _: &Source, _: &Path) -> ProcessOutput {
-        panic!("not implemented")
+        unimplemented!("this backend does not support compilation")
     }
 
     fn execute(&self, source: &Source, target: &Path) -> ExecResult {


nihalpasham:
Thanks for the PR. LGTM

PR #436, opened 2026-07-21
Test unsupported CheckedBinaryOp rejection

### LINKED ISSUE #435: Add regression coverage for unsupported CheckedBinaryOp variants
## Problem

`crates/mir-importer/src/translator/rvalue.rs` supports `CheckedBinaryOp` for `Add`, `Sub`, and `Mul`, but its fail-closed rejection path has no focused regression coverage.

A source-level `error_*` example is not suitable for this case. The rustc MIR bridge only exposes checked binary operations for addition, subtraction, and multiplication. Likewise, `custom_mir::Checked` rejects operators without an overflowing form before cuda-oxide's MIR importer runs.

Consequently, a source example using an unsupported checked operator would test rustc validation rather than the cuda-oxide importer.

## Proposed change

* Extract the supported checked-operation classification into a small private helper in `rvalue.rs`.

* Use that helper from `translate_rvalue`.

* Add a unit test using an unsupported operator such as `BinOp::Div`.

* Assert the existing diagnostic:

  ```text
  CheckedBinaryOp Div not yet implemented
  ```

* Preserve the current handling of checked `Add`, `Sub`, and `Mul`.

## Scope

This issue does not include:

* new checked arithmetic operations
* changes to `mir-lower`
* dialect changes
* LLVM or NVVM lowering changes
* a source-level `error_*` example

## Acceptance criteria

* Unsupported checked binary operators have focused importer-level regression coverage.
* The existing fail-closed diagnostic remains unchanged.
* Checked `Add`, `Sub`, and `Mul` continue to use their existing MIR operations.
* The `mir-importer` unit tests pass.

uurl:
Closes #435 

## Summary

* extract checked binary operation classification from `translate_rvalue`
* preserve the existing checked `Add`, `Sub`, and `Mul` mappings
* add focused unit coverage for the unsupported-operator diagnostic

## Rationale

The unsupported branch cannot be exercised reliably through a source-level negative example with the current rustc MIR contract.

Only addition, subtraction, and multiplication have overflowing MIR forms that are exposed as `CheckedBinaryOp`. Unsupported uses of `custom_mir::Checked` are rejected by rustc before the cuda-oxide importer runs.

A private classification helper allows the importer-owned fail-closed behavior to be tested directly without introducing artificial source fixtures or changing lowering behavior.

## Testing

```text
cargo fmt --all -- --check
cargo test -p mir-importer checked_binary_op
```

## Scope

No changes were made to:

* `mir-lower`
* MIR dialect operations
* LLVM or NVVM lowering
* source-level examples
* error-example categorization

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -236,6 +236,22 @@ fn cast_enum_fields_to_expected_types(
     (result_values, current_prev_op)
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CheckedBinaryOpKind {
+    Add,
+    Sub,
+    Mul,
+}
+
+fn classify_checked_binary_op(bin_op: &mir::BinOp) -> Result<CheckedBinaryOpKind, String> {
+    match bin_op {
+        mir::BinOp::Add => Ok(CheckedBinaryOpKind::Add),
+        mir::BinOp::Sub => Ok(CheckedBinaryOpKind::Sub),
+        mir::BinOp::Mul => Ok(CheckedBinaryOpKind::Mul),
+        _ => Err(format!("CheckedBinaryOp {:?} not yet implemented", bin_op)),
+    }
+}
+
 /// Translates a MIR rvalue to pliron IR operation(s).
 ///
 /// # Returns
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the extraction and the tests. 

I force-pushed a rebase with two changes:

- Resolved the `rvalue.rs` conflict in favor of main's newer result-type code, which translates the checked-op result tuple from the rvalue's rustc type. Keeping the hand-built tuple type would have reintroduced a type-layout mismatch main already fixed. Your `classify_checked_binary_op` helper and both unit tests are kept as-is on top of it.
- Renamed the test module to `checked_binary_op_tests `to avoid clashing with the file's existing test modules.
All 811 mir-importer tests pass, clippy and fmt are clean. Merging.

PR #438, opened 2026-07-21
Translate CopyForDeref as a place read

### LINKED ISSUE #437: Support Rvalue::CopyForDeref in the MIR importer
## Problem

`translate_rvalue` currently ends with a wildcard arm that rejects every MIR `Rvalue` variant not handled by an earlier match arm.

One of the variants hidden behind that wildcard is `Rvalue::CopyForDeref`. Rustc defines `CopyForDeref` as a read from a `Place`, with the resulting value used by a subsequent dereference and projection chain.

The importer already has the required place-reading implementation in `translate_place`. Its slot address-space analysis also treats `CopyForDeref` like `Use(Copy)` and `Use(Move)`, but the actual rvalue translator does not currently provide the corresponding lowering.

The wildcard arm also prevents compiler exhaustiveness checking from identifying new `Rvalue` variants when the pinned Rust toolchain is updated.

## Proposed change

* Translate `Rvalue::CopyForDeref(place)` by delegating to `translate_place`.
* Return no new main operation because `translate_place` inserts and tracks any required load or projection operations.
* Add explicit unsupported arms for `Rvalue::Len` and `Rvalue::ThreadLocalRef`.
* Remove the wildcard arm and the generic rvalue TODO.
* Add `CopyForDeref` to the module's supported-rvalue documentation.

## Rationale

`CopyForDeref` does not require a new MIR dialect operation. It has the same value-producing semantics as an ordinary place read, so reusing `translate_place` keeps projection handling, ZST behavior, slot loads, and operation-order tracking consistent with existing `Copy` and `Move` operands.

Explicitly enumerating `Len` and `ThreadLocalRef` preserves fail-closed behavior while producing more actionable diagnostics. Removing the wildcard also makes future Rust MIR API changes visible as compile-time match-exhaustiveness failures.

## Scope

This issue does not include:

* implementing `Rvalue::Len`
* implementing device-side thread-local storage
* adding MIR dialect operations
* changing `mir-lower`
* changing LLVM or NVVM lowering
* changing place projection semantics

## Acceptance criteria

* `Rvalue::CopyForDeref` is translated through the existing place-read path.
* Any helper operations emitted by `translate_place` remain correctly ordered.
* `Rvalue::Len` receives an explicit unsupported diagnostic.
* `Rvalue::ThreadLocalRef` receives an explicit device-code unsupported diagnostic.
* `translate_rvalue` no longer contains a wildcard match arm.
* The `mir-importer` checks and tests pass.

uurl:
Closes #437 

## Summary

* translate `Rvalue::CopyForDeref` through the existing `translate_place` path
* preserve helper-operation insertion and ordering from place translation
* explicitly reject `Rvalue::Len` and `Rvalue::ThreadLocalRef`
* remove the wildcard rvalue arm and its generic TODO
* document `CopyForDeref` as a supported rvalue

## Rationale

Rustc defines `CopyForDeref` as a read from a `Place`. The importer already implements place reads through `translate_place`, including slot loads, projections, ZST handling, ghost locals, and operation-order bookkeeping.

Delegating to that function avoids introducing a redundant MIR operation or a second place-reading implementation.

The slot address-space classifier already treats `CopyForDeref` like `Use(Copy)` and `Use(Move)`, so this change aligns rvalue emission with the existing analysis behavior.

The remaining unimplemented variants are enumerated explicitly:

* `Len` remains a separate follow-up because arrays and slice metadata require different lowering paths.
* `ThreadLocalRef` remains unsupported because device-side thread-local storage requires an explicit storage and lowering model.

Removing the wildcard makes future additions to rustc's `Rvalue` enum fail during compilation rather than being silently routed to a generic runtime diagnostic.

## Testing

```text
cargo fmt --all -- --check
cargo check -p mir-importer
cargo test -p mir-importer
scripts/smoketest.sh --compile-only --no-color --only '^ref_index_projections$'
```

## Scope

No changes were made to:

* `mir-lower`
* MIR dialect operations
* LLVM or NVVM lowering
* place projection behavior
* array or slice length lowering
* thread-local storage support

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -1601,6 +1601,15 @@ pub fn translate_rvalue(
         mir::Rvalue::CopyForDeref(place) => {
             // CopyForDeref is semantically a place read. Any load or projection
             // operations emitted by translate_place are already inserted and tracked.
+            //
+            // Note: on the pinned toolchain the MIR optimization pipeline
+            // (copy-prop/GVN) eliminates every CopyForDeref before the
+            // optimized MIR reaches this importer; nested-deref kernels
+            // (`**rr`, including through raw pointers) were probed and none
+            // produce one here today. This arm is defensive coverage so a
+            // future toolchain bump or lower mir-opt-level cannot regress
+            // valid nested-deref kernels into an "unsupported construct"
+            // failure.
             let (value, last_inserted) =
                 translate_place(ctx, body, place, value_map, block_ptr, prev_op, loc)?;
 
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -20,6 +20,7 @@
 //! | `Use(operand)`      | `mir.load` of the source slot (no op for constants)   |
 //! | `Aggregate`         | `mir.construct_tuple/struct/enum/array`               |
 //! | `Repeat`            | `mir.construct_array` (array repeat syntax)           |
+//! | `CopyForDeref`      | Same place-read lowering as `Copy`/`Move`             |
 //!
 //! # Key Functions
 //!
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the cleanup, the exhaustive match with named arms is better than the old catch-all. 

I force-pushed two additions:

- Rebased onto current main.
- Added a comment on the new arm documenting a finding from our verification: on the pinned toolchain, the MIR optimization pipeline eliminates every `CopyForDeref` before the optimized MIR reaches the importer (we probed `nested-deref` kernels, including raw-pointer variants, and none produce one). So this arm is defensive coverage for future toolchain bumps rather than a live bug fix, which is worth recording so nobody hunts for a repro later.

All 809 mir-importer tests pass. Merging.

PR #440, opened 2026-07-21
Emit CUDA diagnostics for gpu_assert! messages

### LINKED ISSUE #439: Implement device-side diagnostics for gpu_assert! messages
## Problem

The message form of `gpu_assert!` currently ignores its message and falls back to `debug::trap()`:

```rust
gpu_assert!(condition, "diagnostic message");
```

As a result, failed assertions provide no assertion message or source location.

The existing TODO refers to `llvm.nvvm.assertfail`. CUDA exposes assertion reporting through the device-side `__assertfail` system call rather than an LLVM NVVM intrinsic:

```c
void __assertfail(
    const char *message,
    const char *file,
    unsigned line,
    const char *function,
    size_t char_size
);
```

## Proposed implementation

Implement the message form of `gpu_assert!` by:

1. Generating null-terminated static strings for:

   * the assertion message;
   * `file!()`;
   * `module_path!()` as the available function context.
2. Passing `line!()` and a character size of `1`.
3. Calling a hidden `cuda_device::debug::__gpu_assertfail` placeholder.
4. Translating that placeholder to a typed `nvvm.assertfail` dialect operation.
5. Lowering the operation to an external call to CUDA's `__assertfail` system call.

The no-message form should continue to use `debug::trap()`.

## Scope

This issue covers:

```rust
gpu_assert!(condition, "literal message");
```

It does not cover:

* ordinary Rust `assert!` MIR terminators;
* formatted assertion messages;
* runtime-generated `&str` values;
* exact Rust function-name extraction.

The message must initially be a string literal because CUDA requires a null-terminated C string with storage that remains valid on the device.

## Acceptance criteria

* `gpu_assert!(condition)` retains its current `trap()` behavior.
* `gpu_assert!(condition, "message")` lowers to a call to `__assertfail`.
* The emitted call receives:

  * a null-terminated message;
  * a null-terminated source file;
  * the call-site line number;
  * stable function or module context;
  * `char_size == 1`.
* The condition is evaluated exactly once.
* The custom-message form accepts string literals.
* Generated LLVM or NVVM IR contains a declaration and call to `__assertfail`.
* The existing successful assertion example continues to run.
* A manually invoked failing example reports an assertion failure and terminates without running subsequent CUDA operations.
* The new dialect operation has verifier and lowering tests.

uurl:
Closes #439 

## Summary

Implement device-side assertion diagnostics for the message form of `gpu_assert!`.

Failed assertions now call CUDA's `__assertfail` system call with the supplied message, source file, source line, module context, and character size.

## Changes

* Add the hidden `cuda_device::debug::__gpu_assertfail` compiler placeholder.
* Update the message form of `gpu_assert!` to generate null-terminated static metadata.
* Add a typed `nvvm.assertfail` dialect operation.
* Add MIR importer dispatch and operand translation for the new placeholder.
* Lower `nvvm.assertfail` to:

```llvm
declare void @__assertfail(ptr, ptr, i32, ptr, i64)
```

* Add dialect verification and lowering coverage.
* Extend the debug example with an isolated assertion-failure mode.
* Update the assertion documentation.

## Behavior

Before:

```rust
gpu_assert!(value >= 0, "value must be non-negative");
```

failed through `trap()` and discarded the message.

After:

```rust
gpu_assert!(value >= 0, "value must be non-negative");
```

reports the message and call-site metadata through CUDA's device-side assertion mechanism.

The no-message form remains unchanged and continues to use `trap()`.

## Scope limitations

* Assertion messages must be string literals.
* Formatted messages are not supported.
* Ordinary Rust `assert!` lowering is unchanged.
* `module_path!()` is used as the available function context.

## Validation

* [ ] `cargo fmt --check`
* [ ] `cargo test -p dialect-nvvm`
* [ ] `cargo test -p mir-importer`
* [ ] `cargo test -p mir-lower`
* [ ] `cargo test -p cuda-device`
* [ ] `cargo oxide pipeline debug`
* [ ] Confirm generated IR declares and calls `__assertfail`
* [ ] `cargo oxide run debug`
* [ ] `cargo oxide run debug -- --assert-failure`

### DIFF
--- a/crates/cuda-device/src/debug.rs
+++ b/crates/cuda-device/src/debug.rs
@@ -57,21 +57,23 @@ include!("generated/debug_control.rs");
 /// // Simple assertion
 /// gpu_assert!(x >= 0);
 ///
-/// // With custom message (message is ignored in current impl)
+/// // With a custom string-literal message
 /// gpu_assert!(idx < len, "Index out of bounds");
 /// ```
 ///
 /// # Behavior
 ///
 /// When the condition is false:
-/// - The kernel execution is aborted via `trap()`
+/// - The no-message form aborts kernel execution via [`trap()`]
+/// - The message form reports the message and call-site metadata via CUDA's
+///   device-side `__assertfail` system call
 /// - The CUDA driver reports an error to the host
 /// - Other threads may continue briefly before the error propagates
 ///
 /// # Notes
 ///
 /// - Use sparingly in performance-critical code
-/// - The message argument is currently ignored (will be supported with assertfail)
+/// - Custom messages must be string literals
 /// - For debugging, consider using [`breakpoint()`] instead
 #[macro_export]
 macro_rules! gpu_assert {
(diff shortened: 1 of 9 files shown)


nihalpasham:
LGTM, this is the right mechanism, `__assertfail` is what nvcc uses, and failures now look like native CUDA assert failures. 

I force-pushed a rebase plus one fix and the missing pieces:

- Fixed a crash in the lowering: `replace_operation` panicked with "Replacement values must match the number of results" because `nvvm.assertfail` has no results while the void call carries one. It now uses insert plus `erase_operation`, the same pattern the zero-result cp.async control ops use. The branch as pushed could not compile the debug example.
- Added the verifier shape tests and a lowering test from the issue's acceptance criteria.
- Added a manually invoked `--fail-assert` mode to the debug example and verified on a GPU: the driver prints the message with file and line, and the host gets CUDA_ERROR_ASSERT (710). The passing suite is unchanged.

One follow-up we'll track separately: the call is not marked noreturn yet (nvcc and clang emit noreturn plus unreachable), which is runtime-safe but lets the optimizer keep dead code after the assert.

PR #445, opened 2026-07-22
Publish the backend to the shared cache on setup

### LINKED ISSUE #444: cargo oxide setup does not publish the backend to the shared cache
## Problem

`cargo oxide setup` builds the backend from the local checkout and reports:

```text
✓ Backend built: .../crates/rustc-codegen-cuda/target/.../librustc_codegen_cuda.so

✓ Backend is ready. You can now use:
  cargo oxide run <example>
  cargo oxide build <example>
```

It never writes that build to `~/.cargo/cuda-oxide/`. That cache is step 4 of the discovery order in `backend.rs`, and it is the only backend a project outside the repository can reach: `find_workspace_root` walks up looking for `crates/rustc-codegen-cuda` and finds nothing from an unrelated crate.

So `setup` succeeds for in-repo examples and leaves external projects on whatever the cache last held.

The three existing staleness checks do not catch this. They compare the cached `.so` against the running `cargo-oxide` binary, against the cached source tree, and against the active toolchain fingerprint. A cache that is internally consistent but predates the developer's local checkout passes all three.

Observed: a cache from two months earlier rejected `f64::acos` in device code with

```text
CUDA-OXIDE: FORBIDDEN CRATE IN DEVICE CODE
Device code calls: std::sys::cmath::acos
```

while the in-repo `asin_acos_smoke` example compiled the same call, because it resolved to the local build instead. `cargo oxide setup` and `cargo oxide doctor` both reported success beforehand, the latter naming the local `.so` the external build was not going to load.

## Proposed change

Have `setup` publish the backend it just built to `~/.cargo/cuda-oxide/`, and report the path.

The copy should be best effort: on failure, warn and point at `CUDA_OXIDE_BACKEND` rather than failing the command, since the in-repo build remains usable.

The toolchain fingerprint has to be written alongside the `.so`, as `auto_fetch_and_build` already does. A `.so` installed without one falls back to the mtime checks, which cannot see a toolchain swap.

## Acceptance criteria

* `cargo oxide setup` copies the freshly built backend to `~/.cargo/cuda-oxide/librustc_codegen_cuda.so`.
* The toolchain fingerprint is written next to it.
* The published path is reported on success; a failure warns and mentions `CUDA_OXIDE_BACKEND` without failing the command.
* Installing over an existing cached backend replaces it.
* Unit tests cover both the fresh install and the replacement.
* `cargo test --workspace` and `cargo clippy -- -D warnings` pass.

alejandro-soto-franco:
Closes #444

## Summary

`cargo oxide setup` built the backend and reported "Backend is ready" without writing it to `~/.cargo/cuda-oxide/`. That cache is step 4 of the discovery order and the only backend a project outside the repository can reach, since `find_workspace_root` finds no `crates/rustc-codegen-cuda` above an unrelated crate. `setup` now publishes what it built and reports the path, so external projects use the backend the developer just made rather than whatever the cache last held.

## Changes

- Extract `install_backend_into` from `auto_fetch_and_build`. It copies the `.so` and writes the toolchain fingerprint beside it. The fingerprint has to accompany the `.so`, or the next lookup falls back to the mtime checks, which cannot see a toolchain swap. It takes the cache directory explicitly so it is testable without touching `CARGO_HOME`.
- Add `publish_to_cache`, which resolves the cache directory and delegates.
- Call it from `setup` and report the published path. The copy is best effort: on failure it warns and points at `CUDA_OXIDE_BACKEND` rather than failing the command, since the in-repo build is still usable.
- Add two unit tests: installing writes both artifacts, and installing over an existing cached backend replaces it, which is the path `setup` takes on every run after the first.

No change to the discovery order or to the existing staleness checks.

## Testing

Unit and lint, on this branch off `main`:

```text
$ cargo test -p cargo-oxide
test backend::tests::installing_writes_both_the_backend_and_its_fingerprint ... ok
test backend::tests::installing_replaces_an_existing_cached_backend ... ok
test result: ok. 88 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo clippy -p cargo-oxide --all-targets -- -D warnings
(no warnings)

$ RUSTDOCFLAGS="-D warnings" cargo doc -p cargo-oxide --no-deps
(no warnings)
```

End to end, against the bug in #444. Starting from no cache:

```text
$ rm -rf ~/.cargo/cuda-oxide
$ cargo oxide setup
✓ Backend built: .../crates/rustc-codegen-cuda/target/.../librustc_codegen_cuda.so

✓ Backend is ready. You can now use:
  cargo oxide run <example>
  cargo oxide build <example>

✓ Published to /home/<user>/.cargo/cuda-oxide/librustc_codegen_cuda.so
  Projects outside this repo will now use this build.

$ ls ~/.cargo/cuda-oxide/
librustc_codegen_cuda.so  toolchain-fingerprint.txt
```

A project outside the repository then builds against the cache alone, with `CUDA_OXIDE_BACKEND` unset. Before this change the same build failed with `FORBIDDEN CRATE IN DEVICE CODE` on `f64::acos` against a two month old cache:

```text
$ env -u CUDA_OXIDE_BACKEND cargo oxide run <external-project>
  dim    points   exp max err   log max err
    3      4096     3.220e-15     3.331e-16
   10      4096     3.331e-16     3.886e-16
   50      2048     1.665e-16     2.706e-16

PASSED
```

In-repo examples are unaffected, since they resolve to the local build at step 3:

```text
$ cargo oxide run asin_acos_smoke
SUCCESS: 14 cases × 4 variants within 2 ULP of host libm
```

No example added: the change is in the build tool rather than in codegen, and the behavior is covered by the unit tests plus the end to end run above.

## Checklist

### DIFF
--- a/crates/cargo-oxide/src/backend.rs
+++ b/crates/cargo-oxide/src/backend.rs
@@ -599,12 +599,45 @@ fn auto_fetch_and_build() -> PathBuf {
     let codegen_crate = src_dir.join("crates/rustc-codegen-cuda");
     let built_so = build_backend_from_source(&codegen_crate);
     if built_so.exists() {
-        std::fs::copy(&built_so, &so_path).expect("Failed to copy backend to cache");
-        write_toolchain_fingerprint(&cache_dir);
+        install_backend_into(&cache_dir, &built_so).expect("Failed to copy backend to cache");
         eprintln!("✓ Backend cached at {}", so_path.display());
     }
 
     so_path
+}
+
+/// Copies a freshly built backend into `cache_dir` and records the toolchain
+/// fingerprint beside it.
+///
+/// The fingerprint must be written whenever the `.so` is. A `.so` installed
+/// without one falls back to the mtime checks, which cannot see a toolchain
+/// swap, so the next lookup would load a backend linked against the wrong
+/// `librustc_driver`.
+///
+/// Takes the directory explicitly so it can be exercised without touching
+/// `CARGO_HOME`.
+fn install_backend_into(cache_dir: &Path, built_so: &Path) -> std::io::Result<PathBuf> {
+    std::fs::create_dir_all(cache_dir)?;
+    let so_path = cache_dir.join("librustc_codegen_cuda.so");
+    std::fs::copy(built_so, &so_path)?;
+    write_toolchain_fingerprint(cache_dir);
+    Ok(so_path)
+}
+
+/// Publishes a freshly built backend to the shared cache at
+/// `~/.cargo/cuda-oxide/`.
+///
+/// That path is what step 4 of the discovery order resolves to, and it is the
+/// only one a project outside this repository can reach: `find_workspace_root`
+/// walks up from the current directory looking for `crates/rustc-codegen-cuda`
+/// and finds nothing from an unrelated crate.
+///
+/// Returns `None` when the cache directory cannot be determined or the copy
+/// fails. Callers treat this as best effort: a failure leaves the in-repo build
+/// usable and costs external projects only a rebuild.
+pub fn publish_to_cache(built_so: &Path) -> Option<PathBuf> {
+    let cache_dir = cache_directory()?;
+    install_backend_into(&cache_dir, built_so).ok()
 }
 
 /// Returns the active rustc sysroot path (e.g., `~/.rustup/toolchains/nightly-...`).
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks, this closes the gap for external projects. 

**Verified end to end:** tests and clippy green on the merge with main, and a wiped `~/.cargo/cuda-oxide` gets both the .so and the toolchain fingerprint, with the publish path printed. Closes #444. Merging.

nihalpasham (approved):
Validated end to end; see merge comment.

PR #447, opened 2026-07-22
Report the shared cache in doctor

### LINKED ISSUE #446: doctor does not report the backend external projects will use
## Problem

`cargo oxide doctor` reports the backend the current context resolves to:

```text
Codegen backend... ✓ .../crates/rustc-codegen-cuda/target/.../librustc_codegen_cuda.so
```

Run inside the repository, that is the local build, reached at step 3 of the discovery order. A project outside the repository cannot reach it: `find_workspace_root` walks up looking for `crates/rustc-codegen-cuda` and finds nothing, so it falls through to the cache at `~/.cargo/cuda-oxide/`.

`doctor` never mentions that cache, so the two can disagree with every check reporting success. This is how a two month old cache passed a fully green `doctor` and then rejected `f64::acos` in an external build with

```text
CUDA-OXIDE: FORBIDDEN CRATE IN DEVICE CODE
Device code calls: std::sys::cmath::acos
```

while the in-repo `asin_acos_smoke` example compiled the same call against the local build.

The existing staleness checks do not cover it either. They compare the cached `.so` against the running binary, the cached source tree, and the active toolchain fingerprint. A cache that is internally consistent but predates the developer's checkout passes all three.

## Proposed change

Add a check to `doctor` reporting the cached backend, and warn when it predates the backend built in this checkout.

An empty cache should be reported without a warning, since an external project builds on first use. An unreadable mtime on either side should not warn, since `doctor` cannot substantiate the alarm.

## Acceptance criteria

* `doctor` reports the path of the cached backend as its own check.
* It warns when the cache predates the local build, and names both remedies (`cargo oxide setup`, `CUDA_OXIDE_BACKEND`).
* An absent cache is reported without a warning.
* The comparison is unit tested without requiring a populated cache.
* `cargo test --workspace` and `cargo clippy -- -D warnings` pass.

alejandro-soto-franco:
Closes #446

## Summary

`doctor` reported the backend the current context resolves to and nothing else. Inside the repository that is the local build; a project outside it resolves to `~/.cargo/cuda-oxide/`. The two can disagree with every check still green, which is how a two month old cache passed `doctor` and then rejected `f64::acos` in an external build. `doctor` now reports the cached backend as its own check and warns when it predates the local build.

## Changes

- Add `CacheReport`, `cached_backend_path` and `compare_cache_to_local` to `backend`. The comparison is pure and takes both paths, so it is testable without a populated cache.
- Add a shared cache check to `doctor`, between the backend check and the CUDA headers check.
- Compare by mtime, matching the staleness checks elsewhere in the module. An unreadable mtime on either side reports up to date rather than warning, since `doctor` should not raise an alarm it cannot substantiate. An empty cache is reported without a warning, since an external project builds on first use.
- Four unit tests: absent cache, cache older than the local build, cache newer than it, and nothing built locally to compare against.

Related to #444, which fixes `setup` so the cache is populated in the first place. The two are independent: this one is the diagnostic, that one is the remedy `doctor` now points at. Neither depends on the other to apply.

## Testing

```text
$ cargo test -p cargo-oxide
test backend::tests::cache_report_is_absent_when_nothing_is_cached ... ok
test backend::tests::cache_report_is_older_when_the_local_build_is_newer ... ok
test backend::tests::cache_report_is_up_to_date_when_the_cache_is_newer ... ok
test backend::tests::cache_report_is_up_to_date_when_nothing_is_built_locally ... ok
test result: ok. 90 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo clippy -p cargo-oxide --all-targets -- -D warnings
(no warnings)

$ RUSTDOCFLAGS="-D warnings" cargo doc -p cargo-oxide --no-deps
(no warnings)
```

Cache matching the local build:

```text
$ cargo oxide doctor
Codegen backend... ✓ .../crates/rustc-codegen-cuda/target/.../librustc_codegen_cuda.so
Shared cache (external projects)... ✓ /home/<user>/.cargo/cuda-oxide/librustc_codegen_cuda.so
```

Cache backdated to reproduce the reported state:

```text
$ touch -d '2026-05-27 17:18:00' ~/.cargo/cuda-oxide/librustc_codegen_cuda.so
$ cargo oxide doctor
Shared cache (external projects)... ⚠ /home/<user>/.cargo/cuda-oxide/librustc_codegen_cuda.so
  Older than the backend built here, so projects outside this
  repository would load a different one. Run `cargo oxide setup`
  to publish, or set CUDA_OXIDE_BACKEND to pin an explicit path.
```

No example added: the change is a diagnostic in the build tool rather than codegen, and is covered by the unit tests plus the `doctor` runs above.

## Checklist


nihalpasham:
LGTM. 

**Verified:** on a fresh cache, and after backdating the cached .so, doctor prints the warning with both remedies. 90 tests and clippy green. Closes #446. Merging.

nihalpasham (dismissed):
Validated; see merge comment.

nihalpasham (approved):
Validated on the post-conflict merge; see merge comment.

PR #453, opened 2026-07-23
Mark __assertfail noreturn and emit unreachable

### LINKED ISSUE #452: Mark __assertfail noreturn and terminate assertion blocks with unreachable
## Problem

PR #440  added device-side assertion diagnostics by lowering `nvvm.assertfail` to CUDA's `__assertfail` system call.

The current LLVM lowering emits a normal void call:

```llvm
call void @__assertfail(...)
```

and preserves the control flow that follows the assertion call.

CUDA assertion failure does not return. nvcc and Clang model this with `noreturn` and terminate the block with `unreachable`. Without those properties, runtime behavior remains correct, but LLVM may retain or optimize code under the assumption that execution can continue after `__assertfail`.

This follow-up was identified during review of #440.

## Proposed change

Model the non-returning behavior explicitly:

```llvm
declare void @__assertfail(...) noreturn

call void @__assertfail(...) noreturn
unreachable
```

The lowering should:

1. Mark the generated `__assertfail` declaration as `noreturn`.
2. Mark the generated call as `noreturn`.
3. Emit `llvm.unreachable` immediately after the call.
4. Remove operations following `nvvm.assertfail` in the same block, since they are unreachable.
5. Preserve the existing five-operand ABI and device-side diagnostic behavior.

The LLVM textual exporter should emit the `noreturn` attribute from an explicit LLVM operation marker rather than hardcoding behavior based on the `__assertfail` symbol name.

## Acceptance criteria

* The exported `__assertfail` declaration contains `noreturn`.
* The exported call to `__assertfail` contains `noreturn`.
* `llvm.unreachable` immediately follows the call.
* Operations following `nvvm.assertfail` in the same block are removed.
* The `nvvm.assertfail` operation is fully consumed by lowering.
* The call continues to forward all five operands:

  * message;
  * file;
  * line;
  * function context;
  * character size.
* A lowering test verifies the LLVM dialect shape and exported LLVM IR.
* The passing debug example remains unchanged.
* The manually invoked failing debug example still reports the assertion and returns `CUDA_ERROR_ASSERT`.

uurl:
Closes #452 

## Summary

Model CUDA's device-side `__assertfail` system call as non-returning in the LLVM lowering.

The assertion lowering now emits a `noreturn` declaration and call, terminates the block with `llvm.unreachable`, and removes operations that would otherwise remain after the assertion failure.

## Changes

* Add an internal LLVM operation marker for `noreturn`.
* Export `noreturn` on marked function declarations.
* Export `noreturn` on marked call instructions.
* Return the existing or newly created declaration from `ensure_intrinsic_declared`.
* Mark the `__assertfail` declaration and call as non-returning.
* Emit `llvm.unreachable` immediately after the call.
* Remove the unreachable suffix of the containing block.
* Extend the `__assertfail` lowering test to verify:

  * the declaration marker;
  * the call marker;
  * all five forwarded operands;
  * immediate `llvm.unreachable`;
  * removal of the trailing return;
  * textual LLVM IR containing `noreturn`.

## Generated LLVM IR

Before:

```llvm
declare void @__assertfail(ptr, ptr, i32, ptr, i64)

call void @__assertfail(...)
br label %continue
```

After:

```llvm
declare void @__assertfail(ptr, ptr, i32, ptr, i64) noreturn

call void @__assertfail(...) noreturn
unreachable
```

## Behavior

Runtime assertion behavior is unchanged:

* the CUDA driver prints the assertion message with source location;
* synchronization reports `CUDA_ERROR_ASSERT`;
* execution does not continue after the failing assertion.

The change makes that existing runtime contract explicit in LLVM IR so optimization passes can remove dead control flow after the assertion.

## Validation

* [ ] `cargo fmt --all -- --check`
* [ ] `cargo test -p llvm-export`
* [ ] `cargo test -p mir-lower --test lowering_test test_assertfail_lowers_to_noreturn_call_and_unreachable -- --exact --nocapture`
* [ ] `cargo test -p mir-lower`
* [ ] `cargo check`
* [ ] `cargo oxide pipeline debug`
* [ ] Confirm the exported declaration and call contain `noreturn`
* [ ] Confirm `unreachable` immediately follows the call
* [ ] `cargo oxide run debug`
* [ ] `cargo oxide run debug -- --fail-assert`

### DIFF
--- a/crates/llvm-export/src/export/function.rs
+++ b/crates/llvm-export/src/export/function.rs
@@ -195,6 +195,7 @@ impl<'a> ModuleExportState<'a> {
         // Check for kernel attribute
         let kernel_key: pliron::identifier::Identifier = "gpu_kernel".try_into().unwrap();
         let attrs = &func.get_operation().deref(self.ctx).attributes;
+        let is_noreturn = crate::ops::op_noreturn(self.ctx, func.get_operation());
         let is_kernel = attrs
             .get::<pliron::builtin::attributes::StringAttr>(&kernel_key)
             .is_some();
@@ -290,6 +291,10 @@ impl<'a> ModuleExportState<'a> {
                 }
             }
             write!(output, ")").unwrap();
+
+            if is_noreturn {
+                write!(output, " noreturn").unwrap();
+            }
 
             // Check if this is a known convergent intrinsic
             let is_convergent_intrinsic = Self::is_convergent_intrinsic(&fixed_func_name);
--- a/crates/llvm-export/src/export/ops.rs
+++ b/crates/llvm-export/src/export/ops.rs
@@ -1194,6 +1194,7 @@ impl<'a> ModuleExportState<'a> {
         output: &mut String,
     ) -> Result<(), String> {
         let op_ref = op.get_operation().deref(self.ctx);
+        let is_noreturn = crate::ops::op_noreturn(self.ctx, op.get_operation());
         let callee = op.callee(self.ctx);
         let func_ty = op.callee_type(self.ctx);
         let func_ty_ref = func_ty.deref(self.ctx);
@@ -1394,7 +1395,8 @@ impl<'a> ModuleExportState<'a> {
         // performs a barrier / shuffle / vote, `opt -O2` must not sink or
         // duplicate the call across divergent control flow. opt strips the
         // attribute from calls it proves never reach a convergent op.
-        writeln!(output, ") #0").unwrap();
+        let noreturn_attr = if is_noreturn { " noreturn" } else { "" };
+        writeln!(output, "){noreturn_attr} #0").unwrap();
         self.convergent_used = true;
 
         if normalize_pointer_result {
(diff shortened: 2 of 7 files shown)


nihalpasham:
LGTM, I believe this is a follow-up #440, driving it with a generic marker attribute instead of matching on the symbol name is the right shape.

I pushed one addition: erasing the trailing ops can orphan a successor block. The device stub returns `()`, so translation always emits a branch after `nvvm.assertfail`; when that branch was the block's only incoming edge (what a constant-false `gpu_assert!` leaves after MIR opts), the surviving block had arguments but zero predecessors and the exporter rejected the module. 

Lowering now runs pliron's unreachable-block sweep after conversion, with a regression test that fails without it. The debug example compiles through `opt -O2` + `llc` with both noreturn markers in the IR; GPU check (diagnostic + `CUDA_ERROR_ASSERT`) pending with this batch. Closes #452. Merging.

nihalpasham (approved):
Validated; see merge comment.

PR #454, opened 2026-07-23
fix(cargo-oxide): pass -Zalways-encode-mir so cross-crate recursive device fns link

nh13:
## Problem

A `#[kernel]` that calls a **recursive function defined in a dependency crate** fails device codegen with `Symbol <crate>__<fn> not found`. The device collector skips any callee whose MIR isn't cross-crate-available (`collector.rs:1630`), and rustc encodes cross-crate MIR only for `#[inline]`/generic items — so a recursive dependency fn (which can't be inlined away) is *called* but never *emitted*. Non-recursive dependency fns get inlined, hiding the bug; intrinsics are name-filtered earlier (`collector.rs:1569`), so the recursive-dep case is the only legitimate consumer of the 1630 skip.

## Fix

Add `-Zalways-encode-mir` to cargo-oxide's required device rustflags so every reachable dependency function is device-compilable — the established approach for whole-program-MIR tools (e.g. Miri). Adds the `cross_crate_recursion` regression example (fails before, passes after) and updates the flag-list tests. Validated on a T4 (CUDA 13.2, nightly-2026-04-03): the example and a real-world kernel reaching a recursive radix across a crate boundary both go from `Symbol not found` to compiling.

## Workaround (without this patch)

Mark the recursive dependency fn `#[inline]` — that forces rustc to encode its MIR cross-crate. Works, but must be repeated for every non-`#[inline]`, non-generic device-reachable recursive fn.

## Principled long-term replacement

Per-crate device compile + device-link (each crate emits its own definitions as LTOIR, linked via nvJitLink) would remove the need for cross-crate MIR entirely, along with the build-wide flag's rmeta cost. The scaffolding partially exists (`device_externs`, `emit-ltoir`, `requires_ptx_bundle_merge`) but today targets FFI with *external* C++ LTOIR, not automatic per-Rust-crate linking — so it is substantial new work. `-Zalways-encode-mir` is the interim fix.

---


nihalpasham:
Thanks, this makes sense (I believe this is the same mechanism Miri and Kani use). 

Reproduced fail-before/pass-after with your example.

I pushed a small follow-up:

- The two exact-position flag-list tests still asserted a 5-flag required tail, so `cargo test -p cargo-oxide` failed as submitted. Bumped both to 6.
- Documented `-Z always-encode-mir` in the backend `README` and lib.rs manual-flag tables so `non-cargo-oxide` invokers learn about it.

Heads up: the required `RUSTFLAGS` change means a one-time full rebuild for everyone. Merging.

nihalpasham (approved):
Validated; see merge comment.

PR #456, opened 2026-07-23
Translate Len rvalues for arrays and slices

### LINKED ISSUE #455: Support Rvalue::Len in the MIR importer
## Problem

The MIR importer explicitly rejects `Rvalue::Len`.

`Len` produces a `usize` value but requires different lowering depending on the type of its `Place`:

* For `[T; N]`, the result is the compile-time array length `N`.
* For `[T]`, the result is the metadata component of the slice fat pointer.

The importer already has the operations required for both paths:

* `MirConstantOp` for an array length.
* `MirExtractFieldOp` for extracting field 1 from a slice representation.
* `MirSliceType` for the `(data pointer, length)` fat-pointer value.

The existing `UnaryOp::PtrMetadata` implementation already performs the same field-1 extraction required by the slice case.

## Proposed change

* Add a dedicated `translate_len` helper.
* Lower array lengths to a `mir.constant : usize`.
* Lower ordinary slice lengths by recovering the slice fat-pointer value and extracting metadata field 1.
* Keep unsupported place shapes fail-closed with precise diagnostics.
* Add focused unit coverage for the generated constant and metadata-extraction operations.
* Document `Len` as a supported rvalue.

## Array lowering

For `RigidTy::Array(_, count)`:

* Evaluate `count` with `eval_target_usize`.
* Emit a `MirConstantOp` with the importer's `usize` type.
* Preserve the existing operation-ordering contract of `translate_rvalue`.

## Slice lowering

For `RigidTy::Slice(_)`:

* Require the slice place to end in `ProjectionElem::Deref`.
* Translate the place prefix to recover the `MirSliceType` fat-pointer value.
* Verify that the translated value has slice representation.
* Emit `MirExtractFieldOp` with field index 1 and result type `usize`.

## Testing strategy

Ordinary source-level `.len()` expressions may be transformed before optimized MIR reaches the importer, for example into a constant or `PtrMetadata`.

The focused tests therefore validate the importer-owned operation constructors directly:

* array length produces a valid `MirConstantOp : usize`
* slice length produces a valid `MirExtractFieldOp` using metadata field 1

Existing importer tests and length-related examples remain useful as regression coverage for the currently active lowering paths.

## Scope

This issue does not include:

* device thread-local storage
* changes to `mir-lower`
* new MIR dialect operations
* changes to slice ABI representation
* `DisjointSlice::len`, which has a separate intrinsic lowering
* structs with unsized slice tails
* subslice projection support

## Acceptance criteria

* Array `Rvalue::Len` produces a `usize` constant with the evaluated array count.
* Slice `Rvalue::Len` extracts field 1 from a valid slice fat-pointer value.
* Unsupported place types and projection shapes return precise diagnostics.
* The existing `PtrMetadata` path remains unchanged.
* Focused length-operation unit tests pass.
* All `mir-importer` tests pass.
* Clippy passes with warnings denied.

uurl:
Closes #455 

## Summary

* add `Rvalue::Len` translation for arrays and ordinary slices
* emit a `mir.constant : usize` for array lengths
* extract metadata field 1 from slice fat-pointer values
* add precise diagnostics for unsupported place types and projection shapes
* add focused unit tests for both generated operation forms
* document `Len` as a supported rvalue

## Array lowering

For an array place, the importer resolves `RigidTy::Array(_, count)`, evaluates the count with `eval_target_usize`, and creates a `MirConstantOp` using the target `usize` type.

No new dialect operation or lowering rule is required.

## Slice lowering

A slice place refers to the unsized `[T]` behind a reference or raw pointer.

The implementation removes the final `Deref` projection, translates the remaining place to recover the `MirSliceType` fat-pointer value, validates that representation, and creates a `MirExtractFieldOp` for field 1.

This matches the existing slice ABI:

```text
field 0: data pointer
field 1: length
```

It also matches the importer’s existing `UnaryOp::PtrMetadata` lowering.

## Testing

Focused importer-level tests cover the generated operations directly:

* an array length constant has type `usize`, carries the expected value, and passes `MirConstantOp` verification
* a slice length extraction reads field 1, returns `usize`, and passes `MirExtractFieldOp` verification

This avoids relying exclusively on source-level `.len()` expressions, which the MIR optimization pipeline may convert to constants or pointer-metadata operations before the importer observes them.

## Validation

```text
cargo fmt --all -- --check
cargo check -p mir-importer
cargo test -p mir-importer len_rvalue
cargo test -p mir-importer
scripts/smoketest.sh --compile-only --no-color --only '^(compiler_features|disjoint_slice_len)$'
```

The smoketests verify that the existing active slice-length paths remain unchanged; the focused unit tests cover the new `Rvalue::Len` operation construction.

## Scope

No changes were made to:

* `mir-lower`
* MIR dialect definitions
* LLVM or NVVM lowering
* the slice ABI
* `DisjointSlice::len` intrinsic lowering
* device thread-local storage
* unsized-tail struct projections

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -21,6 +21,7 @@
 //! | `Aggregate`         | `mir.construct_tuple/struct/enum/array`               |
 //! | `Repeat`            | `mir.construct_array` (array repeat syntax)           |
 //! | `CopyForDeref`      | Same place-read lowering as `Copy`/`Move`             |
+//! | `Len`               | `mir.constant` for arrays; slice metadata extraction  |
 //!
 //! # Key Functions
 //!
@@ -37,10 +38,10 @@ use dialect_mir::attributes::MirCastKindAttr;
 use dialect_mir::attributes::MirFP16Attr;
 use dialect_mir::ops::{
     MirAddOp, MirBitAndOp, MirBitOrOp, MirBitXorOp, MirCastOp, MirCheckedAddOp, MirCheckedMulOp,
-    MirCheckedSubOp, MirCmpOp, MirConstructArrayOp, MirConstructEnumOp, MirConstructStructOp,
-    MirDivOp, MirEqOp, MirExtractFieldOp, MirGeOp, MirGlobalAllocOp, MirGtOp, MirInsertFieldOp,
-    MirLeOp, MirLoadOp, MirLtOp, MirMulOp, MirNeOp, MirNegOp, MirNotOp, MirPtrOffsetOp, MirRefOp,
-    MirRemOp, MirShlOp, MirShrOp, MirSubOp, MirUndefOp,
+    MirCheckedSubOp, MirCmpOp, MirConstantOp, MirConstructArrayOp, MirConstructEnumOp,
+    MirConstructStructOp, MirDivOp, MirEqOp, MirExtractFieldOp, MirGeOp, MirGlobalAllocOp, MirGtOp,
+    MirInsertFieldOp, MirLeOp, MirLoadOp, MirLtOp, MirMulOp, MirNeOp, MirNegOp, MirNotOp,
+    MirPtrOffsetOp, MirRefOp, MirRemOp, MirShlOp, MirShrOp, MirSubOp, MirUndefOp,
 };
 use dialect_mir::types::MirFP16Type;
 use pliron::basic_block::BasicBlock;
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks, the mechanics here are faithful, the slice path mirrors the `PtrMetadata` lowering correctly. But `Rvalue::Len` is unreachable on the toolchain we pin: rustc removed it from internal MIR in 2025, the `rustc_public` conversion never constructs it, and slice lengths reach the importer as `UnOp::PtrMetadata`, which is already handled. So this arm can never run, and code no Rust source can reach would be permanently untestable through the real pipeline.

Closing this along with #455; I'll add a note on the existing bail-out arm documenting why it's unreachable so nobody has to rediscover this. If you'd like a translator gap that's live, the issues labeled codegen have several that real kernels hit today.

PR #457, opened 2026-07-24
fix: lower message-carrying device panics to a trap instead of rejecting them

nh13:
## Problem

A kernel that reaches a panic carrying a message fails the build:

```
error: device code reaches a panic that builds a message string; panic message formatting is not supported on the GPU
    --> library/core/src/slice/mod.rs:1989:21
     |
1989 |             None => panic!("mid > len"),
     |                     ^^^^^^^^^^^^^^^^^^^
     |
     = note: the panic lives in `core::slice::<impl [T]>::split_at_mut`, reached from kernel `...`
```

That covers `panic!("...")`, `expect("...")`, and any core function ending in one, so it blocks porting non-trivial `no_std` code.

## Why it failed

Two things were already true in the tree:

1. The importer **already** drops a diverging call into `core::panicking` and emits `nvvm.trap` + `mir.unreachable` in its place (`terminator/mod.rs`). The panic *terminator* was never the problem.
2. The collector ran a pre-flight (`check_panic_machinery`) that fatally rejected any reachable panic block materializing a string constant, because translating that block was guaranteed to fail.

(2) held only because the block's *statements* were still translated ahead of the terminator, and a materialized `&str` constant has no device lowering. Those statements exist solely to build what the dropped call would have consumed.

## The fix

**`mir-importer`** — `translate_block` recognizes a block whose terminator is a diverging call into a panic entry point and emits the trap alone, skipping statement translation. The block has no successor and the only consumer of those statements is gone, so nothing they write can ever be read.

**`rustc-codegen-cuda`** — the generic message-panic fatal is removed (kept as a `--verbose` note). The `thread::index_*` stub-marker fatal **stays**: that message can only reach device code when a helper is missing `#[device]`, and trapping it would turn a missing annotation into a thread that silently aborts the kernel.

Runtime semantics are unchanged from what cuda-oxide already documents for panics: panic → trap → kernel abort, message discarded.

The importer's panic-entry predicate is documented to stay in sync with the collector's copy, with a unit test asserting the agreement — the collector deliberately does not collect these callees, so the importer must drop exactly the calls it skips.

## Scope

Verified working: `panic!("literal")`, `panic!("{}", value)` (runtime-formatted), and `expect("msg")`. Each lowers to `trap;` with no message string and no `core::fmt` symbol left in the PTX.

`<[T]>::split_at_mut` still does not compile, but no longer because of the panic. It now hits an unrelated pre-existing gap in `split_at_mut_checked`'s `Option<(&mut [T], &mut [T])>`:

```
enum slot map: `Option` has overlapping pointer and non-identical storage at byte 0;
refusing to erase LLVM pointer provenance
```

That looks worth its own issue; happy to file one.

## Testing

| | baseline | with this PR |
|---|---|---|
| `device_panic_trap` | `error: device code reaches a panic that builds a message string` | builds; **2 `trap;`** in PTX; runtime `SUCCESS` |

Runtime on the T4:

```
in_range: PASS
out_of_range: PASS (thread trapped: DriverError(719, "unspecified launch failure"))
SUCCESS
```

The generated PTX contains zero occurrences of the panic message and zero `core::fmt` symbols.

Existing coverage, re-run against the patched backend:

- `slice_index_bounds_check` — SUCCESS, including its own trap path
- `vecadd` — SUCCESS
- `drop_glue` — SUCCESS
- `error_missing_device_attr` — still errors, with the byte-for-byte identical stub diagnostic
- `error`, `error_heap_alloc` — still error with their existing diagnostics

`cargo test -p mir-importer -p mir-lower` passes (1024 tests), including two new unit tests covering the panic-entry predicate and the trap-block shape (trap carries its `v1:i0295` ABI marker, followed by `unreachable`). `cargo fmt` and `cargo clippy --all-targets -- -D warnings` are clean for `mir-importer`, `mir-lower` and `rustc-codegen-cuda`.

## Commits

1. `fix(mir-importer)` — elide the dead statements, add the predicate + unit tests
2. `fix(rustc-codegen-cuda)` — relax the pre-flight, keep the `#[device]` diagnostic
3. `test(examples)` — `device_panic_trap` regression example
4. `docs(book)` — the book said `panic!("message")` was a compile error

---


nihalpasham:
Thanks, LGTM: the message can never print on-device, so lowering the block to a bare trap past the dead formatting statements makes sense here, and keeping the stub-marker hard error while dropping the generic pre-flight is good as well. 

I pushed one follow-up: `is_panic_entry_path` is now exported from mir-importer and the collector imports it, replacing the comment-synced copy, so the two predicates can't drift (the agreement test only covered one of them). 

Compile gates green across the panic and error examples. Will GPU-verify the error-719 path along with this batch. Merging.

nihalpasham (approved):
Validated; see merge comment.

PR #458, opened 2026-07-24
feat(intrinsics-gen): generate scalar math intrinsics family

honeyspoon:
## Summary

Add 37 unary floating-point math intrinsics through the auto-generation
pipeline, covering six PTX operations with approximate and IEEE-rounding
precision modes.

| Operation | PTX base | Precisions | Formats | Variants |
|-----------|----------|------------|---------|----------|
| `sin` | `sin.approx[.ftz].f32` | approx | f32 | 2 |
| `cos` | `cos.approx[.ftz].f32` | approx | f32 | 2 |
| `lg2` | `lg2.approx[.ftz].f32` | approx | f32 | 2 |
| `rcp` | `rcp.{approx,rn,rz,rm,rp}[.ftz].{f32,f64}` | approx+4 rounding | f32, f64 | 14 |
| `rsqrt` | `rsqrt.approx[.ftz].{f32,f64}` | approx | f32, f64 | 3 |
| `sqrt` | `sqrt.{approx,rn,rz,rm,rp}[.ftz].{f32,f64}` | approx+4 rounding | f32, f64 | 14 |

**Before (user code):**
```rust
let result: f32;
unsafe {
    ptx_asm!("sin.approx.f32 %0, %1;", out("=f") result, in("f") x);
}
```

**After (user code):**
```rust
let result = cuda_device::float::sin_approx_f32(x); // Safe, no unsafe needed
```

## Motivation

ML kernels (activation functions, normalization, loss computation) depend
heavily on transcendental math. The PTX ISA offers single-cycle approximate
instructions and IEEE-rounding variants that bypass multi-instruction
libdevice call sequences. Making these available as safe, first-class
intrinsics removes the need for verbose `ptx_asm!` blocks.

## Implementation

Extends the intrinsics generator to support a new `scalar_math` family.

**Generator source** (`cuda-intrinsics-gen`):
- `model.rs`: `ScalarMath` contract with format/operation/precision/subnormal
  enums
- `resolve.rs`: Canonical variant enumeration, selectionless intrinsic
  whitelisting (sin/cos/lg2/rsqrt lack LLVM DAG selection patterns), evidence
  profile resolution, recipe generation
- `render.rs`: Probe LLVM IR emission (typed NVVM intrinsic calls for
  rcp/sqrt; inline PTX assembly for sin/cos/lg2/rsqrt), device API stubs,
  dialect op rendering

**New data files:**
- `overlay/scalar_math.toml`: 37 reviewed variant admissions (ABI IDs
[...]

PR #459, opened 2026-07-24
feat(cuda-device): add safe approximate math intrinsics (tanh, ex2)

honeyspoon:
## Summary

Add two safe, first-class intrinsic functions for hardware approximate
math instructions that bypass multi-instruction libdevice paths:

| Function | PTX instruction | Min SM | Use case |
|----------|----------------|--------|----------|
| `tanh_approx_f32` | `tanh.approx.f32` | sm_75+ | Activation functions |
| `ex2_approx_ftz_f32` | `ex2.approx.ftz.f32` | all | Fast 2^x |

These are single-cycle hardware instructions. No LLVM NVVM intrinsic
exists for `tanh.approx.f32`, so inline PTX is the only path. `ex2`
has an LLVM intrinsic but the approximate+ftz variant requires inline
PTX for precise control.

**Before (user code):**
```rust
let result: f32;
unsafe {
    ptx_asm!(
        "tanh.approx.f32 %0, %1;",
        out("=f") result,
        in("f") x,
        options(register_only),
    );
}
```

**After (user code):**
```rust
let result = cuda_device::approx::tanh_approx_f32(x); // Safe, no unsafe needed
```

## Implementation

Full 4-layer pipeline, following existing hand-written intrinsic patterns:

- `cuda-device/src/approx.rs` — 2 safe `#[must_use]` stubs
- `dialect-nvvm/src/ops/approx_math.rs` — 2 dialect ops with shared
  `verify_unary_f32()` verifier (exact FP32 type checking)
- `mir-importer/src/translator/terminator/intrinsics/approx_math.rs` —
  generic `emit_approx_math_unary::<O: Op>()` dispatch helper
- `mir-lower/src/convert/intrinsics/approx_math.rs` — shared
  `convert_approx_math_unary()` lowering to inline PTX (`AsmKind::Pure`,
  constraints `"=f,f"`)
- `dialect-nvvm/tests/ops_test.rs` — allowlist updated

## Testing

```
cargo check -p cuda-device -p dialect-nvvm -p mir-importer -p mir-lower  # pass
cargo clippy -p cuda-device -p dialect-nvvm -p mir-importer -p mir-lower -- -D warnings  # pass
cargo fmt --all -- --check  # pass
cargo test -p dialect-nvvm  # 50/50 pass
```

Example crate: `examples/approx_math_intrinsic/` — single-thread kernel
computing both intrinsics + fast sigmoid composition, host-side
verification against reference values within 1% tolerance.


nihalpasham:
Thanks for pushing on this, the API is worth having. 

I rebuilt it on the branch through the generated `scalar_math` family rather than the four hand-written layers, since these ops are squarely in that family's scope (it landed in #458 for this class of instruction):

- `cuda_device::float::tanh_approx_f32`, `ex2_approx_f32`, and `ex2_approx_ftz_f32` (note the module and the extra non-ftz ex2 variant, matching the family's naming and ftz conventions; lg2 is the precedent).
- tanh routes through the family's inline-PTX path as a ptx-native source: LLVM 22.1.2's tblgen has no record for it at all, and llc-21 miscompiles the named intrinsic into an extern funcall (llc-22 does select it, so it's promotable when the llc-21 floor drops). ex2 binds to the imported overloaded records, which carry no selection pattern, so it's inline PTX too.
- Targets below the family's attested sm_80 floor now fail at compile time with a clear diagnostic instead of `CUDA_ERROR_INVALID_PTX` at kernel load (tanh's hardware floor is sm_75; the family contract floors at the arch its evidence was probed on, which also covers your sm_70 gap).
- Executed on an RTX 5090: `tanh.approx.f32`, `ex2.approx.f32`, and `ex2.approx.ftz.f32` all appear in the PTX and produce correct values.

Merging.

PR #460, opened 2026-07-24
fix: lower core::hint::select_unpredictable to an llvm.select

nh13:
## Problem

A `#[kernel]` that reaches `core::hint::select_unpredictable` (stable since 1.88, and pulled in pervasively by libcore's branchless helpers — slice sorting, `Ord` combinators) fails device codegen:

```
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Translation failed:
  _R...4core4hint20select_unpredictablej...: ["library/core/src/hint.rs": line: 862, column: 21]
  Unsupported construct: rustc intrinsic `std::intrinsics::select_unpredictable` is not yet supported on the device
```

`select_unpredictable(cond, a, b)` is the branchless form of `if cond { a } else { b }`; it bottoms out in the `core::intrinsics::select_unpredictable` intrinsic, which the importer did not recognize.

## The fix

Lower the intrinsic to an LLVM `select`, using the same three-touchpoint placeholder idiom the other rustc intrinsics use (`rotate_left`, `saturating_add`, …):

- **`dialect-mir`** — add the placeholder name `CALLEE_SELECT_UNPREDICTABLE`.
- **`mir-importer`** — recognize `core::intrinsics::select_unpredictable` in `try_dispatch_intrinsic` and emit a placeholder `mir.call` carrying `(cond, true_val, false_val)`.
- **`mir-lower`** — recognize that placeholder in `convert()` and emit a `pliron_llvm` `SelectOp`.

The Rust `bool` condition already lowers to `i1`, which is exactly what `SelectOp` requires, so no coercion is needed. The result type is taken from `true_val` (equal to `false_val` and the intrinsic's return type). The `unpredictable` branch-weight hint is an optimization hint with no device semantics and is intentionally dropped (a plain `select` is semantically identical).

## Scope

Verified working for scalar (`i32`) and pointer-width (`usize`) selects — the shapes libcore's sort emits. `select_unpredictable` on those lowers through `llvm.select`; NVVM then folds the compare-and-select idioms into `max.s32` / `min.u64` (see PTX below), which is correct.

`core::hint::select_unpredictable` also drags in an internal `DropOnPanic` guard whose generic `drop_in_place` call trips a **separate** pre-existing gap — a call to an empty drop-glue shim the collector does not emit (`Symbol ...drop_in_place... not found`). This PR eliminates the `select_unpredictable` error; the follow-up PR #461 (elide calls to empty drop-glue shims) fixes that second gap, and with both applied the example below runs to completion.

## Testing

Two clean checkouts of `main` (`8d02eac`) on a Tesla T4 (CUDA 13.2, `nightly-2026-04-03`), differing only by this change (plus the empty-drop-shim follow-up, needed for the example to link — see Scope).

| | baseline | with this PR (+ drop-shim follow-up) |
|---|---|---|
| `select_unpredictable` | `error: ... select_unpredictable is not yet supported` | builds; runtime `PASS` |

Runtime on the T4:

```
=== select_unpredictable ===
PASS: select_unpredictable (i32 max, usize min)
```

The generated PTX contains `max.s32` and `min.u64` — the two `select_unpredictable` calls, lowered through `llvm.select` and folded by NVVM — and computes the correct elementwise max/min.

`cargo test -p mir-importer -p mir-lower` passes; `cargo fmt` and `cargo clippy --all-targets -- -D warnings` are clean for the touched crates.

## Commits

1. `feat(dialect-mir)` — add the `select_unpredictable` placeholder name
2. `fix(mir-importer,mir-lower)` — lower `select_unpredictable` to `llvm.select`
3. `test(examples)` — `select_unpredictable` regression example

---


nihalpasham:
LGTM, this matches rustc's own lowering (select, with the CPU-only `!unpredictable` hint correctly drops for NVPTX), and keeping it off the `libdevice-backed` set. 

Validated with #461 applied: the example passes on hardware and llc pattern-matches the selects into single `max.s32`/`min.u64` instructions. 

One documented edge for a follow-up: aggregate-typed T through the value select. Merging now that #461 is in.

nihalpasham (approved):

PR #461, opened 2026-07-24
fix(mir-importer): elide calls to empty drop-glue shims

nh13:
## Problem

A kernel that reaches an explicit `ptr::drop_in_place::<T>` call for a trivially-droppable `T` — the shape `core::hint::select_unpredictable`'s internal generic `DropOnPanic<T>` guard produces — fails at LLVM verification:

```
  Verification failed for lowered LLVM module:
  Symbol _R...4core3ptr13drop_in_placej... not found
```

(`drop_in_place::<usize>`, reached from `core::hint::select_unpredictable`.)

## Why it failed

The device collector deliberately does **not** emit empty drop glue: `crates/rustc-codegen-cuda/src/collector.rs` skips `InstanceKind::DropGlue(_, None)` ("Empty drop glue has no body to collect", ~line 1562). That is correct — an empty shim has no body worth emitting.

But an *explicit* `ptr::drop_in_place::<T>` call (as opposed to a `Drop` terminator) is a normal `Call` terminator. The importer translated it as an ordinary call and emitted a reference to the shim symbol — which the collector never defines. The two sides drifted, exactly the failure mode `mir-importer`'s `drop_glue.rs` documents: the no-op decision must stay in lockstep across the importer, the collector's `process_drop_place`, and the codegen no-op filter, or "emitted calls reference uncollected symbols."

The importer already elides this for the `Drop` *terminator* path (via `drop_glue_is_noop`). It did not for an explicit `drop_in_place` *call*.

## The fix

In `translate_call`, before emitting a call, resolve the callee instance; if `instance.is_empty_shim()` is true, emit only the branch to the target (drop the call).

`is_empty_shim()` is `matches!(def, InstanceKind::DropGlue(_, None))` — **byte-for-byte** the predicate the collector uses to skip. So it is impossible to elide anything with a body or an effect: a type needing drop resolves to `DropGlue(_, Some(_))` (not an empty shim), and user functions are `Item`, not a shim. An empty drop shim does nothing, so branching straight to the successor is semantics-preserving, and this is now in exact lockstep with the collector's skip.

Factored the "emit a goto in place of an elided call" logic into `emit_elided_call_goto`, shared with the existing `precondition_check` elision.

## Reproducing

This fix is independent code (it compiles on `main` on its own). The bug itself, however, is only reached through `core::hint::select_unpredictable` — a *cross-crate* generic from `core` whose generic MIR the device collector takes as-is, so its empty-shim `drop_in_place` call survives. (A locally-defined generic guard with the same shape is optimized away before device codegen and does not reproduce; I tried both a local and a cross-crate synthetic guard and neither reproduces — the specific `MaybeUninit` + `DropOnPanic` construction in `core`'s `select_unpredictable` is what preserves the call.) So the before/after below is demonstrated on top of the companion PR that lowers `select_unpredictable` (**#460**), whose example reaches this gap.

## Testing

On the `select_unpredictable`-lowering branch (companion PR), with vs. without this change, Tesla T4 (CUDA 13.2, `nightly-2026-04-03`):

| | select fix only | + this PR |
|---|---|---|
| `select_unpredictable` example | `error: Symbol ...drop_in_place... not found` | builds; runtime `PASS` |

The generated PTX then contains **zero** `drop_in_place` references, and the elided drop is genuinely a no-op (the guarded value is intact).

`cargo test -p mir-importer -p mir-lower` passes (1022 tests); `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean for the touched crates.

## Commits

1. `fix(mir-importer)` — elide calls to empty drop-glue shims; share the goto-elision helper

---

### DIFF
--- a/crates/mir-importer/src/translator/terminator/mod.rs
+++ b/crates/mir-importer/src/translator/terminator/mod.rs
@@ -881,6 +881,46 @@ fn translate_drop(
 ///                         helpers::emit_function_call
 /// ```
 #[allow(clippy::too_many_arguments)]
+/// Emit a branch to `target` as the only effect of a call we are eliding:
+/// the callee does nothing observable and has no device definition (a UB
+/// precondition check, or an empty `drop_in_place` shim). `emit_goto` needs a
+/// prior op to anchor after, so if the block has none yet we plant a dead
+/// `false` constant first.
+fn emit_elided_call_goto(
+    ctx: &mut Context,
+    target_idx: usize,
+    block_ptr: Ptr<BasicBlock>,
+    prev_op: Option<Ptr<Operation>>,
+    block_map: &[Ptr<BasicBlock>],
+    loc: Location,
+) -> Ptr<Operation> {
+    let anchor = if let Some(p) = prev_op {
+        p
+    } else {
+        use pliron::builtin::attributes::IntegerAttr;
+        use pliron::utils::apint::APInt;
+        use std::num::NonZeroUsize;
+
+        let bool_ty = IntegerType::get(ctx, 1, Signedness::Signless);
+        let dummy = Operation::new(
+            ctx,
+            MirConstantOp::get_concrete_op_info(),
+            vec![bool_ty.into()],
+            vec![],
+            vec![],
+            0,
+        );
+        dummy.deref_mut(ctx).set_loc(loc.clone());
+        let const_op = MirConstantOp::new(dummy);
+        let false_val = APInt::from_u64(0, NonZeroUsize::new(1).unwrap());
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks, I pushed one follow-up: 

- the collector skips every DropGlue callee that `drop_instance_is_noop` proves no-op, not just empty shims, so the elision now gates on the callee being `core::ptr::drop_in_place` and consults that same shared predicate (its fast path covers is_empty_shim). 
- Also restored `translate_call`'s doc block, which the new helper had been inserted into. 

Validated together with #460 on hardware. Merging.

nihalpasham (approved):

PR #462, opened 2026-07-24
fix: lower core::intrinsics::copy (ptr::copy) to llvm.memmove

nh13:
## Problem

A `#[kernel]` that reaches `core::ptr::copy` (the overlap-safe move — used by `ptr::swap`, slice rotates, and similar routines) fails device codegen:

```
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Translation failed:
  ...: ["library/core/src/ptr/mod.rs": line: 643, column: 9]
  Unsupported construct: rustc intrinsic `std::intrinsics::copy` is not yet supported on the device
```

Unlike `copy_nonoverlapping` — which reaches MIR as a `StatementKind::Intrinsic(CopyNonOverlapping)` and already lowers to `llvm.memcpy` (`examples/copy_nonoverlapping`) — `core::ptr::copy` bottoms out in the `core::intrinsics::copy` *intrinsic call*, which has no statement form and was unhandled.

## The fix

Lower `core::intrinsics::copy` to the overlap-safe `llvm.memmove`:

- **`dialect-mir`** — add `MirMemmoveOp`, a twin of `MirMemcpyOp` (same `(dst, src, count)` operands, same verifier), registered alongside it.
- **`mir-lower`** — factor `convert_memcpy`'s body into a shared `convert_mem_transfer(base)` (element→byte count scaling, `p<dst>.p<src>.i<bits>` overload naming, `isvolatile=false`) and add `convert_memmove` that selects the `memmove` intrinsic family; wire `MirMemmoveOp` through `MirToLlvmConversion`.
- **`mir-importer`** — intercept `core::intrinsics::copy` in `try_dispatch_intrinsic` and emit a `MirMemmoveOp`, **reordering** rustc's intrinsic argument order `(src, dst, count)` to the op's `(dst, src, count)`. Modeled on the existing `emit_typed_swap` (unit result + branch to target).

`copy_nonoverlapping` is intentionally left alone — it is already covered by the statement path; only the overlapping `copy` needed the overlap-safe `memmove`.

## Scope

Verified working for `core::ptr::copy` with a forward-overlapping destination (`dst = src + 1`). Overlap-safety is exercised directly: a plain `memcpy` would corrupt the forward shift, so a correct result requires `memmove` **and** the correct `(src, dst)` argument order.

## Testing

| | baseline | with this PR |
|---|---|---|
| `ptr_copy` | `error: ... intrinsics::copy is not yet supported` | builds; runtime `PASS` |

Runtime on the T4:

```
=== ptr_copy ===
PASS: ptr_copy (overlapping forward move via memmove)
```

The kernel loads `input` into `out`, then does `ptr::copy(p, p.add(1), n-1)` — a forward-overlapping shift-right-by-one — and the host asserts `out[k] == input[k-1]`. This passes only if the move is overlap-safe and the pointers are in the right order.

`cargo test -p mir-importer -p mir-lower -p dialect-mir` passes; `cargo fmt` and `cargo clippy --all-targets -- -D warnings` clean for the touched crates.

## Commits

1. `feat(dialect-mir)` — add `MirMemmoveOp`
2. `fix(mir-lower)` — share memcpy/memmove lowering; add `convert_memmove`
3. `fix(mir-importer)` — lower `core::intrinsics::copy` to `mir.memmove`
4. `test(examples)` — `ptr_copy` overlapping-move regression example

---


nihalpasham:
Thanks.

I pushed two small follow-ups: 
- the shared conversion's error strings said `memcpy` on the `memmove` path (now parameterized by the op), and a doc/attribute placement fix (the new arm had displaced `typed_swap`'s doc block). 
- ran the example on an RTX 5090: overlapping shift produces the right result where a memcpy would corrupt. Plus a formatting pass on the example crate. 
 
Merging.

nihalpasham (dismissed):

nihalpasham (approved):

Issue #413, opened 2026-07-20
Preserve Rust reference kind and exclusivity through MIR import and lowering


nihalpasham:
## Summary

cuda-oxide currently represents Rust references and raw pointers identically once their pointee type, source mutability, and address space match:

```text
&mut T   -> mir.ptr<T, mutable: true, ...>
*mut T   -> mir.ptr<T, mutable: true, ...>

&T       -> mir.ptr<T, mutable: false, ...>
*const T -> mir.ptr<T, mutable: false, ...>
```

Both [`RigidTy::Ref` and `RigidTy::RawPtr`](https://github.com/NVlabs/cuda-oxide/blob/6a92f548f25899352e47e173bfd387b2cd0d1e18/crates/mir-importer/src/translator/types.rs#L525-L535) use the same pointer translation. [`MirPtrType`](https://github.com/NVlabs/cuda-oxide/blob/6a92f548f25899352e47e173bfd387b2cd0d1e18/crates/dialect-mir/src/types.rs#L101-L104) records the pointee, a mutability bit, and the address space, but not whether the value originated as a raw pointer, shared reference, or unique reference.

The mutability bit cannot by itself mean “exclusive”: Rust permits aliasing `*mut T` pointers, while `&mut T` carries unique-borrow requirements. MIR `Retag` statements are also currently skipped, and MIR-to-LLVM pointer conversion retains only the address space.

## Current impact

We do **not** currently have a demonstrated miscompile from this information loss.

The LLVM exporter intentionally does not emit Rust-derived `noalias`, `readonly`, `dereferenceable`, or related pointer attributes. Current lowering also does not use `MirPtrType::is_mutable` as proof of exclusivity. LLVM therefore handles these pointers conservatively today.

Dropping exclusivity information is not itself incorrect for valid programs while every downstream pass remains conservative. The gap becomes actionable if we later:

- emit `noalias` or other Rust-derived LLVM attributes;
- introduce alias-aware optimizations;
- verify reference use in the MIR dialect;
- rely on ordinary Rust reference rules when lowering GPU-specific APIs.

## Motivating example from PR #350

PR [#350](https://github.com/NVlabs/cuda-oxide/pull/350) contains this shared-memory access:

```rust
static mut TILE: SharedArray<[f32; N], 1> = SharedArray::UNINIT;

let i = thread::index_1d().get();

if i < N {
    unsafe { TILE[0][i] = input[i] };
}
```

[...]

PR #463, opened 2026-07-24
fix(mir-importer): coerce from_raw_parts data pointer to the slice element type

nh13:
## Problem

A `#[kernel]` that builds a slice from a **reinterpret-cast raw pointer** fails device codegen at lowering:

```
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Lowering failed:
  MirConstructSliceOp data pointer pointee mismatch.
  Expected: mir.tuple <[u64, u64], ...>, Actual: builtin.integer ui64
  v38 = mir.construct_slice (v36, v37) ...
    <(mir.ptr <ui64, mutable:true, addrspace:0>, u64) -> (mir.slice <mir.tuple <[u64, u64], ...>>)>
```

The trigger is the common type-punning idiom — building a `[(u64, u64)]` slice from a `*mut u64`:

```rust
let pairs = core::slice::from_raw_parts_mut(buf.as_mut_ptr() as *mut (u64, u64), n);
```

(In a real kernel this shows up as e.g. `state.z.add(...) as *mut (u64, u64)` feeding the chain/hitproc spine.)

## Why it failed

`ptr::from_raw_parts`/`from_raw_parts_mut` build the `[T]` fat pointer from a thin data pointer plus a length. The `p as *mut (u64, u64)` cast is a provenance- and address-preserving `PtrToPtr`, so rustc's MIR pipeline propagates the **pre-cast** SSA pointer (`*mut u64`) into the inlined `aggregate_raw_ptr` while the slice element type comes from the monomorphized signature (`(u64, u64)`). The `Rvalue::Cast` lowering itself is correct (it types the cast result from the target type); the stale pointee is an internal type-tracking imprecision at the fat-pointer construction site, not a user type error. `mir.construct_slice`'s verifier then rejects the data-pointer/element mismatch.

The from_raw_parts lowering already normalizes the data pointer's **address space** (`cast_to_generic_addrspace_if_needed`), but that helper only fires when the pointee already matches — so a pointee mismatch slips straight through to the verifier.

## The fix

When the data pointer's pointee differs from the slice element type, coerce it with a `PtrToPtr` cast to a **generic-address-space** pointer to the element type — the same `get_generic` data slot every other fat-pointer construction in the file uses (a no-op when the pointee already matches). Because the target is the generic address space, this also composes with the address-space normalizer for the shared-memory case: a `shared.as_ptr() as *const (u64, u64)` reinterpret (address space 3) is coerced to a generic-addrspace `(u64, u64)` pointer in one cast, rather than leaving a non-generic data slot the fat pointer forbids.

## Testing

| | baseline | with this PR |
|---|---|---|
| `slice_from_raw_parts_cast` | `Lowering failed: MirConstructSliceOp data pointer pointee mismatch` | builds; runtime `PASS` |

The example reinterprets a `u64` buffer as `(u64, u64)` pairs via a pointer cast + `from_raw_parts`, in **global memory (address space 0)** and **shared memory (address space 3)**, and the host checks both. Runtime on the T4:

```
=== slice_from_raw_parts_cast ===
PASS: slice_from_raw_parts_cast (global + shared reinterpret)
```

`cargo test -p mir-importer` passes; `cargo fmt --check` and `cargo clippy -p mir-importer --all-targets -- -D warnings` are clean.

Scope note: the other `MirConstructSliceOp` site (the fat-pointer-field path) builds its data pointer via `MirFieldAddrOp` from the element type, so its pointee matches by construction and it is intentionally untouched.

## Commits

1. `fix(mir-importer)` — coerce the from_raw_parts data pointer to the slice element type
2. `test(examples)` — `slice_from_raw_parts_cast` regression example (global + shared)

---


nihalpasham:
Thanks for the PR. 

**Gist:** The slice builder trusted the data pointer's declared pointee, so a `from_raw_parts` over a casted pointer built a slice of the wrong element type; re-typing at the build site is the right fix and the regression example pins it.

Nothing to change beyond a formatting pass on the example crate. 

Verified on an RTX 5090: the (u64, u64) pair readback from shared memory comes back correct. Related to but not closing #413 (reference-kind provenance is a different, bigger thing). Merging.

nihalpasham (approved):

PR #465, opened 2026-07-24
docs(inline_ptx): add ptx_asm! constraint and syntax reference

honeyspoon:
## Summary

Add a `ptx_asm!` quick-reference section to the `inline_ptx` example's README
covering operand constraints, options, clobbers, and usage examples.

## Motivation

Users writing inline PTX via `ptx_asm!` currently need to discover the
constraint syntax (`"r"`, `"=f"`, `"l"`, etc.), options (`register_only`,
`may_diverge`), and clobber declarations (`"memory"`) from source code or
trial-and-error. A concise reference in the example README makes the macro
self-documenting for common use cases.

## Contents

- **Syntax**: template placeholder format (`%N`, `%%reg`)
- **Operand constraint table**: all 14 constraint strings mapped to PTX
  register class, direction, and Rust type
- **Options table**: `register_only`, `may_diverge`
- **Clobber table**: `"memory"`
- **Examples**: integer add, float approx math, multi-output, memory barrier

### DIFF
--- a/crates/rustc-codegen-cuda/examples/inline_ptx/README.md
+++ b/crates/rustc-codegen-cuda/examples/inline_ptx/README.md
@@ -15,3 +15,112 @@ Run with:
 ```bash
 cargo oxide run inline_ptx
 ```
+
+## `ptx_asm!` Reference
+
+### Syntax
+
+```rust
+unsafe {
+    ptx_asm!(
+        "<template>",
+        out("<constraint>") <place>,   // output operands (0..8)
+        in("<constraint>") <expr>,     // input operands (0..16)
+        clobber("<name>"),             // side-effect declarations
+        options(<option>, ...),        // assembly options
+    );
+}
+```
+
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks. 

I pushed one fix on top:

- The `may_diverge` row had the semantics inverted: it is not a warning that the asm may diverge, it is a promise that the snippet is safe to move across divergent control flow (it drops the LLVM `convergent` guarantee). The old wording would invite using it on `.sync`/`collective` snippets, which is exactly where it miscompiles. Reworded to match the cuda-macros `README`, with the warning against lane-sensitive code.
- Operand limits are inclusive: up to 8 outputs and 16 inputs, not 0..8/0..16.
- Linked the canonical references so this copy cannot drift, and added notes on the raw `clobber("~{...}")` escape hatch and the `sm_75+` floor for `tanh.approx.f32`.

nihalpasham (approved):

PR #466, opened 2026-07-24
fix(mir-lower): lower rounding intrinsics to LLVM instead of libdevice

honeyspoon:
## Summary

Lower Rust math intrinsics with direct LLVM equivalents to their native
LLVM intrinsic names instead of routing through libdevice. The NVPTX
backend lowers each to a single PTX instruction, avoiding the libdevice
function-call overhead.

Partial fix for #390.

## Operations routed to LLVM intrinsics

| Rust intrinsic | LLVM intrinsic | PTX instruction |
|---------------|----------------|-----------------|
| `f32::abs()` / `f64::abs()` | `llvm.fabs.*` | `abs.f32` / `abs.f64` |
| `f32::copysign()` / `f64::copysign()` | `llvm.copysign.*` | `copysign.f32` / `copysign.f64` |
| `f32::floor()` / `f64::floor()` | `llvm.floor.*` | `cvt.rmi.f32.f32` |
| `f32::ceil()` / `f64::ceil()` | `llvm.ceil.*` | `cvt.rpi.f32.f32` |
| `f32::trunc()` / `f64::trunc()` | `llvm.trunc.*` | `cvt.rzi.f32.f32` |
| `f32::round()` / `f64::round()` | `llvm.round.*` | round-ties-away |
| `f32::round_ties_even()` | `llvm.roundeven.*` | `cvt.rni.f32.f32` |

## Motivation

These operations are pure register-to-register transforms with exact LLVM
intrinsic equivalents. Routing them through libdevice adds an unnecessary
function call and forces the kernel onto the heavier libNVVM/nvJitLink
pipeline even when no transcendental math is used.

For rounding specifically, this eliminates the need for hand-written
IEEE-754 bit tricks (e.g., a ~10-instruction `round_ties_away()` function)
in user kernels -- `x.floor()`, `x.ceil()`, etc. now compile to single
PTX instructions.

## Deliberate exclusions

`max`/`min` are intentionally left on the libdevice path. LLVM 21's
`maxnum`/`minnum` propagate signaling NaNs, which contradicts Rust's
contract (`f32::max` returns the non-NaN operand). These need
compare/select lowering instead, as addressed by #391.

## Implementation

Added `RustFloatMathIntrinsic::llvm_intrinsic_name()` in
`crates/mir-lower/src/convert/ops/call.rs` which returns the width-qualified
LLVM intrinsic name for ops that can bypass libdevice. The existing
`convert_rust_float_math_intrinsic()` checks this method first, falling
through to `libdevice_name()` only for transcendentals (sin, cos, exp,
log, pow, etc.) that have no single-instruction PTX equivalent.

## Testing

```
cargo test -p mir-lower --lib   # 136 tests pass
cargo fmt --all -- --check      # clean
```

Unit test `test_llvm_intrinsic_name_routing` verifies:
- fabs returns `llvm.fabs.f32` / `llvm.fabs.f64`
- copysign returns `llvm.copysign.f32` / `llvm.copysign.f64`
- All 10 rounding ops (floor/ceil/trunc/round/roundeven x f32/f64) return correct LLVM names
- `MaxNumNsz`/`MinNumNsz` return `None` (libdevice path, per #390)
- Transcendentals (`SinF32`, `SqrtF64`) return `None` (libdevice path)


honeyspoon:
Note: there is overlap between this PR and #391 from @niklebedenko, which takes a broader approach (also covers `max_nsz` and `min_nsz` via ordered compare/select). If #391 is preferred, this PR can be closed — the fabs/copysign lowering approach is the same in both.

honeyspoon:
Narrowed to rounding only, to remove the overlap with #391.

#391 moves `fabs` and `copysign` off libdevice as part of a broader change that also handles `max`/`min`, so it is the more general treatment of the part we both touched. I have dropped `fabs` and `copysign` from this PR; it now covers only `floor`, `ceil`, `trunc`, `round`, and `roundeven`, which #391 does not touch. The two should no longer conflict, and the test here asserts `fabs`/`copysign` return `None` so the boundary fails loudly if that changes.

While measuring, one correction to what this PR originally claimed. It said every routed intrinsic lowers to a single PTX instruction and listed `round` among the `cvt` conversions. That is wrong for `round`. With LLVM 21.1 `llc -march=nvptx64 -mcpu=sm_80`, body only:

```
floor      -> cvt.rmi             1
ceil       -> cvt.rpi             1
trunc      -> cvt.rzi             1
roundeven  -> cvt.rni             1
round      -> expanded sequence   11 (f32), 8 (f64)
```

`round` rounds halfway cases away from zero, and that is not one of the four hardware `cvt` modes (`rni` is ties-to-even), so no single instruction can express it. It is still worth routing here: LLVM expands it inline into compare/select over `cvt.rzi` rather than emitting a `roundf` libcall, so the libdevice call goes away without leaving an unresolved symbol. The table is now in the doc comment so the next reader does not have to re-measure.

Happy to rebase on top of #391 once it lands if you would prefer these stacked.

nihalpasham:
Thanks, this makes sense. The PTX mapping table in the PR is quite useful. As diffed it could not compile a rounding kernel though, so I pushed four commits on top:

- pliron identifiers reject dots, so the dotted names (`llvm.floor.f32`) failed with MalformedIdentifierErr before anything ran. The table now uses the repo's underscore encoding (`llvm_floor_f32`), which the exporter decodes back to the dotted name (same mechanism as `llvm_ctpop_i*`).
- The pre-lowering libdevice predictor still classified the ten rounding placeholders as libdevice-backed, so every rounding-only kernel died on the pipeline consistency check. They are now a separate backend-dependent class.
- The lowering dispatches on the intrinsic backend: the llc path gets `llvm.*` (single cvt instructions), the LibNvvm path keeps `__nv_*` (the legacy LLVM 7 dialect has no `llvm.roundeven`).
- Added a rounding-only float_rounding example pinning tie-breaking and signed zero on device.

Verified on an RTX 5090: rounding-only kernels now build self-contained PTX with zero libdevice, `--emit-nvvm-ir` still compiles via libNVVM with `__nv_*` rounding, and `libm_math`/`primitive_stress` pass. Related to #390; #391 owns the max/min move and now has the backend-aware pattern to follow.

nihalpasham (approved):

PR #467, opened 2026-07-24
fix(intrinsics-gen): guard duplicate cast_to_shared_addrspace import

honeyspoon:
## Summary

Prevents a duplicate `cast_to_shared_addrspace` import in the generated `generated_intrinsics.rs` when both the `stmatrix` and `cluster_memory` families are present in the catalog.

## Cause

The `stmatrix` and `cluster_memory` lowering blocks in `render_lowering()` both insert `cast_to_shared_addrspace` into the import list via `str::replace`. The `mbarrier_extended` block already guards against this with `if !output.contains("cast_to_shared_addrspace")`, but the `cluster_memory` block does not.

When both families are present (as they are in the current catalog), the import appears twice:

```rust
use crate::convert::intrinsics::common::{
    call_intrinsic, cast_to_shared_addrspace, cast_to_shared_addrspace, create_i32_const,
    //                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^ duplicate
};
```

Nightly `rustfmt` with `--edition 2024` silently deduplicates the import, so CI passes. Stable `rustfmt` preserves the duplicate, causing `cargo run -p cuda-intrinsics-gen -- check` to report a stale file on systems where stable `rustfmt` is on `PATH`.

## Fix

Add the same `if !output.contains("cast_to_shared_addrspace")` guard to the `cluster_memory` block, matching the existing pattern used by the `mbarrier_extended` block.

## Testing

- `cargo test -p cuda-intrinsics-gen` — 281 passed
- `cargo run -p cuda-intrinsics-gen -- check` — passes with both nightly and stable rustfmt
- `cargo clippy -p cuda-intrinsics-gen -- -D warnings` — clean

### DIFF
--- a/crates/cuda-intrinsics-gen/src/render.rs
+++ b/crates/cuda-intrinsics-gen/src/render.rs
@@ -21788,6 +21788,55 @@ mod tests {
     }
 
     #[test]
+    fn lowering_imports_cast_to_shared_addrspace_exactly_once_across_families() {
+        // The stmatrix, cluster_memory, and mbarrier_extended families all
+        // insert the `cast_to_shared_addrspace` helper into the same
+        // `common::{call_intrinsic, ...}` use group. An unguarded second
+        // insert duplicates the name inside one use group, which is a hard
+        // rustc error (E0252) in the raw generator output. The rustfmt pass
+        // run over generated files silently dedupes the duplicate, so assert
+        // on the raw (pre-rustfmt) output to catch a regressed guard loudly.
+        let repo_root = Path::new(env!("CARGO_MANIFEST_DIR")).join("../..");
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM. Confirmed the raw generator output on main really does emit the duplicate import with both families in the catalog (`E0252`; only rustfmt's silent dedup hides it).

I pushed one commit: 
- a regression test rendering a catalog with both `stmatrix` and `cluster_memory`, asserting the import appears exactly once in the raw pre-rustfmt output. 
- It fails with the guard reverted (`left: 2`, `right: 1`) and also catches the opposite failure, over-suppression. 

Verified: 282 generator tests pass, `check` is current, `generate` produces zero drift.

nihalpasham (approved):

PR #470, opened 2026-07-24
fix(mir-importer): lower assignments to arr[i].field (2-level Index -> Field places)

nh13:
## Problem

Assigning through a two-projection place whose first element is an array/slice `Index` (or `ConstantIndex`) and whose second is a struct `Field` — `arr[i].field = value` for a local `[Struct; N]` — fails to lower:

```
2-level projection Index(_) -> Field(_, _) not yet implemented for assignment
```

This is common Rust. It blocks, for example, a straightforward FM-index extension that writes `ok[i].size = size[i]` where `ok: [Sai; 4]`.

## Why it failed

`translate_statement` handles a place with exactly two projections via a `match (&place.projection[0], &place.projection[1])`. That match enumerated `(Deref, Field)`, `(Field, Field)`, `(Deref, Index|ConstantIndex)`, `(Index, Index)`, and `(Field, ConstantIndex|Index)` — but had **no arm for `(Index, Field)` / `(ConstantIndex, Field)`**, so those places fell through to the `_ =>` unsupported-construct error.

## The fix

Add the missing arm, delegating to `store_through_place_address` — the same address-walk-and-store helper the sibling index arms (`(Deref, Index)`, `(Index, Index)`, `(Field, Index)`) and the 3+‑projection fallback already use, and which `Rvalue::Ref` uses to take `&place` for an arbitrary projection chain. It walks the full `Index -> Field` chain to a destination address and emits `mir.store` through it. If the walker cannot form an address it errors loudly (`cannot compute the destination address …`) rather than miscompiling. The assignment target is always a mutable place, so the helper's `is_mutable: true` request is correct.

24-line change, mirroring the existing index arms.

## Testing

New self-contained example `examples/index_field_assign` — each of 256 threads fills a local `[Cell; 4]` via `arr[i].a`/`arr[i].b` (runtime `Index -> Field`) and bumps `arr[0].a` (`ConstantIndex -> Field`), then reduces the array so the **host verifies every thread's writes actually landed** (`14*tid + 106`), not merely that codegen succeeded.

| | Baseline (`8d02eac`) | This PR |
|---|---|---|
| `cargo oxide run index_field_assign` | ✗ `2-level projection ConstantIndex { .. } -> Field(..) not yet implemented for assignment` | ✓ `PASS` (runs on a T4, host-verified) |

Reverting only the new arm reintroduces the failure; reapplying fixes it.

- `cargo test -p mir-importer` → **814 passed, 0 failed**
- `rustfmt --check` on the changed file → clean
- `cargo clippy -p mir-importer` → clean for this change

Beyond the example, this arm is what lets the full minibwa seeding and chaining GPU kernels compile to PTX (they reach `arr[i].field` through the shared FM-index).

## Commits

1. `fix(mir-importer): lower assignments to arr[i].field (2-level Index -> Field places)`
2. `test(examples): add index_field_assign regression example`

---


nihalpasham:
LGTM, delegating to `store_through_place_address` is right, and the generated IR shows the bounds check surviving as an `icmp+trap` guard in front of the `GEP` chain.

I force-pushed a rebase onto current main plus one comment-only commit: the example claimed array indexing desugars to a bounds-checked `index_mut` call, but built-in `[T; N]` indexing emits a MIR Assert terminator plus a direct Index projection; `IndexMut` is never involved. 

Verified: 857 mir-importer tests pass and `index_field_assign` passes on an RTX 5090 (`14*tid+106` across all 256 threads). Related to #397.

nihalpasham (approved):

Issue #399, opened 2026-07-14
Scalarize small local arrays consumed through iterator adapters


niklebedenko:
## Problem

Small fixed-size local arrays can still become NVPTX local memory when ordinary Rust iterator adapters carry their address through a loop.

A representative shape is:

```rust
let coefficients = [4.0_f32, -3.0, 4.0 / 3.0, -0.25];
for (index, coefficient) in coefficients.iter().copied().take(runtime_levels.min(4)).enumerate() {
    consume(index, coefficient);
}
```

After inlining and O2, LLVM retains an `alloca [4 x float]` and pointer-based iterator state. ScalarEvolution reports the loop trip count and constant maximum as unpredictable, so bounded `loop-unroll` with `upperbound` or `runtime` enabled does not expose the array to SROA. This produces `ld.local` / `st.local` despite the four-element bound.

## Relationship to #397 / #398

#398 scalarizes direct runtime indexing and selected addresses for small compiler-owned arrays. Iterator pointer state is a separate case because the eventual load is visible only after helper inlining and its pointer provenance has passed through loop state.

## Desired result

Keep small arrays used through standard `iter`, `copied`, `take`, `skip`, and `enumerate` chains in SSA without requiring source-level loop expansion or indexing rewrites. Add a GPU regression asserting correct runtime-level behavior and no PTX `.local`, `ld.local`, or `st.local`.

nihalpasham:
I built your repro on current main (nightly-2026-08-28, LLVM 23), and the picture has flipped since July.

- `.local` here means per-thread stack memory in PTX (`ld.local` / `st.local`).
- Your iterator form now compiles to register-only code. The index loop with a `break`, the form people usually reach for instead, is the one that keeps the array in memory.

```rust
let coefficients = [4.0_f32, -3.0, 4.0 / 3.0, -0.25];
let n = levels[i] as usize;

// A: iterator adapters (this issue)   -> 0 ld/st.local, 50 PTX instructions
for (k, c) in coefficients.iter().copied().take(n.min(4)).enumerate() { acc += c * (k as f32 + 1.0); }

// B: index loop with break            -> 4 st.local + 1 ld.local, 62 PTX instructions
for k in 0..4 { if k >= n { break; } acc += coefficients[k] * (k as f32 + 1.0); }
```

[...]

PR #471, opened 2026-07-24
fix(mir-lower): emit a distinct GEP for ZST struct fields in convert_field_addr

nh13:
## Problem

Compiling a device kernel that does `iter().map(...).sum()` fails during lowering:

```
field_addr index 1 out of bounds for struct with 0 fields
```

The offending mono-instance is `core`'s composed `map_fold` closure. `Iterator::sum` folds with `Sum::sum`'s closure, and `Map::fold` composes closures via `map_fold` = `move |acc, elt| g(acc, f(elt))`, which captures the map closure `f` and the `Sum::sum` fold closure `g` as upvars. Both are zero-sized, so the composed closure is a **ZST struct with two ZST fields**, and its body borrows *both* upvars off the **same base pointer**.

## Root cause

`convert_field_addr` recovers the base pointer's pointee type from dialect conversion's per-value pointer-type *history* (`lookup_most_recent_of_type::<MirPtrType>`), because the operand's own type has already been converted to an opaque `llvm.ptr`.

Its ZST-field branch forwarded the base SSA value directly as the field-address result:

```rust
Some(None) => {
    rewriter.replace_operation_with_values(ctx, op, vec![ptr_operand]);
    return Ok(());
}
```

That substitutes a `ptr<composed>` value for a result that was typed `ptr<field>`, **poisoning the base value's recorded pointer-type history** with the field's pointee. So when the *sibling* `field_addr` for field 1 looks up its base pointee, it gets the wrong type — field 0's pointee, a zero-field ZST closure — and the index-1 lookup is out of bounds. A struct with a single ZST field never hits this (no sibling access to corrupt).

## The fix

Emit an explicit zero-offset GEP (a **distinct** result) for the ZST field instead of aliasing the base SSA value — **exactly what the `union` branch immediately above already does, and documents, for this same reason**. Physically identical (byte 0 of the struct is the address of a zero-sized field), but the field_addr no longer overwrites the base pointer's recorded type, so sibling accesses resolve correctly. The ZST-**struct**-field branch was simply never given the treatment its `union` sibling already had.

## Testing

New self-contained `examples/map_sum` — an `i64` and a `usize` `iter().map(..).sum()` kernel (both element types that occur in practice), each writing `out[tid]` with the **host verifying the reduction** (fails on a wrong value, not just a codegen abort).

| | Baseline (`fe4791d`) | This PR |
|---|---|---|
| `cargo oxide run map_sum` | ✗ `field_addr index 1 out of bounds for struct with 0 fields` | ✓ `PASS` (runs on a T4, host-verified) |

Reverting only the fix reintroduces the failure.

- `cargo test -p mir-lower` → **211 passed, 0 failed**
- `rustfmt --check` on the changed file → clean
- `cargo clippy -p mir-lower` → clean

The fix also lets a full downstream codebase (minibwa's seeding + chaining GPU kernels, which reach `iter().map(..).sum()` on both `i64` and `usize`) compile to PTX without hand-rewriting the sums into fold loops.

## Commits

1. `fix(mir-lower): emit a distinct GEP for ZST struct fields in convert_field_addr`
2. `test(examples): add map_sum regression example`

---


nihalpasham:
LGTM, and thanks for the clear diagnosis: the `ZST` alias poisoning the base pointer's conversion type history is subtle, and the `zero-offset GEP` mirrors what the union branch already does for the same reason.

I pushed two commits:

- Rewrote the `convert_field_addr` rustdoc, which still described the old forwarding and would have invited restoring the bug.
- Added a unit test taking addresses of two sibling `ZST` fields off one base; with the fix reverted it fails with the exact original "`field_addr` index 1 out of bounds for struct with 0 fields" error.

Verified: 212 mir-lower tests pass and map_sum passes on an RTX 5090. Related to #399 and #303

nihalpasham (approved):

PR #472, opened 2026-07-24
fix(mir-lower): back every payload pointer leaf with its own ptr slot in enum slot map

nh13:
## Problem

A niche-`Option` whose payload carries **more than one pointer** fails to lower:

```
enum slot map: `...` has overlapping pointer and non-identical storage at byte N; refusing to erase LLVM pointer provenance
```

The concrete trigger is `<[T]>::split_at_mut_checked`, which returns `Option<(&mut [T], &mut [T])>` — a payload of two fat slice pointers `{ptr, len, ptr, len}` with the `None` niche in the first data pointer. This blocks any device kernel that splits a mutable slice.

## Root cause

`build_enum_slot_map` models an enum's storage as a flat LLVM struct of byte "claims". For pointer **provenance** to survive the memory round-trip (`convert_construct_enum` stores the payload at its byte offset, then loads the enum-struct type back), every pointer leaf must be backed by a real `ptr` slot at its offset — a pointer that lands in `[N x i8]` filler loads back as integer bytes with no provenance.

The slot map backed at most **one** pointer with a `ptr` slot: the niche carrier. The old `pointer_storage_matches_claims` gate required the payload's pointer leaves to *exactly* equal the overlapping claims' pointer leaves, so a second pointer leaf (with no slot) failed closed. That correctly refused to strip provenance, but also rejected layouts that are perfectly representable — the second pointer just needs its own slot.

## The fix

Replace the exact-equality gate with `analyze_pointer_overlap`, which admits a pointer-bearing payload when:
- every existing pointer leaf overlapping the field is also carried by the field at the same offset/size/address space, **and**
- each remaining ("extra") pointer leaf lands on bytes no other claim covers (a gap).

It returns those extra leaves, and the call site adds **one fresh `ptr` slot per extra leaf**. So `Option<(&mut [T], &mut [T])>` lowers to `{ptr@0, [pad], ptr@16, [pad]}` — both slice pointers backed, provenance intact, the payload still slotless and round-tripping through memory.

A pointer punned against **non-pointer** bits at the same bytes (a genuine `ptr`/`int` union) still fails closed — there is no single LLVM slot type that is both provenance-carrying and integer-exact. That case remains a separate, deliberately-unsolved design question.

## Testing

- `cargo test -p mir-lower` → **136 passed, 0 failed**. Two enum-slot-map tests that asserted a conservative reject for *representable* layouts (a pointer beside a pointer carrier; a pointer beside an integer niche carrier) now assert the accepted `{ptr, …}` shape; a new `enum_slot_map_backs_split_at_mut_slice_pair` covers the exact `Option<(&mut [T], &mut [T])>` layout. The genuine `ptr`/`int` union tests still reject.
- New `examples/split_at_mut` — thread 0 splits a buffer via `split_at_mut_checked`, bumps both halves through the two returned slices, and the **host verifies both** (a dropped/provenance-stripped second pointer fails the check, not just a codegen abort). Runs on a T4:

| | Baseline (`fe4791d`) | This PR |
|---|---|---|
| `cargo oxide run split_at_mut` | ✗ `refusing to erase LLVM pointer provenance` | ✓ `PASS` (host-verified) |

- `rustfmt --check` clean; `cargo clippy -p mir-lower` clean.
- Downstream: this unblocks the last remaining minibwa kernel (`spine_hitproc`, which calls `split_at_mut_checked`) compiling to PTX.

## Commits

1. `fix(mir-lower): back every payload pointer leaf with its own ptr slot in enum slot map`
2. `test(examples): add split_at_mut regression example`

---


nihalpasham:
LGTM, this was quite the find - whichever side of a genuine `ptr`/`int` pun is placed first, the other side still rejects.

Verified on an RTX 5090: `split_at_mut` passes; `error_enum_pointer_overlap` still fails compilation with the preserved provenance diagnostic; the other five enum-layout examples pass; and after `opt -O2`, SROA promotes both slice-half pointers to phi ptr values with zero `ptrtoint`/`inttoptr` left in the module. Related to #306.

nihalpasham (approved):

PR #473, opened 2026-07-25
feat: implement core::intrinsics::exact_div (unblocks slice::as_chunks)

honeyspoon:
`core::intrinsics::exact_div` is unimplemented in the codegen, and that one gap blocks `slice::as_chunks`:

```
Translation failed: core::slice::as_chunks::<4>
  [core/src/slice/mod.rs:1345:32] Compilation error: invalid input program
```

Line 1345 is `exact_div(self.len(), N)`. So the idiomatic *safe* way to read N adjacent elements does not compile in device code at all, and kernels fall back to raw pointer reads. Supporting this unlocks `as_chunks`, `as_rchunks`, and `as_chunks_mut`.

## Lowering

The contract is that the divisor is non-zero and divides the dividend with no remainder — both caller obligations, and violating either is UB. That is exactly the contract of a bare LLVM division, so this emits `llvm.sdiv` / `llvm.udiv` with **no** zero check and **no** remainder fixup.

Two details worth review:

- **Signedness** comes from the pre-conversion MIR operand type, the same source `mir.div` uses, because the LLVM integer type is signless by the time the operand is rewritten. Signless falls back to unsigned, matching `convert_div`.
- **Width**: a narrower divisor is cast to the dividend type first, so a `usize` dividend against a `u32` count works.

## Layers

Follows the existing `saturating` intrinsic structure:

| Layer | Change |
|---|---|
| `dialect-mir` | `CALLEE_EXACT_DIV` placeholder, registered in the non-libdevice list |
| `mir-importer` | new `intrinsics::exact_div` module + terminator dispatch |
| `mir-lower` | placeholder recognition + conversion to sdiv/udiv |

## Verification

Against this branch: 147 unit tests pass across `dialect-mir` and `mir-lower`, plus 3 new importer tests. `cargo fmt --check` clean, no warnings.

End to end on an RTX A2000 (sm_86, CUDA 13.3) — one kernel using `as_chunks::<4>()` with a weighted per-chunk sum so a permutation inside a chunk would show, one exercising the intrinsic directly on a signed and an unsigned dividend:

```
as_chunks::<4>  : 256 / 256 correct
exact_div direct: 256 / 256 correct
```

The generated PTX contains **no** `div` or `rem` — with a const divisor the computation folds entirely, so the intrinsic is free in the shape that matters.

Tests cover the negative cases deliberately. **`unchecked_div` must not land here**: it promises only a non-zero divisor and discards the remainder, so routing it through `exact_div` would silently drop the exactness obligation.

## Expectation setting on codegen

`as_chunks` yields `[f32; 4]` at 4-byte alignment, and the four loads stay scalar rather than fusing into `ld.global.v4`. This buys the **safety** — no unsafe, no bounds argument discharged by hand — not the access width, which is governed separately by the element type's alignment. Worth saying explicitly so the change is not read as a vectorization fix.

I did not add a GPU example to the repo; happy to add one under `crates/rustc-codegen-cuda/examples/` if you would like it committed rather than just run.


honeyspoon:
Added the GPU example I offered above, so it is committed rather than only run.

`crates/rustc-codegen-cuda/examples/exact_div_chunks/` — two kernels, `Cargo.lock` committed, run command carries an explicit `--arch sm_86`.

```
as_chunks::<4>  : 256 / 256 correct
exact_div direct: 256 / 256 correct

PASS
```

`chunk_sum` combines each thread's four elements with distinct weights, so a wrong chunk boundary or a permutation inside a chunk gives a wrong value instead of passing. `exact_div_direct` exercises the intrinsic away from `as_chunks` on both a signed and an unsigned dividend, since the lowering picks `sdiv` or `udiv` from the operand's signedness — a placeholder round-trip test cannot distinguish those.

The README notes that `as_chunks` yields `[f32; 4]` at 4-byte alignment and its loads stay scalar, so the example is not mistaken for a vectorization result.

nihalpasham:
Thanks, 

I pushed two commits on top: 
- reattached the saturating conversion's rustdoc that the new function had split, and 
- reworked the example so `exact_div_direct` actually calls `core::intrinsics::exact_div` (the plain / operators were lowering through `mir.div`), which also adds signed negative-dividend coverage for the `sdiv` arm. 

Verified on an RTX 5090: both arms appear in the emitted IR, the example passes, and the full 172-example smoketest is green. Merging.

nihalpasham (approved):

PR #475, opened 2026-07-25
feat(cuda-device): add f32 and f64 warp reduce utilities

honeyspoon:
`warp::redux_sync_*` is integer-only. The surface is `redux_sync_add/and/or/xor` plus `redux_sync_{min,max}_{i32,u32}`, with no float form — which mirrors PTX, where `redux.sync` defines only integer operand types. It is a hardware limit, not a gap in the bindings.

The consequence is that anyone reducing floats across a warp writes the same butterfly shuffle by hand. This adds it once, for both precisions:

```
reduce_sum_f32   reduce_max_f32   reduce_min_f32
reduce_sum_f64   reduce_max_f64   reduce_min_f64
```

Each is five XOR shuffles at offsets 16, 8, 4, 2, 1, built on the existing `shuffle_xor_f32` / `shuffle_xor_f64`.

## Notes for review

**Every lane ends with the reduced value**, not just lane 0. That is deliberate — it lets a caller feed the result straight into a block-level reduction without a second barrier, and it is the property a block reduce built on top of these depends on.

**Safe functions, not `unsafe`.** These are warp-synchronous but take no pointers and require no convergence beyond what the underlying shuffle intrinsics already document, so there is no additional obligation to put behind `unsafe`.

`#[must_use]` on all six: the return value is the entire point, so discarding one is always a mistake.

Uses `f32::max` / `f32::min` for the max and min variants rather than an open-coded compare, so the NaN behaviour follows whatever those lower to rather than being independently specified here.

`fmt --check` clean, no warnings, no new dependencies.

### DIFF
--- a/crates/cuda-device/src/warp.rs
+++ b/crates/cuda-device/src/warp.rs
@@ -918,6 +918,148 @@ pub fn is_elected_sync(mask: u32) -> bool {
     elect_sync(mask).1
 }
 
+// =============================================================================
+// Warp-Level Reductions (f32)
+// =============================================================================
+//
+// Butterfly shuffle reduction utilities for f32 values. These use
+// `shuffle_xor_f32` with a full-warp mask (0xFFFF_FFFF) to reduce a
+// value across all 32 lanes, producing the result in every lane.
+
+/// Reduce-sum a scalar f32 across all 32 lanes using butterfly shuffles.
+///
+/// After this call, every lane holds the sum of all input values.
+///
+/// # Example
+///
+/// ```rust,ignore
+/// let val: f32 = per_lane_value;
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks,  

I pushed two commits: 
- the six doc comments now state the full-warp convergence contract and `NaN` behavior (mirroring `redux_sync_add`), and 
- the `warp_reduce` example exercises `reduce_sum`/`max`/`min_f32` plus reduce_sum_f64 with every lane 

Verified on an RTX 5090. Related to #390: `max`/`min` currently lower through `libdevice`, so these helpers speed up automatically when that lands. 

Merging.

nihalpasham (approved):

PR #476, opened 2026-07-25
feat(cuda-device): add f16x2 and bf16x2 unpacking to f32

honeyspoon:
`cuda_device::convert` has 19 generated conversions and every one runs f32 -> packed narrow type. The inverse is missing, so reading packed f16 pairs means open-coding the bit arithmetic at each call site:

```rust
let w0 = f16::from_bits((quad & 0xFFFF) as u16) as f32;
let w1 = f16::from_bits(((quad >> 16) & 0xFFFF) as u16) as f32;
```

## Added

| Function | Purpose |
|---|---|
| `cvt_f32x2_f16x2(u32) -> (f32, f32)` | inverse of `cvt_f16x2_f32` |
| `cvt_f32_f16x2_lo` / `_hi` | one half only, so the other conversion is not emitted |
| `cvt_f32x2_bf16x2(u32) -> (f32, f32)` | inverse of `cvt_rz_bf16x2_f32` |

## Why hand-written rather than generated

Worth justifying, since the preference is to route intrinsics through the generator. I checked the pinned LLVM metadata: **2569 NVVM intrinsics, none of which unpack.**

- Every `ff2f16x2_*` and `ff2bf16x2_*` record is the *packing* direction.
- The `*_to_f16x2_*` records take fp8/fp4 sources, not f16x2.
- PTX has no single unpacking instruction either: a split is `mov.b32` into two 16-bit registers plus two `cvt.f32.f16`, which is exactly what these casts lower to.

So there is nothing for the generator to admit. These are plain safe functions — no `ptx_asm!`, no `unsafe`.

## Tests

- **Half ordering** — `cvt_f16x2_f32` documents its first argument as the low half, so a swap here would be silent. Pinned in both directions.
- **Sign isolation** — a negative value in one half must not bleed into the other, which is what an arithmetic shift instead of a logical one would do.
- **Exhaustive** — all 65536 f16 bit patterns round-trip, including subnormals and both zeroes, skipping only NaN since it is not `==`-comparable.

4 tests, `fmt --check` clean, zero warnings.

### DIFF
--- a/crates/cuda-device/src/convert.rs
+++ b/crates/cuda-device/src/convert.rs
@@ -9,3 +9,136 @@
 //! are more efficient than scalar Rust casts.
 
 include!("generated/convert.rs");
+
+// =============================================================================
+// Packed f16x2 unpacking
+// =============================================================================
+//
+// The generated conversions above all run f32 -> packed narrow type. The
+// inverse is hand-written because there is nothing to generate it from: the
+// pinned LLVM metadata has no f16x2-to-f32x2 intrinsic (every `ff2f16x2_*`
+// record is the packing direction), and PTX has no single unpacking
+// instruction either. Splitting a packed pair is a `mov.b32` into two 16-bit
+// registers followed by two `cvt.f32.f16`, which is what the casts below
+// lower to.
+//
+// These matter because packed f16 is the layout that gives the only reliably
+// wide global loads: reading weights as `u32`/`u64` and unpacking in registers
+// beats four scalar `f16` loads. Doing that inline at every call site is where
+// the manual `from_bits`/shift/mask arithmetic came from.
+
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks, 

I pushed three commits: 
- cross-referencing docs that resolve the name collision with the generated `tcgen05::cvt_f32x2_bf16x2` packer (your destination-first names stay, they match this module's `cvt_dst_src` convention), 
- `bf16` half-unpackers for API symmetry, and a `cvt_packed` round-trip kernel that is the first coverage of the runtime `u16-to-f16` transmute plus `fpext` device path, 
- another small commit drops two redundant casts the new all-targets clippy gate from #486 flags.

Verified bit-exact on an RTX 5090.  

Merging.

nihalpasham (approved):

PR #482, opened 2026-07-26
fix(mir-importer): compare SwitchInt arms at the discriminant width

### LINKED ISSUE #481: SwitchInt on a 128-bit integer scrutinee is rejected
## What

A `#[kernel]` that matches on a 128-bit integer fails device codegen:

```text
  Translation failed: classify_i128: ["src/main.rs": line: 78, column: 21]
  Unsupported construct: SwitchInt value 340282366920938463463374607431768211455
  does not fit in 64 bits
```

Reproducer:

```rust
#[kernel]
pub fn classify(lo: &[u64], hi: &[u64], mut out: DisjointSlice<u32>) {
    if let Some((slot, idx)) = out.get_mut_indexed() {
        let i = idx.get();
        let value = (u128::from(hi[i]) << 64) | u128::from(lo[i]);
        *slot = match value {
            V1 => 1,   // any arm above u64::MAX
            _ => 0,
        };
    }
}
```

The scrutinee has to be built from runtime values; a constant one is folded before it reaches MIR.

## Where

`crates/mir-importer/src/translator/terminator/mod.rs`. `translate_switch` has two comparison paths, one arm compared against the scrutinee directly and several arms built as a chain of comparison blocks. Both narrow the arm value through a checked `u64` conversion before constructing the comparison constant:

```rust
let switch_val = u64::try_from(val).map_err(|_| {
    input_error!(
        loc.clone(),
        TranslationErr::unsupported(format!(
            "SwitchInt value {} does not fit in 64 bits", val
        ))
    )
})?;
let width_nz = NonZeroUsize::new(width).unwrap();
let apint = APInt::from_u64(switch_val, width_nz);
```

`SwitchInt` arm values are `u128` bit patterns at the discriminant's width, so any arm above `u64::MAX` is rejected. This covers `u128` and `i128` scrutinees, and a `-1` arm on an `i128` reaches the same limit through its sign bit.

## Why the narrowing is wider than it needs to be

The comment above it justifies the limit by the dialect storing enum tags as `u64`, `MirEnumType::variant_discriminants` being a `Vec<u64>`.

That limit is enforced already, where the enum type is built. `translate_adt_type` in `translator/types.rs` rejects a discriminant carrier wider than 64 bits:

```rust
if width > 64 {
    return input_err_noloc!(TranslationErr::unsupported(format!(
        "Enum {} uses a {}-bit discriminant; widths above 64 bits are not yet
         represented losslessly (issue #306)", enum_name, width
    )));
}
```

So an enum that reaches `translate_switch` always carries a tag that fits in 64 bits, and the widths above 64 arriving there belong to integer scrutinees. The guard rejects a case the enum limit never covered.

`APInt::from_u128` exists in pliron and is the same construction as `from_u64`, a zeroed `Awi` at the requested width followed by a write of the value.

## How it was found

The in-tree rustlantis differential fuzzer, seed 212, whose generated case switches on an `i128`:

```text
seed 212: COMPILE_FAIL [backend] Unsupported construct: SwitchInt value
285189434699432138271245212154734059274 does not fit in 64 bits
```

## Scope

Integer scrutinees only. Enum discriminants wider than 64 bits stay rejected at type construction, and #306 remains the tracking issue for representing those losslessly.

I have a fix and will open a PR against this.

alejandro-soto-franco:
Closes #481

## Summary

A `match` on a 128-bit integer fails to translate:

```text
  Translation failed: classify_i128: ["src/main.rs": line: 78, column: 21]
  Unsupported construct: SwitchInt value 340282366920938463463374607431768211455
  does not fit in 64 bits
```

`SwitchInt` arm values are `u128` bit patterns at the discriminant's width. Both comparison paths in `translate_switch`, the single arm and the chain, narrowed each value through a checked `u64` conversion before building the comparison constant, so any arm above `u64::MAX` was rejected. That covers `match` on `u128` and `i128`, and a `-1` arm on an `i128` reaches the same limit through its sign bit.

Found by the in-tree rustlantis fuzzer, seed 212.

## Why the narrowing was there

The comment justified it by the dialect storing enum tags as `u64`:

```rust
// SwitchInt values are u128 bit patterns at the discriminant's
// width; the dialect stores tags as u64 (same limit as
// MirEnumType::variant_discriminants in types.rs), so values
// that need more than 64 bits must fail loudly instead of
// silently truncating.
```

That limit is real, and it is enforced where the enum type is built. `translate_adt_type` rejects a discriminant carrier wider than 64 bits before any switch on it can be translated:

```rust
if width > 64 {
    return input_err_noloc!(TranslationErr::unsupported(format!(
        "Enum {} uses a {}-bit discriminant; widths above 64 bits are not yet
         represented losslessly (issue #306)", enum_name, width
    )));
}
```

So an enum reaching `translate_switch` always has a tag that fits in 64 bits, and the widths above 64 that arrive here belong to integer scrutinees. The guard was refusing a case the enum limit never covered.

## Changes

* Build the comparison constant with `APInt::from_u128` at the discriminant's own width, in place of the checked narrowing plus `APInt::from_u64`.
* Move the construction into `switch_arm_constant`, shared by both paths, so the two cannot drift.
* Add `examples/switchint_128bit`.

## Rationale

Nothing that previously compiled changes. `APInt::from_u64` and `APInt::from_u128` are the same construction, `Awi::zero_with_capacity(width, width)` followed by a write of the value, both truncating below the source width and zero-extending above it. For every value that passed the old `u64::try_from` gate the two produce the same `APInt`, which the unit test pins at 8, 16, 32 and 64 bits. Only the values that used to be rejected behave differently.

The enum limit stays enforced where it belongs, in `translate_adt_type`, and #306 stays the tracking issue for representing those tags losslessly.

The dropped error message was the only user of `input_error!` in that function; the diagnostic for an enum tag wider than 64 bits is unaffected and still names the enum and its width.

## Validation

Ran on an RTX 5060 Laptop GPU (sm_120a, CUDA 13.1, `nightly-2026-04-03`). Two builds of the same tree differing only by this change:

| | baseline (`672d8f4a`) | this PR |
|---|---|---|
| `cargo oxide run switchint_128bit` | ✗ `SwitchInt value ... does not fit in 64 bits` | ✓ `PASS`, host-verified |
| `run_seed.py --seed 212` | ✗ `COMPILE_FAIL [backend]` | ✓ `PASS [run] CPU/GPU traces matched` |

Reverting only the constant construction reintroduces the failure.

```text
$ cargo test -p mir-importer
test translator::terminator::tests::switch_arm_constants_are_built_at_the_discriminant_width ... ok
test result: ok. 858 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ scripts/smoketest.sh switchint_128bit --compile-only
[ 1/ 1] switchint_128bit                 ... PASS (compiled) [23s]
Pass:    1 / 1

$ cargo fmt --all -- --check
$ (cd crates/rustc-codegen-cuda/examples/switchint_128bit && cargo fmt --all -- --check && cargo clippy --all-targets -- -D warnings)
(all clean)
```

The example covers the chain path (three arms), the single-arm path, and a signed scrutinee with a negative arm, since the comparison constant takes the discriminant's signedness. Each scrutinee is assembled from runtime halves so the match reaches MIR intact, and two of the default cases differ from a matched arm in only the low or only the high half, so a comparison that dropped either half would misclassify them. The host checks the arm every thread selected.

## Checklist


nihalpasham:
Thanks, 

**Gist:** arm constants now materialize at the discriminant's width, its the same approach as `rustc_codegen_ssa`'s `const_uint_big`

I pushed one commit -
- moving `switch_arm_constant` out from between `translate_switch`'s rustdoc and the function so each doc block attaches to its own item again. 

Verified on an RTX 5090: `switchint_128bit` passes, and fuzzer seeds 212 and 240, the exact failures behind #481, now PASS. Closes #481. 

Merging.

nihalpasham (approved):

PR #484, opened 2026-07-26
feat(fuzzer): trace f32 and f64 values

alejandro-soto-franco:
## Summary

The trace API had a writer for every integer type plus `bool` and `char`, and none for floats, so the adapter refused any seed whose dumped value or return type was a float:

```text
seed 221: UNSUPPORTED [adapter] unsupported return type for return-value tracing: f32
```

Float codegen therefore sat outside the differential comparison. A float reached the hash only where it happened to pass through an `as` cast, a comparison yielding `bool`, or `transmute_place`, which is a thin and accidental sample of a large part of the backend.

## Changes

* Add `f32` and `f64` to the trace API, folded as their `to_bits()` pattern.
* Accept both in the adapter's supported-type set, and add literals for them so a generated function can take float arguments. The literals are exactly representable in binary floating point, so a seed starts from the value it was given.
* Pass `--no-fmad` when running a seed.
* Update the README's type list and record why contraction is disabled.

## Rationale

Floats are hashed as bit patterns so the trace stays exact `u64` equality. A tolerance inside the trace would compare something other than the value a backend produced, and would hide the differences the fuzzer exists to find. The consequence is that every NaN the two backends produce must agree bit for bit, payload included. That holds for the arithmetic rustlantis emits, and a disagreement in a NaN payload is a result worth reporting on its own.

`--no-fmad` is the part worth arguing about. Device codegen sets `contract` on float arithmetic by default, so the NVPTX backend fuses an `fmul` feeding an `fadd` into one `fma.rn`, matching nvcc's `--fmad=true`. The CPU oracle rounds the multiply and the add separately. Both results are correct, so the difference reports nothing about a miscompile, and once it appears in the hash it would mask the differences that do. Disabling contraction for the fuzzer keeps the oracle and the device comparable. Contraction is worth fuzzing separately, against a contracted reference.

This does widen exposure to the libdevice hazard the README already documents. Previously a float influenced the hash only indirectly; now it is hashed directly, so a seed calling a transcendental is more likely to produce a difference that is within the tolerance libdevice is permitted. That triage guidance is unchanged and still applies: check whether the differing value derives from a transcendental and compare in ULPs before treating a mismatch as a miscompile. The alternative, keeping floats out of the trace to avoid it, is what left this coverage gap.

## Validation

Ran on an RTX 5060 Laptop GPU (sm_120a, CUDA 13.1, `nightly-2026-04-03`).

Seed 221, the case that motivated this:

```text
before: seed 221: UNSUPPORTED [adapter] unsupported return type for return-value tracing: f32
after:  seed 221: PASS [run] CPU/GPU traces matched
```

Seeds 200 to 239 with this change:

```text
summary: COMPILE_FAIL=1, PASS=39
```

The single failure is seed 212, a `SwitchInt` value above `u64::MAX`, which is unrelated to this change and is fixed separately.

A further 70 seeds, 230 to 299, ran green while checking the contraction question. No float seed in either range produced a mismatch.

The checked-in seed 162 regression case still passes:

```text
$ cargo oxide run --no-fmad rustlantis-smoke
Stage 1b CPU hash: ...  Stage 1b GPU hash: ...
PASS: CPU/GPU traces match
```

```text
$ cargo fmt --all -- --check
$ python3 -m py_compile crates/fuzzer/tools/mir_generator.py crates/fuzzer/tools/run_seed.py
(all clean)
```

- [x] `cargo oxide run <example>` passes (`rustlantis-smoke`)

No example added: `rustlantis-smoke` is the harness this changes, and it already carries the checked-in seed.

## Checklist

### DIFF
--- a/crates/fuzzer/README.md
+++ b/crates/fuzzer/README.md
@@ -83,23 +83,33 @@ or function signature that uses a type the adapter cannot rewrite. The trace
 API hashes:
 
 ```text
-bool, i8, i16, i32, i64, i128, isize, u8, u16, u32, u64, u128, usize, char
+bool, i8, i16, i32, i64, i128, isize, u8, u16, u32, u64, u128, usize, char,
+f32, f64
 ```
 
-It does not hash `f32` or `f64`. In many `UNSUPPORTED [adapter]` cases, the MIR
-can probably be patched by widening the adapter and trace API. The adapter stops
-because it does not yet know how to rewrite/hash that dumped type safely.
+In many `UNSUPPORTED [adapter]` cases, the MIR can probably be patched by
+widening the adapter and trace API. The adapter stops because it does not yet
+know how to rewrite/hash that dumped type safely.
 
 ## Floating point and libdevice seeds
 
 The comparison is exact `u64` hash equality, so it assumes the CPU and the GPU
-agree bit for bit. Floats are never hashed directly, since the trace API has no
-`f32` or `f64` arm and the adapter refuses a bare float dump. A float can still
-reach the hash indirectly, through an `as` cast to an integer, through a
-comparison that yields a `bool`, or through rustlantis' `transmute_place`.
+agree bit for bit. Floats are folded as their `to_bits()` pattern, which keeps
+that assumption intact: a tolerance in the trace would compare something other
+than the value a backend produced, and would hide the differences the fuzzer
+exists to find. Floats also still reach the hash indirectly, through an `as`
+cast to an integer, through a comparison that yields a `bool`, or through
+rustlantis' `transmute_place`.
 
-Where that happens on a seed whose device code calls libdevice, a mismatch is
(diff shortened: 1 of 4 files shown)


nihalpasham:
Thanks, 

**Gist:** FMA analysis is correct and routing it through the existing `--no-fmad` flag instead of a tolerance keeps the oracle bit-exact. 

I pushed one commit - 
- canonicalizing `NaN` bits in the two writers before hashing (Cranelift-fuzzgen practice): raw payload bits could produce false `MISMATCH` verdicts that the doc comment promised could not happen, and 
- the README triage guidance now matches. 

Verified on an RTX 5090: seeds 200 through 299 pass, and seed 162 passes both with and without `--no-fmad`. 

Merging.

nihalpasham (approved):

PR #488, opened 2026-07-27
docs: Make mir-importer error examples compile

### LINKED ISSUE #487: Make mir-importer error examples compile as doctests
## Problem

The usage example in `crates/mir-importer/src/error.rs` is currently marked as `rust,ignore` and is not self-contained:

* It imports `input_err!` but also invokes `input_err_noloc!` without importing it.
* It references `crate::error::TranslationErr`, which is not the appropriate path for an external doctest.
* It contains standalone `return` expressions without showing the required `TranslationResult` function context.

Because the example is ignored, documentation drift is not detected by doctests.

## Required changes

* Change the code block from `rust,ignore` to `rust,no_run`.
* Import `TranslationErr` and `TranslationResult` through the public `mir_importer` API.
* Import both `input_err!` and `input_err_noloc!`.
* Wrap the location-aware and location-free examples in functions returning `TranslationResult<()>`.

## Acceptance criteria

* `cargo test -p mir-importer --doc` passes.
* Both documented error-construction patterns compile.
* The example uses only public crate paths.
* No compiler, translation, or runtime behavior is changed.

## Out of scope

* Changing `TranslationErr` variants or messages.
* Changing Pliron error-handling behavior.
* Modifying production translation code.

uurl:
Closes #487 

## Summary

* Convert the ignored `mir-importer` error usage snippet into a compiling `no_run` doctest.
* Import `TranslationErr` and `TranslationResult` through the public crate API.
* Import both Pliron result-producing macros used by the example.
* Add function contexts for the location-aware and location-free error cases.

## Motivation

The previous snippet was not self-contained and could drift because rustdoc did not compile it. Making it a `no_run` doctest ensures that the documented API remains synchronized with the actual error-handling interfaces.

## Validation

* `cargo fmt --check`
* `cargo test -p mir-importer --doc`

No production code or runtime behavior is changed.

### DIFF
--- a/crates/mir-importer/src/error.rs
+++ b/crates/mir-importer/src/error.rs
@@ -10,15 +10,19 @@
 //!
 //! # Usage
 //!
-//! ```rust,ignore
-//! use pliron::input_err;
-//! use crate::error::TranslationErr;
+//! ```rust,no_run
+//! use mir_importer::{TranslationErr, TranslationResult};
+//! use pliron::{input_err, input_err_noloc, location::Location};
 //!
-//! // With location (preferred - shows where in MIR the error occurred):
-//! return input_err!(loc, TranslationErr::unsupported("f16 type"));
+//! // With location (preferred because it identifies the MIR source):
+//! fn reject_unsupported_construct(loc: Location) -> TranslationResult<()> {
+//!     input_err!(loc, TranslationErr::unsupported("f16 type"))
+//! }
 //!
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks, all four changes from #487 as specified, and the error-handling examples now compile as real doctests instead of rotting behind `rust,ignore`. 

Merging.

nihalpasham (approved):

PR #510, opened 2026-07-27
feat(cuda-device): add block-level reduce primitives

honeyspoon:
Follows #475, which added the warp reductions this builds on.

`block_reduce_sum` and `block_reduce_max` get written by hand in every kernel that needs them — warp shuffle, shared memory, `sync_threads`. The hand-written versions also tend to bake in a warp count; an unrolled eight-slot tail is not reusable at another block size.

Provides `reduce_sum_f32`, `reduce_max_f32`, `reduce_min_f32`, generic over block size, built on #475's reductions rather than repeating the butterfly.

## The substantive choice: `&mut SharedArray`, not `*mut f32`

A raw pointer erases the length at the boundary. The helper cannot tell whether it was given enough room, so every access inside it is unchecked and correct sizing is left to convention at each call site.

With the capacity in the type, writes are guarded by `N` and the warp count is clamped to it — an undersized array becomes a reduction over fewer warps rather than memory corruption. Bounds stop being a caller obligation, so the safety contract shrinks to what is irreducible.

Still `unsafe fn`, for the one obligation that remains: these contain a `sync_threads()`, so a thread returning early elsewhere in the block deadlocks the rest.

38 tests in the crate, `clippy --all-targets` clean, `fmt --check` clean.

### DIFF
--- a/crates/cuda-device/src/block.rs
+++ b/crates/cuda-device/src/block.rs
@@ -1,0 +1,182 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Block-level collective primitives.
+//!
+//! These utilities combine warp-level shuffles with shared memory and
+//! `sync_threads()` to perform reductions across an entire thread block.
+//!
+//! # Scratch array
+//!
+//! Each reduction needs one `f32` slot per warp. The scratch is taken as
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks for this but `cuda-device` already ships this primitive as `cooperative_groups::block_reduce` (generic over T and the op, const-generic warp count checked at compile time, documented scratch-reuse barrier, warp indexing correct for 2D/3D blocks, result broadcast to every thread, and pipeline-tested by `coop_groups_demo`), so landing a second f32-only variant would leave users two APIs where the existing one is stronger on every axis. 

I'd recommend reworking this as a thin wrapper over that primitive if the shorter `block:: spelling` is worth having;  

Closing this one as already covered.

PR #511, opened 2026-07-27
feat(mma_frag): predicate accumulator registers against a logical extent

honeyspoon:
Follows #474, which derived the layout.

A tile shape rarely divides the problem, and the accumulator makes the mismatch awkward. Each lane's four values are scattered across the tile, so clipping is **not** a contiguous prefix — two lanes disagree about the same register index:

```
M = 9 against the 16-row tile
lane  0  group 0  ->  mask 0b1111     all four registers inside
lane  4  group 1  ->  mask 0b0011     rows 9+ are padding
lane 31  group 7  ->  mask 0b0011
```

`j` 0 and 1 land on row `group` and are always inside; `j` 2 and 3 land on row `group + 8` and are inside only for group 0. So lanes 0..3 keep four registers and every other lane keeps two: `4*4 + 28*2 = 72 = 9 * 8`. The whole logical tile, nothing more.

## API

A 4-bit mask per lane, not a per-store bounds check — a kernel tests one mask, and the warp stays convergent because every lane evaluates the same shape. `acc_mask_is_uniform` reports all-in or all-out, which is what decides whether an epilogue can take one unpredicated path.

## Masking cannot introduce a race

It removes registers from a bijection, so surviving stores stay pairwise disjoint. Asserted rather than argued: the test claims each element and fails on a double claim.

## Hardware validation

Real `mma.sync` on sm_86 with the destination **prefilled with a sentinel** — that is what distinguishes "computed the right values" from "suppressed the wrong stores":

```
valid rows written: 72 (expect 72)
wrong values      : 0
padding clobbered : 0
```

## Tests

Every extent from `0x0` to `16x8` swept, asserting the valid count equals `rows * cols` — a wrong predicate surfaces as a wrong total rather than as missing output. Column clipping covered separately, because it behaves differently: columns come from `tig`, so a narrow extent removes whole **lanes** rather than registers.

12 tests, `clippy --all-targets` clean.

### DIFF
--- a/crates/cuda-device/src/mma_frag.rs
+++ b/crates/cuda-device/src/mma_frag.rs
@@ -270,8 +270,13 @@ mod tests {
 // and nothing more.
 //
 // The predicate is per register rather than per lane, so it is returned as a
-// 4-bit mask. A kernel tests one mask instead of recomputing bounds per store,
-// and the warp stays convergent because every lane evaluates the same shape.
+// 4-bit mask. A kernel tests one mask instead of recomputing bounds per store.
+// Computing the mask adds no divergent control flow (every lane evaluates the
+// same straight-line expressions), but the masks themselves differ across
+// lanes, so stores guarded on mask bits can still diverge exactly as any
+// bounds check would. The mask buys fewer tests and a warp-uniform fast path
+// when `acc_mask_is_uniform` holds for every lane; it does not by itself keep
+// the warp convergent.
 
 /// Whether one accumulator register lies inside a logical extent.
 ///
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks, 

**landed with two maintainer doc commits on top:** r
- restored the #474 module-doc passages that this branch's context hunks reverted (the replacement linked nonexistent `DisjointBlock`/`DisjointTiling` types and broke rustdoc -D warnings), and 
- corrected the 'warp stays convergent' claim - masks differ across lanes, so guarded stores can still diverge; the mask buys fewer tests and a uniform fast path. 

The six new predication tests, clippy, and rustdoc -D warnings are all clean. Closes no open issue. 

Merging

nihalpasham (approved):

PR #512, opened 2026-07-27
feat(cuda-device): over-aligned element types with a checked slice view

honeyspoon:
Transaction width is set by the alignment of the *element type*, not by how many elements a loop touches. Measured on sm_86, four safe spellings of "read four adjacent f32":

| element types | LDG | STG |
|---|---|---|
| `F32x4` → `F32x4` (align 16) | **LDG.E.128** ×1 | **STG.E.128** ×1 |
| same payload at align 4 | LDG.E ×4 | STG.E ×4 |
| `&[f32]` → `F32x4` | LDG.E ×4 | **STG.E.128** ×1 |
| `&[f32]` → `f32` | LDG.E ×4 | STG.E ×1 |

Row three is the load-bearing one: a single kernel with scalar loads and a wide store, differing only in the types at each end.

So reaching a wide access means *choosing an element type*, and the obstacle is that buffers arrive flat. Copying into an aligned buffer defeats the purpose. This adds the types plus a view — check alignment and length once, then accesses through it are wide.

```rust
let quads: &[F32x4] = vector::as_vectors(input).ok_or(Unaligned)?;
```

Seven types covering the 64- and 128-bit widths for f32, f64, u32, u16.

## Refuses rather than truncates

10 `f32` viewed as `F32x4` is an error, not two vectors and a dropped tail. A misaligned base is likewise refused — that is the case the check exists for: a sub-slice starting at an odd element has the right length and the wrong address.

`Vector` is `unsafe` because a wrong `LANES` would make the length check admit an out-of-bounds view. A test asserts the layout each `unsafe impl` claims, so the invariant is checked rather than trusted.

## The caveat, documented rather than left to be found

**Alignment is necessary but not sufficient for a wide load.** With the type 16-byte aligned throughout, only how the value is *consumed* changes it:

```
never decomposed                 LDG.E.128   STG.E.128
lanes read, whole value stored   LDG.E x4    STG.E.128
lanes read, scalar stored        LDG.E x4    STG.E
```

Touching individual lanes lets the optimiser split the local apart and load them separately regardless of alignment; the store side has no such condition.

So these buy width on paths that **move** data — staging global to shared, copying, packing — and not on paths doing immediate elementwise arithmetic. The module says so up front, because "just use an aligned type" only half holds.

## Pairs with #513

Aligned types are only reachable if a buffer can *be* one. #513 adds the length-adjusting cast that makes that possible without every producer and consumer agreeing at once. Either can land alone; together they are the usable form.

8 tests, `clippy --all-targets` clean.

### DIFF
--- a/crates/cuda-device/src/bf16x2.rs
+++ b/crates/cuda-device/src/bf16x2.rs
@@ -5,5 +5,10 @@
 //!
 //! Each `u32` stores two bf16 values. The first value uses the low 16 bits.
 //! The second value uses the high 16 bits.
+//!
+//! The packing here is the ALU format for SIMD arithmetic. For over-aligned
+//! *memory element* types that make loads and stores single wide transactions,
+//! see [`crate::vector`]. For multi-register *value* groups, see
+//! [`crate::cusimd::CuSimd`].
 
 include!("generated/bf16x2.rs");
--- a/crates/cuda-device/src/cusimd.rs
+++ b/crates/cuda-device/src/cusimd.rs
@@ -17,6 +17,11 @@
 //! N values of type T that travel together through registers. All access
 //! methods use **copy semantics** - we copy values out, not reference them.
 //!
+//! For the *memory* counterpart - over-aligned element types whose alignment
+//! makes loads and stores single wide transactions - see [`crate::vector`]
+//! (e.g. [`crate::vector::F32x4`]). For packed 16-bit ALU pairs, see
+//! [`crate::f16x2`] / [`crate::bf16x2`].
+//!
 //! # Example
 //!
 //! ```rust,ignore
--- a/crates/cuda-device/src/f16x2.rs
+++ b/crates/cuda-device/src/f16x2.rs
@@ -5,5 +5,11 @@
 //!
 //! Each `u32` stores two f16 values. The first value uses the low 16 bits.
 //! The second value uses the high 16 bits.
+//!
+//! The packing here is the ALU format for SIMD arithmetic. For over-aligned
+//! *memory element* types that make loads and stores single wide transactions
+//! (e.g. eight packed halves moving as one 128-bit access), see
+//! [`crate::vector`]. For multi-register *value* groups, see
+//! [`crate::cusimd::CuSimd`].
 
 include!("generated/f16x2.rs");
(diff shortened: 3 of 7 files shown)


nihalpasham:
Thanks, 

**Landing with maintainer commits on top:** 
- the misalignment test now properly exercises the alignment rejection, 
- empty slices get uniform `Some-empty-view` semantics, 
- the Vector size contract is checked at compile time, and 
- the module docs cover the host-side story (flat buffers + in-kernel view, or `DeviceBuffer::cast_chunks` from #513) 
- plus cross-references to `CuSimd`/`f16x2`. 
- wiring the PTX check into CI exposed a compiler gap - `promoted &[]` of a struct element type - fixed in mir-importer;
- the vectorization example now proves the in-kernel `as_vectors::<F32x4>` path emits single 128-bit `global load/store`, 

**Validated end-to-end on an RTX 5090 (sm_120):** where llc-21/22 spells the access `ld`/`st.global.v2.b64`. cuda-device and mir-importer tests plus clippy -D warnings are all clean. Closes no open issue.

Merging

nihalpasham (approved):

honeyspoon:
Follow-up on the SROA finding this PR's module docs record — the one that alignment is necessary but not sufficient for a wide load, because touching individual lanes lets the optimiser split the value apart.

That is confirmed and unchanged. But inventorying CCCL 13.2.27's headers turned up the route around it, which I had wrongly written off as a dead end. CUB does not try to keep the loaded value whole; it changes the layout. From `cub/block/block_load.cuh` on `BLOCK_LOAD_VECTORIZE`:

>
> The utilization of memory transactions (coalescing) remains high regardless of items loaded per thread.

So the width survives because the read and the arithmetic use **different layouts** — striped for the load, where nothing decomposes the vector, then an exchange (four of CUB's six strategies go through shared memory) into the blocked arrangement the arithmetic wants. Decomposition still happens, just on registers that never came from a wide load instruction.

That makes `as_vectors` and a striped/blocked exchange complementary rather than competing: `as_vectors` is the checked way to get a genuine `LDG.E.128` *if* the value stays whole, which is the striped-read half of that pattern. The missing piece for decomposing kernels is the exchange, not a different alignment story.

Recording it here because this PR's docs are where the necessary-but-not-sufficient constraint is written down, and a reader reasonably concludes from them that wide loads are simply unavailable to kernels that do arithmetic. They are — via a layout change, which is a separate feature.

PR #513, opened 2026-07-27
feat(device_buffer): add cast_chunks for a length-adjusting reinterpret

honeyspoon:
`cast_elem` requires the same size **and** alignment, which is right for the layout-identical case it exists for (`u64` → `DeviceAtomicU64`). It cannot express grouping scalars into an over-aligned vector element, because such a type is by construction a different size and alignment.

That turns out to be the practical blocker on adopting wide memory accesses. Transaction width comes from the element type's alignment, so a kernel wanting `LDG.E.128` must take an aligned element — but with no length-adjusting cast, that type has to be agreed by **every producer and consumer of a buffer at once**. Migrating a single buffer pair in our kernels cost six signature changes plus a dummy buffer, which is why the aligned types went unused despite being measured to work.

`cast_chunks` moves the decision to the boundary:

```rust
// Allocate and fill as f32, the way a producer kernel leaves it.
let quads: DeviceBuffer<F32x4> = floats.cast_chunks()?;
```

## Errors rather than panics

Returns the buffer unchanged when the reinterpretation would not be exact: a byte extent that is not a whole number of `A`, a pointer not aligned for `A`, or a zero-sized `A`. A caller that cannot widen normally has a scalar path, so handing the buffer back keeps it available. It **refuses rather than truncates** a non-dividing length — dropping a tail belongs at the call site.

## On the alignment precondition

`cuMemAlloc` returns at least 256-byte-aligned memory, which satisfies any vector type, so the check is expected to pass. It is checked rather than assumed because a buffer can also come from `from_raw_parts` with a hand-computed pointer — exactly the case where being wrong is silent. A test pins that every 256-byte-aligned base satisfies every vector alignment.

Both `cast_chunks` and `can_cast_chunks` route through one pure helper, so they cannot disagree and the decision is testable without a device.

## Verification

On an RTX A2000: 1024 `f32` allocated and filled, viewed as 256 `F32x4` with **no copy**, passed to a kernel taking the aligned element type, data intact on readback. A 3-element buffer correctly refused.

5 tests on the pure helper, `clippy --all-targets` clean.

## Pairs with the aligned types

This is the enabler; the element types it casts *to* are the companion PR. Either can land alone.


nihalpasham:
Thanks, merged together with #512. 

I deliberately kept `cast_chunks` bounds unbounded for parity with `cast_elem`, and added a `cast_chunks` example on top that round-trips 1024 f32 through 256 align(16) quads on an RTX 5090 (sm_120) and pins the refusal path for a 3-float buffer. 

The five new `chunk_cast` unit tests, clippy -D warnings, and the GPU smoke are all clean. Closes no open issue.

Merging

nihalpasham (approved):

PR #514, opened 2026-07-27
Emit `.reqntid` for kernels declaring an exact launch-contract block

alejandro-soto-franco:
`#[launch_contract(block = (x, y, z))]` binds a launch to one block shape, and
that shape was known only to the host. `prepare_*` checked it and returned a
typed error on a mismatch, while a raw `_unchecked` launch, or a module whose
PTX reached the driver by another route, went unchecked.

This carries the shape into the compiled artifact as `.reqntid x, y, z`, which
the CUDA driver enforces per axis at every launch. Preparation and the driver
check the same requirement on independent paths.

## What changes

`#[launch_contract]` injects `__launch_contract_block_config::<X, Y, Z>()`
alongside the existing contract marker. The MIR importer reads it, mir-lower
propagates `reqntid_x/y/z`, and llvm-export emits the matching
`!nvvm.annotations` entries.

`.maxntid` and `.reqntid` cannot appear on one entry; ptxas rejects the pair
with "Conflicting directives: .maxntid and .reqntid cannot both be specified".
A kernel declaring an exact block therefore emits `.reqntid` in place of the
thread maximum. `.minnctapersm` is independent and still emitted.

`KernelLaunchBounds` carries a `KernelBlockGeometry` enum (`Exact` from a
launch-contract block, `Max` from `#[launch_bounds]`, or absent), so the
exporter picks `.reqntid` or `.maxntid` from the geometry's provenance
rather than a bare thread count.

## Correcting the emitted geometry

`.maxntid` bounds the product `x * y * z` and says nothing per axis. Every
non-1-D contract was emitting `.maxntid N, 1, 1`, declaring one thread on Y for
blocks of 8 or 16. The driver checks only the product, so this was harmless,
and it is replaced by the real shape.

## Rejecting a contradictory pair

A `#[launch_bounds(max)]` below the exact block's thread count is now an error.
The exact block displaces the maximum in the emitted PTX, so the maximum would
otherwise be dropped in silence and the kernel would launch at a shape its
author ruled out. `#[cuda_module]` already diagnosed this at the attribute
site. The check added here sits in the MIR importer, which sees both markers
regardless of attribute order, and covers standalone contracted kernels that
reach the device compiler without passing through the module path. A maximum
at or above the required count is redundant and stays allowed.

## Change in behaviour

An `unsafe` raw launch of a contracted kernel with a differing block shape now
fails with a driver error. The checked path already rejected that launch.

## Verification

- `.reqntid` is fail-closed: an out-of-range shape makes ptxas warn and the
  driver reject every launch.
- SASS is byte-identical between `.reqntid` and the `.maxntid` it replaces, at
  identical register counts and zero spills, for all four `gemm_views` kernels.
  The whole cubin difference is an `EIATTR_REQNTID` record in place of
  `EIATTR_MAX_THREADS`.
- The `cuda_module_contract`, `gemm_views` and `proof_carrying_views` PTX
  harnesses assert the expected directive per entry, and that no entry declares
  both.

### DIFF
--- a/crates/cuda-core/src/launch.rs
+++ b/crates/cuda-core/src/launch.rs
@@ -170,6 +170,11 @@ impl KernelLaunchConfig for LaunchConfig3D {
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum BlockRequirement {
     /// The launch must use exactly these `(x, y, z)` block dimensions.
+    ///
+    /// A kernel compiled by cuda-oxide also carries this shape as PTX
+    /// `.reqntid`, so the CUDA driver rejects a differing block on any axis
+    /// independently of this check. Preparation still reports the mismatch
+    /// first, with the kernel name and both shapes.
     Exact((u32, u32, u32)),
     /// The block may contain at most this many threads in total.
     ///
--- a/crates/cuda-device/src/thread.rs
+++ b/crates/cuda-device/src/thread.rs
@@ -982,6 +982,44 @@ const fn validate_launch_bounds(max_threads: u32) {
     );
 }
 
+/// Compile-time marker carrying an exact block shape to the device compiler.
+///
+/// `#[launch_contract(block = (x, y, z))]` already binds the host launch to one
+/// block shape. This marker gives the same fact to the device compiler, which
+/// emits `.reqntid x, y, z`. The CUDA driver then rejects a launch whose block
+/// dimensions differ on any axis, so the exact-block obligation is enforced by
+/// the compiled artifact rather than by the host check alone.
+///
+/// # How it works
+///
+/// 1. `#[launch_contract]` injects `__launch_contract_block_config::<X, Y, Z>()`
+///    at kernel start.
+/// 2. The MIR importer detects the call and extracts the const generics.
+/// 3. The marker call is removed during compilation.
+/// 4. LLVM export emits `!nvvm.annotations` with `reqntidx`, `reqntidy` and
+///    `reqntidz`.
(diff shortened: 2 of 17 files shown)


nihalpasham:
Thanks, 

**Gist:** the exact block shape now reaches the artifact as `.reqntid`, so the driver enforces the contract per axis even on raw launches that bypass preparation. 

**Validated on 2x RTX 5090 (sm_120, CUDA 13.3):** 
- full workspace suite (2997 tests), 
- GPU smoketests plus PTX harnesses for `cuda_module_contract` / `gemm_views` / `proof_carrying_views`, ptxas accepting `.reqntid` alongside `.minnctapersm` (and rejecting `.maxntid`+`.reqntid`, as the exporter enum guarantees by construction), 
- a driver-API harness confirming mismatched blocks fail with a clean `CUDA_ERROR_INVALID_VALUE` on every axis, and libNVVM accepting the `reqntid` annotations on both the modern (sm_120) and legacy LLVM7 (sm_86) dialect paths. 

**I added one maintainer commit on top:** 
- that forwards the `__launch_contract_block_config` marker to generic and explicitly instantiated kernel entries, so contracts written above `#[kernel]` also emit `.reqntid` (`generic_aligned` and `explicit_aligned_u32` expectations flipped accordingly). 

One note: the PR description says `max_threads` becomes `Option<u32>`; the code (rightly) models a `KernelBlockGeometry` enum instead - I updated the description to match. 

Related to #516, which this narrows for exact-block contracts but does not close.

nihalpasham (approved):

PR #515, opened 2026-07-27
Add `Uniform<T>`, a launch-uniform scalar witness, and make `tile_2d32_rt` safe

alejandro-soto-franco:
Depends on #514 and is branched from it. GitHub shows both commits until
that one merges; the change under review here is the second commit,
`Add Uniform<T> and make DisjointSlice::tile_2d32_rt safe`. It touches no
compiler crates, so the two are independent in substance.

`DisjointSlice::tile_2d32_rt` took a `u32` row pitch and was `unsafe`, with the
caller carrying an obligation the type system could not see: the stride has to
be the same value in every thread of the launch. Two threads that disagree
describe two different grids over one buffer, and their "disjoint" tiles can
overlap. With 1x1 tiles, thread (1, 0) at stride 5 and thread (0, 5) at stride
100 both resolve flat index 5, giving one element two `&mut`.

`Uniform<T>` turns that premise into a type.

## The witness

`Uniform<T>` is a `#[repr(transparent)]` newtype with a private field and no
public constructor. rustc materialises it from the launch packet as a kernel
parameter, which is what makes it unforgeable in safe device code: the host
marshals one value for the whole grid, so every thread reads the same bytes.

`#[cuda_module]` maps a `Uniform<T>` parameter to a plain `T` on the host,
since the host is the source of the uniformity. A sealed
`__LaunchContractUniform` bound rejects a local type also named `Uniform`,
which would otherwise take the scalar host ABI while presenting a different
device layout.

`Uniform::new_unchecked` remains for a value proven uniform some other way, and
`block_dim_x()` and its siblings return witnesses with no obligation, since a
launch has one block shape. Uniformity is closed under arithmetic whose
operands are all uniform, so `wrapping_mul` and its siblings preserve the
witness.

## Why the witness is sufficient

The old obligation had two clauses, and one of them is a safety property.
Under a uniform stride the `last_col < stride` check keeps each tile inside one
logical row, which makes the coordinate-to-offset mapping injective, so
distinct threads own disjoint elements. "The stride is the true row width"
governs only whether the kernel computes the intended result: a uniform stride
that misdescribes the layout gives wrong answers inside the slice, never
aliasing or an out-of-bounds access. The documentation separates the two.

## API change

`tile_2d32_rt` changes from `unsafe fn(..., u32)` to `fn(..., Uniform<u32>)`.
Both `gemm_views` safe kernels lose their `unsafe` blocks.

## Cost

None. PTX differs in parameter shape and instruction scheduling, and the
generated SASS is byte-identical at identical register counts.

## Note for review

A `Uniform<T>` parameter takes the aggregate `.param .align N .b8 [N]` path in
place of `.param .u32`. Alignment and size are correct at u8, u16, u32 and u64,
and ptxas absorbs the difference, so this costs nothing at runtime. Teaching
kernel-boundary scalarization to unwrap `repr(transparent)` newtypes would
still tidy the emitted IR.

### DIFF
--- a/crates/cuda-core/src/launch.rs
+++ b/crates/cuda-core/src/launch.rs
@@ -170,6 +170,11 @@ impl KernelLaunchConfig for LaunchConfig3D {
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum BlockRequirement {
     /// The launch must use exactly these `(x, y, z)` block dimensions.
+    ///
+    /// A kernel compiled by cuda-oxide also carries this shape as PTX
+    /// `.reqntid`, so the CUDA driver rejects a differing block on any axis
+    /// independently of this check. Preparation still reports the mismatch
+    /// first, with the kernel name and both shapes.
     Exact((u32, u32, u32)),
     /// The block may contain at most this many threads in total.
     ///
--- a/crates/cuda-device/src/lib.rs
+++ b/crates/cuda-device/src/lib.rs
@@ -36,6 +36,7 @@ pub mod shared;
 pub mod tcgen05;
 pub mod thread;
 pub mod tma;
+pub mod uniform;
 pub mod view;
 pub mod warp;
 pub mod wgmma;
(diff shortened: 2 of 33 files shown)


nihalpasham:
Thanks, 

A per-call witness cannot carry the safety proof: 
- safe code can select between two `Uniform<u32>` values under a thread-varying condition, or use different strides at two call sites on one slice, and 
- two threads then hold `&mut` to the same element (stride 5 at (1,0) and stride 10 at (0,5) both resolve element 5). 

Tile disjointness needs one stride per slice for the slice's whole lifetime, so the sound redesign is to bind the pitch into the `RuntimeRowMajorTiles` slice once at the host boundary, the runtime analog of `ROW_STRIDE` in the type. 

The Uniform-to-bare-scalar host mapping, the sealed anti-spoof bound with its trybuild tests, and the requires-relation unwrapping are all worth keeping in that rework, which will also need a rebase now that #514 lands separately.

PR #518, opened 2026-07-27
fix(mir-importer): translate fully-uninitialized constants as undef

### LINKED ISSUE #517: translate_constant has no handler for fully-uninitialized constant allocations (MaybeUninit::uninit of non-ZST aggregates)
Under `-O`, rustc const-promotes a `MaybeUninit::uninit()` initializer of a non-ZST type into an `Allocated` constant whose init mask is entirely unset (`bytes` all `None`, no provenance). None of the `translate_constant` type-dispatch branches (ZST → ptr_to_array → struct → enum → float → pointer → integer) accepts a constant with no defined bytes, so device codegen fails with:

```
Unsupported construct: Unsupported constant type in translate_constant.
  Rust type : ... std::mem::MaybeUninit<...Align16<...Vec3>> ...
  const repr : MirConst { kind: Allocated(Allocation { bytes: [None, ...], provenance: ProvenanceMap { ptrs: [] }, align: 16, ... }) }
```

Hit in practice compiling a physics shader crate (dimforge nexus, 265 device fns) whose math library (glam-style `Align16<Vec3>`) materializes such constants on aggregate init paths. Minimal repro shape:

```rust
#[repr(C, align(16))]
#[derive(Clone, Copy)]
struct Align16([f32; 3]);

#[kernel]
pub fn write_through_uninit(mut out: DisjointSlice<f32>) {
    let tid = thread::index_1d();
    let t = tid.get() as f32;
    if let Some(out_elem) = out.get_mut(tid) {
        let mut slot: MaybeUninit<Align16> = MaybeUninit::uninit();
        slot.write(Align16([t, t + 1.0, t + 2.0]));
        let v = unsafe { slot.assume_init() };
        *out_elem = v.0[0] + v.0[1] + v.0[2];
    }
}
```

A constant with no defined bytes has exactly one sound value for any type — `undef` — and the zero-sized-union path already emits `MirUndefOp` for its analogous case. PR with the one-branch fix + regression example incoming.

haixuanTao:
## Summary

Fixes #517. `translate_constant` fails with "Unsupported constant type" on a fully-uninitialized `Allocated` constant — the shape rustc const-promotes from a `MaybeUninit::uninit()` initializer of a non-ZST type under `-O` (init mask entirely unset, no provenance). A constant with no defined bytes has exactly one sound value for any type: `undef`. Short-circuit to `MirUndefOp` before the type dispatch, mirroring what the zero-sized-union path already emits for its analogous case.

## Changes

- `mir-importer`: in `translate_constant`, detect an `Allocated` constant with a non-empty byte range, all bytes uninit, and no provenance entries, and emit `MirUndefOp` of the translated type.
- New example `examples/uninit_const`: writes through an uninit 16-byte-aligned aggregate slot (`MaybeUninit<Align16([f32; 3])>` — the glam `Align16<Vec3>` pattern that hits this in real shader crates) and verifies the values on the host.

## Testing

- `cargo oxide build uninit_const` on unmodified `main` fails with `Unsupported constant type in translate_constant`; with this change it builds.
- `cargo oxide run uninit_const` passes on an RTX 5090 (sm_120): `uninit_const: PASS (32 threads)`.
- `cargo test -p mir-importer` green.
- Downstream validation: the dimforge nexus physics shader crate (265 device fns, where the pattern was first hit via its math library) compiles through the unified path with this change and runs a full standing-robot physics scene with results matching its previous builds.

- [x] `cargo oxide run uninit_const` passes
- [x] `cargo test -p mir-importer` passes
- [x] New example added (`uninit_const`)

## Checklist

https://claude.ai/code/session_01B7NC7U2wDx2tSmT9mkeF9h


nihalpasham:
Thanks, 

**Gist:** a fully-uninitialized `Allocated` constant has undef as its only sound value, and the pre-dispatch guard (non-empty bytes, all uninit, no provenance) cannot intercept any previously working defined-byte constant. 

Reproduced the #517 failure verbatim on unmodified main, then 

**Validated the branch on 2x RTX 5090 (sm_120):** 
- `uninit_const` passes (32 threads, undef visibly flowing through the exported IR), 
- mir-importer 861/861 tests green, and t
- the full 173-example compile-only smoketest sweep green, confirming nothing relied on the old zero-filling. 

Merging. 

Closes #517.

nihalpasham (approved):

Issue #485, opened 2026-07-26
Experimental standalone API cannot compile libdevice calls, though the pipeline can now link them


alejandro-soto-franco:
## Summary

`cuda_oxide_codegen::experimental` cannot compile a module that calls `sqrt`, `exp`, `tanh`, or any other float intrinsic that lowers to libdevice. Those calls return `CompileError::UnsupportedLinking`, so a frontend generating ordinary elementwise kernels stops at the first transcendental.

This was a stated limitation of the experimental v1 when #314 merged:

> v1 is revision-pinned and standalone compilation does not yet provide libdevice/device-extern linking, runtime loading.

Since then the shared pipeline gained IR-level libdevice linking, and the standalone path already supplies everything that linking needs. Opening this to ask how you want the gap closed, before writing any code, since it touches the API boundary.

## Current behaviour

`mir-lower` routes float intrinsics to libdevice entry points, for example `crates/mir-lower/src/convert/ops/call.rs`:

```rust
Self::SqrtF32 => Ok("__nv_sqrtf"),
Self::SqrtF64 => Ok("__nv_sqrt"),
```

The standalone request is built with `OutputPolicy::SelfContainedPtx`:

```rust
pub(crate) fn for_standalone_ptx(...) -> Self {
    Self {
        device_externs: &[],
        output_policy: OutputPolicy::SelfContainedPtx,
        ...
    }
}
```

which is checked in `run_module_pipeline`:

```rust
if matches!(request.output_policy, OutputPolicy::SelfContainedPtx) {
    let symbols = unresolved_external_symbols(ctx, module);
    if !symbols.is_empty() {
        return Err(PipelineError::UnsupportedLinking { symbols });
    }
}
```

`collect_unresolved_external_symbols` records every body-less `FuncOp` whose name does not begin with `llvm_`, so the `__nv_*` declarations lowering just emitted are collected and the compilation is rejected.

The rustc frontend takes the other branch, `OutputPolicy::ExternalLinkAllowed`, and reaches the libdevice link further down the same function.

## The capability is already wired to this path

The self-contained check sits at the top of the function; the libdevice link sits below it. The module is rejected before reaching the step that would resolve the symbols it was rejected for.

`can_ir_link_libdevice` is computed from the request's own toolchain:

```rust
let can_ir_link_libdevice = libdevice_path.is_some()
    && match request.toolchain {
        ToolchainPolicy::Discover => libdevice_ir_linking_available(request.backend),
[...]

PR #520, opened 2026-07-27
fix(intrinsics-gen): debug-control libNVVM PTX floor is the mechanism floor, not the probe version

### LINKED ISSUE #519: libNVVM backend floors for trap/breakpoint/pmevent (and tanh_approx_f32) encode PTX 9.3, making them unbuildable on the NVVM-IR route
Any device code whose panic/abort path reaches `trap` cannot compile through the NVVM-IR route (the route every libdevice-using module is forced onto, since `emit_nvvm_ir = needs_libdevice`). The build dies at export with:

```
Export failed: generated intrinsic `trap` (`v1:i0295`) requires PTX 9.3, newer than cuda-oxide can request
```

Cause: in `generated_intrinsic_targets.rs`, the `LibNvvm` backend requirement for `trap` encodes `minimum_ptx: from_encoded(93)`:

```rust
GeneratedBackendRequirement {
    backend: GeneratedIntrinsicBackend::LibNvvm,
    requirement: GeneratedTargetRequirement {
        minimum_ptx: GeneratedPtxVersion::from_encoded(93),
        hardware: GeneratedHardwareTarget::AnyOf(&[
            GeneratedHardwareAlternative::MinimumSm(75),
        ]),
    },
},
```

and `ptx_isa_requirement_for_floor` tops out at `89..=90 => Ptx90`, so the floor is unconditionally rejected — `trap` is unbuildable under libNVVM, period.

The floors look like an evidence-pipeline artifact rather than real ISA minima: the source recipe for the same intrinsic (`cuda-intrinsics-gen/src/resolve.rs`, `debug_control_recipe`) says `minimum_ptx: "1.0"` (PTX ISA §9.7.20.4 — `trap` has existed since PTX 1.0, and the `LlvmNvptx` backend floor for the same target is 3.2 with sm_20). 9.3 with sm_75 reads like the doc/probe version the evidence was collected against, not a floor.

Four generated targets carry `from_encoded(93)` under `LibNvvm`: `trap`, `breakpoint`, `pmevent`, `tanh_approx_f32`. The first three are PTX 1.x-era miscellaneous instructions; `tanh.approx.f32` is PTX 7.0/sm_75.

Hit in practice: a physics shader crate using libdevice math plus ordinary bounds-check panics — every kernel with a panic path fails export. Locally overriding the `trap` floor to `from_encoded(10)` unblocks the build and the resulting cubins run correctly on sm_120, for whatever that's worth as evidence the floor is not real.

Happy to PR the change if you can point at where the fix belongs — the generated file itself, or the upstream evidence data it is rendered from.

haixuanTao:
## Summary

Fixes #519. The compact debug-control admission writes the CUDA 13.3 probe environment (PTX 9.3 / sm_75) into the LibNvvm backend requirement as its `minimum_ptx` floor. PTX 9.3 exceeds the newest version `ptx_isa_requirement_for_floor` can request, so any module containing `trap` — i.e. any panic/abort path — fails export on the NVVM-IR route (the route every libdevice-using module takes). trap/brkpt/pmevent are PTX 1.x-era instructions lowered via inline PTX; the probe version is provenance, not a requirement — the stage validator itself encodes this for libNVVM (`ptx >= expected_ptx`, and the test `libnvvm_stage_may_report_newer_ptx_than_the_native_instruction_floor` locks it in).

This PR sets the debug-control LibNvvm PTX floor to the inline-PTX mechanism floor (3.2), matching the sibling LlvmNvptx entry. The hardware floor stays sm_75, deliberately: backend-codegen evidence must sit exactly at the hardware floor and CUDA 13 cannot probe older targets — and the PTX floor alone was what made the route unbuildable.

## Changes

- `cuda-intrinsics-gen/resolve.rs`: `expand_debug_control_admission` LibNvvm lowering floor `9.3/sm_75` → `3.2/sm_75` (with a comment explaining why the probe version must not be the floor), and the matching pin in the debug-control `ensure_exact_inline_ptx_backends` call.
- Regenerated outputs (`cargo run -p cuda-intrinsics-gen -- generate`): the three debug-control LibNvvm floors change `from_encoded(93)` → `from_encoded(32)`; everything else in the 906-file diff is the restamped catalog hash in generated file headers.

## Testing

- `cargo run -p cuda-intrinsics-gen -- check`: generated outputs current.
- `cargo test -p cuda-intrinsics-gen`: 282 passed.
- `cargo test -p cuda-oxide-codegen`: green (with a CUDA≥12.8 `ptxas` on PATH; the sm_120 spine test fails on a stock CUDA 12.0 ptxas regardless of this change).
- End-to-end repro: a shader crate using libdevice math plus ordinary bounds-check panics fails export on unmodified `main` with the `requires PTX 9.3` error; with this change it compiles, and the resulting sm_120 cubins run a full physics scene (dimforge nexus + vortx, RTX 5090) with results matching prior builds.

- [x] `cargo test` passes on touched crates
- [x] `cargo run -p cuda-intrinsics-gen -- check` passes

## Checklist

- [x] SPDX headers on new source files (none added)

https://claude.ai/code/session_01B7NC7U2wDx2tSmT9mkeF9h


nihalpasham:
Thanks, 

**Gist:** confirmed the debug-control LibNVVM floor was the probe PTX version rather than the inline-PTX mechanism floor: 
- on unmodified main, any kernel whose panic path reaches the generated trap fails NVVM-IR export with the exact 'requires PTX 9.3, newer than cuda-oxide can request' error from #519 

With this PR the same module builds, runs correctly on RTX 5090 (sm_120), and an intentional out-of-bounds still aborts via trap at runtime. `cargo run -p cuda-intrinsics-gen -- check` reproduces every restamped file bit-exact, and the full workspace suite (3011 tests) is green. 

Merging. 

Closes #519; related context in #485.

nihalpasham (approved):

PR #522, opened 2026-07-27
Add 'C' compile-time string support to ptx_asm

### LINKED ISSUE #521: Support the ptx_asm! "C" compile-time string constraint
## Problem

`ptx_asm!` currently supports register constraints and immediate integer operands, but it does not support CUDA's `"C"` input constraint.

The `"C"` constraint allows a compile-time character sequence to be substituted directly into the inline PTX template. This is useful for selecting instruction modes, suffixes, rounding modes, cache operators, and other PTX fragments without creating runtime operands.

For example:

```rust
const MODE: &[u8; 6] = b".wide\0";

let product: u64;

unsafe {
    ptx_asm!(
        "mul%1.u32 %0, %2, %3;",
        out("=l") product,
        in("C") MODE,
        in("r") lhs,
        in("r") rhs,
        options(register_only),
    );
}
```

The expected inline assembly after processing is equivalent to:

```text
template:    mul.wide.u32 $0, $1, $2;
constraints: =l,r,r
```

The `"C"` operand must be removed from the runtime operand list and from the LLVM constraint string.

## Required behavior

* Accept `in("C")` as a valid `ptx_asm!` input constraint.
* Require the operand to be a compile-time byte-backed constant.
* Interpret the operand bytes as UTF-8.
* Remove one trailing zero byte when present.
* Substitute the resulting text into every matching template placeholder.
* Remove the `"C"` constraint and operand before creating `nvvm.inline_ptx`.
* Renumber runtime operands that appear after removed `"C"` operands.
* Handle multi-digit operand references such as `$10` atomically.
* Preserve existing escaped dollar sequences such as `$$`.
* Escape dollar characters introduced by the substituted compile-time text.
* Preserve output constraints, clobbers, and inline PTX options.
* Reject `"=C"` and `"+C"` because `"C"` is input-only.

## Testing

Add coverage for:

* Macro validation accepting `"C"`.
* Compile-fail coverage for genuinely unsupported constraints.
* Compile-time string substitution.
* Runtime operand renumbering.
* Repeated references to the same `"C"` operand.
* Multi-digit operand references.
* Removal of exactly one trailing zero byte.
* Dollar escaping.
* Out-of-range template operands.
* An end-to-end CUDA kernel using `"C"` to generate a valid PTX instruction suffix.

## Documentation

Document the new constraint in the `ptx_asm!` README, including:

* Compile-time constant requirements.
* UTF-8 requirements.
* Trailing-zero behavior.
* Input-only semantics.
* Removal from the runtime LLVM inline assembly operands.
* Interaction with the existing 16-input limit.

## Out of scope

* Output or read-write `"C"` constraints.
* Runtime strings.
* Arbitrary LLVM inline assembly constraint modifiers.
* Changes to the NVVM dialect or LLVM lowering representation.

uurl:
Closes #521 

## Summary

Adds support for CUDA's `"C"` compile-time string input constraint to `ptx_asm!`.

A `"C"` operand is evaluated as a compile-time byte string, substituted directly into the PTX template, and removed before the inline assembly operation reaches the NVVM and LLVM lowering layers.

For example:

```rust
const MODE: &[u8; 6] = b".wide\0";

unsafe {
    ptx_asm!(
        "mul%1.u32 %0, %2, %3;",
        out("=l") product,
        in("C") MODE,
        in("r") lhs,
        in("r") rhs,
        options(register_only),
    );
}
```

is normalized to the equivalent LLVM inline assembly representation:

```text
template:    mul.wide.u32 $0, $1, $2;
constraints: =l,r,r
```

### `cuda-macros`

* Adds `"C"` to the supported input constraints.
* Keeps `"C"` input-only and rejects output/read-write forms such as `"=C"` and `"+C"`.
* Updates unit tests for constraint validation.
* Updates the compile-fail unsupported-constraint fixture to use a genuinely unsupported constraint.

### `mir-importer`

* Separates compile-time string inputs from runtime inline assembly operands.
* Reads `"C"` operands from byte-backed MIR constants.
* Validates the substituted bytes as UTF-8.
* Removes one trailing zero byte when present.
* Substitutes compile-time text into the LLVM-style PTX template.
* Removes `"C"` entries from the LLVM constraint string.
* Renumbers runtime operands after removed compile-time operands.
* Handles repeated and multi-digit operand references without partial replacement.
* Preserves existing `$$` escapes.
* Escapes dollar characters introduced by compile-time substitutions.
* Preserves output constraints and clobbers.

### Tests

Adds importer tests covering:

* Compile-time text substitution.
* Runtime operand renumbering.
* Repeated `"C"` references.
* Multi-digit references such as `$10`.
* Trailing-zero removal.
* Existing and introduced dollar escaping.
* Out-of-range template references.

### Documentation and example

* Documents `"C"` behavior and limitations in the `cuda-macros` README.
* Extends the `inline_ptx` example with an end-to-end kernel using `".wide\0"` to generate `mul.wide.u32`.
* Verifies the generated kernel output on the GPU.

## Validation

```text
cargo fmt --all
cargo test -p cuda-macros
cargo test -p mir-importer
cargo clippy -p cuda-macros -p mir-importer --all-targets -- -D warnings
cargo oxide run inline_ptx
```

All commands pass. The end-to-end example reports:

```text
SUCCESS: inline PTX results are correct
```


nihalpasham:
Thanks, 

**Validated on 2x RTX 5090 (sm_120):** 
- the C-constraint kernel splices `.wide` into the template (`call i64 asm ..., "=l,r,r" in the .ll`) and 
- executes correctly, with the workspace suite (3020 tests) green. 

**Two maintainer commits on top:** 
- a formatting pass on the example, and 
- hardening so `in("C")` only accepts `&'static [u8; N]` byte strings - previously `in("C") 42` silently spliced raw integer bytes; now it is a clear type error at the call site plus an importer backstop, with trybuild coverage. 

Closes #521.

Merging

nihalpasham (approved):

PR #523, opened 2026-07-27
feat(cuda-intrinsics): admit f16x2 and packed-FP8 conversions

vyncint:
## Summary

Admits the eight `f16x2` ↔ packed-FP8 conversions, completing **Phase 1** of #337. The status note on that issue records Phase 1 as partial (4 of 10), with "the f16x2-source packs, the unpacks back to f16x2" still missing — this covers exactly those.

Four are the unchecked checklist entries:

| Checklist entry | LLVM intrinsic | Emitted PTX |
| --- | --- | --- |
| `cvt.rn.satfinite.e4m3x2.f16x2` | `llvm.nvvm.f16x2.to.e4m3x2.rn` | `cvt.rn.satfinite.e4m3x2.f16x2` |
| `cvt.rn.satfinite.e5m2x2.f16x2` | `llvm.nvvm.f16x2.to.e5m2x2.rn` | `cvt.rn.satfinite.e5m2x2.f16x2` |
| `cvt.rn.f16x2.e4m3x2` | `llvm.nvvm.e4m3x2.to.f16x2.rn` | `cvt.rn.f16x2.e4m3x2` |
| `cvt.rn.f16x2.e5m2x2` | `llvm.nvvm.e5m2x2.to.f16x2.rn` | `cvt.rn.f16x2.e5m2x2` |

The four `.relu` forms come from the same pinned LLVM intrinsics and the same generator machinery, so they are admitted alongside — the scalar-`f32` FP8 family already carries its ReLU variants, and leaving these out would make the two halves of the family asymmetric. Happy to split them out if you would rather land only the four listed.

ABI ids `i0822`–`i0829`, appended; `check-abi-history` confirms the 821 base entries are preserved.

The remaining two Phase 1 entries, scalar `cvt.f32.e4m3` and `cvt.f32.e5m2`, are **not** included: pinned LLVM 22 exposes no intrinsic for them, so they need an inline-asm contract rather than this data-driven path. Left for a separate change.

## Why the family needed structural work

`packed_conversion` assumed a two-operand scalar-`f32` source throughout. These conversions take one already-packed register, so operand arity now follows the source format instead of being fixed at two:

- `PackedConversionSourceFormat` gains `f16x2`, `e4m3x2`, `e5m2x2`, and reports its own operand count and PTX source token.
- `PackedConversionAdapter` gains `identity` — PTX orders `cvt` operands as `d, a`, so one source needs no high/low reversal.
- The recipe table dispatches on source format. This is load-bearing: unpacking to `f16x2` has the same destination, rounding, and saturation as the existing `cvt.rn.f16x2.f32`, so without the source in the key the two recipes collide.
- Rust/dialect/LLVM operand lists, inline-asm constraints (`=h,r` packing, `=r,h` unpacking), the dialect operand interface (`NOpdsInterface<1>`), probe fixtures, and the expected-PTX operand count all follow the source format.
- The Ada floor keys on FP8 appearing on **either** side, so unpacking to the older `f16x2` still reports PTX ISA 8.1 / sm_89 rather than inheriting f16x2's 7.0 / sm_80.

The existing ten scalar-`f32` recipes are untouched — the new dispatch delegates to them unchanged.

## Lowering choice

These route through inline PTX rather than typed NVVM. The typed intrinsics are declared over `<2 x half>`, while MIR carries both halves in one integer register, so the typed route would need a packed-half MIR type and bitcasts. The recorded evidence assembles the typed and inline routes to a **byte-identical cubin** (`7d86e0e3…`), so this costs nothing in emitted code and the typed route can be switched on later without changing the PTX contract. The sibling `f32x2`→`f16x2`/`bf16x2` conversions already use inline PTX.

## Evidence

Recorded with the pinned rust-llvm 22.1.2 `llc` (SHA-256 `ff1dcdc1…`, matching the profile already in the repo) and CUDA 13.3.33 `ptxas` / libNVVM / nvJitLink. The `ptxas` binary is byte-identical to the one this repo already records (`7fdd01a4…`), and the libNVVM artifact matches the recorded build-ID prefix `2330c0`.

Floor behaviour is recorded from real runs, not asserted:

- `llc` at `+ptx80` or `sm_87` → `Cannot select: intrinsic %llvm.nvvm.f16x2.to.e4m3x2.rn`
- `ptxas` on the `compute_87` PTX → `Feature 'cvt.e4m3x2.f16x2' requires .target sm_89 or higher`
- `nvJitLink` linking sm_89 LTOIR for `sm_87` → `LTOIR arch newer than target arch`

Two notes for reviewers:

1. Recorded `tool_path` values use the repository-standard `/usr/local/cuda-13.3` prefix. My toolkit was installed to a user prefix, so paths are normalised to the convention every other evidence file uses; the byte-identical `tool_sha256` is what pins the tool identity.
2. Unrelated, spotted while working here: `intrinsics/evidence/cuda-13.3-libnvvm-13.3.33-packed-conversion-fp8.json` records `backend_sha256` `b6a5db233133fb0ed693f8cb8aeb627e6f8e2597ba015a8fc53f58f4956dee58`, while the other 48 libNVVM evidence files record `b6a5db233133fb0ed693f8cb8aeb1d405254ce0ab10364478985292ec98cfcf0` for the same artifact — and the latter is what CUDA 13.3.33's `libnvvm.so.4.0.0` actually hashes to. Looks like a one-off typo. Not touched here; happy to send a separate one-line fix.

## Verification

All run without a GPU:

| Gate | Result |
| --- | --- |
| `cuda-intrinsics-gen generate` | 913 outputs |
| `cuda-intrinsics-gen check` | current |
| `cuda-intrinsics-gen probe --all --skip-terminal` | 829 routes (was 821) |
| `check-abi-history` | 821 base entries preserved, 829 total |
| Unit tests, 16 GPU-less crates | 2,610 passing, 0 failing |
| `cargo clippy -- -D warnings` | clean |
| `cargo fmt --check` | clean |

Each of the eight new routes was also probed individually and emits its exact expected instruction at `sm_89 +ptx81`, with real `ptxas` terminal validation.

New generator coverage: `compact_fp8_f16x2_conversion_axes_require_the_exact_closed_product` pins the eight-product expansion, the per-direction signatures, the single-operand arity, and the Ada floor, and rejects every off-axis admission.

Most of the diff is generated: of 922 changed files, 9 are hand-authored and 834 are the one-line catalog-SHA header bump in existing probe fixtures.

## Checklist

- [x] `cargo fmt` clean
- [x] `cargo clippy` clean with `-D warnings`
- [x] Commits signed off (DCO)
- [x] New dependencies: none


nihalpasham:
Thanks, 

**Gist:** This admits all eight `f16x2 <-> packed-FP8` conversions (`e4m3x2`/`e5m2x2`, both directions, base + fused-ReLU) through the closed `packed_conversion` admission, with the new single-operand inline-PTX lowering path. 

**Validated locally on CUDA 13.3.33 with an RTX 5090 (`sm_120`):** 
- catalog/generated outputs bit-exact, 
- ABI history append-only (821 preserved + 8 new), 
- 3028 workspace tests plus fmt/clippy clean, 
- each new intrinsic terminal-revalidated with ptxas at `sm_89` +`ptx81`, and 
- a GPU round-trip kernel confirmed exact FP8 quantization behavior (`e4m3` 1.3->1.25, `e5m2` 240->256 RNE, ReLU clamping) with `cvt.rn.satfinite.e4m3x2.f16x2` and the other seven forms in the emitted PTX. 

Merging; 

this completes the Phase 1 `f16x2<->FP8` rows of #337, which stays open with its checkboxes updated.

nihalpasham (approved):

PR #525, opened 2026-07-28
feat(cuda-macros): support `ptx_asm!` inout operands

### LINKED ISSUE #524: Support read-write `inout` operands in `ptx_asm!`
## Problem

`ptx_asm!` currently supports write-only `out` operands and explicit `in` operands, but it does not provide a source-level representation for PTX operands that are both read and written.

This makes common accumulator-style PTX patterns awkward or impossible to express correctly. For example, the following operation needs `%0` to be initialized from a Rust value and then written back after the assembly executes:

```rust
let mut accumulator = initial;

unsafe {
    ptx_asm!(
        "add.u32 %0, %0, %1;",
        inout("+r") accumulator,
        in("r") increment,
        options(register_only),
    );
}
```

## Proposed syntax

Add an `inout` operand form:

```rust
inout("+r") place
```

Supported read-write constraints should be:

```text
+h
+r
+l
+q
+f
+d
```

The `n` immediate constraint is input-only and must not be accepted for `inout`.

## Required semantics

An `inout("+r") value` operand should:

1. Read the initial value from the Rust place.
2. Initialize the corresponding PTX output register with that value.
3. Execute the PTX snippet.
4. Write the final register value back to the same Rust place.

LLVM inline assembly represents this as a write-only output plus a numeric tied input. For example:

```text
inout("+r") accumulator
in("r") increment
```

should lower to constraints equivalent to:

```text
=r,r,0
```

with marker arguments ordered as:

```text
increment, accumulator_initial_value
```

The hidden tied input must not affect user-visible `%N` template numbering.

## Operand ordering and limits

* `out` and `inout` operands must appear before explicit `in` operands.
* Declaration order must be preserved when mixing `out` and `inout`.
* The combined limit for `out` and `inout` operands should remain 8.
* The explicit `in` limit should remain 16.
* `options(register_only)` should accept either an `out` or an `inout` operand.
* A complex `inout` place must be evaluated exactly once.

Because each `inout` adds a hidden tied marker argument, output markers must support up to 24 arguments:

```text
16 explicit inputs + 8 tied inout inputs
```

Void markers can remain limited to 16 arguments because an `inout` operand always produces an output.

## Acceptance criteria

* [ ] Parse `inout("<constraint>") <place>`.
* [ ] Accept `+h`, `+r`, `+l`, `+q`, `+f`, and `+d`.
* [ ] Reject missing `+` prefixes and unsupported read-write constraints.
* [ ] Reject `inout` operands placed after explicit inputs.
* [ ] Preserve declaration order when mixing `out` and `inout`.
* [ ] Lower read-write operands using numeric tied-input constraints.
* [ ] Keep hidden tied inputs out of `%N` placeholder validation.
* [ ] Evaluate each read-write place exactly once.
* [ ] Support the maximum marker arity of 24.
* [ ] Add compile-pass and compile-fail coverage.
* [ ] Add importer coverage for 24-input output markers.
* [ ] Update macro documentation and the supported-feature matrix.
* [ ] Add an end-to-end GPU example that verifies initial-value read and final-value write-back.

uurl:
Closes #524 

## Summary

This PR adds read-write `inout` operands to `ptx_asm!`.

The new syntax allows a Rust place to initialize a PTX output register and receive the final register value after the assembly executes:

```rust
let mut accumulator = initial;

unsafe {
    ptx_asm!(
        "add.u32 %0, %0, %1;",
        inout("+r") accumulator,
        in("r") increment,
        options(register_only),
    );
}
```

Supported read-write constraints are:

```text
+h
+r
+l
+q
+f
+d
```

## Implementation

The macro now models `out` and `inout` operands in a single ordered output list. This preserves source declaration order for mixed output forms.

Read-write constraints are lowered to LLVM inline assembly using an output constraint followed by a numeric tied-input constraint. For example:

```text
inout("+r") accumulator
in("r") increment
```

produces:

```text
constraints: =r,r,0
arguments:   increment, accumulator_initial_value
```

Explicit inputs remain before generated tied inputs. As a result, hidden initialization operands do not alter the user-visible `%N` numbering in the PTX template.

Each `inout` place is evaluated once. The generated expansion stores its address and initial value before invoking the marker, then writes the returned result back through the saved address.

## Behavior and validation

This PR:

* Adds parsing for `inout("<constraint>") <place>`.
* Validates `+h`, `+r`, `+l`, `+q`, `+f`, and `+d`.
* Rejects missing `+` prefixes and unsupported constraints such as `+n`.
* Rejects `inout` operands after explicit `in` operands.
* Preserves declaration order when mixing `out` and `inout`.
* Keeps the combined `out` and `inout` limit at 8.
* Keeps the explicit input limit at 16.
* Allows `options(register_only)` with an `inout` operand.
* Keeps generated tied inputs invisible to `%N` placeholder validation.
* Supports complex read-write places without evaluating them twice.

## Marker arity

Each read-write operand adds one hidden tied input. The maximum output-marker arity is therefore:

```text
16 explicit inputs + 8 read-write inputs = 24 arguments
```

This PR adds `__ptx_asm_out_17` through `__ptx_asm_out_24`.

Void markers remain capped at 16 arguments because an `inout` operand necessarily produces an output.

## Tests

Added compile-pass coverage for:

* A single read-write operand.
* `register_only` with only an `inout` operand.
* Multiple `inout` operands.
* Mixed `out` and `inout` declaration orders.
* All supported read-write constraints.
* Complex places.
* The maximum 24-argument marker arity.

Added compile-fail coverage for:

* Missing read-write constraint syntax.
* Unsupported read-write constraints.
* An `inout` operand placed after an explicit input.

Added importer coverage for parsing `__ptx_asm_out_24`.

The `inline_ptx` example now includes an end-to-end GPU kernel that:

1. Initializes an accumulator in Rust.
2. Reads and updates it inside PTX.
3. Writes the final value back to Rust.
4. Verifies the result on the host.

## Documentation

Updated:

* The `ptx_asm!` proc-macro documentation.
* The `cuda-macros` README.
* The inline PTX example reference.
* The supported-features matrix.
* Marker-arity documentation.

## Validation

```text
cargo fmt
cargo test -p cuda-macros --test ptx_asm
cargo test -p mir-importer
cargo oxide run inline_ptx
```

Results:

```text
ptx_asm trybuild tests: passed
mir-importer tests: 862 passed
inline_ptx GPU example: SUCCESS
```


nihalpasham:
Thanks 

**Validated on 2x RTX 5090 (sm_120):** 
- the tied `"=r,r,0"` constraint reaches llc and opt promotes the address-taken accumulator (it stays in registers, no .local traffic for it). 

**Two maintainer commits on top:** 
- an extended example kernel exercising multiple read-write operands at mixed widths (`+r`/`+l`/`+f` next to an `=r ou`, emitting `"=r,=l,=f,=r,r,l,f,0,1,2"`) which executes correctly with no pipeline gaps, and 
- a doc fix listing `*T` for `+l` after verifying pointer inouts work end-to-end. 

Merging 

Closes #524.

nihalpasham (approved):

PR #527, opened 2026-07-28
Decode scalar float constants from MIR bytes

### LINKED ISSUE #526: Remove Debug parsing for scalar f32 and f64 constants
## Problem

The MIR importer currently decodes scalar `f32` and `f64` constants by formatting `constant.const_` with `Debug` and parsing the resulting string.

This is fragile because rustc's `Debug` representation is not a stable interface. The current implementation also contains silent fallbacks to `0.0` when parsing fails, which can turn an unsupported or changed constant representation into an apparently valid value.

The `f16` path already reads the constant allocation bytes directly, while `f32` and `f64` use separate Debug-string parsing logic.

## Proposed change

Decode scalar floating-point constants directly from their MIR bytes:

* Read `f16`, `f32`, and `f64` constant bytes through the existing `constant_bytes` helper.
* Interpret the bytes using the target-aware `read_uint_from_bytes` helper.
* Reconstruct the values with `from_bits`.
* Create the corresponding `MirFloatConstantOp`.
* Dispatch floating-point constants before constructing the temporary `const_str` used by non-floating constant paths.
* Remove the existing `f32` and `f64` Debug parsing branches.

## Regression coverage

Extend `fp_special_literal_smoke` to verify exact scalar float bit preservation for both `f32` and `f64`.

In addition to NaN, positive infinity, negative infinity, and a finite control value, cover:

* Negative zero.
* The representable value immediately above `1.0`.

The new cases should be checked with `to_bits()` so sign and mantissa bits are validated exactly.

## Acceptance criteria

* Scalar `f32` and `f64` constants no longer depend on rustc `Debug` output.
* `f16`, `f32`, and `f64` use a shared byte-based translation path.
* Parse failures cannot silently produce `0.0`.
* Negative zero and exact finite mantissa bits survive the importer and backend pipeline.
* Existing MIR importer tests pass.
* The `fp_special_literal_smoke` example compiles and runs successfully.

## Out of scope

This issue does not remove Debug-string parsing used by other constant categories such as pointers, integers, allocation identifiers, or enum discriminants.

uurl:
Closes #526 

## Summary

This PR removes the MIR importer's dependency on rustc `Debug` output when translating scalar `f32` and `f64` constants.

Floating-point constants are now decoded directly from their MIR allocation bytes and reconstructed from their exact IEEE-754 bit patterns.

### MIR importer

* Added `read_float_constant_bits` to:

  * retrieve constant bytes through `constant_bytes`;
  * validate the required byte width;
  * decode the value using the target-aware `read_uint_from_bytes` helper.

* Added `translate_float_constant` to centralize scalar float operation creation for:

  * `f16`;
  * `f32`;
  * `f64`.

* Reconstructs scalar values using:

  * `MirFP16Attr::from_bits`;
  * `f32::from_bits`;
  * `f64::from_bits`.

* Dispatches scalar floating-point constants before creating the Debug-formatted `const_str`.

* Removed the previous `f32` and `f64` implementations that:

  * searched the Debug string for `bytes: [...]`;
  * parsed individual `Some(byte)` entries;
  * attempted decimal parsing;
  * silently fell back to `0.0`.

### Regression coverage

Extended `fp_special_literal_smoke` with exact-bit cases for both `f32` and `f64`:

* negative zero;
* the representable value immediately above `1.0`.

The test validates these values with `to_bits()` to ensure that sign and mantissa bits survive the MIR importer and LLVM export pipeline unchanged.

The existing NaN, positive infinity, negative infinity, and finite control cases remain covered.

## Rationale

rustc's `Debug` representation is not a stable serialization format and should not be used as an importer interface.

Reading the MIR constant bytes directly:

* removes the dependency on formatting details;
* avoids silent corruption through `unwrap_or(0.0)`;
* preserves exact IEEE-754 representations;
* reuses the same mechanism already used for `f16`;
* respects target endianness through the existing byte decoder.

## Validation

```text
cargo fmt --all -- --check
cargo test -p mir-importer
scripts/smoketest.sh --compile-only --only '^fp_special_literal_smoke$'
scripts/smoketest.sh --only '^fp_special_literal_smoke$'
```

Results:

* 861 MIR importer tests passed.
* MIR importer doctests passed.
* Clippy completed with warnings denied.
* `fp_special_literal_smoke` passed in compile-only mode.
* `fp_special_literal_smoke` passed on an NVIDIA RTX 5060 Laptop GPU targeting `sm_120`.

## Scope

This PR intentionally leaves the remaining Debug-based handling for non-floating constant categories unchanged.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -1689,6 +1689,73 @@ pub fn translate_rvalue(
     }
 }
 
+fn read_float_constant_bits(
+    constant: &mir::ConstOperand,
+    kind_name: &str,
+    byte_width: usize,
+    loc: Location,
+) -> TranslationResult<u128> {
+    let bytes = constant_bytes(constant, kind_name, loc.clone())?;
+
+    if bytes.len() < byte_width {
+        return input_err!(
+            loc,
+            TranslationErr::unsupported(format!(
+                "{kind_name} constant needs {byte_width} bytes, found {}",
+                bytes.len()
+            ))
+        );
+    }
+
+    Ok(read_uint_from_bytes(&bytes[..byte_width]))
+}
+
+fn translate_float_constant(
+    ctx: &mut Context,
+    constant: &mir::ConstOperand,
+    const_ty: TypeHandle,
+    block_ptr: Ptr<BasicBlock>,
+    prev_op: Option<Ptr<Operation>>,
+    loc: Location,
+) -> TranslationResult<(Value, Option<Ptr<Operation>>)> {
+    use dialect_mir::ops::MirFloatConstantOp;
+    use pliron::builtin::attributes::{FPDoubleAttr, FPSingleAttr};
+
+    let op = Operation::new(
+        ctx,
+        MirFloatConstantOp::get_concrete_op_info(),
+        vec![const_ty],
+        vec![],
+        vec![],
+        0,
+    );
+    op.deref_mut(ctx).set_loc(loc.clone());
+
+    let float_op = MirFloatConstantOp::new(op);
+
+    if const_ty.deref(ctx).is::<MirFP16Type>() {
+        let bits = read_float_constant_bits(constant, "f16", 2, loc.clone())? as u16;
+        float_op.set_attr_float_value_f16(ctx, MirFP16Attr::from_bits(bits));
+    } else if const_ty.deref(ctx).is::<FP32Type>() {
+        let bits = read_float_constant_bits(constant, "f32", 4, loc.clone())? as u32;
+        float_op.set_attr_float_value(ctx, FPSingleAttr::from(f32::from_bits(bits)));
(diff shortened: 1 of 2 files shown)


nihalpasham:
Thanks, 

**Gist:** the Debug-string parsing (and its silent 0.0 fallback) for scalar `f32`/`f64` constants by decoding the MIR constant bytes directly, unifying `f16`/`f32`/`f64`. 

**Validated on 2x RTX 5090 (sm_120, CUDA 13):** 
- `fp_special_literal_smoke` passes with the new exact-bit -0.0 and one-ULP cases, 
- the mir-importer suite (862 tests) is green, and 
- the full 172-example compile-only smoketest sweep passes. 

**I added one maintainer commit:** 
- that decodes the bytes before allocating the `MirFloatConstantOp` so a decode error can no longer leave an orphan op. 

Closes #526.

Merging

nihalpasham (approved):

PR #529, opened 2026-07-28
fix(mir-importer): preserve device-static pointer addends

### LINKED ISSUE #528: Preserve byte addends in pointers to device statics
## Problem

Stable MIR represents a constant pointer into a Rust static as two pieces of information:

1. Provenance identifying the target static.
2. A byte addend identifying the addressed subobject.

`static_target_from_constant` already extracts both values and stores the addend in `StaticPointerTarget::byte_offset`. However, the importer currently rejects every non-zero addend before emitting MIR dialect operations:

```text
constant pointer into device static <name> has byte offset <N>;
cuda-oxide does not yet preserve interior-static pointer addends
```

As a result, constants such as references to array elements or struct fields inside a device static cannot be imported.

## Example

```rust
static TABLE: [[f32; 2]; 4] = [
    [0.25, 0.5],
    [1.0, 2.0],
    [4.0, 8.0],
    [16.0, 32.0],
];

const PAIR: &'static [f32; 2] = &TABLE[2];
```

`PAIR` points to `TABLE` with a non-zero byte addend.

## Expected lowering

The importer should preserve the addend using byte-addressed pointer arithmetic:

```text
device static base pointer
    -> cast to u8 pointer in the same address space
    -> offset by the encoded byte addend
    -> cast to the constant's final pointee type
```

The resulting pointer must retain:

* The static's CUDA address space, either global memory or constant memory.
* The original pointer mutability.
* The exact byte addend encoded by rustc.

The addend must not be passed directly to `mir.ptr_offset` while the pointer still points to the complete static, because that operation scales offsets by the pointee size.

## Scope

* Remove the explicit rejection of non-zero static-pointer addends.
* Reuse the existing `mir.cast`, `mir.constant`, and `mir.ptr_offset` operations.
* Preserve the current zero-addend behavior.
* Do not broaden support for unsized coercions or fat pointers.
* Keep null pointers and anonymous allocations on their existing translation paths.
* Add an end-to-end regression to the `device_global` example.

No new MIR dialect operation or LLVM lowering should be required.

## Acceptance criteria

* A constant reference to a sized subobject of an ordinary device static compiles.
* The generated pointer addresses the correct byte offset within the static.
* Global and constant CUDA address spaces are preserved.
* Zero-addend pointers continue to use the static base address.
* Existing unsupported unsized coercions continue to fail explicitly.
* `cargo oxide run device_global` validates the interior pointer at runtime.

uurl:
Closes #528 

## Summary

Preserve non-zero byte addends carried by constant pointers and references to device statics.

The importer previously resolved the target static and decoded the addend correctly, but rejected the pointer before emitting MIR dialect operations. This prevented references to static array elements and other sized static subobjects from compiling.

## Implementation

* Pass `StaticPointerTarget::byte_offset` through static-pointer translation.
* Emit the device-global base address as before.
* For non-zero addends:

  * Cast the base pointer to an `u8` pointer.
  * Preserve the original CUDA address space and mutability.
  * Emit a `mir.ptr_offset` using the byte addend.
  * Cast the resulting byte pointer to the constant's final pointee type.
* Keep the existing direct base-pointer path for zero addends.
* Keep unsized pointer coercions outside the scope of this change.

Using an `u8` pointer is necessary because `mir.ptr_offset` uses element-scaled pointer arithmetic. Applying the rustc byte addend to the full static pointer would scale the offset by `sizeof(static_type)` and produce an incorrect address.

## Regression coverage

Extend the `device_global` example with a constant reference to an interior element of `STATIC_WEIGHTS`:

```rust
const STATIC_WEIGHT_PAIR: &'static [f32; 2] = &STATIC_WEIGHTS[2];
```

The kernel reads through both:

* The existing base pointer to `STATIC_WEIGHTS`.
* The new interior pointer carrying a 16-byte addend.

## Test plan

```text
cargo oxide run device_global
cargo fmt --check
```

The runtime result must confirm that the interior pointer resolves to `[4.0, 8.0]` rather than the beginning of `STATIC_WEIGHTS`.


nihalpasham:
Thanks, 

**Gist:** Interior pointers into device statics now keep their byte addends, and `device_global` exercises a 16-byte addend, 

**I added two maintainer commits:** 
- the DST guard only caught layout() errors, but layout() succeeds for `[f32]` with an unsized shape, so slice-typed addends slipped through as mistyped thin pointers; 
- the guard now rejects unsized pointees explicitly, and a new `error_static_slice_addend` example pins that shape as a clean diagnostic in the smoketest. 

**Validated:** Full 173-example compile-only, mir-importer, and workspace suites are green. 

Closes #528.

Merging

nihalpasham (approved):

PR #537, opened 2026-07-28
feat(cargo-oxide): doctor optional compute-sanitizer probe

### LINKED ISSUE #532: Doctor: probe optional `compute-sanitizer` (mirror cuda-gdb)
## Summary

\`cargo oxide sanitize\` requires \`compute-sanitizer\`, and doctor already optionally probes \`cuda-gdb\` for \`cargo oxide debug\`. Doctor does not check for \`compute-sanitizer\`, so missing sanitizer tooling is discovered only when \`sanitize\` fails.

## Proposed scope

1. Add an optional doctor check for \`compute-sanitizer\` (PATH + toolkit bin fallbacks via existing helpers).
2. Print not-found as non-fatal, like cuda-gdb.
3. Unit-test lookup behavior if helpers are reusable.

## Acceptance criteria

- [x] Doctor mentions compute-sanitizer
- [x] Absence does not fail doctor
- [x] Presence prints a version/path line when available

Ayush7614:
## Summary
- Probe `compute-sanitizer` in `cargo oxide doctor` using the same PATH/toolkit fallbacks as `sanitize`.
- Absence is non-fatal (like `cuda-gdb`); presence prints version/path when available.
- Unit test covers toolkit-root discovery; README + book doctor table updated.

Closes #532

## Test plan
- [x] `cargo test -p cargo-oxide doctor_compute_sanitizer`
- [x] Confirmed absence path does not set doctor `ok = false`

### DIFF
--- a/crates/cargo-oxide/README.md
+++ b/crates/cargo-oxide/README.md
@@ -283,7 +283,9 @@ Formats all crates in the workspace: root workspace, `rustc-codegen-cuda`, and a
 Validates that your environment is correctly set up: Rust nightly toolchain,
 CUDA headers (`cuda.h`), CUDA toolkit (`nvcc`, libNVVM, nvJitLink,
 libdevice), LLVM (`llc`), clang/libclang, the NVIDIA driver / GPU, and the
-codegen backend `.so`. Every check reports what was found or how to fix it.
+codegen backend `.so`. Optional probes cover `cuda-gdb` (for `debug`) and
+`compute-sanitizer` (for `sanitize`). Every check reports what was found or
+how to fix it.
 
 `cargo-oxide` itself builds and runs without the CUDA toolkit and without an
 NVIDIA driver, and `doctor` never builds anything first, so it works on a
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks, 

**Gist:** doctor now probes compute-sanitizer with the same discovery order as cargo oxide sanitize. 

**I added one maintainer commit:** 
- the new unit test was non-hermetic (it probed the real `PATH` and failed on any `CUDA` machine), so it now uses a fake tool name like the sibling test; the fallback paths are hoisted into one shared const used by sanitize, doctor, and the test and
- doctor reports the actual `Version ...` line instead of the banner. 

**Validated on an RTX 5090 / CUDA 13.3 host:** 
- full cargo-oxide test suite, 
- clippy -D warnings, and 
- doctor runs both with compute-sanitizer on `PATH` and with `PATH` scrubbed (fallback discovery). 

Closes #532.

Merging

nihalpasham (approved):

PR #541, opened 2026-07-28
Preserve address spaces when lowering static subobject pointers

### LINKED ISSUE #540: Support references to static subobjects without losing the physical address space
## Problem

Taking a reference to a field or indexed subobject of an immutable Rust `static` can produce invalid LLVM IR.

Whole-static references compile correctly:

```rust
static VALUE: PaddedStatic = PaddedStatic {
    tag: 0xab,
    value: 0x1234_5678,
};

fn get_static() -> &'static PaddedStatic {
    &VALUE
}
```

However, references to static subobjects can fail:

```rust
fn get_static_tag() -> &'static u8 {
    &VALUE.tag
}

fn get_static_value() -> &'static u32 {
    &VALUE.value
}
```

A similar case exists for indexed array subobjects:

```rust
static WEIGHTS: [[f32; 2]; 4] =
    [[0.25, 0.5], [1.0, 2.0], [4.0, 8.0], [16.0, 32.0]];

fn get_weight_pair() -> &'static [f32; 2] {
    &WEIGHTS[2]
}
```

## Current behavior

The generated LLVM IR may return an `addrspace(1)` pointer from a function whose declared result type is an `addrspace(0)` pointer:

```text
llvm.func @get_static_tag:
    llvm.func <llvm.ptr (0)() variadic = false>

%global = llvm.addressof @VALUE : llvm.ptr (1)
%field = llvm.gep ... (%global)[0, 0] : llvm.ptr (1)
llvm.return %field
```

LLVM verification then fails with:

```text
ReturnOp operand type does not match the function's result type
```

## Root cause

Address-producing place projections do not consistently preserve the address space of their base pointer.

In particular, a field projection may be represented in MIR as returning a generic `mir.ptr<..., addrspace:0>` even when its base aggregate pointer is in CUDA global memory, `addrspace:1`.

When lowered to LLVM, `getelementptr` necessarily preserves the address space of its base pointer. This produces an `addrspace(1)` LLVM pointer despite the MIR result being declared as `addrspace(0)`.

There is also no final normalization step ensuring that the result of a projected `Rvalue::Ref` or `Rvalue::AddressOf` matches the exact pointer type declared by Rust.

## Expected behavior

Address arithmetic should remain in the physical storage address space:

```text
global address       : addrspace(1)
field or element GEP : addrspace(1)
```

The pointer should only be converted to Rust's generic pointer address space at the semantic boundary where the reference or raw pointer value is produced:

```text
%field_generic = llvm.addrspacecast %field
    : llvm.ptr (1) to llvm.ptr (0)

llvm.return %field_generic
```

## Proposed changes

* Preserve the base pointer address space when constructing field-projection pointer types.
* Keep the place-read classifier consistent with the actual address walker.
* Normalize projected `Rvalue::Ref` results to the exact translated Rust reference type.
* Apply the same normalization to projected `Rvalue::AddressOf` results.
* Add regression tests covering:

  * field projections from global and constant address spaces;
  * normalization from a physical pointer to the expected generic Rust pointer type.
* Add an end-to-end CUDA example covering struct fields and indexed array subobjects.

## Acceptance criteria

* References to fields of immutable device statics compile successfully.
* References to indexed subobjects of static arrays compile successfully.
* LLVM verification succeeds.
* libNVVM accepts the generated module.
* Runtime values read through the subobject references are correct.
* Existing `mir-importer` and `mir-lower` tests continue to pass.

uurl:
Closes #540 

## Summary

This PR adds support for references and raw pointers to subobjects of ordinary Rust statics in CUDA global and constant memory.

It fixes LLVM verification failures produced by expressions such as:

```rust
&STATIC_VALUE.field
&STATIC_ARRAY[index]
&raw const STATIC_VALUE.field
```

## Problem

Static allocations are materialized in their physical CUDA address spaces:

* global memory: `addrspace(1)`;
* constant memory: `addrspace(4)`.

Field and element projections are address calculations and must remain in the address space of their base pointer. LLVM `getelementptr` does not change address spaces.

Previously, field projections could be assigned a generic MIR pointer type in `addrspace(0)` even when the base pointer was in `addrspace(1)`. Lowering then produced an LLVM GEP in `addrspace(1)`, while the enclosing function still expected an `addrspace(0)` return value:

```text
llvm.func @get_padded_static_tag:
    llvm.func <llvm.ptr (0)() variadic = false>

%static = llvm.addressof @PADDED_STATIC : llvm.ptr (1)
%field = llvm.gep ... (%static)[0, 0] : llvm.ptr (1)
llvm.return %field
```

This failed verification with:

```text
ReturnOp operand type does not match the function's result type
```

### Preserve physical address spaces during projections

Adds a shared `projected_pointer_type` helper that creates a pointer to the projected pointee while retaining the address space of the base pointer.

The helper is used by:

* the place-read strategy classifier;
* the actual field-address projection path.

This keeps the classifier and the IR-producing walker type-consistent.

### Normalize projected Rust pointer values

After computing an address through `translate_place_address`, both:

* `Rvalue::Ref`;
* `Rvalue::AddressOf`;

now normalize the result to the exact translated Rust rvalue type.

Address arithmetic therefore remains in the physical address space, and the `mir.cast PtrToPtr` is introduced only when producing the final generic Rust reference or raw pointer.

The resulting LLVM shape is:

```text
%static = llvm.addressof @PADDED_STATIC : llvm.ptr (1)
%field = llvm.gep ... (%static)[0, 1] : llvm.ptr (1)
%generic = llvm.addrspacecast %field
    : llvm.ptr (1) to llvm.ptr (0)
llvm.return %generic
```

### Static constant pointer materialization

Static constants now preserve:

* the full evaluated initializer;
* the exact byte offset of the referenced subobject;
* the physical global or constant address space;
* the exact pointer type expected by the Rust MIR rvalue.

### Regression coverage

Adds unit tests for:

```text
projected_pointer_type_preserves_base_address_space
projected_address_normalization_matches_expected_pointer_type
```

The `device_global` example now also exercises:

* a padded `#[repr(C)]` static;
* a reference to its first field;
* a reference to a nonzero-offset field;
* a reference to an indexed array subobject;
* exact NaN initializer bits;
* persistent mutable device globals.

## Validation

The following commands pass:

```bash
cargo fmt --all
cargo test -p mir-importer
cargo test -p mir-lower
scripts/smoketest.sh --compile-only -o '^device_global$'
cargo oxide run device_global
```

The compile-only smoketest is verified and compiled by libNVVM.

The runtime CUDA test produces:

```text
Launch 1: result = 0xc0ffef
Launch 2: result = 0xc0ffec
Static table: result = 12
NaN payload: bits = 0x7fc01234
Padded static: result = 0x12345678ab
Static subobjects: result = [171, 305419896, 1082130432, 1090519040]

SUCCESS: device globals preserved their storage, initializer bytes, and subobject addresses.
```


nihalpasham:
Thanks, 

**Gist:** field and element references into device statics (`&PADDED_STATIC.tag`, `&STATIC_WEIGHTS[2]`) now keep the static's address space through the projection walk, with a single addrspacecast at the moment the Rust reference is formed. The backend folds the cast away; the PTX still loads via `ld.global`/`ld.const` directly (verified).

**One heads-up on the rebase you'll see in the branch:** 
- this PR predated #529 (which rewrote the same constant-materialization path), so I reconciled the conflict toward main's #529 guards (the zero-addend `pointee-mismatch` and `unsized-pointee` rejections stay) and 
- grafted the new pieces on top - the exact-result-type retyping of materialized constant pointers, the allocation bounds check (relaxed to allow legal one-past-the-end constants), and the projection/boundary fixes with your tests. 
- also pinned the zero-addend slice coercion as a clean diagnostic (`error_static_slice_unsize`) alongside #529's fixture.

**Validated on 2x RTX 5090 (sm_120, CUDA 13):** 
- `device_global` prints the exact subobject values [171, 305419896, 1082130432, 1090519040], 
- the shared-memory-heavy examples (`TMA`, `cp.async`, `mbarrier`, `atomics`) all execute clean since the boundary cast is global, 
- full 175/176 smoketest (sole failure is a known local-tooling issue unrelated to this PR), and 
- 102 fuzzer seeds with matching CPU/GPU traces. 
 
Closes #540.

Merging

nihalpasham (approved):

PR #542, opened 2026-07-28
feat(warp): add predicate-returning shuffles

honeyspoon:
All eight `shfl.sync.*.p` forms are present in the pinned NVVM metadata and none are generated: the `scalar_math` shapes assume a single result and these return two. This adds six of them by hand through multi-output `ptx_asm!` — `bfly`/`down`/`up`/`idx` for f32, plus `bfly`/`down` for u32.

## What the predicate actually is

I wrote these docs first describing `p` as reporting whether the source lane was *active*, so that a reduction over a partially converged warp could tell a real contribution from a stale one. Then I ran it on hardware, and that is not what it does.

Measured on sm_86:

```
shfl.sync.up   delta=1                      p false on lane 0 only
shfl.sync.up   delta=4                      p false on lanes 0..3
shfl.sync.bfly partner outside membermask   p TRUE on every lane
```

The third row is decisive. A butterfly with membermask `0x0000FFFF` and a partner in the upper half **still sets the predicate**, even though that partner is not participating. The *value* did behave as expected — lane 0 received its own input back — so that information is in the value, not in `p`.

So `p` is a **range** predicate: true when the computed source lane lies inside the clamp segment, false when the index was out of range. It does not detect convergence and cannot separate a live contribution from a stale one. The per-function docs say this explicitly, so nobody reaches for them expecting otherwise.

## What they are good for

Segment boundaries, which is where `up` and `down` need them. In a prefix scan the first `delta` lanes have no predecessor and would otherwise combine their own value twice:

```rust
let (prev, valid) = warp::shuffle_up_f32_sync_pred(mask, acc, delta);
acc += if valid { prev } else { 0.0 };
```

For `bfly` and `idx` over a whole-warp segment the predicate is always true and carries no information; they are included for segmented use with a narrower clamp, and for completeness.

## `extern crate self as cuda_device;`

`ptx_asm!` expands to absolute `cuda_device::ptx::__ptx_asm_*` marker paths — see `crates/cuda-macros/src/ptx_asm.rs`, which emits `cuda_device::ptx::__ptx_asm_c` — and a proc-macro cannot emit `$crate`. Those paths resolve in downstream kernel crates but not inside `cuda-device` itself, so this is the one line that lets modules in this crate use the macro. It is a prerequisite for any in-crate `ptx_asm!` use, not just this PR.

## Verification

Expansion checked: `%%p` lowers to a literal `%p`, operands map to `$0..$4`, constraint string is `"=f,=r,f,r,r"` — two results through the tuple form, the multi-output `ptx_asm!` case.

Semantics were validated against hardware rather than the ISA prose, which is how the original wrong claim got caught.

`fmt --check`, `clippy --workspace --all-targets -D warnings`, `RUSTDOCFLAGS='-D warnings' cargo doc`, and `cargo test -p cuda-device` all clean on `nightly-2026-04-03`.

### DIFF
--- a/crates/cuda-device/src/lib.rs
+++ b/crates/cuda-device/src/lib.rs
@@ -12,6 +12,11 @@ pub use cuda_macros::{
 };
 
 // Re-export for convenience
+// `ptx_asm!` expands to absolute `cuda_device::ptx::__ptx_asm_*` marker paths so
+// that downstream kernel crates resolve them; a proc-macro cannot emit `$crate`.
+// Aliasing this crate to its own name lets modules here use the macro too.
+extern crate self as cuda_device;
+
 pub mod async_copy;
 pub mod atomic;
 pub mod barrier;
--- a/crates/cuda-device/src/warp.rs
+++ b/crates/cuda-device/src/warp.rs
@@ -1114,6 +1114,207 @@ pub fn reduce_min_f64(mut val: f64) -> f64 {
     val
 }
 
+// =============================================================================
+// Predicate-returning shuffles
+// =============================================================================
+//
+// PTX `shfl.sync` has an optional second destination: `shfl.sync.op.b32 d|p,
+// a, b, c, membermask`. LLVM has the intrinsics
+// (`int_nvvm_shfl_sync_{bfly,down,idx,up}_{f32,i32}p`) and they are in the
+// pinned metadata, but the generator does not admit them: the scalar_math
+// shapes assume one result and these return two. Written by hand here through
+// the multi-output form of `ptx_asm!`.
+//
+// # What the predicate actually reports
+//
+// It is a **range** predicate, not an activity one: `p` is true when the
+// computed source lane lies inside the clamp segment, meaning the shuffle
+// delivered another lane's value. When `p` is false the source index was out of
+// range and the lane received its **own** input back unchanged.
+//
+// This is worth stating plainly because the natural guess is wrong. Measured on
+// sm_86:
+//
+// ```text
+// shfl.sync.up   delta=1                      p false on lane 0 only
+// shfl.sync.up   delta=4                      p false on lanes 0..3
+// shfl.sync.bfly partner outside membermask   p TRUE on every lane
+// ```
+//
+// The last row is the one that matters. Running a butterfly with membermask
+// `0x0000FFFF` and a partner in the upper half still sets `p`, even though the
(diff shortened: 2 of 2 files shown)


honeyspoon:
Renamed three of the six to match the neighbouring convention in `warp.rs` — the u32 forms take no type infix (`shuffle_xor_sync`), and the indexed form is plain `shuffle_f32_sync`, so each new function is now exactly its existing counterpart plus `_pred`:

| was | now | mirrors |
|---|---|---|
| `shuffle_idx_f32_sync_pred` | `shuffle_f32_sync_pred` | `shuffle_f32_sync` |
| `shuffle_xor_u32_sync_pred` | `shuffle_xor_sync_pred` | `shuffle_xor_sync` |
| `shuffle_down_u32_sync_pred` | `shuffle_down_sync_pred` | `shuffle_down_sync` |

The other three (`shuffle_xor_f32_sync_pred`, `shuffle_down_f32_sync_pred`, `shuffle_up_f32_sync_pred`) already matched. Gates re-run green after the rename.

honeyspoon:
Agreed, and closing was right. `ptx_asm!` is the escape hatch for *users*; reaching for it inside `cuda-device` to work around the generator is backwards — it hand-maintains what the pinned metadata already describes.

The metadata does describe it. All 16 `shfl.sync.*.{i32,f32}p` siblings are imported, each with `results: ["i32", "i1"]` and asm `shfl.sync.<mode>.b32 $dst|$pred, $src, $offset, $mask, $threadmask;` — the same `d|p` shape `elect.sync` and `match.all.sync` already generate through `register_predicate_pair`. So the two-result form is a supported concept in the catalog schema; the shfl path just never admitted it.

I'm redoing this as a generator change instead. What was actually blocking it, concretely: the closed warp-shuffle recipe assumed a single result in six places (rust_result, ptx_result, dialect_results, llvm_results, the expected-PTX destination pattern, and the selection asm), plus the imported-declaration class check — the `*p` siblings carry only `SDPatternOperator`/`Intrinsic` because CUDA C has no `__shfl_sync` spelling that returns the predicate. With those threaded through `predicate: bool` on the contract, the whole policy layer admits it and `gen check` gets as far as evidence. The remaining work is render-side: the probe/dialect/binding renderers still hardcode one result, which LLVM catches immediately (`Intrinsic has incorrect return type!` — it emitted a single-result declaration).

Will open that when it probes clean end to end rather than posting another partial.

Issue #339, opened 2026-07-03
Tracking: SM100+ (Blackwell) device intrinsics — tcgen05 FP8/FP6/FP4 variants


honeyspoon:
## What problem does this solve?

cuda-oxide has comprehensive tcgen05 support on `main` for bf16/f16/tf32, including TMEM allocation, MMA, fence, commit/wait, and 2CTA variants. However, FP8 variants and the new FP6/FP4 narrow floating-point types introduced on Blackwell are missing. This tracks adding those.

Split from #335 for per-architecture tracking. See also: SM80 (#274), SM89 (#337), SM90 (#338).

## Status (2026-07-27)

Partially landed via the #406 intrinsics autogen, from two directions:

- **Warp-specialized tcgen05 narrow-type MMA**: `crates/cuda-device/src/generated/tcgen05.rs` now has per-input-kind entry points `tcgen05_mma_ws_{e4m3,e5m2,e3m2,e2m3,e2m1}` (`kind::f8f6f4`, collector `b0::discard`, TMEM A operand).
- **Generic descriptor-driven path**: the standard entry points (`tcgen05_mma_shared` / `tcgen05_mma_tensor`, plus `_sp_` and 2CTA forms) are generic over a `KIND` const and take a caller-built instruction descriptor (`idesc`), so FP8/FP6/FP4 operands are already *expressible* through them today.
- **Register-level FP6/FP4** (not tcgen05, but relevant to the narrow-type groundwork): `mma_m16n8k32` variants with `e2m1`/`e2m3`/`e3m2` inputs (`kind::f8f6f4`) landed in `generated/register_mma.rs`, plus `sp::ordered_metadata` sparse forms.

What this issue still tracks: typed, checked tcgen05 MMA variants for FP8/FP6/FP4 on the standard (non-ws) path (like the existing `tcgen05_mma_f16` convenience), mixed-input FP8 forms, first-class FP6/FP4 type and conversion support, and end-to-end examples/tests. Boxes stay unchecked until those exist. New items go through the #406 workflow: append to `intrinsics/overlay/*.toml` + `intrinsics/abi-v1.toml` and regenerate.

## Scope

Each item follows the established 4-layer pattern (`cuda-device` > `dialect-nvvm` > `mir-importer` > `mir-lower`) plus lowering tests. Requires SM89 FP8 type support (#337 Phase 1) as a prerequisite for the FP8 variants.

---

### Phase 1: tcgen05 FP8 variants (sm_100+)

[...]

Issue #338, opened 2026-07-03
Tracking: SM90/SM90a (Hopper) device intrinsics — FP8 WGMMA, DPX, setmaxnreg


honeyspoon:
## What problem does this solve?

cuda-oxide has substantial Hopper support on `main` — WGMMA (bf16/f16/tf32), TMA (1D–5D g2s/s2g + multicast), thread block clusters, mbarrier, and tcgen05. However, FP8 WGMMA variants, additional WGMMA shapes, DPX instructions, and several Hopper-specific instructions are missing. This tracks filling those gaps.

Split from #335 for per-architecture tracking. See also: SM80 (#274), SM89 (#337), SM100 (#339).

## Status (2026-07-27)

Two Phase 3 items landed via the #406 intrinsics autogen (boxes checked below):

- `barrier.cluster.arrive` / `barrier.cluster.wait` are generated in `crates/cuda-device/src/generated/cluster_barrier.rs`, including `relaxed` and `aligned` variants.
- `fence.proxy.async` is generated in `crates/cuda-device/src/generated/mbarrier_extended.rs` (`fence_proxy_async_shared_cta` plus generic acquire/release cluster-scope variants).

On the redux item: the integer `redux.sync` add/min/max/and/or/xor forms shipped earlier (#223, #258); the f32 hardware forms are still missing. #475 added shuffle-based f32/f64 warp reduce utilities, which are a software alternative rather than `redux.sync` on f32.

Everything else (FP8 WGMMA, additional WGMMA shapes, `setmaxnreg`, DPX, `cp.async.bulk.prefetch.tensor`, `cp.reduce.async.bulk`) remains unimplemented. WGMMA on `main` still covers only m64n64k16 for f16/bf16/tf32. New items go through the #406 workflow: append to `intrinsics/overlay/*.toml` + `intrinsics/abi-v1.toml` and regenerate.

## Scope

Each item follows the established 4-layer pattern (`cuda-device` > `dialect-nvvm` > `mir-importer` > `mir-lower`) plus lowering tests.

---

### Phase 1: FP8 WGMMA variants (sm_90a)

Hopper's warpgroup-level MMA with FP8 inputs. cuda-oxide already has WGMMA for bf16/f16/tf32 but not FP8. Requires SM89 FP8 type support (#337 Phase 1) first.

- [ ] `wgmma.mma_async.*.f32.e4m3.e4m3` (f32 accumulator, e4m3 inputs)
[...]

PR #543, opened 2026-07-28
feat(intrinsics-gen): add a coverage command for the pinned metadata

honeyspoon:
`intrinsics/imported.json` pins 2569 NVVM intrinsics; the generator catalog covers 714. The remaining 1855 are invisible unless you count them by hand, and any figure written into a doc goes stale the next time the metadata is re-pinned or a family is admitted. This adds the count as a command instead.

```
cargo run -p cuda-intrinsics-gen -- coverage
cargo run -p cuda-intrinsics-gen -- coverage --family shfl
```

```
pinned NVVM intrinsics: 2569 imported, 714 generated, 1855 ungenerated
ungenerated in compute families: 1290  (the rest is surface/texture)

family               generated  ungenerated
wmma                         0          414
sust                         0          210   [not compute]
suld                         0          165   [not compute]
tex                          0          152   [not compute]
mma                        243          147
cp                          28          114
mbarrier                     9           70
tcgen05                    144           56
shfl                         8           24
```

## The split is the point

A single "72% uncovered" overstates the backlog badly. Surface and texture (`sust`/`suld`/`suq`/`tex`/`tld4`/`txq`) are **565 entries — 30% of the total** — and irrelevant to compute kernels. Separating them takes the actionable number from 1855 to **1290**, and the per-family table points a contributor at a family rather than leaving them to guess.

Two things the aggregate hides:

- `wmma` is 414 ungenerated with **zero** generated — the largest single block, and untouched.
- `mma` and `tcgen05` are already majority-covered (243/390 and 144/200), so they are consolidation rather than greenfield.

The non-compute prefix list is a single `const` in the module, so re-classifying a family is a one-line change and the classification is visible rather than implied.

## Relationship to the tracking issues

This is meant to serve the per-architecture tracking issues (#274, #337, #338, #339), which track admission by checklist. `coverage --family <name>` answers "what is left in this family" directly against the pinned metadata, so a status note can be regenerated instead of hand-reconciled.

## Verification

Counts are computed by walking `imported.json` and the generated catalog — no figure is hardcoded, so the command cannot disagree with the tree it runs in.

`fmt --check`, `clippy --workspace --all-targets -D warnings`, and `cargo test -p cuda-intrinsics-gen` clean on `nightly-2026-04-03`.

### DIFF
--- a/crates/cuda-intrinsics-gen/src/coverage.rs
+++ b/crates/cuda-intrinsics-gen/src/coverage.rs
@@ -1,0 +1,301 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Report how much of the pinned LLVM metadata the generator actually covers.
+//!
+//! `extract` pulls every NVVM intrinsic record LLVM knows about into
+//! `intrinsics/imported.json`. `generate` then emits bindings for the subset
+//! described by `intrinsics/catalog.json`. The difference is intrinsics that
+//! reached the repository and stopped there - visible in a JSON file nobody
+//! reads, rather than in any error.
+//!
+//! That difference is large, and most of it is not worth closing. Surface and
+//! texture operations account for roughly a third of it and are irrelevant to
+//! compute kernels, so a raw "72% uncovered" figure overstates the real backlog
+//! by a wide margin. The useful output is therefore per family, not a single
+//! percentage: it tells a contributor which families are worth aiming at and
+//! which to ignore.
+//!
+//! ```text
+//! cargo run -p cuda-intrinsics-gen -- coverage
+//! cargo run -p cuda-intrinsics-gen -- coverage --family shfl
+//! ```
+//!
+//! This is a reporting command. It reads the two JSON files and writes nothing.
+
+use anyhow::{Context, Result};
+use serde_json::Value;
+use std::collections::{BTreeMap, BTreeSet};
+use std::path::Path;
+
+/// Families that exist in the metadata but are not compute functionality.
+///
+/// Surface, texture, and their query forms. Counted separately so the headline
+/// backlog reflects work someone might actually want to do.
+const NON_COMPUTE_PREFIXES: &[&str] = &["sust", "suld", "suq", "tex", "tld4", "txq"];
+
+/// One family's coverage.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FamilyCoverage {
+    /// Family name, taken from the intrinsic's leading path segment.
+    pub family: String,
+    /// Records present in `catalog.json`.
+    pub generated: usize,
+    /// Records in `imported.json` with no catalog entry.
+    pub ungenerated: usize,
+    /// Whether this family is compute-relevant.
+    pub compute: bool,
+}
+
+impl FamilyCoverage {
+    /// Records seen in either file.
+    #[must_use]
+    pub fn total(&self) -> usize {
+        self.generated + self.ungenerated
+    }
+}
+
+/// The whole report.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Coverage {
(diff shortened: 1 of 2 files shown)


honeyspoon:
Concrete worked examples for this command, from cross-referencing CCCL 13.2.27's device surface against the pinned metadata. Every number below came from walking `intrinsics/imported.json` by hand, which is the work this subcommand exists to remove.

| family | imported | generated | note |
|---|---:|---:|---|
| `applypriority` | 2 | **0** | L2 eviction priority on a resident line |
| `discard` | 2 | **0** | `discard.global.L2`, frees a line you are done with |
| `ldu` | 3 | **0** | load-uniform, for warp-invariant values |
| `griddepcontrol` | 2 | **0** | programmatic dependent launch |
| `setmaxnreg` | 2 | **0** | device-side register-count control |
| `createpolicy` | **0** | 0 | **no LLVM intrinsic at all** |

Two things this makes visible that a whole-file grep does not:

**The last row is the important one.** `cuda::annotated_ptr` / `cuda::access_property` — CUDA C++'s L2 residency API, with its `persisting`/`streaming`/`normal` tags — is built on `createpolicy`, and there is no imported intrinsic for it. So that family splits: `applypriority` and `discard` are ordinary generator-coverage work, while `createpolicy` is not reachable by generating at all. Distinguishing "ungenerated" from "unimportable" is exactly the classification the per-family table gives you, and getting it wrong would send a contributor after an intrinsic that does not exist.

**Two of those rows are features nobody had noticed were missing.** `griddepcontrol` and `setmaxnreg` are both imported, ungenerated, and absent from any gap list I had. `setmaxnreg` is also the device-side counterpart to `-maxrregcount`, which makes it interesting independently.

I also hit the trap the description warns about, in the opposite direction: a naive `grep -rl discard crates/cuda-device/src/generated/` returns a hit, but it is a doc comment about tcgen05 collector modes (`0=discard, 1=lastuse`), not `discard.global.L2`. Counting from the metadata rather than from source text is what avoids that.

honeyspoon:
Pushed a correctness fix, found by having six agents audit the CUDA C++ device surface against this metadata and comparing their independent counts against the tool's.

**The tool was over-counting coverage by one, and in a way that inverted a record's meaning.** The catalog does not only record what it binds — it also records intrinsics it explicitly *rejects*. The live case is `gridid`:

```json
"source": { "kind": "ptx_native", "instruction": "mov.u64 %gridid" },
"special_register": { "llvm_exclusion": {
    "source_record": "int_nvvm_read_ptx_sreg_gridid",
    "reason": "result_width_mismatch" } }
```

The imported intrinsic returns `b32` while the register is 64-bit, so it is unusable by construction. `collect_names` walked every `int_nvvm_*` string, so it counted that rejection as generated:

| | generated | ungenerated |
|---|---:|---:|
| before | 722 | 1847 |
| after | **721** | **1848** |

Now it binds only `source.source_record` where `source.kind == "llvm_imported"`, and tracks exclusions separately — "deliberately rejected" and "nobody has got to it yet" are different facts, and keeping them apart is precisely what this command is for. A regression test pins it using the `gridid` shape.

**Second, the figures in the description are already stale**, which is the argument for the command rather than against it. It says 714 generated / 1855 ungenerated; the tree today gives 721 / 1848. The catalog gained entries in between (#523's f16x2↔FP8 conversions among them). A hardcoded figure in a doc would still read 714. Correct current output:

```
pinned NVVM intrinsics: 2569 imported, 721 generated, 1848 ungenerated
ungenerated in compute families: 1283  (the rest is surface/texture)
```

Also worth recording from the same audit: `wmma` is 414 ungenerated with **zero** generated, still the largest single untouched block, and the compute/non-compute split moved 1290 → 1283 as families were admitted. Gates green on `nightly-2026-04-03`.

nihalpasham:
Thanks, this makes things easier. 

**I layered one commit on top:** 
- the collected exclusion set is now surfaced in the report ("excluded (deliberately rejected): 1") instead of being folded into the read family's backlog, unknown `--family` exits non-zero for scripts, and the dead `texsurf` test arm asserts. 

288 crate tests green. 

Merging.

nihalpasham (approved):

PR #544, opened 2026-07-28
feat(swizzle): XOR swizzles for shared memory, with the conflict arithmetic

honeyspoon:
The swizzle surface in this crate today selects a **hardware** pattern for an engine to apply during a transfer — `Tcgen05SwizzleMode`, the WGMMA descriptor fields. There is nothing for the SIMT path, where a kernel lays shared memory out itself and needs the index arithmetic. The book discusses bank conflicts in `advanced/shared-memory-and-synchronization.md` without giving users a way to compute them. This adds that.

## The problem, quantified

Shared memory is 32 banks of 4 bytes. A warp reading one column of a row-major tile whose row stride is a multiple of 32 elements puts every lane on the same bank:

```
column access, f32 tile      worst conflict   banks used
row stride 32                32-way            1 of 32
row stride 64                32-way            1 of 32
row stride 16                16-way            2 of 32
row stride 33 (padded)        1-way           32 of 32
```

Padding to 33 fixes it, wastes a column per row, and breaks the alignment that wide accesses need. The alternative is to *permute* rather than space out:

```
row stride 32 with a swizzle
Swizzle<5, 0, 5>              1-way           32 of 32
Swizzle<3, 0, 5>              4-way            8 of 32
Swizzle<2, 0, 5>              8-way            4 of 32
```

Five bits is exactly the 32 banks, so `Swizzle<5, 0, 5>` is the default worth reaching for. The tests pin this entire table, so the numbers cannot drift from the documentation.

## Why it is safe over a disjoint access pattern

A swizzle is not an affine function of the index digits, so the compact-layout reasoning used elsewhere in the crate says nothing about it directly. It composes anyway, for a simpler reason: `x ^ (((x >> S) & ((1 << B) - 1)) << M)` is a **bijection** on the index space, and applying a bijection to pairwise-disjoint indices leaves them pairwise disjoint. So swizzling a disjoint pattern cannot introduce aliasing. The only thing to get right is that reads and writes use the same swizzle.

Both properties are tested directly — bijection over a 1024-offset span, and self-inverse.

## The involution constraint is compile-enforced

`M + B <= S` is required. Without it the XOR target bits overlap the bits being read; it stays a bijection but stops being its own inverse, so a store and a load through it would disagree. A `const` assertion rejects the bad case:

```
error[E0080]: evaluation panicked: Swizzle<B, M, S> requires M + B <= S,
otherwise applying it twice does not restore the original offset
```

with a trybuild case pinning that rejection. The check earned its keep immediately — it rejected `Swizzle<5, 0, 4>` in my own first draft of the tests, which I had written assuming it was valid.

## Scope

Pure `const` index arithmetic: `bank_of`, `conflict_degree`, `banks_used`, `column_offsets`, `row_offsets`. It does not allocate, does not touch shared memory, and does not interact with the TMA/tcgen05 descriptor modes — those remain the right tool when the engine is doing the transfer. This is for the case where the kernel is doing it.

## Verification

`fmt --check`, `clippy --workspace --all-targets -D warnings`, `RUSTDOCFLAGS='-D warnings' cargo doc`, and `cargo test -p cuda-device` clean on `nightly-2026-04-03`. The trybuild fixture was generated on that toolchain.

### DIFF
--- a/crates/cuda-device/src/lib.rs
+++ b/crates/cuda-device/src/lib.rs
@@ -34,6 +34,7 @@ pub mod mma_frag;
 pub mod prmt;
 pub mod ptx;
 pub mod shared;
+pub mod swizzle;
 pub mod tcgen05;
 pub mod thread;
 pub mod tma;
--- a/crates/cuda-device/src/swizzle.rs
+++ b/crates/cuda-device/src/swizzle.rs
@@ -1,0 +1,305 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! XOR swizzles for shared memory, and the bank-conflict arithmetic to pick one.
+//!
+//! Shared memory is 32 banks of 4 bytes, and the bank is
+//! `(byte_offset / 4) % 32`. A warp that reads one column of a row-major tile
+//! whose row stride is a multiple of 32 elements has every lane land on the same
+//! bank, which serialises the access 32 ways:
+//!
+//! ```text
+//! column access, f32 tile          worst conflict   banks used
+//! row stride 32                    32-way            1 of 32
+//! row stride 64                    32-way            1 of 32
+//! row stride 16                    16-way            2 of 32
+//! row stride 33 (padded)            1-way           32 of 32
+//! ```
+//!
+//! Padding the stride to 33 fixes it and wastes a column per row, which also
+//! breaks the alignment that wide accesses need. The alternative is to permute
+//! the offsets instead of spacing them out: XOR part of the address with bits
+//! taken from higher up, so lanes that shared a bank no longer do.
+//!
+//! ```text
+//! row stride 32 with a swizzle     worst conflict   banks used
+//! Swizzle<5, 0, 5>                  1-way           32 of 32
+//! Swizzle<3, 0, 5>                  4-way            8 of 32
+//! Swizzle<2, 0, 5>                  8-way            4 of 32
+//! ```
+//!
+//! Fewer scattered bits leaves proportionally more conflict, so five bits -
+//! exactly the 32 banks - is the choice worth defaulting to. Combinations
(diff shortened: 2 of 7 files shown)


honeyspoon:
Found a concrete consumer for this while inventorying CCCL 13.2.27 (read from the shipped headers, `cuda_cccl-13.2.27/include`), and it is a better motivation than the column-access table currently in the description.

CUB's `BlockLoad`/`BlockStore` reach wide, coalesced loads *without* relying on the loaded value staying whole. From `cub/block/block_load.cuh` on `BLOCK_LOAD_VECTORIZE`:

>
> The utilization of memory transactions (coalescing) remains high regardless of items loaded per thread.

The six strategies are `DIRECT`, `STRIPED`, `VECTORIZE`, `TRANSPOSE`, `WARP_TRANSPOSE`, `WARP_TRANSPOSE_TIMESLICED`. Four of them perform that transpose **through shared memory** — which is precisely where a row-stride-32 column access puts every lane on one bank, the 32-way case this module computes.

So the swizzle is not only for hand-written tiles: it is what a striped-to-blocked exchange needs in order not to serialise. That also makes the bijection argument load-bearing rather than incidental — an exchange writes striped and reads blocked, i.e. two different access patterns over the same buffer, and `Swizzle` being a bijection is what guarantees the read sees what the write placed. The involution constraint (`M + B <= S`) is what lets the same swizzle serve both directions.

No change to the code; recording the intended consumer so the module's justification does not rest on a synthetic column-access example.

nihalpasham:
Thanks, 

**Gist:** the involution proof, the compile-time `M + B <= S` rejection, and the pinned conflict tables all check out (verified by hand). 

- I added a device-pipeline example (`swizzle_smem:` a 32x32 tile transpose swizzled on both sides, verified on hardware; both apply calls survive to PTX as xor), 
- softened the "remains a bijection" claim (false at `M == S`, where the sourced bits get zeroed), and 
- added a CuTe correspondence note since the `S` conventions differ: `PR(B,M,S) = CuTe(B,M,S-M)`. 

Merging.

nihalpasham (approved):

PR #545, opened 2026-07-28
feat(access): derive tile shape from a target transaction width

honeyspoon:
Transaction width is a consequence of element alignment, so a kernel cannot request it. That makes the useful question the inverted one — *given* the width I want, what must the tile shape be? This adds the `const` arithmetic for that direction.

## Why

The `gemm_sol` example already does this derivation by hand, in a comment (`crates/rustc-codegen-cuda/examples/gemm_sol/src/main.rs:313`):

```text
Each K-group: 128 rows x 8 elements x 2 bytes = 2048 bytes
```

A 128-bit transaction of f16 is 8 elements, 128 rows move 1024 elements per group, and run against a `K = 32` tile that forces `M >= 1024 / 32 = 32` — which is how CUTLASS states the constraint for a row-major A operand at 128 threads and 8-element access. Every step is arithmetic on width, element size, and launch shape, so it can be computed rather than re-reasoned per kernel. A test reproduces exactly that derivation, so the module's stated justification cannot drift from its code.

## API

```rust
const PLAN: AccessPlan = access::plan::<f16>(access::TXN_128, 128).unwrap();
const MIN_M: usize = PLAN.min_extent(32).unwrap();   // 32
```

| Item | Purpose |
|---|---|
| `plan::<T>(txn_bytes, threads)` | elements per thread, per block, required alignment |
| `min_extent(other)` | smallest extent of one axis given the other |
| `fits_tile` / `passes_for_tile` | whether a tile divides into whole block-wide accesses |
| `widest::<T>(tile_elems, threads)` | the auditing direction — widest whole fit for an existing tile |

Everything is `const`, so a plan can drive a `const` generic or sit in a `const {}` assertion and turn a tile shape that cannot reach the intended width into a compile error rather than a silent narrowing.

## Scope — what this does not claim

It answers "what shape does this width **need**", not "will I get this width". Measured on sm_86 with the element type 16-byte aligned throughout:

```text
how the element is used                     LDG          STG
never decomposed                            LDG.E.128    STG.E.128
lanes read, whole value assembled+stored    LDG.E x4     STG.E.128
lanes read, scalar stored                   LDG.E x4     STG.E
```

So alignment is a necessary condition and, for loads, not a sufficient one: touching individual lanes lets the optimiser split the local apart and load the lanes separately whatever the alignment. Stores have no such condition. A plan therefore describes the shape a width needs, not a width the kernel is guaranteed to get — the data has to move as whole elements as well. The module docs say this, and point at the `vectorization` example for the alignment half.

## Verification

`fmt --check`, `clippy --workspace --all-targets -D warnings`, `RUSTDOCFLAGS='-D warnings' cargo doc`, and `cargo test -p cuda-device` clean on `nightly-2026-04-03`.

### DIFF
--- a/crates/cuda-device/src/access.rs
+++ b/crates/cuda-device/src/access.rs
@@ -47,6 +47,17 @@
 //! So `align` is a necessary condition, and for loads not a sufficient one:
 //! touching individual lanes lets the optimiser split the local apart and load
 //! the lanes separately whatever the alignment. Stores have no such condition.
+//!
+//! That limit is specific to a *direct blocked* load, where the wide value
+//! itself feeds the per-element arithmetic. CUB's `BLOCK_LOAD_VECTORIZE`
+//! reaches the width for such kernels anyway, by splitting the layouts: load
+//! *striped* with wide loads (where nothing decomposes the value), then
+//! exchange through shared memory into the blocked arrangement the arithmetic
+//! wants. Decomposition still happens, but on registers that never came from
+//! a wide load. cuda-oxide has no striped/blocked exchange primitive yet, so
+//! a decomposing kernel currently cannot reach the width - the gap is the
+//! missing layout change, not anything about alignment.
+//!
 //! A plan therefore describes the shape a width *needs*, not a width the kernel
 //! is guaranteed to get - the data has to move as whole elements as well.
 //! See the `vectorization` example for the alignment half.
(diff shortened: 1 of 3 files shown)


honeyspoon:
Turned the `plan` example into a real doctest. It was ```rust,ignore`, which meant it was never compile-checked — and it wasn't correct: it named `half::f16`, but `half` is not a dependency of this crate and `f16` is the primitive behind `#![feature(f16)]` (as `atomic.rs` uses it).

Since `plan` is host-computable `const` arithmetic there is no reason to skip it, so it now runs and asserts the derivation end to end:

```
test crates/cuda-device/src/access.rs - access::plan (line 122) ... ok
```

asserting `(elems_per_thread, elems_per_block, min_extent(32)) == (8, 1024, 32)`. `cargo test --doc -p cuda-device` is green.

honeyspoon:
An addition to the "Scope — what this does not claim" section, from inventorying CCCL 13.2.27's headers.

The scope note says a plan describes the shape a width *needs*, not a width the kernel is guaranteed to get, because touching individual lanes lets the optimiser split the local apart and load the lanes separately. That is still accurate. What I did not know when writing it is that CUB does not treat this as a dead end — it changes the layout instead of fighting the optimiser.

`cub/block/block_load.cuh`, on `BLOCK_LOAD_VECTORIZE`:

So there are two layouts, not one: **striped** for the read (coalesced, and vectorizable precisely because nothing decomposes it there) and **blocked** for the arithmetic (one thread's items contiguous), bridged by an exchange. The decomposition that defeats a widened load happens after the exchange, on registers that never came from a wide load instruction.

That does not change what this module computes — the arithmetic relating width, element size and launch shape is the same either way, and `min_extent`/`widest` are as useful for planning a striped read as a direct one. It does mean the scope note is slightly pessimistic as phrased: the width is not unreachable for decomposing kernels in general, only unreachable *through a direct blocked load*. Happy to reword that paragraph if you would prefer it state the distinction explicitly rather than leaving it at "the data has to move as whole elements as well".

nihalpasham:
Thanks, 

**Two commits on top:** 
- `min_extent` now also requires the extent to hold a whole per-thread access (the `f16`/`TXN_128` plan blessed a 4-element row that can't fit one 16-byte access; CUTLASS enforces the same divisibility), with the doc reworded to the strided model it implements, and 
- the scope note now carries your own #512 follow-up nuance (the width is unreachable only through a direct blocked load). 

The vectorization example now derives its shape through `access::plan` as an in-repo consumer. 

Merging.

nihalpasham (approved):

PR #558, opened 2026-07-28
feat(mir-importer): support zero-addend static array→slice unsize

### LINKED ISSUE #556: Support zero-addend device-static array→slice unsize (`&[T; N]` → `&[T]`)
## Summary

`const S: &[f32] = &ARRAY_STATIC` (and the equivalent zero-addend unsize of a device `static [T; N]` to `&[T]`) is rejected today with:

> zero-addend pointee reinterpretations and unsized coercions are not supported

Pinned by `error_static_slice_unsize` (STATUS support-gap) after #529/#541. Thin static pointers cannot carry slice length; the importer must build a fat pointer (`mir.construct_slice`) with the array length.

## Proposed scope

1. Detect zero-addend `Array<T, N>` → `Slice<T>` unsize when materializing static pointers.
2. Emit global alloc for the array + `construct_slice(data_ptr, N)`.
3. Convert `error_static_slice_unsize` into a positive example (or fold into `device_global`) and update STATUS / smoketest.
4. Leave interior addend slice unsize (`error_static_slice_addend`) for a follow-up.

## Acceptance criteria

- [x] Zero-addend `&TABLE` as `&[T]` compiles through mir-importer
- [x] Smoketest no longer expects the old rejection diagnostic for that shape
- [x] STATUS.md updated
- [x] Interior addend case still rejected cleanly

## Out of scope

Full pointer-relocation initializers (`static REF: &T = &TARGET`), aggregate provenance, `error_static_slice_addend`.

Ayush7614:
## Summary
- Detect zero-addend `[T; N]` → `[T]` when materializing device-static pointers.
- Emit thin global pointer + `mir.construct_slice(data, N)` fat pointer.
- Convert `error_static_slice_unsize` into positive `static_slice_unsize`; update STATUS + smoketest.
- Interior addend slice unsize (`error_static_slice_addend`) still rejected.

Closes #556

## Test plan
- [x] `cargo oxide run static_slice_unsize` (needs GPU/CUDA toolkit)
- [x] `cargo oxide build error_static_slice_addend` still fails with fat-pointer metadata diagnostic
- [x] STATUS.md / ERROR_EXAMPLES no longer list `error_static_slice_unsize`


nihalpasham:
Thanks, 

**I layered two commits:** 
- the slice length now comes from the constant's own fat-pointer metadata word rather than the array type's `N`, which is what makes zero-addend prefix subslices (`split_at(2).0` over the static) carry length 2 instead of silently claiming 4 (len > N is now a hard diagnostic), plus 
- follow-ups (`static mut`, `len 0`, `nested arrays`, tail subslices - all correct). 

Also restored the example's [workspace] table and fixed its host code to the `cu_deviceptr()` idiom - it could not build as committed, so the test plan is worth re-running before ticking. 

**Verified on hardware:** 
- including the new prefix-subslice kernel; 
- `error_static_slice_addend` still rejects cleanly. 

Merging; 

closes #556.

nihalpasham (approved):

PR #559, opened 2026-07-28
feat(mir-transforms): recognize eq-wrapped range-for unroll guards

### LINKED ISSUE #557: Recognize range-based `for` loops in `#[unroll]` (start with `0..const`)
## Summary

`#[unroll]` / `#[unroll(N)]` are shipped and macros already inject markers on \`for\` loops, but \`mir-transforms\` only recognizes counted \`while\` shapes. Idiomatic \`#[unroll] for i in 0..N\` is therefore a silent no-op (warn + skip) after Range desugars to Option / discriminant exits.

## Proposed scope

1. Canonicalize or extend induction so post-import \`for i in 0..CONST\` (exclusive Range, step +1) is recognized as a counted loop.
2. Reuse existing full-unroll path once \`LoopRecurrences\` is filled.
3. Unit tests in \`mir-transforms\` (fixtures + unroll e2e); update pass docs.
4. Prefer warning clearly when a \`for\` is annotated but still unrecognized.

## Acceptance criteria

- [x] \`mir-transforms\` tests: \`0..const\` for-shaped (or canonicalized) loop fully unrolls
- [x] Existing while unroll tests still pass
- [x] Docs no longer claim for-loops are never recognized (for the shipped subset)

## Out of scope

\`step_by\`, inclusive \`..\=\`, arbitrary iterators, runtime-only bounds without partial unroll follow-up.

Ayush7614:
## Summary
- Extend induction guard matching to peel `MirEq`/`MirNe` wrapping `<`/`<=`/`>`/`>=` vs a bool constant (post-`switchInt` range-`for` shape).
- Unit tests: induction recovers trip count; full `#[unroll]` removes the loop.
- Update `#[unroll]` docs in mir-transforms / cuda-device / cuda-macros.
- Out of scope: `step_by`, inclusive ranges, custom iterators, Option-discriminant-only headers without a recoverable relational cmp.

Closes #557

## Test plan
- [x] `cargo test -p mir-transforms analyzes_range_for_eq_wrapped_guard`
- [x] `cargo test -p mir-transforms full_unroll_range_for_eq_wrapped_removes_the_loop`
- [x] `cargo test -p mir-transforms` (full suite, 30 passed)

### DIFF
--- a/crates/cuda-device/src/thread.rs
+++ b/crates/cuda-device/src/thread.rs
@@ -1094,8 +1094,10 @@ pub fn __unchecked_indexing_config<const ENABLED: bool>() {
 /// }
 /// ```
 ///
-/// The pass currently recognizes explicit counted `while` loops. Range-based
-/// `for` loops are not yet recognized.
+/// The pass recognizes counted `while` loops and range `for i in lo..hi` forms
+/// whose exit test lowers to a relational comparison (including an equality
+/// against that comparison's boolean result). Other iterators (`step_by`,
+/// inclusive ranges, custom `IntoIterator`) are not yet recognized.
 ///
 /// Loops with several `continue` paths are supported. Full `#[unroll]` also
 /// preserves `break` paths and multiple exit targets. Partial `#[unroll(N)]`
--- a/crates/cuda-device/src/thread.rs
+++ b/crates/cuda-device/src/thread.rs
@@ -1094,10 +1094,12 @@ pub fn __unchecked_indexing_config<const ENABLED: bool>() {
 /// }
 /// ```
 ///
-/// The pass recognizes counted `while` loops and range `for i in lo..hi` forms
-/// whose exit test lowers to a relational comparison (including an equality
-/// against that comparison's boolean result). Other iterators (`step_by`,
-/// inclusive ranges, custom `IntoIterator`) are not yet recognized.
+/// The pass recognizes explicit counted `while` loops. An annotated
+/// exclusive-range `for i in lo..hi` is canonicalized by the kernel macro
+/// into that counted `while` form before rustc ever sees it, so the idiomatic
+/// range form unrolls too. Other iterator forms (`step_by`, inclusive ranges,
+/// custom `IntoIterator`) are not recognized; the pass warns and leaves them
+/// as written.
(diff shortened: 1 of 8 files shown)


nihalpasham:
A couple of things: before we talk about the implementation, I want to step back to motivation.

- Do you have a real use case where unrolling an iterator-based for-loop bought you a measurable perf win? If yes, could you elaborate? In general, all such hot loops (register-resident accumulators, staged GEMM k-loops, and the like) are written on the standard while path: an induction variable, a test, an increment. That shape is fully supported today.

For background on why this PR doesn't get us there: rustc desugars `for i in 0..N` through `Iterator::next` into an Option discriminant test, so the eq-wrapped guard this PR matches never actually occurs for range loops in our pipeline. `#[unroll] for i in 0..8` still warns and stays rolled with this change applied. So the real question is whether range-for support is worth building at all, and there are only three ways to do it:

- **Rewrite `#[unroll] for i in lo..hi`:** into the counted while form inside the kernel macro at expansion time. Scoped, but the frontend starts owning loop semantics for user code, and we don't know that **it can ever be made sound and complete.**
- **A dialect-mir canonicalization pass:** that recognizes the imported iterator shape and collapses it back into a counted loop. This is the principled, sound and complete solution, but it needs a full pass implemented and reviewed carefully.
- **Document the limitation:**

Right now we're going with option 3. Rust's for loops aren't simple counted loops, they're iterators, and treating them as anything else honestly requires option 2's machinery. The one narrow case where a rewrite is even conceivable is literal integer ranges like `0..8`, and if your loop is that simple, you're better off writing the while loop: it's the same number of lines and it unrolls today.

Ayush7614:
Thanks @nihalpasham  — you’re right that the original eq-wrapped shape wasn’t what range-for produces in-pipeline; that was a mistaken assumption on my side.

I don’t have a measured GPU perf win for iterator-for vs while; the goal was ergonomics / matching the advertised #[unroll] story.

Happy to close this and leave docs-only for now, or keep your macro desugar commits if the project wants that path. Prefer your call.

nihalpasham:
Thanks for confirming, closing this one. We'll stay docs-only for now;

PR #563, opened 2026-07-28
fix(cargo-oxide): insulate unit tests from the ambient CUDA_OXIDE_* env

### LINKED ISSUE #562: cargo-oxide unit tests fail when CUDA_OXIDE_* variables are exported, and one mutates the environment unsafely
## Summary

Several `cargo-oxide` unit tests read the ambient `CUDA_OXIDE_*` environment
through the production code they exercise, so they fail for developers who have
those variables exported. Related to #560 (same class, different variable), but
these are separate variables and separate lookups.

Reproduced on `main` (`d1cc7da`), each independently:

| Exported | Failing test |
| --- | --- |
| `CUDA_OXIDE_NO_FMA=1` | `sanitize_fingerprint_tracks_output_affecting_settings` |
| `CUDA_OXIDE_UNCHECKED_INDEXING=1` | `sanitize_fingerprint_tracks_output_affecting_settings` |
| `CUDA_OXIDE_EMIT_NVVM_IR=1` | `nvvm_ir_requested_accepts_disabled_project_configuration`, `codegen_mode_changes_rebuild_only_the_tracked_device_owner` |
| `CUDA_OXIDE_DEBUG=1` | `sanitize_interop_codegen_defaults_to_line_tables_and_forwards_no_fmad` |

```bash
# each passes clean, fails with the variable exported
CUDA_OXIDE_NO_FMA=1 cargo test -p cargo-oxide --all-targets
CUDA_OXIDE_EMIT_NVVM_IR=1 cargo test -p cargo-oxide --all-targets
CUDA_OXIDE_DEBUG=1 cargo test -p cargo-oxide --all-targets
```

`CUDA_OXIDE_VERBOSE`, `CUDA_OXIDE_TARGET`, `CUDA_OXIDE_DEVICE_ARCH`,
`CUDA_OXIDE_PTX_DIR`, `CUDA_OXIDE_NO_OPT`, `CUDA_OXIDE_DUMP_LLVM`, and
`CUDA_OXIDE_DUMP_MIR` are clean.

As with #560, CI is green because the `cargo-oxide` matrix entry runs on a stock
runner with none of these set.

## Causes

**1. Fingerprint collapse.** `sanitize_codegen_fingerprint` calls
`passthrough_codegen_fingerprint`, which slurps the whole ambient environment
via `std::env::vars_os()`. An ambient `CUDA_OXIDE_NO_FMA` is already folded into
the digest, and the `if opts.no_fmad` branch then inserts the same key — so
`no_fmad: false` and `no_fmad: true` hash identically and the test's `assert_ne!`
fails.

Worth stressing: the *production* behavior is correct. The child cargo inherits
the ambient variable, so the flag really is on either way and an identical
fingerprint is the right answer. This is purely a test-insulation defect.

Notably `passthrough_fingerprint_tracks_output_affecting_settings` — the direct
sibling, same name suffix — already does this correctly via
`passthrough_codegen_fingerprint_with_env` and `BTreeMap::new()`. The sanitize
variant reads like a copy that missed the injection. The seam already exists;
one caller just doesn't use it.

**2. Precedence shadowing.** `nvvm_ir_requested` and
`apply_default_sanitizer_line_tables` both check the process value before
project config, and both are hardcoded to `std::env::var_os`. The tests
configure the project value and are then overruled. Same shape as #560.

**3. `unsafe set_var` with an incorrect safety justification.**
`detect_run_target_arch_skips_when_env_target_set` mutates process-global state:

```rust
// Test in isolation; the `CUDA_OXIDE_TARGET` env handle is process-wide.
// SAFETY: single-threaded test serialised by the cargo test harness.
unsafe {
    std::env::set_var("CUDA_OXIDE_TARGET", "sm_75");
}
```

The justification is factually wrong: libtest runs tests concurrently on a
thread pool by default, not serialised on one thread. The write therefore races
the `std::env::vars_os()` reads that the fingerprint helpers perform — reachable
from ~31 fingerprint call sites in this same test binary — which is exactly the
data race that makes `set_var` `unsafe` in edition 2024.

In fairness, I could **not** get this to produce an observable failure: 40
consecutive full-suite runs were clean. So this is latent UB and a misleading
comment rather than a live flake. It is still worth removing, and removing it
also lets the test assert the condition it names instead of depending on global
state.

## Proposed fix

Use the seam the crate already prefers (`cuda_toolkit_root(get_env)`,
`cuda_header_candidates(toolkit, arch)`, `passthrough_codegen_fingerprint_with_env`):
keep each public entry point as a thin wrapper that reads the real environment,
and move the body into a `_with_env` counterpart that takes the value. Tests call
the counterpart. Production call sites and behavior stay unchanged.

That covers `sanitize_codegen_fingerprint`, `nvvm_ir_requested`,
`apply_default_sanitizer_line_tables` (plus its
`apply_interop_device_codegen_options` caller), and `detect_run_target_arch` —
letting the `unsafe` block be deleted outright.

## Known remaining item

`codegen_mode_changes_rebuild_only_the_tracked_device_owner` is **not** fixable
this way. It spawns a real `cargo` through `cargo_artifact_freshness`, so the
child inherits the ambient environment regardless of what the in-process
fingerprint is told. Insulating it needs the child's `CUDA_OXIDE_*` scrubbed
*and* the fingerprint injected through `cargo_passthrough_command`, which is a
production-facing signature used in several places. Happy to do that as a
follow-up if you'd like it in the same place.

## Acceptance criteria

- [ ] `cargo test -p cargo-oxide --all-targets` passes with any single `CUDA_OXIDE_*` variable exported, except the documented subprocess test
- [ ] No `std::env::set_var` / `remove_var` in the `cargo-oxide` test suite

vyncint:
## Summary

Fixes #562. Four `cargo-oxide` tests fail on `main` for developers who export
`CUDA_OXIDE_*` variables, because they read the ambient environment through the
production code they exercise:

| Exported | Failing test |
| --- | --- |
| `CUDA_OXIDE_NO_FMA=1` | `sanitize_fingerprint_tracks_output_affecting_settings` |
| `CUDA_OXIDE_UNCHECKED_INDEXING=1` | `sanitize_fingerprint_tracks_output_affecting_settings` |
| `CUDA_OXIDE_EMIT_NVVM_IR=1` | `nvvm_ir_requested_accepts_disabled_project_configuration` |
| `CUDA_OXIDE_DEBUG=1` | `sanitize_interop_codegen_defaults_to_line_tables_and_forwards_no_fmad` |

Same class as #560/#561, different variables and lookups. CI is green because
the `cargo-oxide` matrix entry runs on a stock runner with none of these set.

## What was wrong

**Fingerprint collapse.** `sanitize_codegen_fingerprint` reached
`passthrough_codegen_fingerprint`, which collects the whole environment via
`vars_os`. An ambient `CUDA_OXIDE_NO_FMA` is already folded into the digest, and
the `if opts.no_fmad` branch inserts the same key — so `no_fmad: false` and
`no_fmad: true` hash identically and the `assert_ne!` fails.

To be explicit: **the production behavior is correct.** The child cargo inherits
the ambient variable, so the flag is on either way and a single digest is the
right answer. This is purely test insulation.

The seam already existed and one caller just didn't use it — the direct sibling
`passthrough_fingerprint_tracks_output_affecting_settings` already injects
`BTreeMap::new()` through `passthrough_codegen_fingerprint_with_env`.

**Precedence shadowing.** `nvvm_ir_requested` and
`apply_default_sanitizer_line_tables` check the process value before project
config, with the ambient read hardcoded, so tests configuring the project value
were overruled.

**An `unsafe` environment mutation resting on a false premise.**
`detect_run_target_arch_skips_when_env_target_set` claimed:

```rust
// SAFETY: single-threaded test serialised by the cargo test harness.
```

libtest runs tests concurrently on a thread pool by default, so that `set_var`
raced the `vars_os` reads the fingerprint helpers perform on other test threads —
the data race that makes `set_var` `unsafe` in edition 2024.

I want to be straight about the severity: I could **not** turn this into an
observable failure. 40 consecutive full-suite runs were clean. So it was latent
UB plus a misleading comment, not a live flake. Removing it also lets the test
assert the condition it names rather than depending on global state.

## Approach

The same seam the crate already prefers — `cuda_toolkit_root(get_env)`,
`cuda_header_candidates(toolkit, arch)`,
`passthrough_codegen_fingerprint_with_env`: each entry point stays a thin
wrapper that reads the real environment, and the body moves to a `_with_env`
counterpart that takes the value. Tests call the counterpart.

Applied to `sanitize_codegen_fingerprint`, `nvvm_ir_requested`,
`apply_default_sanitizer_line_tables` (plus its
`apply_interop_device_codegen_options` caller), and `detect_run_target_arch` —
the last of which lets the `unsafe` block be deleted outright.
`inherited_process_env` is factored out so both fingerprint wrappers share one
collection.

**Production call sites and runtime behavior are unchanged.** The only
non-test caller of each wrapper keeps calling the wrapper.

The one visible test refactor: `sanitize_fingerprint_tracks_output_affecting_settings`
had six 10-line positional calls differing only in two bools; since every one
needed the extra argument anyway, they collapse into a local closure so the
varying dimension is legible.

## Validation

No GPU and no CUDA toolkit required.

Per-variable sweep, `cargo test -p cargo-oxide --all-targets`:

| Exported | Before | After |
| --- | --- | --- |
| *(none)* | 145 / 0 failed | 145 / 0 |
| `CUDA_OXIDE_NO_FMA=1` | 144 / **1 failed** | 145 / 0 |
| `CUDA_OXIDE_UNCHECKED_INDEXING=1` | 144 / **1 failed** | 145 / 0 |
| `CUDA_OXIDE_DEBUG=1` | 144 / **1 failed** | 145 / 0 |
| `CUDA_OXIDE_EMIT_NVVM_IR=1` | 143 / **2 failed** | 144 / 1 (see below) |
| `CUDA_OXIDE_VERBOSE`, `_TARGET`, `_DEVICE_ARCH`, `_PTX_DIR`, `_NO_OPT`, `_DUMP_LLVM`, `_DUMP_MIR` | 145 / 0 | 145 / 0 |

Also green: `cargo clippy -p cargo-oxide --all-targets -- -D warnings`,
`cargo fmt --all -- --check`, and the CPU unit-test matrix (`mir-lower`,
`mir-importer`, `dialect-mir`, `dialect-nvvm`, `llvm-export`, `oxide-artifacts`,
`reserved-oxide-symbols`, `cuda-intrinsics`, `cuda-intrinsics-gen`,
`mir-transforms`, `nvvm-transforms`, `cuda-artifact-finalizer`, `cuda-device`,
`libnvvm-sys`, `nvjitlink-sys`) — 2678 tests, 0 failures.

No `std::env::set_var` or `remove_var` remains anywhere in `cargo-oxide`.

## Deliberately not fixed

`codegen_mode_changes_rebuild_only_the_tracked_device_owner` still fails under
`CUDA_OXIDE_EMIT_NVVM_IR`. It spawns a real `cargo` via
`cargo_artifact_freshness`, so the child inherits the ambient environment no
matter what the in-process fingerprint is told. Fixing it needs the child's
`CUDA_OXIDE_*` scrubbed *and* the fingerprint threaded through
`cargo_passthrough_command`, a production-facing signature used in several
places — a different mechanism, so I left it out rather than half-doing it.
Glad to follow up if you want it.

## Note on #561

Independent of #561 and based on `main`, but both touch
`crates/cargo-oxide/src/commands.rs`. The hunks don't overlap (that PR is around
`find_cuda_toolkit_executable` and the sanitizer-lookup tests), so they should
merge in either order; happy to rebase whichever lands second.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -9245,6 +9245,31 @@ edition = "2024"
     }
 
     #[test]
+    fn nvvm_ir_requested_env_disable_overrides_enabled_project_configuration() {
+        let ctx = Context {
+            workspace_root: PathBuf::from("/tmp/project"),
+            codegen_crate: PathBuf::from("/tmp/project"),
+            examples_dir: PathBuf::from("/tmp/project"),
+            backend_so: PathBuf::from("/tmp/backend.so"),
+            is_workspace: false,
+            config: OxideConfig {
+                env: vec![("CUDA_OXIDE_EMIT_NVVM_IR".to_string(), "true".to_string())],
+                ..OxideConfig::default()
+            },
+        };
+
+        // The process environment outranks `cuda-oxide.toml`: an explicit
+        // false in the environment wins over the project's `true`, in either
+        // accepted spelling.
+        for disabled in ["false", "0"] {
+            assert_eq!(
+                nvvm_ir_requested_with_env(&ctx, Some(disabled.into())),
+                Ok(false)
+            );
+        }
+    }
+
+    #[test]
     fn scaffold_sync_template_uses_launch_contract_and_docs() {
         let files = scaffold_files("demo_kernel", false);
         assert!(files.cargo_toml.contains("name = \"demo_kernel\""));
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -59,10 +59,22 @@ fn prepare_materialization(
     )
 }
 
-fn nvvm_ir_requested(ctx: &Context) -> Result<bool, String> {
-    const EMIT_NVVM_IR_ENV: &str = "CUDA_OXIDE_EMIT_NVVM_IR";
+const EMIT_NVVM_IR_ENV: &str = "CUDA_OXIDE_EMIT_NVVM_IR";
 
-    if let Some(value) = std::env::var_os(EMIT_NVVM_IR_ENV) {
+fn nvvm_ir_requested(ctx: &Context) -> Result<bool, String> {
+    nvvm_ir_requested_with_env(ctx, std::env::var_os(EMIT_NVVM_IR_ENV))
+}
+
+/// `nvvm_ir_requested` with the ambient `CUDA_OXIDE_EMIT_NVVM_IR` injected.
+///
+/// The process value outranks project config, so resolution has to be
+/// injectable for unit tests: an exported `CUDA_OXIDE_EMIT_NVVM_IR` would
+/// otherwise decide the answer before the configured value is consulted.
+fn nvvm_ir_requested_with_env(
+    ctx: &Context,
+    env_value: Option<std::ffi::OsString>,
+) -> Result<bool, String> {
+    if let Some(value) = env_value {
         let value = value
             .into_string()
             .map_err(|_| format!("{EMIT_NVVM_IR_ENV} is not valid Unicode"))?;
@@ -2586,6 +2598,20 @@ fn resolve_device_codegen_crates(
         .transpose()
 }
 
+/// The ambient environment, in the shape the fingerprint helpers consume.
+///
+/// Split out so both fingerprint wrappers share one collection, and so their
+/// `_with_env` counterparts stay the only entry points a unit test needs.
+fn inherited_process_env() -> BTreeMap<String, Vec<u8>> {
+    std::env::vars_os()
+        .filter_map(|(key, value)| {
+            key.into_string()
+                .ok()
+                .map(|key| (key, value.as_encoded_bytes().to_vec()))
+        })
+        .collect()
+}
+
 fn passthrough_codegen_fingerprint(
     ctx: &Context,
     opts: &CargoPassthroughOptions<'_>,
@@ -2593,20 +2619,13 @@ fn passthrough_codegen_fingerprint(
     target_arch: Option<&str>,
     materialization: &MaterializationMode,
 ) -> String {
-    let inherited_env: BTreeMap<String, Vec<u8>> = std::env::vars_os()
-        .filter_map(|(key, value)| {
-            key.into_string()
-                .ok()
-                .map(|key| (key, value.as_encoded_bytes().to_vec()))
-        })
-        .collect();
     passthrough_codegen_fingerprint_with_env(
         ctx,
         opts,
         owner_filter,
         target_arch,
         materialization,
-        &inherited_env,
+        &inherited_process_env(),
     )
 }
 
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks, 

**One commit on top:** 
- adds an env-over-config precedence pin through the new `nvvm_ir_requested_with_env` seam. 

Merging; 

closes #562.

nihalpasham (approved):

PR #565, opened 2026-07-29
fix(cargo-oxide): keep materialization and backend-pin lookups out of tests

### LINKED ISSUE #564: cargo-oxide: an exported CUDA_OXIDE_MATERIALIZE_CUBIN aborts the whole test binary; CUDA_OXIDE_BACKEND shadows the update pin test
## Summary

Two more `cargo-oxide` unit-test lookups read the ambient environment through
the production code they exercise. Same class as #560 and #562, and both
variables are ones those issues never covered — but the first has a much worse
failure mode than an assertion failure: it **aborts the entire test binary**, so
the tests that had not run yet never run at all.

Reproduced on `main` (`20a5616`), each independently:

| Exported | Effect |
| --- | --- |
| `CUDA_OXIDE_MATERIALIZE_CUBIN=1` (or `=true`) | whole `cargo-oxide` test binary exits 2 mid-run, **no test-result summary** |
| `CUDA_OXIDE_BACKEND=/any/path` | `update_refuses_when_the_config_pins_a_backend` fails |

```bash
# 149 passed clean
cargo test -p cargo-oxide --bins

CUDA_OXIDE_MATERIALIZE_CUBIN=1 cargo test -p cargo-oxide --bins
CUDA_OXIDE_BACKEND=/tmp/backend.so cargo test -p cargo-oxide --bins
```

As with #560 and #562, CI is green because the `cargo-oxide` matrix entry runs on
a stock runner with neither variable set.

## Cause 1: `CUDA_OXIDE_MATERIALIZE_CUBIN` aborts the suite

`materialization_requested` checks the process value before `opts` and project
config, so an exported value overrides `materialize_cubin: false`. Tests that
build a command through `cargo_passthrough_command` therefore enable
materialization, which calls `discover_materializer_provenance`.

That helper re-executes `std::env::current_exe()` with the
`__materializer-provenance` argument. Under `cargo test`, `current_exe()` is the
**libtest binary**, not `cargo-oxide`. libtest treats the argument as a
test-name filter, matches nothing, prints its own summary, and — the part that
makes this sharp — **exits 0**. So the `output.status.success()` check passes and
libtest's own stdout is validated as a provenance digest:

```
test commands::tests::architecture_and_output_mode_do_not_change_global_rustflags ... Error: CUDA materializer discovery returned an invalid provenance digest: "running 0 tests\n\ntest result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 149 filtered out; finished in 0.00s"
```

The digest is rejected, and `prepare_materialization` responds with
`std::process::exit(2)` — from inside a test thread. That kills the harness:

```
error: test failed, to rerun pass `-p cargo-oxide --bin cargo-oxide`

Caused by:
  process didn't exit successfully: `.../deps/cargo_oxide-b3b5418ee7cf0375` (exit status: 2)
note: test exited abnormally; to see the full output pass --no-capture to the harness.
```

Consequences worth separating out:

- There is **no `test result:` line at all**, so the run cannot be summarised.
- Which tests completed first depends on libtest's thread scheduling, so the set
  of tests that got to run is nondeterministic.
- Because it is `exit`, not `panic`, no test can be blamed and nothing is
  recoverable.

To be clear about severity: cargo's exit status is non-zero, so this is not
invisible to CI. The problem is that the signal is opaque — an aborted suite
looks nothing like a failing test, and an unknown number of tests silently did
not execute.

The *production* path is correct: there, `current_exe()` really is `cargo-oxide`,
which handles `__materializer-provenance` and prints a real digest. This is
purely test insulation.

## Cause 2: `CUDA_OXIDE_BACKEND` shadows the project pin

`update_pin_refusal` checks `std::env::var_os("CUDA_OXIDE_BACKEND")` before the
`.cargo/cuda-oxide.toml` pin. With the variable exported, every input returns the
env refusal — including the unpinned `Context` the test asserts returns `None`,
and the pinned one whose message it matches on. Same shape as #560.

## Proposed fix

The seam the crate already prefers (`cuda_toolkit_root(get_env)`,
`passthrough_codegen_fingerprint_with_env`, `nvvm_ir_requested_with_env`): keep
each entry point a thin wrapper that reads the real environment, move the body
into a `_with_env` counterpart taking the value, and have tests call the
counterpart. Production call sites and behavior stay unchanged.

For cause 1 that chain is `materialization_requested` →
`prepare_materialization_result` → `prepare_materialization` →
`cargo_passthrough_command`. This is the follow-up offered at the end of #562:
threading the value through `cargo_passthrough_command`, which turns out to have
exactly **one** production call site, with the other six in the test module. A
single test helper that passes `None` covers all of them.

For cause 2 it is one `_with_env` counterpart on `update_pin_refusal`.

## Known remaining item

`codegen_mode_changes_rebuild_only_the_tracked_device_owner` still fails under
`CUDA_OXIDE_EMIT_NVVM_IR=1`. That is the item #562 documented and deliberately
left out: it spawns a real `cargo`, so the child inherits the ambient
environment regardless of what the in-process fingerprint is told. Unchanged
here, and still needs the child's environment scrubbed.

## Acceptance criteria

- [ ] `cargo test -p cargo-oxide --bins` passes with `CUDA_OXIDE_MATERIALIZE_CUBIN` or `CUDA_OXIDE_BACKEND` exported
- [ ] No unit test can reach `discover_materializer_provenance`
- [ ] Production behavior and call sites unchanged

vyncint:
## Summary

Fixes #564. Two more ambient reads reach `cargo-oxide` unit tests through the
production code they exercise. Same class as #560/#561 and #562/#563, but
neither variable was covered there — and the first does not merely fail an
assertion, it **aborts the whole test binary**:

| Exported | Effect |
| --- | --- |
| `CUDA_OXIDE_MATERIALIZE_CUBIN=1` (or `=true`) | entire suite exits 2 mid-run, **no test-result summary** |
| `CUDA_OXIDE_BACKEND=/any/path` | `update_refuses_when_the_config_pins_a_backend` fails |

CI is green because the `cargo-oxide` matrix entry runs on a stock runner with
neither set.

## What was wrong

**A unit test re-executing the test binary, then killing it.**
`materialization_requested` checks the process value before `opts` and project
config, so an exported `CUDA_OXIDE_MATERIALIZE_CUBIN` overrides
`materialize_cubin: false`. Every test that builds a command through
`cargo_passthrough_command` therefore enabled materialization and reached
`discover_materializer_provenance`, which re-executes `std::env::current_exe()`
with `__materializer-provenance`.

Under `cargo test`, `current_exe()` is the **libtest binary**, not `cargo-oxide`.
libtest reads the argument as a test-name filter, matches nothing, prints its own
summary, and — the part that makes this sharp — **exits 0**. So the
`output.status.success()` check passes and libtest's own stdout is validated as a
provenance digest:

```
Error: CUDA materializer discovery returned an invalid provenance digest:
"running 0 tests\n\ntest result: ok. 0 passed; 0 failed; ... 149 filtered out ..."
```

`prepare_materialization` answers that with `std::process::exit(2)`, from inside
a test thread:

```
error: test failed, to rerun pass `-p cargo-oxide --bin cargo-oxide`

Caused by:
  process didn't exit successfully: `.../deps/cargo_oxide-<hash>` (exit status: 2)
note: test exited abnormally; to see the full output pass --no-capture to the harness.
```

Three things follow, worth separating:

- There is **no `test result:` line**, so the run cannot be summarised.
- Which tests completed first depends on libtest's thread scheduling, so the set
  that got to run is nondeterministic.
- Because it is `exit` and not `panic`, nothing is attributable or recoverable.

On severity, to be straight: cargo's exit status is non-zero, so this is not
invisible to CI. What makes it bad is that an aborted suite looks nothing like a
failing test, and an unknown number of tests silently never ran.

**The production path is correct.** There `current_exe()` really is
`cargo-oxide`, which handles `__materializer-provenance` and prints a real
digest. This is purely test insulation.

**Precedence shadowing.** `update_pin_refusal` checks `CUDA_OXIDE_BACKEND`
before the `.cargo/cuda-oxide.toml` pin, so an exported value returned the env
refusal for every input — including the unpinned context the test asserts is
`None`, and the pinned one whose message it matches on. Same shape as #560.

## Approach

The seam the crate already prefers — `cuda_toolkit_root(get_env)`,
`passthrough_codegen_fingerprint_with_env`, `nvvm_ir_requested_with_env`: each
entry point stays a thin wrapper that reads the real environment, the body moves
to a `_with_env` counterpart taking the value, and tests call the counterpart.
Production call sites and behavior are unchanged.

For materialization that chain is `materialization_requested` →
`prepare_materialization_result` → `prepare_materialization` →
`cargo_passthrough_command` — the follow-up offered at the end of #562. It turned
out to be contained: `cargo_passthrough_command` has **one** production call
site, and the six test sites now go through one `passthrough_command_for_test`
helper that passes `None`. No test can reach materializer discovery.

`prepare_materialization_with_env` keeps the existing `exit(2)` semantics rather
than converting the error to a `Result`, to keep this change to test insulation.
Its doc comment records that exiting from a test aborts the suite, which is the
reason the injection matters.

## Verification

No GPU and no CUDA toolkit required.

```bash
cargo test -p cargo-oxide --bins                                    # 149 passed
CUDA_OXIDE_MATERIALIZE_CUBIN=1    cargo test -p cargo-oxide --bins  # 149 passed (was: suite aborted)
CUDA_OXIDE_MATERIALIZE_CUBIN=true cargo test -p cargo-oxide --bins  # 149 passed
CUDA_OXIDE_BACKEND=/tmp/b.so      cargo test -p cargo-oxide --bins  # 149 passed (was: 1 failed)
```

Also re-swept the variables #562 cleared (`CUDA_OXIDE_TARGET`, `_DEBUG`,
`_NO_FMA`, `_UNCHECKED_INDEXING`, `_VERBOSE`, `_NO_OPT`, `_OPT`, `_LLC`,
`_LIBDEVICE`, `_PTX_DIR`, `_DEVICE_ARCH`, `_DUMP_MIR`, `_SHOW_RUSTC_MIR`,
`_DEVICE_CODEGEN_CRATE`, plus `CUDA_TOOLKIT_PATH`, `CUDA_HOME`, `RUSTFLAGS`,
`CARGO_ENCODED_RUSTFLAGS`) — all still clean. `cargo fmt` and
`cargo clippy -p cargo-oxide --all-targets` are clean.

The `update_pin_refusal` test also gains a case asserting the env value outranks
the project pin, which the previous shape could not express.

## Out of scope

`codegen_mode_changes_rebuild_only_the_tracked_device_owner` still fails under
`CUDA_OXIDE_EMIT_NVVM_IR` — the item #562 documented and deliberately left out.
It spawns a real `cargo`, so the child inherits the ambient environment
regardless of what the in-process fingerprint is told; insulating it needs the
child's environment scrubbed. Untouched here, and still open.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -59,6 +59,33 @@ fn prepare_materialization(
     )
 }
 
+/// `prepare_materialization` with the ambient `CUDA_OXIDE_MATERIALIZE_CUBIN`
+/// injected, so `cargo_passthrough_command_with_env` can reach
+/// `materialization_requested_with_env`.
+///
+/// Note this still exits the process on error, which inside a unit test aborts
+/// the whole test binary rather than failing one case -- a further reason tests
+/// must not reach the ambient read.
+fn prepare_materialization_with_env(
+    ctx: &Context,
+    cli_requested: bool,
+    cli_arch: Option<&str>,
+    emit_nvvm_ir: bool,
+    materialize_env: Option<std::ffi::OsString>,
+) -> MaterializationMode {
+    prepare_materialization_result_with_env(
+        ctx,
+        cli_requested,
+        cli_arch,
+        emit_nvvm_ir,
+        materialize_env,
+    )
+    .unwrap_or_else(|error| {
+        eprintln!("Error: {error}");
+        std::process::exit(2);
+    })
+}
+
 const EMIT_NVVM_IR_ENV: &str = "CUDA_OXIDE_EMIT_NVVM_IR";
 
 fn nvvm_ir_requested(ctx: &Context) -> Result<bool, String> {
@@ -89,11 +116,27 @@ fn nvvm_ir_requested_with_env(
 }
 
 fn materialization_requested(ctx: &Context, cli_requested: bool) -> Result<bool, String> {
+    materialization_requested_with_env(ctx, cli_requested, std::env::var_os(MATERIALIZE_ENV))
+}
+
+/// `materialization_requested` with the ambient `CUDA_OXIDE_MATERIALIZE_CUBIN`
+/// injected.
+///
+/// The process value outranks project config, so resolution has to be
+/// injectable for unit tests: an exported value would otherwise turn
+/// materialization on for tests that pass `materialize_cubin: false`, sending
+/// them into `discover_materializer_provenance`. Same rationale as
+/// `nvvm_ir_requested_with_env`.
+fn materialization_requested_with_env(
+    ctx: &Context,
+    cli_requested: bool,
+    env_value: Option<std::ffi::OsString>,
+) -> Result<bool, String> {
     if cli_requested {
         return Ok(true);
     }
 
-    if let Some(value) = std::env::var_os(MATERIALIZE_ENV) {
+    if let Some(value) = env_value {
         let value = value
             .into_string()
             .map_err(|_| format!("{MATERIALIZE_ENV} is not valid Unicode"))?;
@@ -113,7 +156,26 @@ fn prepare_materialization_result(
     cli_arch: Option<&str>,
     emit_nvvm_ir: bool,
 ) -> Result<MaterializationMode, String> {
-    let enabled = materialization_requested(ctx, cli_requested)?;
+    prepare_materialization_result_with_env(
+        ctx,
+        cli_requested,
+        cli_arch,
+        emit_nvvm_ir,
+        std::env::var_os(MATERIALIZE_ENV),
+    )
+}
+
+/// `prepare_materialization_result` with the ambient
+/// `CUDA_OXIDE_MATERIALIZE_CUBIN` injected, forwarded to
+/// `materialization_requested_with_env`.
+fn prepare_materialization_result_with_env(
+    ctx: &Context,
+    cli_requested: bool,
+    cli_arch: Option<&str>,
+    emit_nvvm_ir: bool,
+    materialize_env: Option<std::ffi::OsString>,
+) -> Result<MaterializationMode, String> {
+    let enabled = materialization_requested_with_env(ctx, cli_requested, materialize_env)?;
     if !enabled {
         return Ok(MaterializationMode::default());
     }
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM as-is, 

I reproduced both failure modes on main 
- an exported `CUDA_OXIDE_MATERIALIZE_CUBIN` kills the whole test binary mid-run with no summary line; 
- `CUDA_OXIDE_BACKEND` fails the update-pin test) and 

**Verified locally:** 
- 149 tests pass plain and 
- with each variable exported, 
- fmt/clippy clean. 
 
The one residual you already documented - `CUDA_OXIDE_EMIT_NVVM_IR` still reaching the spawned cargo child in one test - stays out of scope per #562.

Merging; closes #564.

nihalpasham (approved):

PR #567, opened 2026-07-29
test(cuda-oxide-codegen): resolve ptxas through the toolkit contract

### LINKED ISSUE #566: spine_kernel_ptx locates ptxas without consulting CUDA_TOOLKIT_PATH/CUDA_HOME, so the suite fails on a non-default toolkit prefix
## Summary

`cuda-oxide-codegen`'s `spine_kernel_ptx` test locates `ptxas` by probing a
hardcoded `/usr/local/cuda/bin/ptxas` and then falling back to bare `ptxas` on
`PATH`. It never consults `CUDA_TOOLKIT_PATH` or `CUDA_HOME`.

So on a checkout whose CUDA toolkit lives at any other prefix — which the build
scripts fully support — `cargo test --workspace` fails, even though this test
needs no GPU:

```
$ echo $CUDA_TOOLKIT_PATH
/home/user/cuda-13.3
$ ls $CUDA_TOOLKIT_PATH/bin/ptxas
/home/user/cuda-13.3/bin/ptxas          # present, and CUDA_TOOLKIT_PATH points at it

$ cargo test -p cuda-oxide-codegen --test spine_kernel_ptx
thread 'spine_add_kernel_emits_entry_and_validates' panicked at
crates/cuda-oxide-codegen/tests/spine_kernel_ptx.rs:348:28:
ptxas runs: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

Reproduced on `main` (`20a5616`). Adding the same toolkit's `bin` to `PATH` makes
it pass unchanged, which isolates discovery as the only cause.

## Why this is inconsistent rather than just unlucky

`CUDA_TOOLKIT_PATH` → `CUDA_HOME` → `/usr/local/cuda` is the toolkit contract
the rest of the repo already implements:

- `crates/cuda-bindings/build.rs` resolves `cuda.h` through it, and its failure
  message states it outright: *"Set CUDA_TOOLKIT_PATH or CUDA_HOME to a CUDA
  Toolkit install root; when neither is set, /usr/local/cuda is used."*
- `cargo oxide doctor`'s `cuda_toolkit_root` implements exactly that precedence,
  and its doc comment says it is kept in lockstep by hand with that build script.

This test implements only the last step of that chain. `.github/workflows/unit-tests.yml`
even sets `CUDA_TOOLKIT_PATH` for this job — the test just ignores it and happens
to pass because `Jimver/cuda-toolkit` also puts `ptxas` on `PATH`.

Worth being clear on scope: this is a **test-only** defect. No production
discovery path is affected, and CI is correct today. The cost is local: a
contributor who installs the toolkit anywhere but `/usr/local/cuda` cannot run
the workspace suite. That prefix needs root, so a user-prefix install is the
normal arrangement on a shared or unprivileged machine — and `ptxas` is CPU-only,
so this is otherwise exactly the kind of test that *should* run there.

## Secondary: the diagnostic gives nothing to act on

`.expect("ptxas runs")` renders the raw `io::Error`:

```
ptxas runs: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

That names neither the path tried nor the variables that would fix it, and
`NotFound` here is easy to misread as the *PTX* file being missing rather than
`ptxas` itself.

## Proposed fix

Resolve `ptxas` in the documented order — `CUDA_TOOLKIT_PATH`, then `CUDA_HOME`
(first non-empty wins), then `/usr/local/cuda`, then bare `ptxas` for `PATH` —
and report the attempted path plus the two variables when the spawn fails. The
`PATH` fallback stays last, so CI behavior is unchanged.

## Acceptance criteria

- [ ] The test passes with a toolkit at a non-default prefix named only by `CUDA_TOOLKIT_PATH` or `CUDA_HOME`
- [ ] Still passes when neither is set and `ptxas` is on `PATH` (current CI shape)
- [ ] A blank/whitespace variable does not shadow the later candidates
- [ ] Failure names the path tried and how to fix it
- [ ] No GPU required

vyncint:
## Summary

Fixes #566. `spine_kernel_ptx` located `ptxas` by probing a hardcoded
`/usr/local/cuda/bin/ptxas` and then falling back to bare `ptxas` on `PATH`,
never consulting `CUDA_TOOLKIT_PATH` or `CUDA_HOME`. A checkout whose toolkit
lives at any other prefix failed the workspace suite even with the variable
pointing straight at a working `ptxas`:

```
$ ls $CUDA_TOOLKIT_PATH/bin/ptxas
/home/user/cuda-13.3/bin/ptxas

$ cargo test -p cuda-oxide-codegen --test spine_kernel_ptx
ptxas runs: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

Putting that same toolkit's `bin` on `PATH` made it pass unchanged, which
isolates discovery as the only cause.

## Why this is an inconsistency, not just bad luck

`CUDA_TOOLKIT_PATH` → `CUDA_HOME` → `/usr/local/cuda` is the contract the rest of
the repo already implements:

- `crates/cuda-bindings/build.rs` resolves `cuda.h` through it, and says so in
  its own failure message: *"Set CUDA_TOOLKIT_PATH or CUDA_HOME to a CUDA Toolkit
  install root; when neither is set, /usr/local/cuda is used."*
- `cargo oxide doctor`'s `cuda_toolkit_root` implements that precedence, with a
  doc comment noting it is kept in lockstep by hand with that build script.

This test implemented only the last step. `unit-tests.yml` even sets
`CUDA_TOOLKIT_PATH` for this job — the test ignored it and passed only because
`Jimver/cuda-toolkit` also puts `ptxas` on `PATH`.

To be clear on scope: **test-only.** No production discovery path is affected and
CI is correct today. The cost is local — `/usr/local/cuda` needs root, so a
user-prefix install is the normal arrangement on a shared or unprivileged
machine, and `ptxas` is CPU-only, making this exactly the kind of test that
should run there.

## Approach

Resolve in the documented order, with the bare-`ptxas` `PATH` fallback kept
**last** so CI behavior is unchanged, and blank values skipped so an empty
variable cannot shadow the later candidates.

The failure message now names the path tried and the variables that would fix
it. The previous raw `io::Error` named neither, and its `NotFound` reads as
easily as the PTX file being missing as `ptxas` itself:

```
could not run ptxas: No such file or directory (os error 2)
This test needs `ptxas` from a CUDA toolkit (no GPU required). Point
CUDA_TOOLKIT_PATH or CUDA_HOME at the install root, or put `ptxas` on PATH.
```

## Verification

No GPU required. Toolkit at `~/cuda-13.3`, nothing at `/usr/local/cuda`:

| Case | Result |
| --- | --- |
| `CUDA_TOOLKIT_PATH` only | passes (**was failing**) |
| `CUDA_HOME` only | passes |
| neither set, `ptxas` on `PATH` (current CI shape) | passes |
| `CUDA_TOOLKIT_PATH="  "` + `ptxas` on `PATH` | passes — blank does not shadow |
| nothing anywhere | fails with the actionable message above |

`cargo fmt` and `cargo clippy -p cuda-oxide-codegen --all-targets` are clean.

### DIFF
--- a/crates/cuda-oxide-codegen/tests/spine_kernel_ptx.rs
+++ b/crates/cuda-oxide-codegen/tests/spine_kernel_ptx.rs
@@ -271,6 +271,31 @@ fn build_unused_helper(module: &mut CodegenModule) {
     });
 }
 
+/// Locate `ptxas`. Discovery order, mirroring the toolkit contract that
+/// `crates/cuda-bindings/build.rs` and `cargo oxide doctor`'s
+/// `cuda_toolkit_root` already implement:
+///
+/// 1. `CUDA_TOOLKIT_PATH`, then `CUDA_HOME` — first non-empty one wins
+/// 2. `/usr/local/cuda`, the conventional default prefix
+/// 3. bare `ptxas`, leaving resolution to `PATH`
+///
+/// Step 3 is what CI relies on, since the toolkit action puts `ptxas` on `PATH`.
+/// Steps 1 and 2 are what a local checkout relies on: the build scripts fully
+/// support a toolkit installed at any prefix, so this test should not be the one
+/// place that ignores where it was told the toolkit lives.
+fn find_ptxas() -> std::path::PathBuf {
+    ["CUDA_TOOLKIT_PATH", "CUDA_HOME"]
+        .iter()
+        .filter_map(|var| std::env::var(var).ok())
+        .filter(|root| !root.trim().is_empty())
+        .map(|root| std::path::PathBuf::from(root).join("bin/ptxas"))
+        .chain(std::iter::once(std::path::PathBuf::from(
+            "/usr/local/cuda/bin/ptxas",
+        )))
+        .find(|candidate| candidate.exists())
+        .unwrap_or_else(|| std::path::PathBuf::from("ptxas"))
+}
+
 #[test]
 fn spine_add_kernel_emits_entry_and_validates() {
     let mut module = CodegenModule::new("spine_module").unwrap();
@@ -327,14 +352,10 @@ fn spine_add_kernel_emits_entry_and_validates() {
     let cubin_path = dir.join("spine.cubin");
     std::fs::write(&ptx_path, &ptx).unwrap();
 
-    let ptxas = if std::path::Path::new("/usr/local/cuda/bin/ptxas").exists() {
-        "/usr/local/cuda/bin/ptxas"
-    } else {
-        "ptxas"
-    };
+    let ptxas = find_ptxas();
     // Capture the Result before cleanup so the scratch dir is always removed,
     // even when ptxas is absent and `.output()` would otherwise panic.
-    let ptxas_result = std::process::Command::new(ptxas)
+    let ptxas_result = std::process::Command::new(&ptxas)
         .arg("-arch=sm_120")
         .arg("--compile-only")
         .arg(&ptx_path)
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM as-is. 

**One wrinkle in verification:** 
- on my host the real toolkit sits at the hardcoded fallback path, so passing there proves nothing.  
- I pointed `CUDA_TOOLKIT_PATH` at a scratch prefix with a logging ptxas wrapper (`CUDA_HOME` unset, `PATH` stripped) and the marker confirmed the env-var candidate resolves first. 
- Blank-variable fall-through 

Verified: 
- the normal CI shape both pass; 
- fmt/clippy clean.

Merging; 

closes #566.

nihalpasham (approved):

PR #555, opened 2026-07-28
docs(cargo-oxide): sync README Usage and Flags with shipped CLI

Ayush7614:
## Summary
- Add `inspect`, `clean`, and `update [--force]` to the cargo-oxide README Usage block.
- Update the Flags table for `inspect` / `test` applicability, `--unchecked-indexing`, and `--force`.

Closes #550

## Test plan
- [x] Cross-checked against `Commands` in `crates/cargo-oxide/src/main.rs`

PR #552, opened 2026-07-28
feat(cargo-oxide): forward codegen flags through cargo oxide test

Ayush7614:
## Summary
- Add `--no-fmad` and `--unchecked-indexing` to `cargo oxide test` and forward them into passthrough options.
- Update README Flags table; add CLI parser unit tests.

Closes #547

## Test plan
- [x] `cargo test -p cargo-oxide test_parser_accepts_codegen_flags`
- [x] `cargo test -p cargo-oxide empty_test_and_explicit`
- [x] `cargo oxide test --help` shows both flags

PR #568, opened 2026-07-29
feat(cargo-oxide): expose the device debug policy as --lineinfo / --device-debug

honeyspoon:
The device debug policy is already plumbed end to end; only a CLI flag was missing, so device line info was reachable from an env var but not from `cargo oxide`.

| Layer | Already present |
|---|---|
| artifact | `ArtifactCompileOptions` debug bits (`OPTION_DEBUG_LINE_TABLES` / `_FULL`) |
| sidecar | `compile-options-v2` line-table and full-debug constants |
| finalizer | `FinalizationOptions::with_debug_policy` |
| libNVVM | `-g` + `-opt=0` for full debug |
| nvJitLink | `-lineinfo` or `-g` |
| selection | `CUDA_OXIDE_DEBUG` (`off` / `line` / `full`) |

Adds `--lineinfo` and `--device-debug` to the seven subcommands carrying `--no-fmad`, including `test` since #552 gave it the codegen flags.

---

## 1. Two boolean flags, not one `--device-debug=<level>`

**Chosen:** `--lineinfo` and `--device-debug` as independent booleans, resolved by `DeviceDebug::from_flags`.

**Alternative:** a single valued flag, `--device-debug=off|line|full` (clap `ValueEnum`).

The valued form is tidier and I would prefer it in isolation. I chose two booleans to mirror **nvcc**, where these are `-lineinfo` and `-G` — separate switches that a user already has in muscle memory and in existing build scripts. The cost is having to define what both-at-once means (see below) and two flags where one would do.

## 2. Both flags together yields `Full`, rather than erroring

**Chosen:** ordered resolution — `device_debug` wins.

**Alternative:** clap `conflicts_with`, making the combination a usage error.

Full debug already *carries* line tables, so the combination is redundant rather than contradictory; rejecting it would fail a script that passes both defensively. `from_flags` encodes the order and a test pins all four combinations.

## 3. Absent flags export **nothing**, not `CUDA_OXIDE_DEBUG=off`

**Chosen:** `DeviceDebug::Off.env_value() -> None`.

**Alternative:** always export, so the flag fully determines the level.

Always-exporting would make an *absent* flag silently cancel a level the environment or `cuda-oxide.toml` had already requested — turning an omission into an active opt-out. This follows the existing `apply_default_sanitizer_line_tables`, which documents the same deference ("An explicit process or project setting remains authoritative, including an intentional `CUDA_OXIDE_DEBUG=off`"). A test pins `Off -> None`.

## 4. An explicit flag outranks the environment

**Chosen:** flag wins where given.

Matches `--no-fmad` vs `CUDA_OXIDE_NO_FMA`. Combined with (3): the flag is authoritative when present, invisible when absent — which is the only combination where neither source can surprise you.

## 5. The policy participates in the codegen fingerprint

**Chosen:** threaded through `standard_`, `sanitize_`, `pipeline_` and `interop_codegen_fingerprint`, and the `_with_env` variant added by #563.

**Alternative:** leave it out — it is "just debug info".

It is not: full debug passes `-opt=0` to libNVVM, so it changes generated code. Sharing a fingerprint with the default would let Cargo hand back device code built *without* debug info, which is the silent-stale-artifact failure the fingerprint exists to prevent. A test asserts the three policies produce three distinct fingerprints, using an empty inherited env for the same reason the sibling tests do — an ambient `CUDA_OXIDE_DEBUG` is folded in on its own and would collapse them onto the base.

## 6. `DeviceDebug` mirrors `ArtifactDebugPolicy` rather than reusing it

**Chosen:** a small CLI-side enum in `commands.rs`.

**Alternative:** have the CLI take `oxide_artifacts::ArtifactDebugPolicy` directly.

They differ in one way that matters: `DeviceDebug::Off` means "say nothing", whereas `ArtifactDebugPolicy::None` means "no debug info". Collapsing them would lose decision (3). The conversion happens where the env var is written, and `env_value` is pinned by a test against the strings `device_debug_kind_with_override` accepts — a typo there would silently fall through to the profile-derived default rather than fail.

## 7. `cargo oxide debug` and interop builds pass `Off`

Neither is driven by these flags: `debug` sets its own dump environment, and interop device builds take policy from config. The sanitizer path keeps its existing line-table default, which decision (3) is what makes possible.

# Note on the rebase

Not a clean rebase — `cargo-oxide` moved underneath. **#552** gave `test` the codegen flags, **#555** rewrote the README table, and **#563** refactored the fingerprint tests around `sanitize_codegen_fingerprint_with_env`, which did not exist on my branch. Checking the three files out wholesale showed 367 deletions, i.e. it would have reverted all of that, so I re-applied change by change instead. The rework is strictly better: `test` is now covered rather than deliberately skipped.

# Gates

fmt, clippy `--all-targets -D warnings`, rustdoc `-D warnings` clean on `nightly-2026-04-03`. `cargo test -p cargo-oxide`: 152 passed, 1 failed — `cargo_json_selects_custom_bin_in_configured_target_dir`, which **reproduces identically on an untouched `main`**.

+354/−5; all five deletions are my own `apply_common_codegen_env` call lines gaining a parameter, so no upstream content is reverted.

### DIFF
--- a/crates/cargo-oxide/README.md
+++ b/crates/cargo-oxide/README.md
@@ -63,6 +63,8 @@ cargo oxide update --force          # inside the workspace, run setup via update
 | `-v, --verbose`              | run, sanitize, build, test, emit-ltoir, inspect | Show detailed compilation output           |
 | `--no-fmad`                  | run, sanitize, build, test, emit-ltoir, pipeline, inspect | Keep ordinary multiply and add/subtract operations separate |
 | `--unchecked-indexing`       | run, sanitize, build, test, emit-ltoir, pipeline, inspect | Elide device slice/array bounds checks (UB on OOB) |
+| `--lineinfo`                 | run, sanitize, build, test, emit-ltoir, pipeline, inspect | Emit device line-number info for profilers (nvcc `-lineinfo`) |
+| `--device-debug`             | run, sanitize, build, test, emit-ltoir, pipeline, inspect | Emit full device debug info; disables libNVVM optimization (nvcc `-G`) |
 | `--force`                    | update                           | Inside the workspace, run `setup` instead of advising it |
 | `--async`                    | new                              | Use the async template                          |
 | `--cgdb`                     | debug                            | Use cgdb instead of cuda-gdb                    |
@@ -82,6 +84,20 @@ For NVVM IR and LTOIR, cuda-oxide records the policy in matching `.options`
 and versioned `.target` files and passes `-fma=0` through libNVVM and
 nvJitLink. Keep both sidecars with the artifact when another build system
 consumes it.
+
+`--lineinfo` and `--device-debug` select the device debug policy, mirroring nvcc's
+`-lineinfo` and `-G`. `--lineinfo` preserves source line mappings with optimization
+intact and reaches nvJitLink as `-lineinfo`; `--device-debug` emits full debug
+information and additionally passes `-g` with `-opt=0` to libNVVM, so device
+optimization is disabled. Asking for both yields full debug, because full debug
+already carries line tables.
+
+Both are equivalent to `CUDA_OXIDE_DEBUG=line` and `CUDA_OXIDE_DEBUG=full`, and an
+explicit flag outranks that variable. Omitting the flags exports nothing rather than
+`off`, so an absent flag never cancels a debug level the environment or
+`cuda-oxide.toml` already requested. Because the policy changes what the CUDA compiler
+stages are asked to do, it participates in the codegen fingerprint: switching it
+rebuilds device code instead of reusing artifacts compiled under a different policy.
 
 `--materialize-cubin` moves the remaining libNVVM and nvJitLink work into the
 host build. The executable embeds a native cubin, so deployment does not need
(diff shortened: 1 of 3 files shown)


honeyspoon:

```
```

nihalpasham:
LGTM. 

**Two commits on top:**
- `pipeline` accepted the new flags and re-fingerprinted the build but never exported `CUDA_OXIDE_DEBUG` to the child, so the flag changed nothing. It now forwards like no-fmad/unchecked-indexing: pipeline should show the program your build actually produces, and full debug turns optimization off entirely.
- `sanitize`'s line-tables default was applied over the flag, silently downgrading --device-debug back to line tables. The flag now wins, with a unit test checking sanitize `--device-debug` really sets `CUDA_OXIDE_DEBUG=full`.

**Verified on hardware:** 
- `vecadd --lineinfo` emits `.file`/`.loc` in the PTX, 
- `--device-debug` carries `-g -opt=0` through finalization, 
- 154/154 crate tests. 
 
Your reported pre-existing failure (`cargo_json_selects_custom_bin_in_configured_target_dir`) doesn't reproduce here, on main or on your branch.

Merging.

nihalpasham (approved):

PR #570, opened 2026-07-29
fix(mir-importer): support device-static slice addends

### LINKED ISSUE #569: Support interior addends in device-static array-to-slice unsizing
# Body

## Problem

cuda-oxide supports two related static-pointer cases:

1. Non-zero byte addends for pointers to sized subobjects of device statics.
2. Zero-addend array-to-slice unsizing, where the importer constructs a fat slice value from a static data pointer and the length stored in the constant.

These paths are not currently combined.

An interior reference that is unsized from an array reference to a slice is rejected:

```rust
static TABLE: [[f32; 2]; 4] = [
    [0.25, 0.5],
    [1.0, 2.0],
    [4.0, 8.0],
    [16.0, 32.0],
];

const PAIR_SLICE: &[f32] = &TABLE[2];
```

The data pointer in `PAIR_SLICE` refers to `TABLE` with a 16-byte addend, while the second word of the constant stores the slice length `2`.

The importer currently rejects this because the pointee type `[f32]` is unsized, even though it already has all the information required to construct the fat pointer.

## Current behavior

Compilation fails with a diagnostic similar to:

```text
constant pointer into device static TABLE has byte offset 16,
but pointee type [f32] is unsized; cuda-oxide does not yet
preserve the fat-pointer metadata an interior slice or DST
pointer needs
```

## Expected lowering

For a same-element static array-to-slice unsize, the importer should:

```text
device-static base pointer
    -> cast to a byte pointer in the same CUDA address space
    -> apply the rustc byte addend
    -> cast to a pointer to the slice element type
    -> read the stored slice length metadata
    -> construct the slice value from { data pointer, length }
```

The existing static-pointer addend lowering and `MirConstructSliceOp` should be reused. No new MIR dialect operation should be required.

## Scope

* Support non-zero addends for same-element array-to-slice unsizing.
* Recognize array regions nested inside outer arrays.
* Validate that the addend identifies an element-aligned array region.
* Read the slice length from the constant's metadata word.
* Reject lengths that exceed the selected array region.
* Preserve the static's CUDA address space and pointer mutability.
* Keep other unsized pointees, trait objects, `str`, and unrelated pointer reinterpretations unsupported.
* Convert `error_static_slice_addend` into a positive runtime regression example.

## Acceptance criteria

* `const PAIR_SLICE: &[f32] = &TABLE[2]` compiles for device code.
* The generated data pointer retains the 16-byte addend.
* The generated fat pointer stores length `2`.
* A runtime test reads `[4.0, 8.0]`, produces sum `12.0`, and reports length `2`.
* Zero-addend static array-to-slice behavior remains unchanged.
* Unsupported non-array DST pointers continue to fail explicitly.
* `error_static_slice_addend` is removed from the expected-error classification and from `STATUS.md`.

uurl:
Closes #569 

## Summary

Support array-to-slice unsizing when a constant slice points to an interior region of a device static.

The importer already preserves non-zero byte addends for sized static pointees and constructs fat slices for zero-addend static arrays. This change combines those paths for same-element array regions while preserving the existing fail-closed behavior for unrelated DST pointers.

### Detect interior array regions

Add a conservative classifier for non-zero static-pointer addends whose final pointee is `[T]`.

The classifier:

* walks nested array types using rustc layout information;
* verifies that the offset identifies an element-aligned region;
* requires the source array element type to match the slice element type;
* computes the number of elements remaining in the selected array region.

Struct, tuple, enum, trait-object, `str`, and unrelated reinterpretation cases remain unsupported.

### Preserve slice metadata

Read the slice length from the constant's fat-pointer metadata word using the existing `slice_len_from_constant` path.

Reject the constant if its stored length exceeds the selected array region's remaining capacity.

### Materialize the interior data pointer

Extend `translate_static_array_as_slice` to accept the static byte addend.

For a non-zero addend, reuse `translate_static_global_pointer` to:

1. Materialize the device-static base address.
2. Cast it to an `u8` pointer in the same CUDA address space.
3. Apply the byte addend with `mir.ptr_offset`.
4. Cast the result to a pointer to the slice element type.
5. Normalize the pointer to the exact Rust operand type.

Pair that data pointer with the stored length using `MirConstructSliceOp`.

No new MIR or LLVM operation is introduced.

### Promote the regression example

Rename:

```text
error_static_slice_addend
```

to:

```text
static_slice_addend
```

Convert it from an expected compilation failure into a runtime test that verifies:

* `&TABLE[2]` addresses `[4.0, 8.0]`;
* the resulting slice has length `2`;
* reading both elements produces `12.0`.

Remove the old fixture from:

* `crates/rustc-codegen-cuda/STATUS.md`;
* `ERROR_EXAMPLES` in `scripts/smoketest.sh`.

Update the zero-addend example documentation to reference the new positive interior-addend example.

## Test plan

```text
cargo fmt --all -- --check
cargo test -p mir-importer
cargo oxide run static_slice_addend
cargo oxide run static_slice_unsize
scripts/smoketest.sh --compile-only -o '^(static_slice_addend|static_slice_unsize)$'
```

Expected runtime result:

```text
static_slice_addend: PASS (sum 12, len 2)
```


nihalpasham:
LGTM. 

**One commit on top:** dropped the comment-only rewording hunks in `rvalue.rs` - four other PRs in flight touch the same file, and two of the rewordings were grammar related. The surviving diff is purely the feature.

**Verified on hardware:** 
- `static_slice_addend` prints PASS (sum 12, len 2), 
- `static_slice_unsize`'s zero-addend path is unchanged, and 
- `error_static_slice_addend` is retired from the error fixtures now that the pattern compiles.

Merging; 

closes #569.

nihalpasham (approved):

PR #571, opened 2026-07-29
feat(disjoint): derive multi-element and grid-stride index proofs

honeyspoon:
`DisjointSlice::get_mut` takes a `ThreadIndex` and yields **one** element. A thread that owns *N contiguous* elements has no way to say so, so multi-element and grid-stride kernels fall back to `get_unchecked_mut`.

This adds two derivations, each producing a proof the next step consumes:

```rust
let block  = thread::index_1d().scale::<4>();     // owns 4 contiguous elements
let tiling = block.repeat::<8>();                 // ... 8 times, at a derived period
let chunk: &mut [f32; 4] = slice.get_block_mut(block)?;
slice.for_each_tile::<4, 8>(tiling, |pass, chunk| { /* ... */ });
```

* `ThreadIndex::scale::<N>() -> DisjointBlock<'kernel, N, IndexSpace>` — multiplying a disjoint index by a constant keeps it disjoint.
* `DisjointBlock::repeat::<K>() -> DisjointTiling<'kernel, N, K>` — grid-stride, period derived.
* `DisjointSlice::get_block_mut::<N>(block) -> Option<&mut [T; N]>`, plus `for_each_tile::<N, K>` returning the number of passes actually visited.

---

## 1. New proof types, not a length on `ThreadIndex`

**Chosen:** `DisjointBlock<'kernel, N, IndexSpace>` and `DisjointTiling<'kernel, N, K>` as distinct types.

**Alternative:** add a length parameter to `ThreadIndex` and keep one type.

`ThreadIndex` means "this thread owns element *i*". Overloading it to mean "owns a run" changes what every existing consumer can assume, and `get_mut` would have to reject the run case at runtime. Separate types keep each proof exactly as strong as its name, and make `get_block_mut` unable to accept a bare index by construction.

## 2. `const N` rather than a runtime length

**Chosen:** const generic, so `get_block_mut` returns `&mut [T; N]`.

**Alternative:** `scale(n: usize)` returning a slice-shaped proof.

The array type is the point — it is what lets the caller index without further bounds checks and what an over-aligned element type would need. A runtime length can only give `&mut [T]`, which re-introduces a length the optimiser must track. Cost: the tile width must be a compile-time constant, which is true of every tiled kernel I have seen but is a real restriction; a runtime variant can be added later without changing this one.

## 3. `repeat` **derives** the period instead of accepting a stride

**Chosen:** `period = gridDim.x * blockDim.x * N`, computed inside `repeat`.

**Alternative:** `repeat(stride: usize, passes: usize)`.

This is the decision I would most like reviewed. A passed-in stride is the classic grid-stride bug: it silently disagrees with the launch geometry and threads overlap. Deriving it means the proof cannot be constructed wrongly. The cost is that `repeat` reads `gridDim`/`blockDim`, so it is only meaningful inside a kernel — and it forecloses deliberately non-standard strides. If you want those reachable, the shape would be a separate `repeat_with_stride` marked `unsafe`, not a parameter here.

## 4. Infallible construction, `is_valid()` queried later

**Chosen:** `scale`/`repeat` always return a value; `is_valid()` reports whether the derived range is usable, and `get_block_mut` returns `None` when it is not.

**Alternative:** return `Option<DisjointBlock<N>>` from `scale`, making an invalid proof unrepresentable.

The `Option` version is more principled and I nearly went that way. I chose infallible because the interesting failure — the tile running past the slice — depends on the *slice*, which `scale` has never seen, so an `Option` there would only catch overflow in the index arithmetic and give false confidence about the rest. Validity is checked once, at the point that knows the length. Happy to flip this if you would rather have the stricter constructor.

## 5. `for_each_tile` takes a closure and returns a count

**Chosen:** `f: impl FnMut(usize, &mut [T; N])`, returning passes visited.

**Alternative:** an iterator of `&mut [T; N]`.

An iterator is the more idiomatic Rust, but each item borrows the slice mutably, so a lending iterator is required — which means either GATs plumbing or handing out raw pointers. The closure keeps the borrow trivially sound today. The returned count matters because a grid-stride loop legitimately visits fewer than `K` passes at the tail, and silently doing fewer is exactly the bug worth surfacing.

## 6. Overflow is a stop, never a wrap

Every offset uses `checked_*`: a pass whose offset or end would wrap breaks the loop rather than folding into a passing bounds test, and `get_block_mut` uses `start.checked_add(N)?` so a tile straddling the end of the address space cannot alias low memory. Unusual to state, but this is index arithmetic guarding an `unsafe` block, so wrap-on-overflow would be a soundness hole rather than a wrong answer.

---

# What this is *not*

**It makes no performance claim.** An earlier fork PR of mine proposed 32-bit index arithmetic on the theory that it would cut bounds-check cost; the measurement did not reproduce and I closed it. This is a different change. On the codebase that motivated it, `get_block_mut` on a `DisjointSlice<f32>` yields align-4 arrays that stay scalar, so it is instruction-neutral. The value is that a multi-element write stops needing `unsafe`.

# Gates

fmt, clippy `--workspace --all-targets -D warnings`, rustdoc `-D warnings`, tests and doctests clean on `nightly-2026-04-03`. +617/−1 against `main`; the single deletion is a `use crate::thread::{...}` line being rewritten.

The period counterexample in the tests routes through an `at(k, t, period, n)` helper rather than literal `0 *` / `1 *` terms, because clippy's `erasing_op`/`identity_op` reject those under `-D warnings`.

### DIFF
--- a/crates/cuda-device/src/disjoint.rs
+++ b/crates/cuda-device/src/disjoint.rs
@@ -679,17 +679,26 @@ mod tiling_algebra_tests {
     #[test]
     fn a_period_below_grid_span_collides() {
         let (grid, n) = (4usize, 1usize);
-        let bad_period = 2; // G·N would be 4.
-        // (k=0, t=2) and (k=1, t=0) both land on element 2.
-        let a = 0 * bad_period + 2 * n;
-        let b = 1 * bad_period + 0 * n;
-        assert_eq!(a, b, "the counterexample must actually collide");
 
-        // The derived period keeps them apart.
+        // `phi` with the period supplied rather than derived, so a wrong period
+        // can be expressed at all. Written as a call so the (k, t) digit pair
+        // stays visible at each use.
+        fn at(k: usize, t: usize, period: usize, n: usize) -> usize {
+            k * period + t * n
+        }
+
+        let bad_period = 2; // G·N would be 4.
+        assert_eq!(
+            at(0, 2, bad_period, n),
+            at(1, 0, bad_period, n),
+            "with a period below G·N, (k=0,t=2) and (k=1,t=0) must collide"
+        );
+
+        // The derived period keeps the same two apart.
         let good_period = grid * n;
         assert_ne!(
-            0 * good_period + 2 * n,
-            1 * good_period + 0 * n,
+            at(0, 2, good_period, n),
+            at(1, 0, good_period, n),
             "period G·N must separate the pass and thread digits"
         );
     }
--- a/crates/cuda-device/src/disjoint.rs
+++ b/crates/cuda-device/src/disjoint.rs
@@ -258,9 +258,22 @@ impl<'a, T, IndexSpace> DisjointSlice<'a, T, IndexSpace> {
     /// ```
     ///
     /// Returning `&mut [T; N]` rather than `&mut [T]` is deliberate: the length
-    /// is static, so this is also the shape the codegen can fuse into a single
-    /// wide access when `[T; N]` carries the matching alignment (see the
-    /// `vectorization` example).
+    /// is static, so the tile is a value the caller can assign in one statement.
+    ///
+    /// It does **not** widen the access, and it is worth being explicit about
+    /// that. Measured on sm_86, whether an access becomes `LDG.E.128` /
+    /// `STG.E.128` is decided by the *alignment of the element type*, not by the
+    /// static length: a 16-byte-aligned aggregate fuses, the same payload at
+    /// 4-byte alignment does not. Since `T` here is the element type of a
+    /// `DisjointSlice<T>`, a `DisjointSlice<f32>` yields `&mut [f32; N]` with
+    /// 4-byte alignment, which stays scalar.
+    ///
+    /// To get wide accesses, make the element type aligned and use
+    /// [`Self::get_mut`] on a `DisjointSlice<F32x4>` where
+    /// `#[repr(C, align(16))] struct F32x4([f32; 4])`. That path emits one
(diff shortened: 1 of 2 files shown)


honeyspoon:

```
```

honeyspoon:
Closing — the design is not ready, and the reason is specific rather than a general doubt.

This invented a **parallel vocabulary** for something #477 already established in `view.rs`:

| existing | what this added |
|---|---|
| `LocalIndex32<N>` (bounded index proof) | `DisjointBlock<N>` (bounded run proof) |
| `InBounds32` / `InBoundsMut32` | `Option<&mut [T; N]>` |
| `StaticView32::at(..)`, `at_const::<I>()` | `get_block_mut::<N>(block)` |
| `StaticTileMut32<T, ROWS, COLS, ROW_STRIDE>` | `DisjointTiling<N, K>` |

Two vocabularies for one idea is worse than the gap. The established pattern is a view type parameterised by extents, indexed by a bounded-index proof, returning an `InBounds` wrapper — and a solution should extend that (a thread obtaining a sub-view it owns) rather than bolt `Disjoint*` types onto the slice.

It also carried no measured benefit: `get_block_mut` on a `DisjointSlice<f32>` yields align-4 arrays that stay scalar in SASS, so the value would have been purely that a multi-element write stops needing `unsafe` — worth having, but not at the cost of a second vocabulary.

The gap is real and now filed with worked examples, including the grid-stride-mismatch trap and the partial-tail problem: **#583**. Happy to attempt it again against that framing if the composition question gets an answer.

PR #573, opened 2026-07-29
feat(disjoint): predicate derived tiles against the slice extent

honeyspoon:
**Stacked on #571** — its diff includes that PR's commits. Review #571 first; this adds the last 244 lines.

A derived index proof covers the case where the extent divides the launch. Real launches round up, so the last block owns a partial run, and today the only way to handle that is to abandon the proof and index manually.

This adds the predicated forms:

```rust
pub enum Tile<'a, T, const N: usize> { Full(&'a mut [T; N]), Partial(&'a mut [T]) }

slice.get_block_clipped_mut(block)      // -> Tile: Full when it fits, Partial at the tail
slice.for_each_tile::<N, K>(tiling, |tile| ...)
```

`Tile` makes the tail a type-level distinction rather than a runtime `if`: `Full` hands back `&mut [T; N]`, which keeps the fixed-size array the arithmetic wants, and `Partial` degrades to a slice only where it must. `len`, `is_full`, `is_empty` and `as_mut_slice` cover the uniform cases.

The disjointness argument is unchanged by clipping: intersecting each thread's run with the slice extent can only shrink runs that were already pairwise disjoint, so no clipped run can overlap another. That is why this is a safe method rather than an `unsafe` one — the proof survives the clip.

## Why it is stacked rather than standalone

`get_block_clipped_mut` takes a `DisjointBlock`, so it cannot compile without #571. Splitting them would mean inventing a second, weaker parameter type to be replaced immediately. Happy to squash both into one PR if you would rather review them together.

## Gates

fmt, clippy `--workspace --all-targets -D warnings`, rustdoc `-D warnings`, tests and doctests clean on `nightly-2026-04-03`. Own content +244/−0 on top of #571.

### DIFF
--- a/crates/cuda-device/src/disjoint.rs
+++ b/crates/cuda-device/src/disjoint.rs
@@ -679,17 +679,26 @@ mod tiling_algebra_tests {
     #[test]
     fn a_period_below_grid_span_collides() {
         let (grid, n) = (4usize, 1usize);
-        let bad_period = 2; // G·N would be 4.
-        // (k=0, t=2) and (k=1, t=0) both land on element 2.
-        let a = 0 * bad_period + 2 * n;
-        let b = 1 * bad_period + 0 * n;
-        assert_eq!(a, b, "the counterexample must actually collide");
 
-        // The derived period keeps them apart.
+        // `phi` with the period supplied rather than derived, so a wrong period
+        // can be expressed at all. Written as a call so the (k, t) digit pair
+        // stays visible at each use.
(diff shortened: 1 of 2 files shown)


honeyspoon:
Closing — this was stacked on #571 and its diff carried that PR's commits, which makes it awkward to review independently. Reopening as a standalone PR once #571 lands, so each change is reviewed on its own diff.

The content is unchanged and parked on `up2/tile-predication`: `Tile<'a, T, N>` with `Full`/`Partial`, `get_block_clipped_mut`, and `for_each_tile` — 244 lines on top of #571.

PR #574, opened 2026-07-29
feat(cuda-core): RAII CUDA graph capture and replay

honeyspoon:
Graph replay is worth a fixed ~1 µs per launch, and there is no wrapper at all today, so it is driven by raw FFI with `assert_eq!` on every status.

```text
path    direct P50   graph P50   saved    nodes   µs/node
FP16    3.867 ms     3.697 ms    0.170    173     0.98
INT8    4.670 ms     4.491 ms    0.179    222     0.81
W8A16   9.180 ms     9.002 ms    0.178    293     0.61
```

The saving is near-constant while node counts differ by 70%, which identifies it as per-launch overhead rather than anything kernel-related.

```rust
let capture = stream.begin_capture(CaptureMode::Global)?;   // RAII guard
enqueue_work(&stream)?;                                      // recorded, not run
let graph = capture.end()?;                                  // CapturedGraph
let mut exec = unsafe { graph.instantiate() }?;               // GraphExec
exec.launch(&stream)?;
```

---

## 1. `instantiate` is `unsafe` — the central decision

**Chosen:** `pub unsafe fn instantiate(&self) -> Result<GraphExec, DriverError>`, with the buffer-lifetime obligation documented.

**Alternatives, both real:**
- **(a) capture-scope closure** — `stream.capture(|cap| { ... })` where the closure owns the buffers, so the exec cannot outlive them.
- **(b) graph holds `Arc` clones** of every buffer touched during capture.

Why neither, yet: capture records raw device *addresses*, not Rust borrows, so the API cannot observe which buffers were touched. (a) only helps if the closure's captured set is exactly the referenced set, which the driver never reports; (b) requires an allocation registry the crate does not have, and changes ownership semantics for every buffer used in a captured region.

Why `unsafe` rather than a safe signature with docs: the violation is **silent**. Measured on an RTX A2000 (`tests/graph_capture.rs`, the `#[ignore]`d probe), replaying a graph whose captured buffer had been dropped gave:

```
launch = Ok(())   sync = Ok(())   read = Ok(())
dst    = [7, 8, 9, 10]   (captured source was [7, 8, 9, 10])
diagnosed = false
```

The replay read freed memory that still held the old bytes, so it "worked". Stream-ordered deallocation returning pages to a pool makes that the likely presentation, so the failure mode is "passes in testing, corrupts once pages are reused". A safe signature would promise a check neither the driver nor the type system performs.

**This is the decision I most want overruled if you have a preference** — (a) is nicer if you are willing to accept "buffers must be owned by the closure" as a documented convention rather than a proof.

## 2. `launch(&mut self)` rather than `&self`

**Chosen:** `&mut self`.

**Alternative:** `&self`, letting one exec be launched from several places.

Concurrent replay of a single `CUgraphExec` is not permitted. `&mut` makes that a compile error instead of a documented rule. The cost is that an exec cannot be shared behind an `Arc` without a lock — which is the correct constraint, just occasionally inconvenient.

## 3. `StreamCapture` guard that terminates in `Drop`

**Chosen:** RAII guard; `end()` consumes it, `Drop` terminates and discards.

**Alternative:** `begin_capture` / `end_capture` as free calls on the stream.

Not a style preference. The driver documents that after an error the stream sits in `CU_STREAM_CAPTURE_STATUS_INVALIDATED` and

> The capture sequence must be terminated with `cuStreamEndCapture` on the stream where it was initiated in order to continue using `hStream`.

and separately that a capturing blocking stream makes the legacy null stream unusable too. So an early `?` or a panic between begin and end leaves the stream permanently broken — and the existing raw-FFI code, which `assert_eq!`s in that window, has exactly that bug. Two hardware tests cover the remedy: after an abandoned capture the stream is `CaptureStatus::None` and still usable, and it can be recaptured.

## 4. `CaptureMode` enum, defaulting to `Global`

**Chosen:** three-variant enum, `#[default] Global`.

**Alternative:** take the raw `CUstreamCaptureMode`.

`Global` is the conservative mode (prohibits unsafe driver actions in *any* thread). Defaulting to it means the easy call is the safe one; `Relaxed` has to be asked for by name.

## 5. Node mutation deliberately **not** wrapped

`cuGraphExecKernelNodeSetParams` and friends are omitted, and not merely for scope. `requires` launch contracts run on the host *at launch*, so replay never executes them. That is sound only while replay reuses the arguments recorded at capture — each configuration gets its own capture and therefore its own validation — which is precisely what mutation would break. Wrapping it needs that interaction resolved first, so it belongs in its own PR with that argument made explicitly.

Also omitted: graphs containing memory-allocation nodes (the driver permits at most one live exec per graph then) and device-side launch.

## 6. Scope: capture only, no graph *construction*

**Chosen:** stream capture as the only way to build a graph.

**Alternative:** also wrap `cuGraphCreate` + `cuGraphAddKernelNode` for explicit construction.

Capture is the path that carries the measured win and the one every existing user is on. Explicit construction needs typed node handles and a dependency-edge API — a much larger surface, and easier to design once capture's lifetime story is settled.

For context: **CCCL does not wrap graphs either** — no graph type under `cuda/`, and cudax is not in the `cuda_cccl` package — so there is no upstream design to copy here.

# Gates

fmt, clippy `--workspace --all-targets -D warnings`, rustdoc `-D warnings` and doctests clean on `nightly-2026-04-03`. `graph_capture`: 4 passed / 1 ignored on an RTX A2000. `cargo test -p cuda-core` has one failure, `device_copy_derive_compile_pass`, which **reproduces identically on an untouched `main`**.

### DIFF
--- a/crates/cuda-core/src/graph.rs
+++ b/crates/cuda-core/src/graph.rs
@@ -1,0 +1,375 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! CUDA graph capture and replay (RAII), **exploratory**.
+//!
+//! This module exists to find out what a safe graph API can and cannot promise.
+//! It wraps the smallest useful slice of the graph surface — stream capture,
+//! instantiation, replay — and records, in prose next to each item, which
+//! obligations the type system discharges and which it merely documents.
+//!
+//! # Why bother
+//!
+//! Graph replay is worth a fixed ~1 µs per launch. Measured on this repo's
+//! benchmark harness (RTX A2000, sm_86, 500 iterations):
+//!
+//! ```text
+//! path    direct P50   graph P50   saved    nodes   µs/node
+//! FP16    3.867 ms     3.697 ms    0.170    173     0.98
+//! INT8    4.670 ms     4.491 ms    0.179    222     0.81
+//! W8A16   9.180 ms     9.002 ms    0.178    293     0.61
+//! ```
+//!
+//! The saving is near-constant while node counts differ by 70%, which is what
+//! identifies it as per-launch overhead rather than anything kernel-related.
+//!
+//! # What the types actually enforce
+//!
+//! **Capture must be terminated even when it fails.** This is the obligation
+//! most easily missed and the main reason to prefer a guard. The driver
+//! documents that after an error, the stream enters
+//! `CU_STREAM_CAPTURE_STATUS_INVALIDATED` and
+//!
+//! > The capture sequence must be terminated with `cuStreamEndCapture` on the
+//! > stream where it was initiated in order to continue using `hStream`.
+//!
+//! So an early return or panic between begin and end leaves the stream
+//! permanently unusable — and, because a capturing blocking stream also puts
+//! the legacy null stream in "an unusable state", it can take the default
+//! stream down with it. [`StreamCapture`] terminates in [`Drop`], so unwinding
+//! cannot poison the stream.
+//!
+//! **Replays of one executable graph are serialised.** [`GraphExec::launch`]
+//! takes `&mut self`. Two concurrent replays of the same `CUgraphExec` are not
+//! permitted, and `&mut` is exactly that constraint, checked at compile time
+//! rather than documented.
+//!
+//! # What the types do NOT enforce — read before relying on this
+//!
+//! **Captured graphs freeze device pointers, and nothing here ties their
+//! lifetimes together.** Capture records the *argument values* that were
+//! enqueued, including raw device addresses. If a buffer used during capture is
+//! dropped while a [`GraphExec`] built from that capture is still alive, replay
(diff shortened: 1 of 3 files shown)


honeyspoon:

```
```

honeyspoon:
Closing to pursue the prerequisite properly rather than landing `unsafe instantiate` with a documented obligation.

On re-reading my own justification, I think it is wrong. I wrote that a capture-scope closure "only helps if the closure's captured set is exactly the referenced set, which the driver never reports" — but the driver does not need to report anything. What keeps a buffer alive is Rust's *closure capture*, and if the returned exec carries a lifetime tied to those captures, the borrow checker rejects the use-after-free at compile time.

Exploring that on `gen/graph-scope`. Will reopen with whichever design survives.

honeyspoon:
Superseded by #585, which takes the design this PR's own justification said was unavailable.

Both alternatives I dismissed here got prototyped. The lifetime-only form does work for the hazard (`E0597`, zero runtime cost) but holds a shared borrow of the captured buffers, so writing new input between replays is `E0502` — sound and unusable. Ownership resolves both, and drops `unsafe` from the public surface entirely.

PR #575, opened 2026-07-29
feat(access): count the cache lines a warp's global accesses touch

honeyspoon:
`swizzle::conflict_degree` (#544) answers "how many shared-memory transactions does this static pattern cost". Nothing answers the global-memory version — which is the same computation over a different memory space, and global bandwidth is what most kernels are actually limited by.

```rust
pub const fn lines_touched(offsets: &[usize; WARP_LANES], elem_size: usize) -> usize;
pub const fn minimum_lines(total_bytes: usize) -> usize;
pub const fn wasted_lines(offsets: &[usize; WARP_LANES], elem_size: usize) -> usize;
pub const fn is_fully_coalesced(offsets: &[usize; WARP_LANES], elem_size: usize) -> bool;
```

## What it adds over `AccessPlan`

Two patterns with **identical tile shapes**, which `AccessPlan` therefore cannot separate:

| pattern | bytes requested | lines touched | wasted |
|---|---:|---:|---:|
| 32 lanes × 16 B, contiguous | 512 | 4 | **0** |
| 32 lanes × 4 B, 16 B apart | 128 | 4 | **3** |

Both are in the tests. The second is the case worth catching: same shape, same plan, four lines fetched to deliver one line's worth of bytes.

---

## 1. Same signature shape as `conflict_degree`

**Chosen:** `(offsets: &[usize; WARP_LANES], elem_size: usize)` — per-lane **element** offsets, element size in bytes.

**Alternative:** take byte offsets directly, which reads more naturally for global memory.

Byte offsets would be a unit mismatch with the sibling function in `swizzle`, and mixing the two up is a silent wrong answer rather than a compile error. Consistency wins here; the cost is that a sub-element byte shift is inexpressible, which one test documents by reaching for a 12-byte element to demonstrate a boundary crossing.

## 2. Integer counts, not a float efficiency

**Chosen:** `lines_touched` / `minimum_lines` / `wasted_lines` as `usize`, plus a `bool`.

**Alternative:** a `coalescing_efficiency -> f32` (or a `Ratio`) in `[0, 1]`.

Floating point in `const fn` is still awkward, and — more importantly — `wasted_lines == 0` is the assertable predicate, while `0.97` invites a threshold nobody can justify. `conflict_degree` made the same choice: an integer degree where 1 is the good case. A ratio can be computed by the caller from the two counts if they want to report it.

## 3. `wasted_lines` rather than a raw ratio against the floor

`minimum_lines` is exposed separately so the floor is inspectable, but `wasted_lines` is the number that means something operationally: lines fetched and discarded. It is also the one that stays meaningful for a broadcast, where touched == floor == 1.

## 4. O(lanes²) dedup, no scratch buffer

**Chosen:** for each line a lane spans, scan earlier lanes for an overlap.

**Alternative:** a `[bool; N]` seen-array, as `swizzle::banks_used` uses.

`banks_used` can do that because banks are bounded at 32. Line indices are unbounded (they scale with the offset), so a fixed seen-array does not work. 32 lanes makes the quadratic scan trivial, and it needs no allocation, which is what keeps the whole thing `const`-evaluable. `line_span` saturates rather than wrapping so a pathological offset cannot fold back onto a low line and fake a hit.

## 5. Lines, with sectors exposed but not modelled

`SECTOR_BYTES` is a public constant, but the metrics count 128-byte lines. Sector-level accounting is the finer and more accurate model on recent architectures, and I did not implement it because it needs per-arch behaviour I have not measured. Adding `sectors_touched` later is additive; guessing at it now would be a number nobody can check.

---

# The caveat, stated in the module docs rather than implied

**These are necessary-condition checkers.** They bound what the *layout* permits; they cannot report what codegen delivers. This module already contains the proof: it reports `TXN_128` for a tile shape whose measured SASS is four scalar `LDG.E`, because SROA splits any decomposed value. `wasted_lines == 0` likewise proves the pattern does not *forbid* a coalesced access — never that one happened.

A green assertion read as a guarantee is worse than no assertion, so the docs now say this for both halves of the module.

# One thing worth reporting about the tests

The `const` assertion rejected my first test pattern, correctly. I had modelled "each lane takes 4 contiguous f32" as a 4-element stride with a 4-byte element — which is 32 *strided scalar* reads, not one contiguous vector read per lane. The right model is element size 16 with unit stride. Both are now tests, because that misreading is precisely what the metric exists to catch.

# Context

This is E1 of six extensions identified while researching what `swizzle` and `access` have in common: both are `const fn` from a static access description to a hardware cost metric — what the literature calls static/analytical performance models (GPUDrano solves this exact problem for global coalescing as an LLVM pass; CUPL polyhedrally, requiring affine access functions, which ours are). The unusual part here is delivery: as `const fn`, a violated prediction is a build error rather than an analyser's report.

# Gates

fmt, clippy `--workspace --all-targets -D warnings`, rustdoc `-D warnings`, tests and doctests clean on `nightly-2026-04-03`. +260/−0 against `main`, one file.

### DIFF
--- a/crates/cuda-device/src/access.rs
+++ b/crates/cuda-device/src/access.rs
@@ -61,6 +61,21 @@
 //! A plan therefore describes the shape a width *needs*, not a width the kernel
 //! is guaranteed to get - the data has to move as whole elements as well.
 //! See the `vectorization` example for the alignment half.
+//!
+//! # Coalescing is the same kind of claim
+//!
+//! [`lines_touched`] and friends answer the global-memory question that
+//! [`crate::swizzle::conflict_degree`] answers for shared memory: how many
+//! hardware transactions a static access pattern costs. They carry the identical
+//! caveat, and it is worth stating rather than implying.
+//!
+//! **These are necessary-condition checkers.** They bound what the *layout*
+//! permits; they cannot report what codegen delivers. This module already
+//! contains the proof: it reports `TXN_128` for a tile shape whose measured SASS
+//! is four scalar `LDG.E`, because SROA splits any decomposed value. Likewise
+//! `wasted_lines == 0` proves the pattern does not forbid a coalesced access; it
+//! never proves one happened. A green assertion that is read as a guarantee is
+//! worse than no assertion.
 
 /// One 32-bit transaction (`LDG.E` / `STG.E`).
 pub const TXN_32: usize = 4;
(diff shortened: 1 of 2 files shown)


honeyspoon:
Also wired this into the `vectorization` example, since a `const`-arithmetic module that nothing consumes reads as speculative — and #545 already put an `access::plan` `const` block there, so this extends the same block rather than adding a new example:

```rust
// The same shape checked for global coalescing, which `plan` cannot see:
// one 16-byte access per lane laid end to end covers 512 contiguous bytes
// and touches the four lines that is the floor for them.
assert!(access::lines_touched(&lanes, elem) == 4);
assert!(access::is_fully_coalesced(&lanes, elem));
```

Which is the point of the metric in one line: `f32x4_view_copy` and a strided variant have the **identical** `AccessPlan`, and only this can tell them apart. Verified the block actually evaluates by compiling the example, not just the crate.

+20 lines there; the module itself is unchanged.

honeyspoon:

```
```

nihalpasham:
LGTM. 

**Two commits on top:**
- `lines_touched` now states its base-alignment assumption (element offset 0 is assumed 128-byte line-aligned - true for allocation bases, not for arbitrary sub-slice bases), and
- the one non-saturating multiply in `wasted_lines` now saturates like the rest of the module.

**Validated:** 
- Crate tests and doctests pass; 
- the vectorization example still compiles through the device pipeline.

Merging.

nihalpasham (approved):

PR #576, opened 2026-07-29
fix(mir-importer): implement Rvalue::Len for arrays and slices

deepakshirkem:
`Rvalue::Len` was unimplemented in `mir-importer`, so any kernel calling `.len()` on a slice or array failed to compile for the device.

This adds support for both cases:
- **Arrays** `[T; N]` — the length is a compile-time constant, emitted as `mir.constant`.
- **Slices** `&[T]` — the length is read from the slice's `{ ptr, len }` pair using `mir.extract_field`, following the same pattern used elsewhere in the file for DST tail lengths.

Added a `len_test` example — a kernel where every thread reads `input.len()` and writes it to the output. Verified on real hardware (GTX 1650, sm_75):


nihalpasham:
Thanks for the PR. 

Unfortunately the premise doesn't hold on our pinned toolchain: 
- `rustc` removed `Rvalue::Len` from MIR back in 2025 (rust-lang/rust#134330), so the bridge can never construct that variant. 
- Slice `.len()` already works in kernels - it reaches the backend as `PtrMetadata`, which we've lowered since the first release (`float_rounding` exercises it on device in CI on every PR).

You can verify both halves quickly: 
- build the `len_test` example against unpatched main (it compiles and passes), and 
- run any `.len()` kernel with `CUDA_OXIDE_DUMP_MIR=1` - you'll see PtrMetadata and no `Len`.

Closing this one.

PR #578, opened 2026-07-29
fix(mir-importer): preserve pointer provenance in direct tuple constants

### LINKED ISSUE #577: Direct tuple constants with pointer fields fail MIR import
## Bug

Direct tuple value constants containing references or raw pointers are rejected during MIR import, even when every pointer relocation can be resolved without losing provenance.

```rust
static DATA: [u8; 16] = [
    0, 1, 2, 3, 4, 5, 6, 7,
    8, 9, 10, 11, 12, 13, 14, 15,
];

const DIRECT: (&[u8; 16], bool) = (&DATA, true);

#[kernel]
pub unsafe fn kernel(output: *mut u8) {
    let (data, flag) = DIRECT;
    unsafe {
        output.write(data[0] + flag as u8);
    }
}
```

Current result:

```text
Tuple constant contains 1 pointer relocation(s); cuda-oxide cannot yet preserve tuple pointer provenance
```

Expected result:

```text
The tuple is constructed with a real pointer to the device static, preserving
the allocation identity, address space, and byte addend recorded by rustc.
```

## Root cause

`translate_tuple_constant` currently rejects the entire constant whenever its allocation contains any pointer relocation. The layout-aware tuple decoder therefore receives only raw bytes and cannot associate a relocation with the pointer field at the corresponding rustc field offset.

The importer already has provenance-preserving paths for standalone pointer constants:

* Device statics lower through `MirGlobalAllocOp`.
* Interior static pointers preserve their byte addend.
* Promoted constant allocations are materialized and referenced through `MirRefOp`.

The tuple decoder should reuse those paths per field rather than interpreting relocation placeholder bytes as pointer integers.

## Required behavior

For each direct tuple field:

1. Use rustc's tuple field offset and stored size.
2. Detect a relocation beginning at that field offset.
3. Require the field to be a supported thin reference or raw pointer.
4. Resolve the relocation target and byte addend.
5. Materialize the pointer through the existing static or promoted-constant path.
6. Continue decoding non-pointer fields from their bytes.
7. Verify that every relocation was consumed by exactly one pointer field.

## Correctness boundaries

This issue covers direct thin pointer fields targeting:

* Rust `static` or `static mut` allocations.
* Anonymous promoted constant allocations.
* Interior addresses with non-zero byte addends.
* Multiple independent pointer fields in one tuple.

The importer must continue to reject:

* Relocations in padding or non-pointer fields.
* Relocations partially overlapping a field.
* Fat pointers and DST metadata not explicitly reconstructed.
* Function and vtable allocations.
* Pointee allocations that themselves contain unsupported nested relocations.
* Pointer-bearing arrays, structs, and enums, which have separate support boundaries.

No relocation may be converted through `PointerWithExposedProvenance` using its placeholder bytes.

## Acceptance criteria

* `(&STATIC, scalar)` direct tuple constants compile and execute correctly.
* Interior static pointers preserve their byte addend.
* References to promoted literals preserve their allocation identity.
* Multiple direct thin-pointer fields can be materialized independently.
* Null raw-pointer fields without provenance continue to use the existing byte path.
* Unsupported relocation shapes fail with an actionable diagnostic.
* The negative fixture is converted into a runtime regression example.
* `STATUS.md`, the smoketest classification, and supported-feature documentation are updated.
* Existing array, struct, enum, standalone static-pointer, and promoted-pointer behavior remains unchanged.

uurl:
Closes #577 

## Summary

Preserve rustc pointer relocations when materializing direct tuple value constants.

Tuple constants no longer fail solely because their backing allocation contains provenance. The importer now associates each relocation with the direct tuple field at the corresponding rustc layout offset and lowers supported pointer fields through the existing provenance-preserving static or promoted-constant paths.

## Root cause

`translate_tuple_constant` previously counted pointer relocations and rejected the entire tuple before field decoding:

```text
Tuple constant contains N pointer relocation(s); cuda-oxide cannot yet preserve tuple pointer provenance
```

The subsequent decoder only received a byte image. Pointer bytes in a rustc allocation are placeholders carrying an addend, not an exposed machine address, so decoding them through an integer-to-pointer cast would fabricate an unrelated pointer.

Standalone pointer constants already preserve the required information through `MirGlobalAllocOp` for device statics and `MirRefOp` for promoted constant allocations. The missing piece was making those paths reusable from aggregate field decoding.

### MIR importer

* Extracted reusable lowering for resolved static-pointer targets.
* Extracted reusable lowering for pointers to promoted constant allocations.
* Added a shared constant-allocation accessor for `Allocated` and `TyConstKind::Value` constants.
* Added relocation lookup by allocation byte offset.
* Changed direct tuple decoding to inspect relocations per rustc field offset.
* Materialized supported thin pointer and reference fields with their original target allocation and byte addend.
* Continued decoding fields without relocations through the existing layout-aware byte decoder.
* Added validation that every relocation is consumed exactly once.
* Added explicit diagnostics for relocations in padding, scalar fields, fat-pointer fields, unsupported allocation kinds, and pointees with nested relocations.

### Regression example

* Renamed `error_tuple_constant_provenance` to `tuple_constant_provenance`.
* Converted the expected-failure fixture into an end-to-end runtime test.
* Covered:

  * A pointer to a complete device static.
  * An interior pointer with a non-zero addend.
  * A pointer to an anonymous promoted literal.
  * Scalar fields stored alongside pointers.
* Removed the obsolete exact-diagnostic assertion from `scripts/smoketest.sh`.

### Documentation

* Removed the resolved support-gap row from `STATUS.md`.
* Documented support for thin pointer fields in direct tuple constants.
* Kept pointer-bearing arrays, structs, enums, fat pointers, and nested relocations explicitly unsupported.

## Correctness

Relocation placeholder bytes are never interpreted as pointer integers.

Each supported pointer field is reconstructed from:

* The rustc provenance target.
* The addend stored in the pointer word.
* The translated Rust pointer type.
* The physical CUDA address space of the target allocation.

Relocations that cannot be associated with a supported direct pointer field fail closed.

## Validation

* [ ] `cargo oxide fmt --check`
* [ ] `cargo test -p mir-importer`
* [ ] `cargo clippy -p mir-importer --all-targets -- -D warnings`
* [ ] `scripts/check-error-example-status.sh`
* [ ] `scripts/smoketest.sh --compile-only -o '^tuple_constant_provenance$'`
* [ ] `cargo oxide run tuple_constant_provenance`
* [ ] Remaining array, struct, and enum provenance fixtures still fail with their expected diagnostics


nihalpasham:
LGTM. 

**Two commits on top:**
- fixed a regression inside the PR: a zero-sized tuple like `((), ())` has no allocation to hand the decoder, and the rewrite treated missing bytes as a hard error in every case. Missing bytes for a zero-sized tuple decodes as empty again; missing bytes for a tuple that should contain data stays the error. Unit tests cover both sides plus padding zeroing.
- regenerated the example lockfile to the snapshot the sibling examples use.

**Verified on hardware:** 
- `tuple_constant_provenance` passes - both relocations, the +7 addend, and 
- global dedup for two references to one static.

Promoted-literal tuples (`&5` in a const tuple) stay rejected by design, so this partially addresses #577 ; re-scoping that issue to the remainder rather than closing it.

Merging.

nihalpasham (approved):

PR #579, opened 2026-07-29
feat(mir-importer): materialize thin pointers in aggregate const values

Ayush7614:
## Summary

- Struct, tuple, and array **const** values whose thin pointer fields relocate to device statics are decoded via `MirGlobalAllocOp` per field (instead of rejecting the whole constant).
- Converts the former support-gap fixtures `error_struct_constant_provenance`, `error_tuple_constant_provenance`, and `error_tuple_array_provenance` into positive smoke examples.
- Enum constant pointer provenance and device-static **initializer** pointer relocations stay fail-closed (unchanged).

This closes the STATUS support-gaps for aggregate const thin-pointer fields. It does **not** claim that `static REF: &T = &TARGET` initializers work — that remains `error_static_initializer_provenance`.

## Test plan

- [x] `cargo test -p mir-importer` (880 passed + 1 ignored doctest)
- [x] `cargo oxide run struct_constant_provenance` (needs GPU / toolkit)
- [x] `cargo oxide run tuple_constant_provenance`
- [x] `cargo oxide run tuple_array_provenance`
- [x] Confirm `error_enum_constant_provenance` and `error_static_initializer_provenance` still fail closed


nihalpasham:
LGTM on the direction - after integration this is the shared machinery all the aggregate-provenance PRs route through. 

**Seven commits on top**; it needed some work:
- the fail-closed guards #581/#587 require: the scalar/default branch now rejects relocations in range, and a per-aggregate audit errors on any relocation no field consumed;
- bare arrays of structs/enums stay on the documented contract instead of silently expanding it (arrays nested inside struct constants keep working - they already did on main);
- a regression on zero-sized constants: function items and non-capturing closures are (empty) struct types with no ADT metadata, and the reworked decoders demanded ADT metadata before decoding, which broke `function_item_call` and `map_sum`. Both decoders now early-out to the ZST synthesis path;
- deleted the dead `build_const_from_bytes`, added the relocation-matching/addend/field-kind unit tests #581 asked for, and cleaned fmt/clippy, which failed as submitted.

**Verified on hardware:** 
- struct/tuple/array provenance examples pass, 
- the full example sweep is 177/177, 
- workspace tests green.

Merging as the top of the provenance stack; together with #582 and #588 

this closes #581 and #587.

nihalpasham (approved):

PR #580, opened 2026-07-29
feat(mir-importer): support ConstantIndex from_end on arrays

Ayush7614:
## Summary

- Resolve `ProjectionElem::ConstantIndex { from_end: true }` as `len - offset` for known-length arrays across place loads, SSA extracts, address walks, and element writes.
- Slices / unknown-length bases still get a clear unsupported or punt path (no silent wrong index).
- Adds unit tests for the resolver and a `constant_index_from_end` example using `let [.., last] = arr` / `let [.., ref mut last] = arr`.

## Test plan

- [x] `cargo test -p mir-importer` (883 passed + 1 ignored doctest), including `resolve_constant_index_*` unit tests
- [ ] `cargo oxide run constant_index_from_end` (needs GPU / toolkit)

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -3306,17 +3306,9 @@ fn translate_place_value_fallback(
                     // `MirConstantOp` + `MirArrayElementAddrOp` and load.
                     // `mem2reg` collapses the resulting load-after-store pairs
                     // back into SSA extracts for promotable arrays.
-
-                    let index = if *from_end {
-                        return input_err!(
-                            loc,
-                            TranslationErr::unsupported(
-                                "ConstantIndex with from_end=true not yet supported"
-                            )
-                        );
-                    } else {
-                        *offset as usize
-                    };
+                    //
+                    // `from_end` resolves via [`resolve_constant_index`] using
+                    // the known `MirArrayType` length.
 
                     // Load the current array value if we don't have a slot (ZST/edge case)
                     // so that we fall back to the SSA extract-field behaviour.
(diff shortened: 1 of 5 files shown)


nihalpasham:
Thanks Ayush, but this has the same problem as #559: the code that runs is never reached, and the case that's reached is not implemented.

The "count from the end" index only exists when the length isn't known at compile time:

```rust
    // array - length known: rustc computes 4 - 1 itself and hands us
    // a plain "index 3". "From the end" never reaches our pipeline.
    let arr = [10, 20, 30, 40];
    let [.., last] = arr;

    // slice - length known only at runtime: this is the ONE case where
    // "1 from the end" actually arrives, and it needs s.len() - 1
    // computed at runtime from the slice's pointer+length pair.
    fn k(s: &[u32]) { if let [.., last] = *s { ... } }
```

This PR implements the array case - which rustc already resolved before we see the MIR (from_end is documented "always false when indexing an array") - so the new code can never execute, and your own example compiles on unpatched main for exactly that reason. The slice case is the real feature, and it's the part left unimplemented.

There's also a hazard: the guard decides "is this an array?" from the pointee type alone, so for a slice of rows it misfires:

```rust
    fn k(s: &[[u32; 3]]) { if let [.., row] = *s { ... } }
    // pointee is [u32; 3], so the guard computes 3 - 1 = 2 and reads
    // s[0][2] - an element inside the FIRST row - instead of selecting
    // the last row, s[s.len() - 1].
```

Today that pattern stops with a clean "not supported" error; with this change it would compile and silently produce wrong data.

If you want to pick up the real feature: extract the length from the slice fat pointer at runtime and subtract, the way `rustc_codegen_ssa` does, including the slice-of-rows case above. 

A check that will save you time on future PRs: build the example against unpatched main first, and dump the MIR (`CUDA_OXIDE_DUMP_MIR=1`) to confirm the shape you're handling actually occurs.

Closing this one.

PR #582, opened 2026-07-29
Preserve pointer provenance in struct constants

### LINKED ISSUE #581: Support pointer provenance in struct constants
## Problem

Struct constants containing pointer or reference fields are currently rejected during MIR import.

For example:

```rust
static FIRST: [u8; 16] = [
    1, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0,
];

struct Holder {
    pointer: &'static [u8; 16],
    flag: bool,
}

const DIRECT: Holder = Holder {
    pointer: &FIRST,
    flag: true,
};
```

Using `DIRECT.pointer` from device code fails with:

```text
Struct constant contains 1 pointer relocation(s); cuda-oxide cannot yet preserve struct pointer provenance
```

## Current behavior

`translate_struct_constant` converts the constant allocation into a raw byte image before translating its fields.

This is sufficient for integer and floating-point fields, but not for pointer fields. In rustc's constant representation, the bytes stored at a pointer field contain only the byte addend into the target allocation. The allocation identity is stored separately in `Allocation::provenance.ptrs`.

Flattening the allocation to bytes would therefore replace a valid pointer with a null or otherwise invalid integer-derived pointer. The importer currently prevents that miscompilation by rejecting every struct allocation containing relocations.

## Expected behavior

The importer should preserve the allocation's provenance table while decomposing the struct.

For every direct pointer or reference field:

1. Locate the relocation at the field's rustc layout offset.
2. Decode the pointer addend using the target pointer width.
3. Resolve the relocation target.
4. Materialize references to Rust statics through the existing device-global pointer path.
5. Preserve the static's physical CUDA address space and normalize the result to the exact Rust field type.
6. Use the existing raw integer-to-pointer path only when the field has no relocation.

The importer must continue to fail closed when a relocation overlaps a non-pointer field, cannot be resolved, or remains unconsumed after all fields have been translated.

## Proposed changes

* Split struct constant translation into allocation resolution and allocation-aware field translation.
* Retain `Allocation::provenance.ptrs` alongside the raw byte image.
* Extract the relocation-resolution logic currently used by top-level static pointer constants into a reusable helper.
* Materialize relocated pointer fields through `translate_static_global_pointer`.
* Replace the hardcoded eight-byte pointer decoder with the target pointer width.
* Add unit tests for relocation matching, addends, invalid field kinds, and unconsumed relocations.
* Convert `error_struct_constant_provenance` into a successful runtime example named `struct_constant_provenance`.
* Remove the resolved support gap from `STATUS.md` and `ERROR_EXAMPLES`.

## Scope

This issue covers direct pointer and reference fields in struct constants.

Pointer relocations embedded inside tuple, array, and enum storage remain tracked by their existing support-gap fixtures.

## Acceptance criteria

* A struct constant containing a reference to an immutable Rust static compiles for the CUDA target.
* Reading through the reference produces the correct runtime value.
* Non-zero pointer addends are preserved.
* Static pointers retain their global or constant CUDA address space during address calculation.
* Relocation placeholder bytes are never interpreted as literal pointer addresses.
* Unsupported or unmatched relocations produce a precise diagnostic.
* `error_struct_constant_provenance` is replaced by a successful `struct_constant_provenance` example.
* `scripts/check-error-example-status.sh` passes.
* `cargo test -p mir-importer` passes.
* `cargo test -p mir-lower` passes.
* The new example passes the compile-only smoketest and the runtime CUDA test.

uurl:
Closes #581 

## Summary

Add support for direct pointer and reference fields in struct constants without flattening rustc relocation records into integer pointer bits.

Struct constant translation now keeps the source allocation and its provenance table available while decomposing fields. Pointer relocations targeting Rust statics are materialized through the existing device-global pointer path, preserving their byte addends and CUDA address spaces.

## Problem

rustc represents a pointer inside a constant allocation using two components:

* the pointer's byte addend in the allocation bytes;
* the target allocation identity in `Allocation::provenance.ptrs`.

The struct constant importer previously discarded the provenance side table and retained only the byte image. Because those bytes are not an address, decoding them as an integer and emitting `PointerWithExposedProvenance` would produce an invalid device pointer.

To prevent silent miscompilation, the importer rejected every struct constant containing a pointer relocation.

This prevented valid constants such as:

```rust
static FIRST: [u8; 16] = [
    1, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0,
];

struct Holder {
    pointer: &'static [u8; 16],
    flag: bool,
}

const DIRECT: Holder = Holder {
    pointer: &FIRST,
    flag: true,
};
```

### Preserve the source allocation

Split struct constant translation into:

* resolution of the effective allocation;
* allocation-aware field decomposition.

The field translator now receives both the allocation bytes and its provenance entries.

### Materialize relocated pointer fields

For each pointer field, the importer:

1. Uses rustc's field layout offset.
2. Looks for a relocation beginning at that offset.
3. Reads the relocation addend using the target pointer width.
4. Resolves static allocation targets.
5. Materializes the target through `translate_static_global_pointer`.
6. Produces the exact MIR pointer type required by the struct field.

Pointer fields without relocations continue to use the raw integer-to-pointer path for null or exposed-address constants.

### Fail closed on unsupported layouts

The importer reports an error when:

* a relocation overlaps a non-pointer field;
* a relocation does not start at the pointer field boundary;
* multiple relocations claim one field;
* the relocation target cannot be materialized;
* any relocation remains unconsumed after translating the struct.

This prevents relocation placeholder bytes from being interpreted as literal addresses.

### Reuse static-pointer infrastructure

Extract the allocation-relocation resolution logic from `static_target_from_constant` into a reusable helper.

The implementation reuses the existing device-global materialization and pointer address-space normalization. No changes are required in `mir-lower`, `dialect-mir`, or `llvm-export`.

### Regression coverage

Rename the negative fixture:

```text
error_struct_constant_provenance
```

to the successful runtime example:

```text
struct_constant_provenance
```

The example launches a kernel that reads through a pointer stored in a struct constant and verifies the resulting values on the host.

The resolved entry is removed from:

* `crates/rustc-codegen-cuda/STATUS.md`;
* `ERROR_EXAMPLES` in `scripts/smoketest.sh`.

## Validation

```bash
cargo fmt --all -- --check
cargo test -p mir-importer
cargo test -p mir-lower
cargo clippy --workspace -- -D warnings

scripts/smoketest.sh --compile-only \
    -o '^struct_constant_provenance$'

cargo oxide run struct_constant_provenance
```

Expected runtime result:

```text
SUCCESS: struct constant pointer provenance preserved.
```


nihalpasham:
LGTM. 

**Two commits on top:**
- a unit test proving the promoted (by-ref) struct-with-relocation path still fails closed - #581 asked for negative coverage, and the rejection closure is now a named helper the test can target, and
- regenerated the example lockfile to match the sibling snapshot.

**Verified on hardware:** 
- `struct_constant_provenance` passes, and 
- the emitted IR shows the pointer field as an addrspacecast of the named global - no inttoptr. 
- The near-duplicate static-target helper this shares with #578 was unified at integration.

Merging; 

closes #581.

nihalpasham (approved):

Issue #583, opened 2026-07-29
DisjointSlice cannot express a thread owning N contiguous elements


honeyspoon:
## Summary

`DisjointSlice::get_mut` takes a `ThreadIndex` and yields **one** element. A thread that owns *N contiguous* elements — the normal shape for a coarsened or grid-stride kernel — has no way to say so, so those kernels drop to `get_unchecked_mut` and lose the disjointness proof along with the bounds check.

## The gap, concretely

A kernel where each thread handles 4 contiguous elements has to write this today:

```rust
#[kernel]
pub fn scale_by_four(mut data: DisjointSlice<f32>, n: usize) {
    let tid = thread::index_1d().get();
    let base = tid * 4;
    // No `ThreadIndex` describes "I own base..base+4", so the proof is gone and
    // every access is unchecked. The disjointness argument now lives in a
    // comment instead of the type system.
    for k in 0..4 {
        if base + k < n {
            unsafe { *data.get_unchecked_mut(base + k) *= 2.0; }
        }
    }
}
```

`get_mut` cannot help: it resolves one `ThreadIndex` to one element. The thread genuinely does own a disjoint *run*, and nothing can express that.

Grid-stride is the same problem with an extra trap — the stride must equal the launch geometry:

```rust
// If this stride disagrees with the actual launch, threads silently overlap.
let stride = grid_dim_x() * block_dim_x();
let mut i = tid;
while i < n {
    unsafe { *data.get_unchecked_mut(i) *= 2.0; }
    i += stride;
}
```

## What a solution needs

1. **A proof that a thread owns a contiguous run**, derived from an existing `ThreadIndex` rather than constructed by the user. Multiplying a disjoint index by a constant preserves disjointness, so the derivation is sound; the point is that the user must not be able to state a run that is not.
2. **A grid-stride period derived from the launch**, not passed in. A passed stride is the bug above.
3. **Partial-tail handling.** Real launches round up, so the last thread's run is clipped. Whatever the API returns must distinguish a full run from a clipped one without forcing the caller back to raw indexing — otherwise the tail undoes the whole thing.

## Why my own attempt was the wrong shape

I proposed this in #571 (now closed) as `ThreadIndex::scale::<N>() -> DisjointBlock<N>`, `DisjointBlock::repeat::<K>() -> DisjointTiling<N, K>`, and `DisjointSlice::get_block_mut::<N>() -> Option<&mut [T; N]>`.

[...]

PR #346, opened 2026-07-06
feat: fp and cuda-graph support

devillove084:
Summary

Adds **FP8 (E4M3/E5M2), FP6 (E3M2/E2M3), and FP4 (E2M1)** type support through the full compiler pipeline (types → dialect → MIR translation → PTX lowering), plus a complete **CUDA Graph** API with stream capture, manual node construction, executable graph update, and an auto-warmup strategy inspired by llama.cpp. This is foundational work for Blackwell (sm_100a+) tensor core operations and low-overhead kernel launch.

### Sub-byte float type system (4 crates)
- `DeviceExternType` gains `BFloat16`, `Float8E4M3`, `Float8E5M2`, `Float4E2M1`, `Float6E3M2`, `Float6E2M3` — all sub-byte float types map to `i8`/`i16` in LLVM IR since no native types exist in pliron/LLVM
- All match arms updated: `module.rs`, `ops.rs`, `lower.rs`, `export_test.rs`

### FP8/FP6/FP4 tcgen05 device intrinsics + compiler pipeline (5 crates)
- `Tcgen05ElementType` extended with `E4M3`, `E5M2`, `E2M3`, `E3M2`, `E2M1` + `idesc_value()` method for kind-dependent idesc encoding (enum discriminants are identifiers; `build()` maps to correct PTX ISA bit patterns per kind family)
- New device intrinsics: `tcgen05_mma_ws_e4m3/e5m2/e2m3/e3m2/e2m1()`, plus `new_e4m3()`/`new_e5m2()`/`new_e2m3()`/`new_e3m2()`/`new_e2m1()` descriptor builders
- Dialect ops: `Tcgen05MmaWsE2M3Op`, `Tcgen05MmaWsE3M2Op`, `Tcgen05MmaWsE2M1Op` registered in `dialect-nvvm`
- MIR translation: `emit_tcgen05_mma_ws_e2m3/e3m2/e2m1()` in `mir-importer` + dispatch entries for all f8f6f4 types
- PTX lowering: all f8f6f4-family types (FP8/FP6/FP4) use `kind::f8f6f4`; F16/BF16 use `kind::f16`; TF32 uses `kind::tf32` — the element type is encoded in the instruction descriptor (idesc), not the PTX mnemonic

### CUDA Graph API (1 crate, 1 new module)
- **Stream capture**: `CudaStreamCaptureExt` trait — `begin_capture`, `end_capture`, `is_capturing`
[...]

PR #585, opened 2026-07-29
feat(cuda-core): safe CUDA graph capture and replay

honeyspoon:
Replaces #574, which I closed. That PR shipped `unsafe instantiate` with the buffer-lifetime obligation documented; this one makes the obligation unrepresentable and exposes no `unsafe` at all.

Graph replay is worth a fixed ~1 µs per launch — measured on the harness that motivated this, 0.170 / 0.179 / 0.178 ms across FP16 / INT8 / W8A16, near-constant while node counts differ by 70%, which is what identifies it as per-launch overhead. There is no wrapper today, so it runs through raw FFI with `assert_eq!` on every status.

```rust
let mut exec = stream.capture_owning(CaptureMode::Global, bufs, |b| {
    b.dst.copy_from_device_async(&b.src, &stream)
})?;

for input in batches {
    exec.state_mut().src.copy_from_host(&stream, input)?;
    exec.launch(&stream)?;
    stream.synchronize()?;
}
let bufs = exec.into_state();
```

Public surface: `capture_owning`, `OwnedGraphExec` (`launch` / `state` / `state_mut` / `into_state` / `node_count`), `CaptureMode`, `CaptureStatus`, `capture_status`. Nothing else, and no `unsafe`.

---

## 1. The exec **owns** the captured state — this is the whole design

**Chosen:** `capture_owning(mode, state, |&mut state| ...) -> OwnedGraphExec<S>`.

**Alternatives, both prototyped rather than argued about:**

**(a) Lifetime-only.** An exec carrying `PhantomData<&'captures ()>` derived from the closure. This *works* for the hazard — zero runtime cost, no ownership change, and the use-after-free becomes:

```
error[E0597]: `src` does not live long enough
```

But it holds a *shared* borrow of every captured buffer, so writing new input between replays — the thing graph replay exists for — is:

```
error[E0502]: cannot borrow `src` as mutable because it is also borrowed as immutable
```

Sound, zero-cost, and unusable. I only found that by building it.

**(b) `unsafe instantiate` + documented obligation** (what #574 did). Usable and unsound, and unsound in the worst way: the violation is **silent**. Measured on an RTX A2000, replaying a graph whose captured buffer had been freed gave

```
launch = Ok(())   sync = Ok(())   read = Ok(())
dst    = [7, 8, 9, 10]   (captured source was [7, 8, 9, 10])
```

The replay read freed device memory that still held the old bytes. Stream-ordered deallocation returning pages to a pool makes that the *likely* presentation, so the failure mode is "passes in testing, corrupts once pages are reused" rather than a fault.

Owning resolves both. Capture records raw device *addresses*, not Rust borrows, so the driver can never report which allocations a graph references — but it does not need to, because nothing outside the exec can name the buffers. The hazard is unrepresentable rather than rejected, and `state_mut` gives the write-then-replay loop back.

**Cost, stated plainly:** the caller must group the buffers into one value and hand ownership over. That is a real ergonomic imposition, and the reason I would still take it is that the alternative is a silent-corruption footgun.

## 2. One public type, not five

The prototype exposed `StreamCapture`, `CapturedGraph`, `GraphExec`, a borrow-based exec and an owning one. That is five vocabularies for one concept, which is the mistake #583 documents. All the intermediates are now private, which also removes `unsafe instantiate` from the API surface entirely.

Consequence worth flagging: `node_count` moved onto the exec and is recorded at capture time, because the `CUgraph` is released once the executable graph exists.

## 3. Internal RAII guard, kept private

Not a style choice. The driver documents that after an error the stream sits in `CU_STREAM_CAPTURE_STATUS_INVALIDATED` and

> The capture sequence must be terminated with `cuStreamEndCapture` on the stream where it was initiated in order to continue using `hStream`.

and separately that a capturing blocking stream makes the legacy null stream unusable too. So a `record` closure returning `Err` would otherwise strand the stream permanently — which is a bug the existing raw-FFI code has, since it `assert_eq!`s inside that window. Two tests probe the remedy through the public API.

## 4. `launch(&mut self)`

Concurrent replay of one `CUgraphExec` is not permitted, so `&mut` makes it a compile error rather than a documented rule. Cost: an exec cannot be shared behind an `Arc` without a lock, which is the correct constraint.

## 5. Capture only — no explicit graph construction

`cuGraphCreate` + `cuGraphAddKernelNode` would need typed node handles and a dependency-edge API. Capture is the path that carries the measured win and the one existing users are on; explicit construction is easier to design once the lifetime story is settled.

## 6. Node mutation deliberately absent

`cuGraphExecKernelNodeSetParams` interacts with launch contracts: `requires` clauses run on the host *at launch*, so replay never executes them. That is sound only while replay reuses the arguments recorded at capture — precisely what mutation would break. It needs that resolved first, in its own PR.

For context, CCCL does not wrap graphs either — no graph type under `cuda/`, and cudax is not in the `cuda_cccl` package — so there is no upstream design to copy here.

# What remains on the caller

`launch` *enqueues*; it does not wait. A write issued after it returns is only safe if ordered behind the replay, which holds when both go on the same stream — `DeviceBuffer`'s copy helpers are all stream-ordered, so the natural usage is correct. Writing from a different stream, or from the host without synchronising, races the in-flight graph. I do not see a way to encode that in types, since the API cannot observe which stream a future write will use.

# Tests

Six, all on the public API only, on an RTX A2000:

- capture → replay a device copy, and that capture **records rather than executes** (`dst` still zero after capture)
- `node_count == 1`
- write-then-replay across three iterations, each observing the input written before it
- a failing `record` leaves the stream `CaptureStatus::None` and usable
- a stream can be recaptured after a failed capture
- an idle stream reports no capture

Note what is *absent*: there is no test for replaying a graph whose captured buffer was freed. That program cannot be written through this API.

# Gates

fmt, clippy `--workspace --all-targets -D warnings`, rustdoc `-D warnings`, doctests clean on `nightly-2026-04-03`. `cargo test -p cuda-core` has one failure, `device_copy_derive_compile_pass`, which **reproduces identically on an untouched `main`**.

### DIFF
--- a/crates/cuda-core/src/graph.rs
+++ b/crates/cuda-core/src/graph.rs
@@ -1,0 +1,468 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! CUDA graph capture and replay (RAII), **exploratory**.
+//!
+//! This module exists to find out what a safe graph API can and cannot promise.
+//! It wraps the smallest useful slice of the graph surface — stream capture,
+//! instantiation, replay — and records, in prose next to each item, which
+//! obligations the type system discharges and which it merely documents.
+//!
+//! # Why bother
+//!
+//! Graph replay is worth a fixed ~1 µs per launch. Measured on this repo's
+//! benchmark harness (RTX A2000, sm_86, 500 iterations):
+//!
+//! ```text
+//! path    direct P50   graph P50   saved    nodes   µs/node
+//! FP16    3.867 ms     3.697 ms    0.170    173     0.98
+//! INT8    4.670 ms     4.491 ms    0.179    222     0.81
+//! W8A16   9.180 ms     9.002 ms    0.178    293     0.61
+//! ```
+//!
+//! The saving is near-constant while node counts differ by 70%, which is what
+//! identifies it as per-launch overhead rather than anything kernel-related.
+//!
+//! # What the types enforce
+//!
+//! **Captured buffers cannot be dropped while the graph can replay.** Capture
+//! records raw device *addresses*, not Rust borrows, so the driver can never
+//! report which allocations a graph references. [`OwnedGraphExec`] sidesteps that
+//! entirely: the state moves into the exec, so nothing outside can name the
+//! buffers, let alone free them. The use-after-free is not rejected -- it is
+//! unrepresentable.
+//!
+//! That was worth doing rather than documenting, because the violation is
+//! **silent**. Measured on an RTX A2000, replaying a graph whose captured buffer
+//! had been freed returned `Ok(())` from launch, synchronize *and* read-back,
+//! with the expected data -- the replay read freed memory that still held the old
+//! bytes. Stream-ordered deallocation returning pages to a pool makes that the
+//! likely presentation, so the failure mode is "passes in testing, corrupts once
+//! pages are reused" rather than a fault.
+//!
+//! **Capture is terminated even when recording fails.** The driver documents that
+//! after an error the stream sits in `CU_STREAM_CAPTURE_STATUS_INVALIDATED` and
+//!
+//! > The capture sequence must be terminated with `cuStreamEndCapture` on the
+//! > stream where it was initiated in order to continue using `hStream`.
+//!
+//! and separately that while a blocking stream is capturing, the legacy null
+//! stream is unusable too. So a `record` closure returning `Err` would otherwise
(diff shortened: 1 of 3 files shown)


nihalpasham:
A note: since PRs have been submitted for cuda-graphs in the past. 

I'm closing this for direction, not quality 
- same reasoning as #346: we're moving cuda-oxide's host-side crates (cuda-bindings/cuda-core/cuda-async) onto the crates.io-published crates cutile-rs uses, so a Graph API added to cuda-core today lands on a foundation we're about to swap out. 
- This design belongs upstream in those crates, where the current graph surface is the raw `begin`/`end_capture` + `CUgraph` FFI your PR body describes. 
- #107 tracks how that unfolds.

PR #586, opened 2026-07-29
docs(view): by-value readers must document the initialization obligation

honeyspoon:
Four methods in `view.rs` copy their element out **by value**, and none documents that the element must already be initialized:

| method | signature |
|---|---|
| `InBounds32::read` | `fn read(&self) -> T where T: Copy` |
| `InBoundsMut32::read` | `fn read(&self) -> T where T: Copy` |
| `RowView32::get` | `fn get(&self, i: u32) -> Option<T> where T: Copy` |
| `ColView32::get` | `fn get(&self, i: u32) -> Option<T> where T: Copy` |

A slot that was never written holds an indeterminate bit pattern, so producing a `T` from it is undefined behaviour for any `T` whose values are not all bit patterns — `bool`, `char`, enums, references. `T: Copy` does not help: `bool` is `Copy`.

## Why this is reachable, not theoretical

`DeviceBuffer::zeroed` and `from_host` produce initialized memory, but `uninitialized_async` does not until written. Nothing in these four doc comments points at that distinction, so a caller who allocates with `uninitialized_async` and reads before writing has no signal that the read is the problem. Searching `view.rs` for "initiali" returns zero hits today.

## Scope

Docs only, +28 lines in one file, no signature or behaviour change.

Deliberately **not** touched: `InBounds32::get` and `InBoundsMut32::get_mut` hand back a reference rather than a copy, so they carry no such requirement — the asymmetry is the point, and conflating them would overstate the obligation. `DisjointSlice::get_unchecked_mut` likewise returns `&mut T`.

## How it was found

Writing out a per-operation soundness argument for a (now closed) PR that added by-value `read`/`write` to `DisjointSlice`. The obligation applied to the *existing* by-value readers too, which is the part worth keeping regardless of what happens to that API.

## Gates

fmt, clippy `--workspace --all-targets -D warnings`, rustdoc `-D warnings`, tests and doctests clean on `nightly-2026-04-03`.

### DIFF
--- a/crates/cuda-device/src/view.rs
+++ b/crates/cuda-device/src/view.rs
@@ -18,6 +18,24 @@
 //! The pointer fields are private. Safe code can obtain a view only through a
 //! checked slice constructor or a `DisjointSlice` method that consumes a
 //! thread-unique [`crate::thread::ThreadIndex32`].
+//!
+//! # Initialization
+//!
+//! The by-value readers ([`InBounds32::read`], [`InBoundsMut32::read`],
+//! [`RowView32::get`], [`ColView32::get`], [`RowViewIter32::next`],
+//! [`ColViewIter32::next`] and [`ZipView32::next`]) copy the element out of
+//! the underlying buffer, so every slot they touch must already be
+//! **initialized**. Reading a slot that was never written is undefined
+//! behavior for any element type: Rust treats even integers and floats read
+//! from uninitialized memory as invalid. Beyond being written at all, the
+//! slot must hold a valid value of the type, which zeroed memory does not
+//! guarantee for types with invalid bit patterns such as references or enums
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM. 

**One commit on top:** 
- the four duplicated paragraphs became a single module-level `# Initialization` section linked from all seven by-value readers (the three iterator next() methods were missing from the original four), and 
- the "not all bit patterns" carve-out is gone - reading a never-written slot is UB for every element type, integers and floats included; bit-pattern validity is the additional obligation on top. 
- Wording now matches `uninitialized_async`'s own safety contract.

**Validated:** 
- Docs build with warnings denied; 
- generated kernels are byte-identical.

Merging.

nihalpasham (approved):

PR #588, opened 2026-07-30
Support pointer relocations in tuple-array constants

### LINKED ISSUE #587: Preserve pointer provenance in tuple-array constants
## Problem

Bare array constants are currently lowered from their allocation bytes. When an array element is a tuple containing a pointer, rustc stores placeholder bytes in the pointer field and records the actual pointer identity in the allocation provenance map.

Treating those placeholder bytes as an integer pointer value would discard provenance and could materialize an invalid pointer.

For example:

```rust
static FIRST: u32 = 11;
static VALUES: [u32; 3] = [11, 17, 23];

const TABLE: [(&u32, bool); 2] = [
    (&FIRST, false),
    (&VALUES[2], true),
];
```

The second element also requires preserving a non-zero byte addend into a device static.

## Proposed behavior

Extend array constant lowering so tuple elements can retain and consume the original rustc allocation provenance map.

Pointer fields should be materialized from their relocation entry rather than reconstructed from placeholder bytes.

Supported relocations should include:

* pointers to complete device statics,
* pointers to sized subobjects of device statics,
* non-zero static byte addends,
* nested tuple and array ranges that share the original allocation.

## Validation requirements

* Associate every relocation with exactly one aggregate field.
* Reject relocations that cross array-element or tuple-field boundaries.
* Reject relocations located in tuple padding.
* Reject multiple relocations for a single thin-pointer field.
* Reject unsupported allocation targets.
* Do not generate `inttoptr` from pointer placeholder bytes.
* Preserve existing layout handling for padded, reordered, nested, and over-aligned tuple arrays.
* Add runtime and LLVM code-shape regression coverage.
* Remove the obsolete `error_tuple_array_provenance` negative fixture once the positive regression passes.

## Non-goals

This issue does not add support for pointer relocations in:

* direct tuple constants,
* struct constants,
* enum constants,
* device-global initializers,
* unsized or fat-pointer subobjects.

Those cases should continue to produce their existing diagnostics.

uurl:
Closes #587 

## Summary

This PR preserves rustc pointer provenance when lowering array constants whose elements are tuples containing pointers.

Previously, array constants were decoded only from their raw allocation bytes. Pointer fields therefore could not be supported safely because those bytes contain an addend placeholder while the target allocation identity is stored separately in rustc's provenance map.

The new allocation-aware lowering path keeps the original allocation and absolute byte offset while recursively decoding arrays and tuples.

## Implementation

* Add allocation-aware array and tuple constant decoders.
* Carry absolute allocation offsets through nested aggregate lowering.
* Locate pointer relocations within the current aggregate range.
* Validate that relocations are fully contained within exactly one tuple field.
* Reject relocations that cross aggregate boundaries or fall into padding.
* Materialize thin pointer fields from their rustc relocation target.
* Support device-static pointers with zero and non-zero byte addends.
* Preserve sized static-subobject projections through pointer-offset operations.
* Reject unsupported allocation targets, unsized pointees, and malformed relocation layouts.
* Keep direct tuple pointer constants rejected by the existing diagnostic.
* Add tuple-array static-pointer coverage to `array_constants`.
* Add LLVM shape checks for the non-zero static subobject projection and absence of `inttoptr`.
* Remove the obsolete `error_tuple_array_provenance` fixture and its smoketest classification.
* Update the supported-features documentation.

## Scope

This PR supports pointer-bearing tuple elements inside array constants.

It intentionally does not enable pointer relocations in:

* direct tuple constants,
* struct constants,
* enum constants,
* device-global initializers,
* fat or unsized pointer fields.

## Validation

```text
git diff --check
cargo fmt --all -- --check
cargo check -p mir-importer
cargo clippy -p mir-importer --all-targets --all-features -- -D warnings
./scripts/check-error-example-status.sh
cargo oxide build array_constants
```

The runtime regression validates both:

* a zero-addend pointer to a complete device static,
* a non-zero pointer addend into a device-static array.

`error_tuple_constant_provenance` continues to fail with:

```text
Tuple constant contains 1 pointer relocation(s); cuda-oxide cannot yet preserve tuple pointer provenance
```


nihalpasham:
LGTM. 

**Two commits on top:** 
- restored the sorted import order in the example (root cargo fmt doesn't cover the standalone example crates, so this slips through), and 
- the 0/1/2/3 value-kind magic numbers in the new decoder are now a function-local enum matching the file's convention.

**Verified on hardware:** 
- `array_constants` passes including its code-shape check - `getelementptr` on the named global, no `inttoptr`.

Merging; 

closes #587.

nihalpasham (approved):

PR #590, opened 2026-07-30
Support pointer relocations in device global initializers

### LINKED ISSUE #589: Support pointer relocations in device-global static initializers
# Description

## Problem

cuda-oxide currently preserves the literal byte image of initialized Rust statics, but static initializers that contain pointers also carry rustc allocation provenance.

For example:

```rust
static TARGET: u32 = 0x1234_5678;
static REFERENCE: &u32 = &TARGET;
```

The initializer for `REFERENCE` is not just eight literal bytes. It contains a pointer-sized addend together with provenance identifying `TARGET`.

Flattening the allocation to bytes would discard that provenance and could silently emit a null or otherwise invalid pointer.

## Expected behavior

Preserve static-to-static pointer relocations through the complete compilation pipeline:

1. Extract relocation records from rustc allocation provenance in `mir-importer`.
2. Carry the relocation metadata through MIR lowering.
3. Represent pointer slots without losing the exact surrounding initializer bytes.
4. Resolve relocation targets independently of global declaration order.
5. Emit valid LLVM/NVVM constant expressions for both modern opaque pointers and legacy LLVM 7 typed pointers.

## Initial scope

Support thin pointers between Rust device statics located in:

* CUDA global memory, address space 1
* CUDA constant memory, address space 4

The implementation should support:

* direct static-to-static references;
* multiple pointer fields in one initializer;
* repeated references to the same target;
* references to different targets;
* forward references;
* zero and non-zero byte addends;
* interior pointers;
* statics reachable only through another static initializer.

## Validation requirements

Relocation handling must fail closed when encountering:

* unsupported pointer widths;
* malformed or overlapping relocation slots;
* unaligned pointer slots;
* out-of-bounds source offsets;
* unknown target identities;
* mismatched target address spaces;
* target addends beyond the target allocation;
* functions, vtables, or anonymous allocations;
* fat pointers or unsized pointees;
* packed or otherwise unaligned pointer storage.

## Acceptance criteria

* `static REFERENCE: &u32 = &TARGET` compiles and executes correctly on the GPU.
* Multiple relocations in one initializer are supported.
* Non-zero interior-pointer addends are preserved.
* Global declaration order does not affect relocation resolution.
* Modern and legacy NVVM IR exporters emit valid relocation expressions.
* Existing pointer-free static initializers remain byte-exact.
* Unit tests cover encoding, decoding, lowering, export, and malformed metadata.
* The `device_global` example validates the behavior end to end.
* The obsolete negative example and support-gap status entry are removed.

uurl:
Closes #589 

# Description

## Summary

This PR adds support for thin pointer relocations stored inside initialized Rust device statics.

Rustc represents these pointers as pointer-sized addend bytes plus allocation provenance identifying the target static. The previous initializer path preserved literal bytes but could not represent that provenance.

The new pipeline carries relocation records from rustc allocation data through MIR lowering and reconstructs them as LLVM/NVVM constant expressions during export.

### `mir-importer`

* Extracts pointer relocation records from evaluated static allocations.
* Records:

  * source byte offset;
  * pointer width;
  * target address space;
  * target byte addend;
  * stable rustc target identity.
* Accepts only relocations targeting Rust statics in CUDA global or constant memory.
* Recursively materializes statics reachable through initializer relocations.
* Registers globals before traversing dependencies so self-references and cyclic static graphs terminate.
* Preserves deterministic zero bytes for undefined Rust padding.
* Keeps promoted-array provenance outside the supported scope and fails closed.

### `mir-lower`

* Propagates relocation metadata to LLVM globals.
* Uses segmented physical storage for relocated initializers:

  * `[N x i8]` fields preserve literal byte spans;
  * `i64` fields represent pointer-width relocation slots.
* Validates pointer width, alignment, bounds, overlap, target address space, and storage size.
* Preserves the stable rustc source key on each lowered global.

### `llvm-export`

* Indexes globals by stable rustc source key before emitting definitions.
* Resolves relocation targets independently of textual global order.
* Reconstructs pointer constants using:

  * `getelementptr`;
  * `addrspacecast`;
  * `ptrtoint`.
* Supports:

  * modern opaque-pointer NVVM IR;
  * legacy LLVM 7 typed-pointer NVVM IR.
* Rejects unknown targets, conflicting source keys, address-space mismatches, malformed storage, and invalid addends.

## Supported scope

This PR supports thin pointers from one device static to another device static in:

* address space 1, CUDA global memory;
* address space 4, CUDA constant memory.

It includes:

* direct references;
* multiple relocations;
* repeated targets;
* distinct targets;
* forward references;
* zero and non-zero byte addends;
* interior pointers;
* transitively reachable statics.

## Out of scope

The following cases continue to fail closed:

* functions and vtables;
* slices and other fat pointers;
* trait-object metadata;
* unsized pointees;
* packed or unaligned pointer slots;
* relocation targets outside device static storage.

## Tests

Added coverage for:

* relocation metadata encode/decode round trips;
* truncated metadata rejection;
* segmented storage lowering;
* overlapping relocation rejection;
* direct static pointer export;
* multiple relocations;
* non-zero addends;
* global and constant address spaces;
* forward-reference resolution;
* unknown target rejection;
* modern and legacy NVVM IR output;
* end-to-end GPU execution in `device_global`.

Validated with:

```text
cargo fmt --all
cargo test -p mir-importer
cargo test -p mir-lower
cargo test -p llvm-export
scripts/smoketest.sh --compile-only -o '^device_global$'
cargo oxide run device_global
```

The GPU test validates:

```text
Static initializer relocations:
[305419896, 305419896, 3405691582, 1090519040]
```

The result covers direct references, repeated targets, a second target, and an interior pointer with a non-zero byte addend.

## Repository cleanup

* Removes `error_static_initializer_provenance`.
* Removes the corresponding support-gap row from `STATUS.md`.
* Removes the obsolete error-example entry from `scripts/smoketest.sh`.
* Extends the `device_global` example and documentation.


nihalpasham:
LGTM.

**Rebased:** the branch predated yesterday's constant-provenance stack, so it conflicted in `rvalue.rs`, STATUS.md, and the smoketest error list. The resolution keeps the new aggregate decoders and this PR's static materialization side by side; the error list becomes main's minus `error_static_initializer_provenance`. Also folded in the lint fixes the pliron bump exposed in the new tests.

**Verified on hardware:**
- `device_global` passes on GPU including the new `static_initializer_relocations` kernel - base target, second reference to the same target, addend into a static array, and the one-past-the-end offset all read back correctly.
- NVVM verification passes; error fixtures 12/12 with `error_static_initializer_provenance` retired; full compile-only sweep 177/177.
- 891 mir-importer / 147 mir-lower / 76 llvm-export tests; fmt and clippy clean.

Merging;

closes #589.

nihalpasham (approved):

PR #595, opened 2026-07-30
Fix enum constant pointer provenance during MIR import

### LINKED ISSUE #594: Support pointer provenance in enum constants
# Description

Enum constants containing references or thin pointers cannot currently be imported without losing rustc pointer provenance.

For example:

```rust
static TARGETS: [u64; 2] = [
    0x1122_3344_5566_7788,
    0x8877_6655_4433_2211,
];

const VALUE: Option<&'static u64> = Some(&TARGETS[1]);
```

rustc represents the pointer field using two separate pieces of information:

1. the bytes stored in the allocation, which encode the byte addend;
2. a relocation entry identifying the target allocation.

The enum constant importer currently treats the enum as a byte image without preserving its enclosing allocation and provenance map. Following or discarding the relocation can therefore replace the pointer with the pointee contents, misclassify a niche-encoded variant, or otherwise lose the target allocation identity.

This affects both:

* niche-encoded enums such as `Option<&T>`;
* direct-tagged enums containing a reference payload.

## Expected behavior

The importer should:

* retain the enum's outer allocation while decoding the active variant;
* use relocation overlap to distinguish a valid niche pointer from the null niche;
* reconstruct direct thin-pointer fields from rustc relocation entries;
* preserve non-zero byte addends into device statics;
* reject unsupported or ambiguous relocation layouts instead of decoding them as ordinary bytes.

## Scope

This issue covers direct thin references to device statics in enum constants.

The following cases may remain unsupported:

* pointer relocations nested inside aggregate enum fields;
* pointer and integer payloads sharing overlapping storage;
* address-space-3 pointer layouts whose physical width depends on the selected exporter mode;
* fat pointers with provenance.

## Acceptance criteria

* `Option<&'static T>` constants correctly distinguish `Some` from `None`.
* Interior pointers into device statics preserve their byte addend.
* Direct-tagged enums with reference payloads execute correctly on the GPU.
* Unconsumed, partial, or ambiguous relocations fail closed.
* A runtime regression example validates the generated GPU result.

uurl:
Closes #594 

# Description

## Summary

Preserve rustc pointer provenance while importing enum constants.

This changes enum constant translation to retain the enclosing allocation and its relocation map instead of decoding the enum exclusively from raw bytes.

## Problem

Pointer-bearing constants are represented by rustc as allocation bytes plus relocation metadata. The bytes beneath a relocation contain only the pointer addend; they do not contain a complete exposed address.

The previous enum constant path did not retain this provenance information. As a result, it could not safely:

* distinguish a relocated niche pointer from an all-zero `None` carrier;
* reconstruct a direct pointer payload;
* preserve an interior byte offset into a device static.

The existing negative example therefore rejected every enum constant containing pointer relocations.

## Implementation

This PR:

* retains the outer allocation when translating enum constants;
* adds allocation-aware enum variant decoding;
* detects pointer relocations overlapping a niche carrier;
* treats one exact, full-width relocation as the untagged pointer variant;
* reconstructs direct thin-reference fields through the existing static-global pointer path;
* preserves non-zero relocation addends using checked `u64` conversion;
* audits enum relocations after translating active variant fields;
* rejects partial, ambiguous, nested, or otherwise unconsumed relocations;
* adds unit coverage for relocation-range overlap detection.

## Runtime regression

The former `error_enum_constant_provenance` fixture is converted into the successful `enum_constant_provenance` runtime example.

The example validates:

* a niche-encoded `Option<&T>` pointing at an interior device-static element;
* the `None` niche variant;
* a direct-tagged empty variant;
* a direct-tagged variant with a reference payload.

Expected output:

```text
enum_constant_provenance: PASS
```

## Unsupported cases

* pointer relocations nested inside aggregate enum fields;
* pointer/integer overlapping enum storage;
* address-space-3 pointer layouts;
* fat pointers with provenance.

These cases continue to fail closed.

## Validation

The following checks pass:

```text
cargo fmt --all -- --check
git diff --check
cargo test -p mir-importer
cargo test -p mir-lower
cargo clippy --manifest-path crates/rustc-codegen-cuda/examples/enum_constant_provenance/Cargo.toml --all-targets -- -D warnings
scripts/smoketest.sh --compile-only -o '^enum_constant_provenance$'
cargo oxide run enum_constant_provenance
```

Results:

* `mir-importer`: 892 tests passed;
* `mir-lower`: 145 unit tests and 76 integration tests passed;
* libNVVM verification and compilation passed;
* the runtime regression passed on an NVIDIA GeForce RTX 5060 Laptop GPU.


nihalpasham:
LGTM as-is.

**Verified on hardware:**

- `enum_constant_provenance` prints PASS on GPU - the niche relocation selects Some, and the addend into the static array reads the right element.
- `error_enum_pointer_overlap` and `error_enum_shared_pointer_layout` still fail closed; 
- 892 mir-importer tests pass.

Merging;

closes #594.

nihalpasham (approved):

PR #596, opened 2026-07-30
codegen: opt-in libdevice linking for the standalone PTX API

### LINKED ISSUE #485: Experimental standalone API cannot compile libdevice calls, though the pipeline can now link them

alejandro-soto-franco:
Closes #485.

A frontend driving `cuda_oxide_codegen::experimental` stops at the first `sqrt`. Float intrinsics lower to libdevice `__nv_*` calls, and v1 rejects every unresolved external symbol. The rejection at `pipeline.rs:364` sits directly above the IR-level libdevice link at `pipeline.rs:417`, so a module is refused for symbols the step below it would have resolved.

```rust
pub enum Linking { SelfContained, Libdevice }

impl CompileOptions {
    pub fn with_linking(self, linking: Linking) -> Self;
    pub fn linking(&self) -> Linking;
}
```

`Linking::SelfContained` is the default and changes no compilation that succeeds today: no link runs, and the symbol filter is the identity.

### 1. The option belongs in experimental v1

**Chosen:** a default-off option on the existing surface.

**Alternative:** hold it for a v2 that redesigns linking properly.

No existing consumer changes, and a v2 stays free to supersede it. It is the smallest shape that unblocks an out-of-tree frontend and the cheapest to withdraw.

### 2. A `CompileOptions` field

**Chosen:** a field, with discovery inside the pipeline.

**Alternative:** the caller passes a path to `libdevice.10.bc`.

Discovery reuses the `libnvvm_sys::find_libdevice()` call `compile_translated_module` already makes on this path, so the API gains no filesystem search it did not already perform. A caller-supplied path adds a parameter that duplicates work already done.

### 3. No NVVM IR fallback for standalone callers

**Chosen:** one path in, one artifact kind out.

**Alternative:** fall back to NVVM IR when the link is unavailable, as `ExternalLinkAllowed` does.

`should_emit_nvvm_ir` returns `false` for `SelfContainedPtx` whatever the flag holds, and a test loops all eight combinations of the flag, strict-libdevice use and link availability to pin it. When the link is needed and unavailable, compilation fails with a new `CompileError::LibdeviceUnavailable` naming which piece is missing, since an absent `libdevice.10.bc` and an absent `llvm-link` have different fixes.

## What the filter admits

Symbols beginning `__nv_`. The predicate sits beside `unresolved_external_symbols`, so the libdevice detector, the surviving-declaration scan and the filter read one definition. Unresolved symbols that are not libdevice are still rejected, and a module carrying both loses only the libdevice half.

This admits any `__nv_*` symbol, including one a frontend names itself. Several libdevice functions have no Rust intrinsic and so no placeholder callee for `mir-lower` to recognise, `__nv_erff` and `__nv_rhypotf` among them. The alternative for such a frontend is a polynomial approximation, which gives up bit-exactness against nvcc for a function CUDA already ships. Both routes have tests.

## Fixing a precision defect

The first test asserting which instruction libdevice produces failed. libdevice selects an implementation by reading reflect variables, and `NVVMReflectPass` resolves an unset one to 0.

| reflect variable | unset | nvcc default | this PR |
|---|---|---|---|
| `__CUDA_PREC_SQRT` | 0, selects `sqrt.approx.f32` | `-prec-sqrt=true` | passes 1 |
| `__CUDA_FTZ` | 0 | `-ftz=false`, already matches | leaves unset |
| `__CUDA_ARCH` | resolved from `-mcpu` | | leaves unset |

Nothing in the tree passed a reflect variable to `llc`, so every `sqrt` reaching libdevice compiled to `sqrt.approx.f32` where nvcc produces `sqrt.rn.f32`.

**This reaches the rustc frontend too.** The arguments are keyed on a link having happened rather than on the output policy, and `sqrt` has no entry in `llvm_intrinsic_name`, so any kernel calling `f32::sqrt` moves to `sqrt.rn.f32` wherever IR-level linking is available. That is the result Rust specifies for `f32::sqrt` and the one nvcc produces by default, and it removes a disagreement between this path and the libNVVM fallback. Callers wanting the fast path have `cuda_device::float::sqrt_approx_f32`. Nothing in the tree asserts `sqrt.approx.f32` as a libdevice-path output, and `rustc-codegen-cuda` is a separate workspace with no PTX assertions, so no written expectation changes.

Happy to split this into its own pull request if you would rather judge it on numerics. It touches only `ptx.rs` and applies cleanly to `main`.

## A caveat

**`llvm-link --only-needed` resolves what it has and stays silent about the rest.** A `__nv_*` name libdevice does not define survived into the PTX as `.extern .func` and failed at `cuModuleLoad`; `ptxas` accepts such PTX, so nothing downstream caught it. The pipeline now scans for surviving `__nv_*` declarations after linking and reports them through `UnsupportedLinking`.

That scan is scoped to `SelfContainedPtx`. `can_ir_link_libdevice` is computed from tool availability alone, so `ExternalLinkAllowed { request_nvvm_ir: false }` can also reach the direct-PTX-with-linked-libdevice route on a machine holding both pieces, and the same class of unresolved symbol stays reachable there. That predates this branch and is untouched by it.

## Verification and testing

160 tests across 6 suites, debug and release. fmt, clippy `--workspace --all-targets -D warnings` and rustdoc clean. +915/−26 against `main`, seven files.

`ptxas -arch=sm_120` accepts every generated artifact. On device (RTX 5060 Laptop, sm_120) the `sqrt` kernel is bit-exact against `f32::sqrt` over 256 values; before the precision fix it computed the approximate result.

The integration tests require `llvm-link` and a discoverable `libdevice.10.bc`, which are new requirements for this crate. The `cuda-oxide-codegen` job sets `needs_cuda: true` and `rust-toolchain.toml` includes `llvm-tools`, so CI has both.

### DIFF
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -96,6 +96,21 @@ pub enum DebugInfo {
     Full,
 }
 
+/// Whether compilation may resolve libdevice calls at the LLVM IR level.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Linking {
+    /// Reject every unresolved external symbol. The generated PTX is
+    /// self-contained and the CUDA driver can load it with no further step.
+    #[default]
+    SelfContained,
+    /// Link `libdevice.10.bc` into the module at the LLVM IR level, resolving
+    /// `__nv_*` calls before `llc` runs. Unresolved symbols that are not
+    /// libdevice are still rejected, and the output is still a single
+    /// self-contained PTX artifact.
+    Libdevice,
+}
+
 /// Typed options for one standalone PTX compilation.
 #[derive(Clone, Debug)]
 #[non_exhaustive]
@@ -105,6 +120,7 @@ pub struct CompileOptions {
     fma_contraction: bool,
     debug_info: DebugInfo,
     verbose: bool,
+    linking: Linking,
 }
 
 impl CompileOptions {
@@ -116,6 +132,7 @@ impl CompileOptions {
             fma_contraction: true,
             debug_info: DebugInfo::None,
             verbose: false,
+            linking: Linking::SelfContained,
         }
     }
 
@@ -136,6 +153,16 @@ impl CompileOptions {
     /// [`DebugInfo::Full`] must be paired with [`Optimization::None`].
     pub fn with_debug_info(mut self, debug_info: DebugInfo) -> Self {
         self.debug_info = debug_info;
+        self
+    }
+
+    /// Select whether libdevice calls may be resolved by an IR-level link.
+    ///
+    /// [`Linking::SelfContained`], the default, rejects any unresolved
+    /// external symbol. [`Linking::Libdevice`] resolves `__nv_*` calls
+    /// against `libdevice.10.bc`.
+    pub fn with_linking(mut self, linking: Linking) -> Self {
+        self.linking = linking;
         self
     }
 
@@ -168,6 +195,11 @@ impl CompileOptions {
     /// Requested device debug-information tier.
     pub fn debug_info(&self) -> DebugInfo {
         self.debug_info
+    }
+
+    /// Requested libdevice linking policy.
+    pub fn linking(&self) -> Linking {
+        self.linking
     }
 
     /// Whether progress and tool-selection diagnostics were requested.
@@ -632,6 +664,7 @@ impl Compiler {
             &backend_options,
             debug_kind,
             &self.toolchain.inner,
+            options.linking == Linking::Libdevice,
             OutputFiles {
                 llvm_ir: &ll_path,
                 ptx: &ptx_path,
@@ -977,6 +1010,17 @@ mod tests {
         assert!(
             cloned.try_deref(&module.context).is_err(),
             "the cloned operation should be erased even though the guarded closure panicked"
+        );
+    }
+
+    #[test]
+    fn linking_defaults_to_self_contained() {
+        let options = CompileOptions::new(Target::parse("sm_90").unwrap());
+        assert_eq!(options.linking(), Linking::SelfContained);
+        assert_eq!(Linking::default(), Linking::SelfContained);
+        assert_eq!(
+            options.with_linking(Linking::Libdevice).linking(),
+            Linking::Libdevice
         );
     }
 }
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -677,7 +677,15 @@ impl Compiler {
         (|| {
             let generated =
                 compile_translated_module(ctx, module, &request).map_err(CompileError::from)?;
-            debug_assert_eq!(generated.artifact_kind, ModuleArtifactKind::Ptx);
+            if generated.artifact_kind != ModuleArtifactKind::Ptx {
+                return Err(CompileError::Codegen {
+                    message: format!(
+                        "the standalone pipeline produced {:?} where PTX was required; \
+                         this path never emits NVVM IR",
+                        generated.artifact_kind
+                    ),
+                });
+            }
             let ptx = std::fs::read(&ptx_path).map_err(|source| CompileError::Io {
                 action: "read generated PTX",
                 path: ptx_path.clone(),
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -160,7 +160,9 @@ impl CompileOptions {
     ///
     /// [`Linking::SelfContained`], the default, rejects any unresolved
     /// external symbol. [`Linking::Libdevice`] resolves `__nv_*` calls
-    /// against `libdevice.10.bc`.
+    /// against `libdevice.10.bc` and fails with
+    /// [`CompileError::LibdeviceUnavailable`] when the toolchain cannot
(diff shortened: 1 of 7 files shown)


alejandro-soto-franco:
Re-verified on CUDA 13.3.1, having been developed against 13.1.

- 160 tests and 6 suites are unchanged.
- `sqrt.rn.f32` in the emitted PTX.
- On device, sm_120: bit-exact against `f32::sqrt` over 256 values.

13.3's libdevice also references only `__CUDA_ARCH`, `__CUDA_FTZ` and `__CUDA_PREC_SQRT`, so `__CUDA_PREC_DIV` is inert on that build too.

nihalpasham:
LGTM.

**Verified on hardware:**

- all `cuda-oxide-codegen` suites pass locally (162 tests) with llvm-link, libdevice, and ptxas present.
- the other fix reaches the rustc frontend as intended: `libdevice_math` passes on GPU and its PTX carries `sqrt.rn.f32` for `f32::sqrt`; the only approx form left is rsqrt, which has no precision knob. `sqrt_approx_f32` remains the explicit fast path.
- **one correction to the PR's write-up:** clang doesn't actually match nvcc here. I checked clang-21 and 22 - both leave `__CUDA_PREC_SQRT` unset and ship approximate `sqrt`, same hole we had. nvcc and libNVVM are the ones that default precise, and Rust documents `f32::sqrt` as correctly rounded, so this is the right call. Will release-note the `sqrt` change.

Merging;

closes #485.

nihalpasham (approved):

PR #598, opened 2026-07-31
feat(wgmma): add deferred BF16 accumulator lowering

### LINKED ISSUE #597: Support deferred accumulator lowering for BF16 WGMMA m64n64k16
## Problem

The public WGMMA API exposes:

```rust
wgmma_mma_m64n64k16_f32_bf16(
    acc: &mut [[f32; 8]; 4],
    desc_a: u64,
    desc_b: u64,
)
```

However, this operation cannot be lowered safely as an isolated pointer-based MMA call.

Each thread owns 32 `f32` accumulator elements. Those values must remain in registers across the asynchronous WGMMA sequence:

```text
wgmma.fence
one or more wgmma.mma_async instructions
wgmma.commit_group
wgmma.wait_group 0
```

Loading and storing the accumulator around each individual MMA call would violate the WGMMA lifetime contract because the updated values are not available until the corresponding wait completes.

The current fail-closed behavior therefore rejects BF16 WGMMA MMA calls instead of generating incorrect PTX.

## Proposed solution

Add a deferred-accumulator lowering path for:

```text
wgmma.mma_async.sync.aligned.m64n64k16.f32.bf16.bf16
```

The compiler should:

1. Import BF16 WGMMA MMA calls as pointer-form dialect operations.
2. Recognize a complete statically linear sequence consisting of:

   * one `wgmma_fence`;
   * one or more BF16 MMA calls using the same accumulator;
   * exactly one `wgmma_commit_group`;
   * one `wgmma_wait_group::<0>`.
3. Replace the sequence with a deferred WGMMA group operation.
4. Lower the group into one convergent inline-PTX scope that:

   * loads all 32 accumulator values;
   * emits the fence;
   * emits all MMA instructions;
   * commits the group;
   * waits for group zero;
   * stores the accumulator only after the wait.

## Initial scope

The initial implementation should support only:

```text
m64n64k16.f32.bf16.bf16
```

The sequence must satisfy all of the following:

* all operations form one straight-line control-flow chain;
* every MMA uses the same accumulator;
* the accumulator is a mutable generic MIR pointer;
* both descriptors are unsigned `u64` values;
* exactly one commit operation is present;
* the sequence ends with `wait_group<0>`;
* no unsupported operation appears between the fence and wait.

The implementation should reject:

* F16 and TF32 MMA variants;
* partial waits;
* multiple accumulator objects;
* multiple commits;
* MMA instructions after the commit;
* nested fences;
* branches and control-flow joins;
* loops;
* incomplete fence/MMA/commit/wait sequences;
* malformed operation shapes or descriptor types.

## Acceptance criteria

* [ ] A valid BF16 sequence is fused before final MIR-to-LLVM lowering.
* [ ] The generated group lowers to one convergent inline-PTX scope.
* [ ] All accumulator loads occur before the WGMMA instructions.
* [ ] All accumulator stores occur after `wgmma.wait_group.sync.aligned 0`.
* [ ] Multiple MMA instructions using the same accumulator are supported within one group.
* [ ] Unsupported or malformed sequences fail with structured compiler errors rather than panics.
* [ ] F16 and TF32 entry points remain fail-closed.
* [ ] Pointer-form and deferred-group verifiers enforce exact operand, result, pointer and descriptor contracts.
* [ ] A compile-only `sm_90a` example exercises the supported BF16 path.
* [ ] Dialect, lowering, codegen and WGMMA smoketests pass.

## Hardware note

Functional execution requires a Hopper `sm_90a` GPU and valid shared-memory descriptors.

A compile-only example may use zero descriptors to validate compiler lowering and PTX generation, but that kernel must not be launched.

uurl:
Closes #597 

## Summary

This PR adds initial lowering support for:

```text
wgmma.mma_async.sync.aligned.m64n64k16.f32.bf16.bf16
```

The implementation preserves the 32 per-thread `f32` accumulator values across the asynchronous WGMMA fence/MMA/commit/wait sequence.

Instead of lowering each pointer-form MMA independently, `mir-lower` recognizes a complete supported sequence and replaces it with a deferred WGMMA group. The group is then lowered into one convergent inline-PTX scope, keeping the accumulator in registers until `wait_group<0>` completes.

## Motivation

WGMMA accumulator values are asynchronous outputs.

A direct pointer-based lowering would need to load and store the accumulator around an individual MMA call, but the updated values are not valid until the committed group has completed. Storing them before `wgmma.wait_group.sync.aligned 0` would therefore produce incorrect semantics.

This PR models the complete asynchronous region explicitly and keeps unsupported sequences fail-closed.

### Public API and importer

* Enables importing `wgmma_mma_m64n64k16_f32_bf16`.
* Keeps the existing pointer-form dialect operation as an intermediate carrier.
* Continues rejecting the F16 and TF32 variants with an explicit unsupported diagnostic.
* Documents the required fence/MMA/commit/wait lifetime contract.

### NVVM dialect

* Adds:

```text
nvvm.wgmma_mma_group_m64n64k16_f32_bf16
```

* Represents one accumulator and one or more descriptor pairs:

```text
[acc_ptr, desc_a_0, desc_b_0, ..., desc_a_n, desc_b_n]
```

* Strengthens both pointer-form and deferred-group verification.
* Requires the accumulator to be a mutable generic MIR pointer.
* Requires every descriptor to be an unsigned `u64`.
* Rejects malformed operand and result counts.

### Deferred-accumulator fusion

Adds a preprocessing pass that recognizes:

```text
wgmma_fence
one or more BF16 WGMMA MMA calls using the same accumulator
wgmma_commit_group
wgmma_wait_group::<0>
```

The pass supports a straight-line sequence across unconditional basic-block transitions and rejects ambiguous or incomplete regions.

It fails closed for:

* multiple accumulators;
* multiple commits;
* MMA operations after the commit;
* nested fences;
* nonzero or non-`u64` wait-group constants;
* malformed MMA descriptor operands;
* intervening unsupported operations;
* branches, joins and loops;
* regions that end before `wait_group<0>`.

### LLVM lowering

The deferred group lowers to one convergent inline-PTX scope that:

1. declares 32 `f32` accumulator registers;
2. loads the existing accumulator values;
3. emits `wgmma.fence.sync.aligned`;
4. emits one or more BF16 WGMMA MMA instructions;
5. emits `wgmma.commit_group.sync.aligned`;
6. emits `wgmma.wait_group.sync.aligned 0`;
7. stores all accumulator values after the wait.

The original pointer-form operation remains illegal at final lowering unless it has been consumed by the fusion pass.

### Tests and examples

* Adds positive coverage for fusing and lowering a valid WGMMA sequence.
* Adds negative coverage for:

  * branches and joins;
  * intervening operations;
  * nested fences;
  * missing or duplicate commits;
  * MMA after commit;
  * multiple accumulators;
  * unsupported wait groups;
  * malformed pointers and descriptors.
* Adds verifier coverage for immutable, non-generic and non-pointer accumulators.
* Replaces the previous negative WGMMA MMA example with a compile-only BF16 example.
* Adds the BF16 example to the standard WGMMA smoketest set.
* Updates the WGMMA documentation and support status.

## Current limitations

This PR intentionally supports only:

```text
m64n64k16.f32.bf16.bf16
```

The following remain unsupported:

* F16 and TF32 WGMMA MMA variants;
* partial waits;
* multiple accumulator objects;
* multiple commit groups;
* branches and control-flow joins;
* loops;
* arbitrary operations inside the deferred region.

These cases continue to produce structured compiler errors.

## Validation

The following commands pass:

```bash
cargo fmt --all
cargo test -p dialect-nvvm
cargo test -p mir-lower
cargo test -p cuda-oxide-codegen
scripts/smoketest.sh -x -v '^(wgmma|wgmma_mma_bf16)$'
```

Observed results:

* `dialect-nvvm`: 50 tests passed.
* `mir-lower`: 146 unit tests and 88 integration tests passed.
* `cuda-oxide-codegen`: unit, compilation, concurrency, PTX and target-provenance suites passed.
* WGMMA smoketest: 2 of 2 examples passed.

## Hardware validation

The compile and smoketest validation was performed on a Blackwell GPU, where Hopper WGMMA instructions cannot execute.

The BF16 example therefore validates compilation and compiler lowering only. Functional execution still requires:

* a Hopper `sm_90a` GPU;
* valid WGMMA shared-memory descriptors;
* a complete 128-thread warpgroup.


nihalpasham:
LGTM, with one thing I want to be explicit about before merging.

**On the design:** two ways to lower this exist. CUTLASS and Triton thread the accumulator through registers across separate `asm` statements; this PR fuses everything into one `asm` block and round-trips the accumulator through memory:

```
CUTLASS / Triton (value-threaded)        This PR (single fused block)
---------------------------------        ----------------------------
acc lives in 32 vregs                    acc lives in memory (&mut [[f32; 8]; 4])

asm { wgmma.fence }                      asm {
asm { wgmma.mma_async }  <- acc crosses    ld.f32  x32   acc -> regs
asm { wgmma.mma_async }     asm            wgmma.fence
asm { commit_group }        boundaries     wgmma.mma_async
  ... issue more groups,                   wgmma.commit_group
      overlap other work ...               wgmma.wait_group 0
asm { wait_group<N> }    <- partial        st.f32  x32   regs -> acc
read acc                    waits        }
                                         read acc   (already waited)
 ```
The CUTLASS/Triton approach on the left is what makes WGMMA fast (multiple groups in flight, register-resident K-loops), but it has a real soundness hole for us at the LLVM layer:

```
asm { mma_async %acc0..%acc31 }    accumulator now in flight
       |
       |   <- LLVM may spill/reload these vregs right here;
       |      the PTX ISA forbids touching in-flight accumulators
       v
asm { wait_group }
```

The fused block can't hit that by construction, and Rust MIR keeps an address-taken accumulator in memory anyway (no aggregate-SSA machinery yet). 

So: this sound bridge is okay for now but a pipelined, value-threaded path as a tracked follow-up would make sense.

**Two commits on top:**

- The docs said loops are rejected outright. Actually only a fence-to-wait sequence crossing a branch, join, or loop back-edge is rejected; a complete sequence inside a loop body fuses fine, paying the 32-value round-trip plus a full wait per iteration. Docs, README, and book now say that.
- `fmt` and `clippy` for CI: import order in the example (root `cargo fmt` doesn't reach standalone example crates), and `manual_is_multiple_of` in the verifier.

**Verified:** 
- 891 mir-importer and 146+88 mir-lower tests, 
- `wgmma` and `wgmma_mma_bf16` smoketest-pass, 
- error-example status clean. 

Compile-verified only - I don't have Hopper hardware, so descriptor handling and end-to-end numerics on sm_90a remain unexecuted.

Merging;

closes #597.

nihalpasham (approved):

PR #600, opened 2026-07-31
test(cuda-oxide-codegen): make the libdevice re-exec check prove it ran

### LINKED ISSUE #599: libdevice_linking re-exec check passes when the child runs no test, and CUDA_OXIDE_LLVM_LINK failures do not name the variable
## Summary

Two problems in `crates/cuda-oxide-codegen/tests/libdevice_linking.rs` (new in
#596). Both are about the test misleading its reader rather than about the
libdevice feature, which works.

1. The re-exec guard passes when the child ran **no** test, so the check can
   silently stop running while staying green.
2. When an exported `CUDA_OXIDE_LLVM_LINK` makes libdevice unavailable, four
   tests fail without naming the variable that caused it.

## 1. The re-exec guard can verify nothing

`libdevice_linking_reports_unavailable_when_llvm_link_cannot_run` re-execs itself
in a child so a broken `CUDA_OXIDE_LLVM_LINK` cannot race its siblings — a nice
piece of isolation. But the parent only checks the exit status:

```rust
let output = std::process::Command::new(exe)
    .arg("--exact")
    .arg("libdevice_linking_reports_unavailable_when_llvm_link_cannot_run")
    ...
assert!(output.status.success(), ...);
```

libtest exits 0 when a filter matches nothing:

```
$ ./libdevice_linking-3896ba24ec1514e7 --exact this_test_does_not_exist --nocapture
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 6 filtered out
$ echo $?
0
```

So the `--exact` string and the function name are coupled only by convention. If
the test is renamed — or the literal is mistyped — the child runs nothing, exits
0, the parent asserts success, and the suite reports green while the
LibdeviceUnavailable path is no longer covered at all. Nothing in the file would
reveal that.

The check itself is worth protecting: it is the only thing proving a broken
`llvm-link` cannot silently emit PTX.

## 2. A `CUDA_OXIDE_LLVM_LINK` failure does not name `CUDA_OXIDE_LLVM_LINK`

`Toolchain::discover()` reads `CUDA_OXIDE_LLVM_LINK` from the environment (the
file's own comment says so). Exporting it — the documented way to point at an
`llvm-link` that is not a sibling of `llc` — at a missing or major-mismatched
binary fails four tests:

```
$ CUDA_OXIDE_LLVM_LINK=/nonexistent/llvm-link cargo test -p cuda-oxide-codegen
libdevice_linking_resolves_a_rust_intrinsic_kernel        FAILED
libdevice_linking_resolves_a_frontend_declared_symbol     FAILED
libdevice_linking_rejects_a_symbol_libdevice_does_not_define  FAILED
libdevice_linking_still_rejects_a_device_extern           FAILED

libdevice linking resolves __nv_sqrtf: LibdeviceUnavailable { message:
  "no `llvm-link` sharing the selected `llc`'s LLVM major was found" }
```

Nothing there points at the override. The same file already does this well for
the other external tool:

```rust
"could not run {}: {error}\nThis test needs `ptxas` from a CUDA toolkit (no GPU
 required). Point CUDA_TOOLKIT_PATH or CUDA_HOME at the install root, or put
 `ptxas` on PATH."
```

`llvm-link` deserves the same treatment. To be clear about severity: an override
pointing at a broken tool *should* fail — the complaint is only that the failure
does not say which knob to turn. The realistic case is a machine with several
LLVMs where the override is valid for the user's own builds but its major does
not match the pinned `llc`.

## Proposed fix

1. Have the child print a sentinel once the check has actually executed, and have
   the parent assert on that in addition to the exit status. Exit status alone
   cannot distinguish "passed" from "never ran".
2. Add a small helper that appends the `CUDA_OXIDE_LLVM_LINK` guidance to a
   `LibdeviceUnavailable` a test did not expect, used at the two `.expect(...)`
   sites and the two `other =>` arms. Empty for every other error, so unrelated
   failures gain no noise.

Test-only; no production code and no behaviour change. No GPU required.

## Acceptance criteria

- [ ] Renaming the re-exec'd test fails the suite instead of silently skipping the check
- [ ] A `LibdeviceUnavailable` from an ambient `CUDA_OXIDE_LLVM_LINK` names the variable
- [ ] All six tests still pass on a correctly configured machine

vyncint:
## Summary

Fixes #599. Two ways `crates/cuda-oxide-codegen/tests/libdevice_linking.rs` (new
in #596) misleads its reader. The libdevice feature itself is fine — this is
test-only, no production code, no behaviour change, no GPU.

## 1. The re-exec guard could verify nothing

`libdevice_linking_reports_unavailable_when_llvm_link_cannot_run` re-execs itself
in a child so a broken `CUDA_OXIDE_LLVM_LINK` cannot race its siblings. Good
isolation — but the parent asserted only on the child's exit status, and libtest
exits 0 when a filter matches nothing:

```
$ ./libdevice_linking-<hash> --exact this_test_does_not_exist --nocapture
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 6 filtered out
$ echo $?
0
```

The `--exact` literal and the function name were coupled only by convention.
Rename the test, or mistype the literal, and the child runs nothing, exits 0, the
parent asserts success, and the suite stays green — with the only check that a
broken `llvm-link` cannot silently emit PTX no longer running.

The child now prints a sentinel once the check has actually executed, and the
parent asserts on that too.

**Verified**: pointing the `--exact` argument at a name that does not exist now
fails with

```
the child exited 0 but never ran the check -- the `--exact` filter above no
longer matches this test's name
```

where before it passed.

## 2. A `CUDA_OXIDE_LLVM_LINK` failure did not name `CUDA_OXIDE_LLVM_LINK`

`Toolchain::discover()` reads that variable from the environment — the file's own
comment says so. Exporting it (the documented way to point at an `llvm-link` that
is not a sibling of `llc`) at a missing or major-mismatched binary fails four
tests:

```
$ CUDA_OXIDE_LLVM_LINK=/nonexistent/llvm-link cargo test -p cuda-oxide-codegen
libdevice_linking_resolves_a_rust_intrinsic_kernel            FAILED
libdevice_linking_resolves_a_frontend_declared_symbol         FAILED
libdevice_linking_rejects_a_symbol_libdevice_does_not_define   FAILED
libdevice_linking_still_rejects_a_device_extern               FAILED

libdevice linking resolves __nv_sqrtf: LibdeviceUnavailable { message:
  "no `llvm-link` sharing the selected `llc`'s LLVM major was found" }
```

Nothing pointed at the override. The file already handles the other external tool
well (`assert_ptxas_accepts` says "Point CUDA_TOOLKIT_PATH or CUDA_HOME at the
install root, or put `ptxas` on PATH"), so `llvm-link` now gets the same
treatment via `libdevice_unavailable_hint`, used at the two `expect` sites and the
two `other =>` arms. It returns nothing for any other error, so unrelated
failures gain no noise.

After:

```
libdevice linking resolves __nv_sqrtf: libdevice linking is unavailable: no
`llvm-link` sharing the selected `llc`'s LLVM major was found
This needs an `llvm-link` whose LLVM major matches the selected `llc`.
`Toolchain::discover()` honours CUDA_OXIDE_LLVM_LINK, so an exported override
pointing at a missing or mismatched binary produces exactly this error: unset it,
or point it at the `llvm-link` beside your `llc`.
```

To be explicit about severity: an override aimed at a broken tool **should** fail,
and still does. The fix is only that the failure now says which knob to turn. The
realistic case is a machine with several LLVMs, where the override is valid for
the user's own builds but its major does not match the pinned `llc`.

Side effect worth noting: the two `expect` sites become `unwrap_or_else`, which
renders the error through `Display` instead of `Debug` — hence
`libdevice linking is unavailable: ...` rather than
`LibdeviceUnavailable { message: ... }`.

## Validation

All CPU-only; no GPU and no driver.

- `cargo test -p cuda-oxide-codegen --test libdevice_linking`: 6 passed, 0 failed
- Drifted-filter experiment: fails before the fix would have passed (above)
- `CUDA_OXIDE_LLVM_LINK=/nonexistent/llvm-link`: the four failures now name the variable
- `cargo clippy -p cuda-oxide-codegen --all-targets --locked -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `cuda-intrinsics-gen check`: generated outputs current (untouched)
- 18-crate CPU matrix: **3017 passed, 0 failed**

## How I found it

Sweeping the suite against every `CUDA_OXIDE_*` / toolkit variable the repo reads.
`CUDA_OXIDE_LLVM_LINK` was the only one that broke anything — the rest
(`CUDA_OXIDE_LLC`, `_OPT`, `_LIBDEVICE`, `_DUMP_LLVM`, `_DUMP_MIR`,
`_SHOW_RUSTC_MIR`, `_DEVICE_CODEGEN_CRATE`, `_DEBUG_CONST`, `CUDA_PATH`) are
clean, which is what led me into this file and then to the re-exec guard.

### DIFF
--- a/crates/cuda-oxide-codegen/tests/libdevice_linking.rs
+++ b/crates/cuda-oxide-codegen/tests/libdevice_linking.rs
@@ -227,18 +227,20 @@ fn assert_ptxas_accepts(ptx: &[u8], label: &str) {
 /// Extra guidance for a `LibdeviceUnavailable` a caller did not expect.
 ///
 /// `Toolchain::discover()` honours `CUDA_OXIDE_LLVM_LINK` straight from the
-/// environment, so an exported override pointing at a missing binary -- or at
-/// one whose LLVM major differs from the selected `llc`'s -- turns every
-/// libdevice test in this file into `LibdeviceUnavailable`. The error names
-/// neither the variable nor the override, so the tests do, the same way
-/// `assert_ptxas_accepts` points at `CUDA_TOOLKIT_PATH`. Empty for every other
-/// error, where the variable is irrelevant.
+/// environment, so an exported override pointing at a missing or unrunnable
+/// binary turns every libdevice test in this file into
+/// `LibdeviceUnavailable`. (A runnable override is trusted as-is; an LLVM
+/// major mismatch surfaces later as a link error, not as this variant.) The
+/// error names neither the variable nor the override, so the tests do, the
+/// same way `assert_ptxas_accepts` points at `CUDA_TOOLKIT_PATH`. Empty for
+/// every other error, where the variable is irrelevant.
 fn libdevice_unavailable_hint(error: &CompileError) -> &'static str {
     match error {
         CompileError::LibdeviceUnavailable { .. } => {
-            "\nThis needs an `llvm-link` whose LLVM major matches the selected \
-             `llc`. `Toolchain::discover()` honours CUDA_OXIDE_LLVM_LINK, so an \
-             exported override pointing at a missing or mismatched binary \
+            "\nThis needs a runnable `llvm-link` (auto-discovery only accepts \
+             one sharing the selected `llc`'s LLVM major). \
+             `Toolchain::discover()` honours CUDA_OXIDE_LLVM_LINK, so an \
+             exported override pointing at a missing or unrunnable binary \
              produces exactly this error: unset it, or point it at the \
              `llvm-link` beside your `llc`."
         }
--- a/crates/cuda-oxide-codegen/tests/libdevice_linking.rs
+++ b/crates/cuda-oxide-codegen/tests/libdevice_linking.rs
@@ -224,6 +224,28 @@ fn assert_ptxas_accepts(ptx: &[u8], label: &str) {
     );
 }
 
+/// Extra guidance for a `LibdeviceUnavailable` a caller did not expect.
+///
+/// `Toolchain::discover()` honours `CUDA_OXIDE_LLVM_LINK` straight from the
+/// environment, so an exported override pointing at a missing binary -- or at
+/// one whose LLVM major differs from the selected `llc`'s -- turns every
+/// libdevice test in this file into `LibdeviceUnavailable`. The error names
+/// neither the variable nor the override, so the tests do, the same way
+/// `assert_ptxas_accepts` points at `CUDA_TOOLKIT_PATH`. Empty for every other
+/// error, where the variable is irrelevant.
+fn libdevice_unavailable_hint(error: &CompileError) -> &'static str {
+    match error {
+        CompileError::LibdeviceUnavailable { .. } => {
+            "\nThis needs an `llvm-link` whose LLVM major matches the selected \
+             `llc`. `Toolchain::discover()` honours CUDA_OXIDE_LLVM_LINK, so an \
+             exported override pointing at a missing or mismatched binary \
+             produces exactly this error: unset it, or point it at the \
+             `llvm-link` beside your `llc`."
+        }
+        _ => "",
+    }
+}
+
 #[test]
 fn self_contained_linking_still_rejects_a_libdevice_kernel() {
     let mut module = CodegenModule::new("sqrt_module").unwrap();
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM, with one wording fix I pushed

The re-exec guard could pass vacuously: libtest exits 0 when a `--exact` filter matches nothing, so renaming the child test would silently stop the only check proving a broken llvm-link can't produce PTX. The sentinel closes that hole for good.

Verified locally:

- All 6 tests pass; sabotaging the `--exact` name now fails loudly with the new message.
- With `CUDA_OXIDE_LLVM_LINK=/nonexistent/llvm-link`, all four failures print the hint naming the variable.
- The pushed commit corrects one claim in the hint: a runnable-but-major-mismatched override is trusted at discovery and fails later as a link error, not as `LibdeviceUnavailable`. Only the missing/unrunnable case (which you verified) produces this error.

Merging;

closes #599.

nihalpasham (approved):

PR #602, opened 2026-07-31
test(smoketest): pin the diagnostic for three error fixtures

### LINKED ISSUE #601: smoketest accepts any compile error for three error fixtures that pin an exact diagnostic
## Summary

`verdict_error` in `scripts/smoketest.sh` pins an exact diagnostic for six of the
ten `error*` fixtures. The other four accept **any** compile error, so three
fixtures that pin an exact message in their own source or README are satisfied by
a failure that has nothing to do with the contract they protect.

## The gap

`verdict_error` special-cases six examples before falling through to a generic
"did it fail to compile" check:

| fixture | pinned diagnostic asserted? |
| :-- | :-- |
| `error_enum_pointer_overlap` | yes |
| `error_enum_shared_pointer_layout` | yes |
| `error_generated_intrinsic_abi` | yes |
| `error_generated_intrinsic_unknown_id` | yes |
| `error_generated_intrinsic_fn_pointer` | yes |
| `error_generated_intrinsic_callable` | yes |
| `error_heap_alloc` | **no** |
| `error_missing_device_attr` | **no** |
| `error_set_discriminant_uninhabited` | **no** |
| `error` | **no** |

The script already states the principle behind the six:

```sh
# The generated-intrinsic fixtures protect fail-closed compiler contracts,
# so merely observing an unrelated compile error is not enough.
```

That reasoning applies just as much to three of the four that were left out,
because those three pin an exact message in their own documentation:

- `error_heap_alloc/src/main.rs` — "Expected: the build FAILS with this exact
  diagnostic (pinned): `heap allocation is not supported in kernels (no device
  allocator); use fixed-size arrays or SharedArray`"
- `error_missing_device_attr/src/main.rs` — same wording, for
  "``thread::index_1d`` only works inside `#[kernel]` / `#[device]` functions; …"
- `error_set_discriminant_uninhabited/README.md` — "The build must fail with:
  `SetDiscriminant cannot select uninhabited variant 1`"

The harness pins none of them.

## Demonstration

Feeding `verdict_error` a log whose only error is unrelated — an unresolved
import — passes all three:

```
$ cat unrelated.log
error[E0432]: unresolved import `foo::bar`
error: could not compile `error_heap_alloc` (bin "error_heap_alloc") due to 1 previous error

error_heap_alloc                       PASS (expected compile failure, exit=101)
error_missing_device_attr              PASS (expected compile failure, exit=101)
error_set_discriminant_uninhabited     PASS (expected compile failure, exit=101)
```

The six pinned fixtures reject that same log, which is the asymmetry:

```
error_enum_pointer_overlap             FAIL (missing overlapping enum pointer-provenance diagnostic)
error_generated_intrinsic_abi          FAIL (missing intrinsic ABI-v2 diagnostic)
error_generated_intrinsic_callable     FAIL (missing direct-call-only intrinsic diagnostic)
```

So if one of those three examples stopped compiling for an unrelated reason — a
dependency bump, a renamed API, a typo in the fixture — the smoketest would keep
reporting PASS while the diagnostic it exists to protect went unexercised. The
`error_missing_device_attr` and `error_heap_alloc` fixtures exist precisely
because the *previous* behaviour was an opaque unrelated error; those are the
failures most likely to creep back, and they are the ones the check cannot
currently tell apart.

## Proposed fix

Add the same kind of `grep -Fq` assertion for the three that pin a message, using
substrings of the real format strings
(`crates/rustc-codegen-cuda/src/collector.rs` and
`crates/mir-importer/src/translator/statement.rs`).

`error` should stay on the generic path: unlike the others it pins no single
message — it carries two kernels to show a valid one still compiles while
`core::fmt` machinery is refused — so there is no exact string to assert without
inventing one. Worth stating in a comment so the omission reads as deliberate.

## Acceptance criteria

- [ ] An unrelated compile error no longer satisfies the three pinning fixtures
- [ ] Each fixture's genuine diagnostic still passes
- [ ] `error` keeps its current generic behaviour, documented as intentional
- [ ] Shell-only; no GPU and no toolchain build required

vyncint:
## Summary

Fixes #601. `verdict_error` asserted an exact diagnostic for six of the ten
`error*` fixtures and let the other four fall through to a generic "did it fail to
compile" check. Three of those four pin an exact message in their own source or
README, so **any** unrelated compile error satisfied them.

## Demonstration

Feeding `verdict_error` a log whose only error is an unresolved import:

```
$ cat unrelated.log
error[E0432]: unresolved import `foo::bar`
error: could not compile `error_heap_alloc` (bin "error_heap_alloc") due to 1 previous error

error_heap_alloc                       PASS (expected compile failure, exit=101)
error_missing_device_attr              PASS (expected compile failure, exit=101)
error_set_discriminant_uninhabited     PASS (expected compile failure, exit=101)
```

The six pinned fixtures reject that same log — that asymmetry is the bug:

```
error_enum_pointer_overlap             FAIL (missing overlapping enum pointer-provenance diagnostic)
error_generated_intrinsic_abi          FAIL (missing intrinsic ABI-v2 diagnostic)
error_generated_intrinsic_callable     FAIL (missing direct-call-only intrinsic diagnostic)
```

So an example that stopped compiling for an unrelated reason — a dependency bump,
a renamed API, a typo in the fixture — kept reporting PASS while the diagnostic it
exists to protect went unexercised.

That matters most for these three specifically: `error_missing_device_attr` and
`error_heap_alloc` exist *because* the previous behaviour was an opaque unrelated
error ("the user got an inscrutable const-translation error spanned into
`alloc/src/boxed.rs`"). A regression to that state is exactly what the check could
not tell apart from success.

## Change

The script already stated the principle for the six:

```sh
# The generated-intrinsic fixtures protect fail-closed compiler contracts,
# so merely observing an unrelated compile error is not enough.
```

This extends it to the three whose own docs pin a message:

| fixture | pins, per its own source/README | asserted substring |
| :-- | :-- | :-- |
| `error_heap_alloc` | "the build FAILS with this exact diagnostic (pinned)" | `heap allocation is not supported in kernels` |
| `error_missing_device_attr` | same wording | ``only works inside `#[kernel]` / `#[device]` `` |
| `error_set_discriminant_uninhabited` | "The build must fail with" | `cannot select uninhabited variant` |

Substrings come from the real format strings in
`crates/rustc-codegen-cuda/src/collector.rs` and
`crates/mir-importer/src/translator/statement.rs` — not from the doc comments,
which are line-wrapped. Both `only works inside` call sites (collector.rs:2183
and :2372) wrap the literal at different points but render the same single line,
so one substring covers both.

`error` deliberately stays on the generic path: unlike the others it pins no
single message — it carries two kernels to show a valid one still compiles while
`core::fmt` machinery is refused — so there is no exact string to assert without
inventing one. A comment records that, so the omission reads as deliberate rather
than forgotten.

## Validation

Shell-only — no Rust code, no GPU, no toolchain build. I exercised
`verdict_error` directly against synthetic logs.

| log | before | after |
| :-- | :-- | :-- |
| unrelated `error[E0432]`, all three fixtures | **PASS** (wrong) | FAIL, each naming its missing diagnostic |
| genuine `heap allocation is not supported in kernels …` | PASS | PASS |
| genuine ``\`thread::index_1d\` only works inside …`` | PASS | PASS |
| genuine `SetDiscriminant cannot select uninhabited variant 1` | PASS | PASS |
| unrelated error, `error` fixture | PASS | PASS (unchanged by design) |

Also: `bash -n scripts/smoketest.sh` clean, `scripts/check-error-example-status.sh`
still reports OK, and `smoketest.sh --help` unchanged.

I could not run the full smoketest end to end — it builds the backend and every
example — so the guarantee here is over `verdict_error`'s logic and the substrings
being real format-string text, verified against the compiler sources rather than
the wrapped doc comments.

## Relationship to #600

Independent and disjoint: #600 touches
`crates/cuda-oxide-codegen/tests/libdevice_linking.rs`, this touches
`scripts/smoketest.sh`. Same theme, though — a check that reports green without
verifying what it claims.

### DIFF
--- a/scripts/smoketest.sh
+++ b/scripts/smoketest.sh
@@ -394,6 +394,30 @@ verdict_error() {
                 return 1
             fi
             ;;
+        # These three pin an exact diagnostic in their own source or README
+        # ("Expected: the build FAILS with this exact diagnostic (pinned)"),
+        # so an unrelated compile error must not satisfy them either.
+        error_heap_alloc)
+            if ! grep -Fq 'heap allocation is not supported in kernels' "${log}"; then
+                echo "FAIL (missing heap-allocation diagnostic)"
+                return 1
+            fi
+            ;;
+        error_missing_device_attr)
+            if ! grep -Fq 'only works inside `#[kernel]` / `#[device]`' "${log}"; then
+                echo "FAIL (missing host-only-stub diagnostic)"
+                return 1
+            fi
+            ;;
+        error_set_discriminant_uninhabited)
+            if ! grep -Fq 'cannot select uninhabited variant' "${log}"; then
+                echo "FAIL (missing uninhabited-variant diagnostic)"
+                return 1
+            fi
+            ;;
+        # `error` stays on the generic check below: unlike the fixtures above it
+        # pins no single message, carrying two kernels to show that a valid one
+        # still compiles while `core::fmt` machinery is refused.
     esac
 
     if grep -qE 'Device codegen failed|Translation failed|Compilation error|Unsupported construct' "${log}"; then


nihalpasham:
LGTM.

Verified locally:

- `error_heap_alloc`, `error_missing_device_attr`, `error_set_discriminant_uninhabited` all 3/3 PASS with their diagnostics present in the captured logs

Merging;

closes #601.

nihalpasham (approved):

PR #608, opened 2026-08-01
fix(mir): lower dead uninhabited-enum paths to typed undef

### LINKED ISSUE #607: Reachable dead MIR for uninhabited enum variants fails device import
# Reachable dead MIR for uninhabited enum variants fails device import

## Problem

Rust can retain reachable control-flow blocks whose values are nevertheless
impossible to construct. In particular, infallible fixed-array helpers retain
residual branches involving uninhabited enums such as
`NeverShortCircuitResidual`.

The device importer currently translates those dead branches into ordinary
enum construction, discriminant, and payload operations. The MIR dialect
correctly rejects those operations because no value can inhabit the referenced
enum or variant, so otherwise valid uses of `core::array::from_fn` and
`array::map` fail during device compilation.

## Why this matters

`array::from_fn` and `array::map` are ordinary, safe Rust APIs for fixed-size
arrays. Rejecting them because of impossible residual paths in rustc-generated
MIR prevents common array code from compiling on the device even though the
paths cannot execute.

## Expected behavior

The importer should keep dynamically dead uninhabited-enum paths representable
without weakening verification of real enum operations. Infallible fixed-array
helpers should compile and execute normally, while attempts to construct or
inspect an actually uninhabited value should remain invalid.

## Acceptance criteria

- Dead construction, discriminant reads, and payload reads involving
  uninhabited enum variants are represented with correctly typed undefined
  values.
- MIR dialect verification continues to reject real operations on
  uninhabited enums and variants.
- Undefined values remain exportable when an earlier-printed PHI references
  one from a later-printed block.
- An end-to-end device regression verifies the computed results of
  `array::from_fn` followed by `array::map`.

niklebedenko:
Closes #607.

## Summary

Rustc retains impossible residual-enum arms in the MIR for infallible
fixed-array helpers. The importer was emitting real enum operations for those
dead paths, which the MIR dialect correctly rejected. This change represents
dead construction, discriminant, and payload paths with typed `mir.undef`
values, preserving strict verification while allowing `array::from_fn` and
`array::map` to compile and execute on the device.

Before:

```text
MirEnumPayloadOp cannot extract a field from uninhabited variant 1
```

After: the focused GPU regression computes and validates the mapped array for
all 32 launched threads.

### Preserve dead uninhabited-enum paths without weakening verification

- Inspect the authoritative `MirEnumType::variant_inhabited` layout facts at
  enum construction, discriminant-read, and payload-read lowering sites.
- Emit a `MirUndefOp` with the operation's declared result type when rustc's
  MIR names an enum or variant that cannot be inhabited.
- Leave `MirConstructEnumOp`, `MirGetDiscriminantOp`, and
  `MirEnumPayloadOp` verification unchanged. Unit coverage confirms those
  operations still reject uninhabited inputs.

### Make undef values available to later-block PHIs

- Register an undef result's literal LLVM spelling during the exporter's
  whole-function value-name pre-pass.
- Cover a join block printed before both predecessors, with a PHI incoming
  value defined by `undef` in a later-printed block.

### Focused end-to-end coverage

- Extend `array_for_loop` with a `[u32; 4]` kernel that composes
  `core::array::from_fn` and `array::map` and checks the exact host result.
- The regression uses ordinary scalar arrays and is independent of
  over-aligned union storage and bare enum-array constants.
- This extracts the enabling correctness fix previously carried inside draft
  PR #398. That PR is on hold over its always-on array-optimization pipeline;
  this branch contains none of that optimization, unrolling, or SSA
  scalarization work.

## Known separate limitations

- This only substitutes undef when imported `MirEnumType` layout facts prove
  an enum or variant uninhabited. Unsupported or malformed inhabited enum
  operations continue to fail verification.
- The upstream-main smoke reaches the payload verifier failure first, so that
  one run does not independently demonstrate red failures for construction and
  discriminant lowering. Focused tests cover all three paths, and the patched
  generated IR exercises undefined discriminants flowing through PHIs.
- Empty-enum layout reconstruction, over-aligned union storage, bare
  enum-array constants, and multi-artifact loading are deliberately outside
  this patch.
- The regression exercises current rustc MIR for `array::from_fn` and
  `array::map`; it does not attempt to generalize unrelated dead-code
  elimination.

## Testing

  `scripts/smoketest.sh --only '^array_for_loop$' --no-color --keep-logs`
  failed with cargo exit 101 because `MirEnumPayloadOp` attempted to extract
  uninhabited variant 1 inside `core::array::try_from_fn`.
- Patched green: the same command passed 1/1 on an NVIDIA GeForce RTX 3080
  (sm_86), including exact results for all 32 threads.
- `cargo test -p dialect-mir test_uninhabited_enum_construct_and_discriminant_fail_verification`
  passed.
- `cargo test -p llvm-export phi_can_reference_undef_from_a_later_block`
  passed.
- `cargo test -p mir-importer --lib` passed all 891 tests.
- `cargo clippy -p dialect-mir -p llvm-export -p mir-importer --all-targets -- -D warnings`
  passed.
- Root and standalone-example formatting checks and `git diff --check`
  passed.

## Checklist

### DIFF
--- a/crates/dialect-mir/tests/ops_test.rs
+++ b/crates/dialect-mir/tests/ops_test.rs
@@ -1333,3 +1333,90 @@ fn test_mir_enum_payload_rejects_uninhabited_variant() {
     payload.set_attr_payload_field_index(&ctx, FieldIndexAttr(0));
     assert!(payload.verify(&ctx).is_err());
 }
+
+/// Reachable-but-dead MIR (e.g. the residual arms of `array::try_from_fn`)
+/// can name uninhabited variants. The importer must lower such reads and
+/// constructions to typed undefs; if it ever emits the real ops again, these
+/// verifiers are the loud stop.
+#[test]
+fn test_uninhabited_enum_construct_and_discriminant_fail_verification() {
+    use dialect_mir::types::{EnumCarrierKind, EnumEncoding, EnumLayoutKind};
+
+    let mut ctx = Context::new();
+    dialect_mir::register(&mut ctx);
+    let i8_ty = IntegerType::get(&ctx, 8, Signedness::Unsigned);
+    let i32_ty = IntegerType::get(&ctx, 32, Signedness::Unsigned);
+
+    // Constructing an uninhabited variant must fail verification.
+    let partial = MirEnumType::get_with_encoding(
+        &mut ctx,
+        "HasImpossibleVariant".into(),
+        i8_ty.into(),
+        vec![0, 1],
+        vec![
+            EnumVariant::unit("Live".into()),
+            EnumVariant::new_with_layout(
+                "Impossible".into(),
+                vec![i32_ty.into()],
+                vec![0],
+                vec![4],
+            ),
+        ],
+        EnumEncoding {
+            tag_offset: 0,
+            total_size: 8,
+            abi_align: 4,
+            layout_kind: EnumLayoutKind::Direct,
+            carrier_kind: EnumCarrierKind::Integer,
+            carrier_width: 8,
+            variant_inhabited: vec![1, 0],
+            ..EnumEncoding::default()
+        },
+    );
+    let block = BasicBlock::new(&mut ctx, None, vec![i32_ty.into()]);
+    let field = block.deref(&ctx).get_argument(0);
+    let construct = Operation::new(
+        &mut ctx,
+        MirConstructEnumOp::get_concrete_op_info(),
+        vec![partial.into()],
+        vec![field],
+        vec![],
+        0,
+    );
+    let construct = MirConstructEnumOp::new(construct);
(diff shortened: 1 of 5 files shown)


nihalpasham:
LGTM.

Infallible helpers like `core::array::from_fn` + `array::map` failed device compilation because rustc keeps an error-handling branch that names an enum variant which can never exist, and we translated that dead branch into real enum ops the verifier rightly rejects. Substituting a typed undef on provably-dead paths is what rustc's own backend does at the same layer, and live enum ops stay strictly verified.

**Verified on hardware:**

- `array_for_loop` passes on GPU including the new `from_fn`/`map` kernel.
- 892 mir-importer tests, plus dialect-mir and llvm-export suites (the PHI-ordering fix included); fmt and clippy clean.

Merging;

closes #607.

nihalpasham (approved):

PR #610, opened 2026-08-01
fix(mir-lower): preserve over-aligned union size and stride

### LINKED ISSUE #609: Over-aligned Rust unions are rejected even when size and stride are representable
# Over-aligned Rust unions are rejected even when size and stride are representable

## Problem

MIR lowering rejects every Rust union whose ABI alignment exceeds 16 bytes,
even when its storage size is known, is a multiple of that alignment, and can
be represented exactly by an LLVM aggregate.

For example, a `#[repr(C, align(32))]` union containing a `u32` has a
well-defined 32-byte size and 32-byte array stride, but device compilation
fails before code generation with an alignment-limit diagnostic.

## Why this matters

Rust uses over-aligned unions directly and inside standard-library storage
types. Rejecting an otherwise byte-representable union prevents valid aggregate
code from compiling and makes arrays of those values unusable even though
cuda-oxide already carries rustc's explicit alignment facts on memory
operations.

## Expected behavior

MIR lowering should preserve the exact Rust union size and resulting array
stride. The LLVM storage carrier may use the strongest natural alignment
NVPTX can represent, while accesses retain the authoritative Rust ABI
alignment through their explicit memory-operation alignment.

## Acceptance criteria

- A 32-byte-sized, 32-byte-aligned union lowers to exactly 32 bytes of LLVM
  storage.
- A three-element array of that union lowers to exactly 96 bytes, preserving
  the Rust element stride.
- Full MIR-to-LLVM lowering preserves 32- and 64-byte union size/stride in the
  LLVM carrier and records the corresponding explicit alignment on allocations
  and stores.
- Existing size, field-layout, and pointer-provenance validation remains
  fail-closed.
- A direct `#[repr(C, align(32))]` union-array GPU regression compiles,
  executes, and validates its result.

niklebedenko:
Closes #609.

## Summary

Over-aligned Rust unions were rejected solely because LLVM/NVPTX aggregate
types have no natural alignment above 16 bytes. The actual correctness rule is
that the storage carrier must preserve Rust's byte size and array stride while
memory operations carry the stronger ABI alignment explicitly. This change
uses a 16-byte natural-alignment anchor for over-aligned carriers, preserves
the exact Rust size, and admits representable `#[repr(C, align(32))]` unions.

Before:

```text
union `OverAlignedWord` requires 32-byte alignment; cuda-oxide currently supports union alignments up to 16 bytes
```

After: hardware-free MIR-to-LLVM lowering tests preserve the carrier size,
array stride, and explicit memory-operation alignment for both 32- and
64-byte-aligned unions. The direct GPU regression also compiles and returns the
expected values from all three 32-byte-stride elements.

### Preserve authoritative Rust size and stride

- Continue using `MirUnionType::total_size` and `abi_align`, which come from
  rustc layout, instead of deriving layout from the declared fields.
- Cap only the LLVM carrier's natural-alignment anchor at 16 bytes, the
  strongest scalar alignment available for this NVPTX representation.
- Require the lowered carrier size to equal rustc's union size exactly and
  reject a carrier whose natural alignment exceeds the Rust ABI alignment.
- Keep the existing requirement that a non-empty union's size be a multiple
  of its Rust ABI alignment. Arrays therefore inherit the exact Rust stride.

### Focused coverage and related paths

- Replace the old blanket-rejection unit test with exact layout assertions:
  `Some((32, 16))` for the union carrier and `Some((96, 16))` for a
  three-element array.
- Add a full MIR-to-LLVM code-shape regression for 32- and 64-byte-aligned
  unions. It asserts three-element LLVM carrier sizes of 96 and 192 bytes,
  32- and 64-byte element strides, and matching explicit alignment on the
  resulting `llvm.alloca` and `llvm.store`.
- Extend `union_aggregate` with a direct
  `#[repr(C, align(32))] union OverAlignedWord` array and validate distinct
  values from all three elements at runtime.
- Existing pointer-carrier selection, incompatible-address-space rejection,
  non-byte-faithful carrier rejection, and ordinary union layout tests are
  unchanged and pass.
- Draft PR #398 carried an earlier form of this enabling fix alongside an
  on-hold, always-on array optimization. Its maintainer review explicitly
  identified union over-alignment as independently valuable; this branch
  extracts only that layout correction.

## Known separate limitations

- LLVM storage still has at most 16-byte natural alignment. Correct
  over-alignment relies on the existing explicit rustc-derived alignment on
  allocations and memory accesses; this patch does not add a new LLVM type
  alignment mechanism.
- Union shapes that lose pointer provenance, mix incompatible pointer
  representations, have invalid size/alignment relationships, or otherwise
  lack a byte-faithful carrier remain rejected.
- Empty-enum lowering, `array::map`, uninhabited-enum paths, enum-array
  constants, and multi-artifact loading are deliberately outside this patch.

## Testing

  `cargo test -p mir-lower
  convert_ref_preserves_over_aligned_union_array_layout_and_alignment`
  failed during lowering with the 32-byte union alignment-limit diagnostic.
- Upstream GPU red at the same revision:
  `scripts/smoketest.sh --only '^union_aggregate$' --no-color --keep-logs`
  failed with cargo exit 101 and the 32-byte union alignment-limit diagnostic.
- Patched hardware-free green: the focused lowering/code-shape test passed for
  both 32- and 64-byte alignment.
- Patched GPU green: the same smoke command passed 1/1 on an NVIDIA GeForce RTX 3080
  (sm_86), including the three-element over-aligned union result.
- `cargo test -p mir-lower union_storage_preserves_over_aligned_size_and_stride`
  passed with exact `Some((32, 16))` and `Some((96, 16))` assertions.
- `cargo test -p mir-lower --all-targets` passed all 222 tests.
- `cargo clippy -p mir-lower --all-targets -- -D warnings` passed.
- Standalone-example clippy, root and standalone-example formatting checks,
  and `git diff --check` passed.

## Checklist

### DIFF
--- a/crates/mir-lower/src/convert/ops/memory.rs
+++ b/crates/mir-lower/src/convert/ops/memory.rs
@@ -1012,7 +1012,7 @@ mod tests {
     use super::*;
     use crate::convert::ops::test_util::*;
     use dialect_mir::ops as mir;
-    use dialect_mir::types::{MirArrayType, MirPtrType, MirStructType, MirTupleType};
+    use dialect_mir::types::{MirArrayType, MirPtrType, MirStructType, MirTupleType, MirUnionType};
     use llvm_export::op_interfaces::PointerTypeResult;
     use llvm_export::ops as llvm;
     use llvm_export::types::{PointerType, address_space as llvm_addr};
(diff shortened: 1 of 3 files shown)


nihalpasham:
LGTM. 

A `#[repr(align(32))]` union was rejected outright even though its size and stride are perfectly representable. Capping the alignment anchor at 16 (the strongest NVPTX scalar alignment) and letting rustc's real alignment flow through the existing explicit-alignment stamping is the same mechanism structs and enums already use — the carrier under-promises, the explicit attribute tells the truth.

**Verified on hardware:**

- `union_aggregate` passes on GPU; case 21 returns 0x0033_2211.
- The generated `.ll` shows the 32-byte stride (`[3 x {...}] = 96 bytes`) with explicit align 32 on alloca/store/load, and llc accepts it.
- 237 mir-lower tests pass.

**One follow-up worth its own issue (pre-existing on main, not introduced here):** 
- putting an over-aligned type inside a `SharedArray` creates an alignment promise the storage doesn't keep. 
- The shared global's alignment comes only from the `ALIGN` const generic (default 16), while every element load/store is stamped with the type's real alignment, e.g. align 32. 
- So the array may start at shared offset 16, but we tell LLVM each access is 32-byte aligned, and it's allowed to act on that (say, with a wide vector access that faults if the promise is false). 
- The fix is for shared/global allocation sites to take max(`ALIGN`, the element's rustc alignment) so the promise becomes true by construction. 
- Structs and enums already had this gap; this PR just makes unions eligible to hit it too.

Merging;

closes #609.

nihalpasham (approved):

PR #612, opened 2026-08-01
fix(mir): preserve wrapping pointer-offset semantics

### LINKED ISSUE #611: Wrapping pointer offsets cannot be represented through MIR lowering
# Wrapping pointer offsets cannot be represented through MIR lowering

## Problem

Safe raw-pointer methods such as `wrapping_add` and `wrapping_offset` lower
through rustc's `arith_offset` intrinsic. Device MIR import does not recognize
that intrinsic, so otherwise valid kernels fail compilation.

There is a second correctness boundary behind the missing import: every
`mir.ptr_offset` currently exports as LLVM `getelementptr inbounds`. Rust's
wrapping pointer methods deliberately permit an intermediate pointer to leave
the allocation or wrap the address space, so attaching LLVM's `inbounds`
promise would turn valid wrapping arithmetic into poison.

## Why this matters

Wrapping pointer arithmetic is a safe Rust operation and is used by standard
raw-pointer APIs. Rejecting it prevents ordinary generic pointer code from
compiling for CUDA. Lowering it as an in-bounds GEP would be worse: code could
compile but be misoptimized even when the wrapped pointer is only compared or
converted to an address and is never dereferenced.

## Expected behavior

Device MIR should represent whether a pointer offset carries Rust's in-bounds
contract. Ordinary pointer offsets should retain their existing in-bounds
lowering, while `arith_offset` should flow through MIR and LLVM export as a
non-`inbounds` GEP with the original pointee type and address space.

## Acceptance criteria

- `core::intrinsics::arith_offset` and its `std` spelling import as wrapping
  pointer offsets.
- Wrapping offsets export without LLVM's `inbounds` marker, while existing
  ordinary offsets continue to export with it.
- GPU regression coverage validates positive and negative offsets, aggregate
  pointee scaling, mutable pointers, zero-sized pointees, and an `isize::MAX`
  address wrap without dereferencing an out-of-allocation pointer.
- Pointer-distance intrinsics and unrelated small-array scalarization behavior
  remain unchanged.

niklebedenko:
Closes #611.

# fix(mir): preserve wrapping pointer-offset semantics

## Summary

Safe raw-pointer wrapping methods previously failed device compilation because
rustc's `arith_offset` intrinsic was unsupported. This change carries the
source pointer-arithmetic contract through MIR and LLVM lowering: ordinary
offsets remain `getelementptr inbounds`, while wrapping offsets become plain
`getelementptr`, preserving Rust's permitted address wrapping without adding
LLVM poison semantics.

```text
Before: wrapping_offset -> unsupported arith_offset
After:  wrapping_offset -> mir.ptr_offset(non-inbounds) -> getelementptr
```

### Pointer-arithmetic contract

- Adds an `inbounds` semantic marker to `MirPtrOffsetOp`. Absence defaults to
  `true`, preserving every existing producer's behavior.
- Imports both rustc spellings of `arith_offset` through one focused importer
  path and stamps those operations as non-inbounds.
- Propagates the marker to the LLVM GEP operation and makes textual export emit
  `inbounds` conditionally.
- Continues to obtain element scaling from the MIR pointer's authoritative
  pointee type rather than reconstructing it from the call name or result.

### Coverage and related paths

- Dialect, lowering, and exporter tests assert both sides of the contract:
  default ordinary offsets are in-bounds and explicit wrapping offsets are not.
- Extends the existing `ptr_offset_from` example so one GPU run checks positive
  and negative wrapping offsets, aggregate scaling, const and mutable pointers,
  zero-sized pointees, and `isize::MAX` address wrapping.
- The extreme and out-of-allocation cases are compared as addresses only. They
  are never dereferenced.
- Merged PR #206 added pointer-distance intrinsics; those paths remain
  unchanged. Open PR #398's `DeferredGep` and small-array scalarization changes
  are deliberately absent from this branch.

## Known separate limitations

- A wrapped pointer may be dereferenced only when the resulting pointer is
  valid for that access under Rust's normal rules.
- This does not change exposed-provenance reconstruction or pointer/integer
  cast semantics.
- The compatibility default on operations without the marker is in-bounds.
  New wrapping producers must set the marker explicitly.
- This change adds no array scalarization, speculative memory access, or
  optimization-pipeline policy.

## Testing

- Upstream red at `bead880c` with only the regression present:
  `scripts/smoketest.sh --only '^ptr_offset_from$' --no-color` failed
  (`exit=101`) because `std::intrinsics::arith_offset` was unsupported.
- Prepared branch:
  `scripts/smoketest.sh --only '^ptr_offset_from$' --no-color` passed on an
  RTX 3080 (`sm_86`), 1/1 example.
- Focused regressions passed:
  `test_mir_ptr_offset_verify`,
  `gep_inbounds_marker_controls_exported_pointer_semantics`, and
  `convert_wrapping_ptr_offset_lowers_to_non_inbounds_gep`.
- `cargo test -p dialect-mir -p llvm-export -p mir-importer -p mir-lower
  --all-targets` passed, including 891 importer tests, 146 mir-lower unit
  tests, and 76 mir-lower integration tests.
- `cargo clippy -p dialect-mir -p llvm-export -p mir-importer -p mir-lower
  --all-targets -- -D warnings` passed.
- The standalone example passes `cargo fmt --check` and
  `cargo clippy --all-targets -- -D warnings`; repository formatting and
  `git diff --check` pass.

## Checklist

### DIFF
--- a/crates/dialect-mir/src/ops/memory.rs
+++ b/crates/dialect-mir/src/ops/memory.rs
@@ -9,13 +9,14 @@
 
 use pliron::{
     builtin::{
-        attributes::IntegerAttr,
+        attributes::{BoolAttr, IntegerAttr},
         op_interfaces::{NOpdsInterface, NResultsInterface, OneOpdInterface, OneResultInterface},
         types::IntegerType,
     },
     common_traits::Verify,
     context::{Context, Ptr},
     derive::op_interface_impl,
+    identifier::Identifier,
     irbuild::{inserter::Inserter, rewriter::Rewriter},
     location::Located,
     op::Op,
@@ -732,10 +733,38 @@ impl Verify for MirRefOp {
 )]
 pub struct MirPtrOffsetOp;
 
+const PTR_OFFSET_INBOUNDS_KEY: &str = "mir_ptr_offset_inbounds";
+
 impl MirPtrOffsetOp {
     /// Create a new MirPtrOffsetOp wrapper.
     pub fn new(op: Ptr<Operation>) -> Self {
         MirPtrOffsetOp { op }
+    }
+
+    /// Set whether this offset carries Rust's in-bounds pointer-arithmetic
+    /// promise. Wrapping offsets must set this to false.
+    pub fn set_inbounds(&self, ctx: &mut Context, inbounds: bool) {
+        let key = Identifier::try_new(PTR_OFFSET_INBOUNDS_KEY.to_string())
+            .expect("valid ptr-offset attribute key");
+        self.get_operation()
+            .deref_mut(ctx)
+            .attributes
+            .set(key, BoolAttr::new(inbounds));
+    }
+
+    /// Return whether this offset may lower to LLVM `getelementptr inbounds`.
+    ///
+    /// Pointer offsets without an explicit attribute carry Rust's ordinary
+    /// in-bounds contract, so absence defaults to true.
+    pub fn is_inbounds(&self, ctx: &Context) -> bool {
+        let key = Identifier::try_new(PTR_OFFSET_INBOUNDS_KEY.to_string())
+            .expect("valid ptr-offset attribute key");
+        self.get_operation()
+            .deref(ctx)
+            .attributes
+            .get::<BoolAttr>(&key)
+            .map(|attr| bool::from(attr.clone()))
+            .unwrap_or(true)
     }
 }
 
--- a/crates/dialect-mir/tests/ops_test.rs
+++ b/crates/dialect-mir/tests/ops_test.rs
@@ -305,6 +305,15 @@ fn test_mir_ptr_offset_verify() {
     );
     let offset_op = MirPtrOffsetOp::new(op);
     assert!(offset_op.verify(&ctx).is_ok(), "Valid MirPtrOffsetOp");
+    assert!(
+        offset_op.is_inbounds(&ctx),
+        "ordinary pointer offsets default to inbounds"
+    );
+    offset_op.set_inbounds(&mut ctx, false);
+    assert!(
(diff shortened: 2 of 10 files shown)


nihalpasham:
LGTM as-is.

**Gist:**
- Two kinds of pointer arithmetic. 
  - `ptr.add(n)` promises the result stays inside the same allocation; 
  - `ptr.wrapping_add(n)` makes no such promise, and it's safe code precisely because you're allowed to compute a wild out-of-bounds address as long as you never dereference it. 
- The second kind didn't compile on device at all: rustc lowers it through an `arith_offset` intrinsic our importer didn't recognize.
- The simplest one-line fix would have been a miscompile. Both forms become LLVM's `getelementptr` (pointer-plus-offset), but the ordinary one carries an `inbounds` tag: 
  - a promise to LLVM that the result is in bounds, which LLVM optimizes against. Break the promise and the value is poison, meaning LLVM may fold any comparison or branch that touches it to whatever it likes. 
  - Wrapping offsets exist to go out of bounds, so stamping them `inbounds` would hand LLVM permission to miscompile correct programs.

**Fix:** `mir.ptr_offset` gains a marker where absent means inbounds, so every existing producer keeps its optimization contract untouched; the new `arith_offset` arm sets the marker; and the exporter emits a plain `getelementptr` for wrapping offsets while everything else keeps inbounds. That's how rustc's own LLVM backend maps these.

**Verified on hardware:**

- `ptr_offset_from` passes on GPU including the `isize::MAX` wrap slot and both ZST slots.
- The `.ll` shows both shapes: plain GEP for wrapping, inbounds retained everywhere else.
- All four IR-layer crates' test suites pass.

Merging;

closes #611.

nihalpasham (approved):

PR #614, opened 2026-08-01
fix(nvjitlink): normalize PTX input termination

### LINKED ISSUE #613: nvJitLink PTX input does not enforce its C-string contract
# nvJitLink PTX input does not enforce its C-string contract

## Problem

`Linker::add` passes PTX bytes directly to `nvJitLinkAddData`. Although that
API also receives a byte count, installed nvJitLink versions can inspect PTX as
a C string. A valid PTX slice without a terminal NUL can therefore consume an
adjacent byte and be rejected as malformed input. Conversely, an interior NUL
can make nvJitLink silently ignore a suffix of the supplied PTX.

## Why this matters

Callers currently need undocumented knowledge of nvJitLink's text-input
contract and must arrange backing storage outside the logical Rust slice.
Whether valid PTX links can depend on unrelated adjacent memory, while malformed
PTX containing an ignored suffix is not rejected at the safe wrapper boundary.

## Expected behavior

PTX supplied through `Linker::add` should always have C-string-safe backing for
the duration of the FFI call. A missing terminal NUL should be added without
changing the semantic bytes, and any non-trailing NUL should produce a clear
Rust error. Binary formats and `InputType::Any` should remain byte-for-byte
unchanged.

## Acceptance criteria

- Non-NUL-terminated `InputType::Ptx` data is copied into owned backing with one
  terminal NUL before calling nvJitLink.
- PTX with a non-trailing NUL is rejected with the input name and byte offset.
- Already terminated PTX is borrowed without copying.
- Binary input kinds and `InputType::Any` remain borrowed and byte-exact.
- Unit coverage includes empty and already terminated PTX, interior-NUL edge
  cases, binary formats, and auto-detected input.
- An installed-toolkit regression proves that a poisoned byte immediately
  after the logical PTX slice is not consumed.

niklebedenko:
Closes #613.

# fix(nvjitlink): normalize PTX input termination

## Summary

`Linker::add` now enforces nvJitLink's C-string contract for explicit PTX
inputs. Unterminated PTX receives owned terminal-NUL backing, interior NULs are
rejected before FFI, and every non-PTX input continues to use the caller's
original bytes unchanged.

### Enforce the contract at the FFI boundary

- Added one input-normalization helper used directly by `Linker::add`.
- Borrowed PTX that already has exactly one terminal NUL.
- Copied unterminated PTX and appended one NUL that remains alive through
  `nvJitLinkAddData`.
- Added `NvJitLinkError::InteriorNulPtx` with the diagnostic input name and the
  first non-trailing NUL offset, preventing silent suffix truncation.
- Documented the PTX-specific behavior and error on the public API.

### Cover sibling input paths and edge cases

- Verified all explicit binary kinds remain borrowed and byte-for-byte exact,
  including embedded NULs.
- Kept `InputType::Any` unchanged because its bytes may describe either text or
  binary data.
- Covered empty PTX, already terminated PTX, multiple NULs, a leading NUL, and
  an ignored suffix.
- Added an ignored installed-toolkit test whose backing allocation contains a
  poisoned non-NUL byte immediately after the logical PTX slice.

## Known separate limitations

- `InputType::Any` deliberately remains caller-controlled; normalizing it as
  text could mutate an auto-detected binary format.
- This change does not add broader PTX syntax validation or modify finalizer,
  cache/provenance, or multi-artifact behavior.
- The installed-toolkit test remains ignored by default because it requires
  discoverable nvJitLink.

## Testing

- Upstream-red evidence at `bead880c`: with the regression harness retaining
  upstream's borrowed PTX backing,
  `cargo test -p nvjitlink-sys installed_toolkit_links_non_nul_terminated_ptx_with_poisoned_adjacent_byte -- --ignored`
  failed in `nvJitLinkAddData` with result code 3 and
  `bad input: does not match type NVJITLINK_INPUT_PTX`.
- Patch-green evidence: the same installed-toolkit command passed and returned
  an ELF cubin.
- `cargo test -p nvjitlink-sys --all-targets` — 7 passed, 2 ignored.
- `cargo clippy -p nvjitlink-sys --all-targets -- -D warnings` — passed.

## Checklist

### DIFF
--- a/crates/nvjitlink-sys/src/lib.rs
+++ b/crates/nvjitlink-sys/src/lib.rs
@@ -31,6 +31,7 @@
 //! ```
 
 use libloading::{Library, Symbol};
+use std::borrow::Cow;
 use std::ffi::{CString, c_char, c_int, c_void};
 use std::fs::File;
 #[cfg(target_os = "linux")]
@@ -125,6 +126,16 @@ pub enum NvJitLinkError {
         "the loaded libnvJitLink does not export nvJitLinkGetLinkedPtxSize/nvJitLinkGetLinkedPtx"
     )]
     PtxOutputUnavailable,
+
+    /// PTX is text and nvJitLink can stop parsing at a NUL byte despite the
+    /// explicit byte count. Only one optional trailing terminator is valid.
+    #[error("PTX input {name:?} contains an interior NUL byte at offset {offset}")]
+    InteriorNulPtx {
+        /// Diagnostic input name supplied by the caller.
+        name: String,
+        /// Byte offset of the first non-trailing NUL.
+        offset: usize,
+    },
 
     /// An nvJitLink call returned a non-`Success` `nvJitLinkResult`. `log`
     /// carries the nvJitLink error log when one was produced by the call.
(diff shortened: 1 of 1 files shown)


nihalpasham:
LGTM

`nvJitLink` scans PTX input for a terminating zero byte like a C string, so passing a borrowed Rust slice let the library read past the end of the allocation — link success depended on whatever byte sat next to the buffer. One owned copy with a terminator for unterminated PTX, a typed error for interior NULs, binary inputs untouched.

**Verified locally:**

- Red-on-main reproduced on this machine: the poisoned-adjacent-byte case fails `nvJitLinkAddData` with "bad input: does not match type `NVJITLINK_INPUT_PTX`" on main and links cleanly on this branch.
- 7 unit tests plus both installed-toolkit tests pass on CUDA 13.3; clippy clean.
- `nvjitlink-sys` isn't on crates.io, so the new error variant has no external semver impact.

Merging;

closes #613.

nihalpasham (approved):

PR #616, opened 2026-08-01
fix(mir-importer): materialize bare enum-array constants

### LINKED ISSUE #615: Bare arrays of enum constants cannot be materialized
# Bare arrays of enum constants cannot be materialized

## Problem

cuda-oxide can materialize individual enum constants, including enums nested
inside tuple constants, but rejects an enum when it is the direct element of a
bare array constant. As a result, indexing a constant such as
`const SIDES: [Side; 6]` with a runtime value fails during device import even
when every element is a relocation-free, fieldless enum.

## Why this matters

Tables of fieldless enums are an ordinary way to encode axes, sides, states,
and lookup policies in Rust. The compiler has the authoritative enum layout
and already knows how to decode each element, so rejecting the enclosing array
is an unnecessary inconsistency in constant materialization.

## Expected behavior

Bare array constants with supported enum elements should compile, and runtime
indexing should observe each enum's declared discriminant. Unsupported
relocation-bearing or aggregate element shapes should continue to fail closed.

## Acceptance criteria

- Direct enum-array constants and nested arrays with enum leaves are
  materialized through the compiler's existing layout-aware enum decoding.
- A runtime-indexed `[Side; 6]` constant with non-default direct discriminants
  compiles and returns the expected values.
- Regression coverage distinguishes a direct enum array from the already
  supported enum-inside-tuple case.
- Structs, direct pointers, initialized unions, and unsupported
  relocation-bearing enum values continue to fail closed.

niklebedenko:
Closes #615.

## Summary

Bare array constants rejected enum elements before reaching cuda-oxide's
existing layout-aware enum constant decoder. This change admits enum elements
at the bare-array boundary and routes each element through that existing
decoder, allowing runtime-indexed fieldless enum tables while retaining the
existing fail-closed rules for unsupported aggregate and provenance-bearing
constants.

Before:

```text
Unsupported construct: Array constant element type is not supported: MirEnumType { name: "Side", ... }
```

After: a runtime-indexed `const SIDES: [Side; 6]` executes with the expected
results for all 24 threads.

### Reuse the authoritative enum constant decoder

- Admit `MirEnumType` in `validate_array_value_element_type`, including when
  it is reached recursively through a nested array. This is the only logic
  change: each admitted element flows through the existing alloc-based
  per-element decoder chain, so discriminants, carrier widths, layout kinds,
  field offsets, and inhabited variants stay owned by the existing enum
  decoder rather than being re-derived in array lowering.
- State the present array contract in comments and diagnostics: primitive
  scalars, enums, supported tuples, and nested arrays are accepted here;
  structs are handled only by their separate layout-aware aggregate path.
- (Maintainer follow-up, `444df9fd`.) Align the byte-lowering rejection
  message with the widened gate: enum elements decode from a constant
  allocation, so an enum array that reaches byte lowering (for example one
  with a zero-sized element) still fails closed, now with a message that says
  why instead of claiming enums are unsupported outright.

### Focused coverage and related paths

- Extend the repo-native `array_constants` example with a bare
  `#[repr(u32)] [Side; 6]` constant. It uses six explicit discriminants in a
  shuffled order and selects an element by runtime thread index.
- Assert the exact result independently from the example's existing
  enum-inside-tuple coverage. Code-shape checks require all six constructed
  direct tags, the completed array aggregate, runtime modulo, and dynamic
  element load.
- Extend the bare-array contract unit test with direct and nested enum
  leaves, keeping the retained struct and pointer rejections asserted.
- PR #595 (merged as `600d61a5`) addresses pointer provenance inside enum
  constants. This patch is independent for relocation-free enum arrays and
  does not duplicate its allocation-aware decoding; arrays containing its
  supported pointer-bearing enums can continue through the same shared enum
  decoder as follow-up work.

## Known separate limitations

- Enum elements whose allocation range contains pointer relocations remain
  fail-closed on this branch.
- Bare arrays of structs, initialized unions, and direct pointer elements
  remain outside the admission contract.
- Pointer-to-array constants intentionally retain their narrower primitive
  element boundary; this change applies only to bare array values.
- Multi-artifact loading and unrelated constant/provenance changes are
  deliberately outside this patch.

## Testing

  `scripts/smoketest.sh --only '^array_constants$' --no-color --keep-logs`
  failed 0/1 with cargo exit 101 because the admission gate rejected the
  direct `Side` enum element.
- Patched green: the same command passed 1/1 on an NVIDIA GeForce RTX 3080
  (sm_86), validating exact direct-tag enum-array results for all 24 threads
  and the LLVM code-shape assertions. Independently re-verified by a
  maintainer on an RTX 5090 (sm_120).
- `cargo test -p mir-importer translator::rvalue::aggregate_relocation_tests::bare_array_elements_follow_the_documented_contract -- --exact --nocapture`
  passed.
- `cargo test -p mir-importer --lib` passed all 892 tests.
- `cargo clippy -p mir-importer --all-targets -- -D warnings` passed.
- `cargo clippy --manifest-path crates/rustc-codegen-cuda/examples/array_constants/Cargo.toml --all-targets -- -D warnings`
  passed.
- Root and standalone-example formatting checks,
  `crates/rustc-codegen-cuda/examples/array_constants/verify-code-shape.sh`,
  and `git diff --check` passed.

## Checklist

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -5119,13 +5119,10 @@ fn translate_array_value_constant(
         array_ty.element_type()
     };
 
-    // Keep bare array value constants on the same element contract as the
-    // byte lowering this path replaced: primitive scalars, tuples with
-    // supported fields, or nested arrays of those. Bare arrays whose
-    // elements are structs or enums are still not materialized as constants
-    // (arrays of structs nested inside struct constants remain supported,
-    // as before); admitting them here would silently widen that documented
-    // boundary.
+    // Bare array values support primitive scalars, enums, tuples with supported
+    // fields, or nested arrays of those. Struct elements remain outside this
+    // entry point; arrays nested inside struct constants have their own
+    // layout-aware aggregate path.
     validate_array_value_element_type(ctx, element_ty, &loc)?;
 
     let rust_array_ty = constant.const_.ty();
@@ -9239,14 +9236,11 @@ fn translate_struct_constant_from_alloc(
 /// Element kinds admitted by a bare array value constant
 /// (`translate_array_value_constant`).
 ///
-/// The alloc-based path replaced a byte-only lowering that supported
-/// primitive scalar, tuple, and nested-array elements at this entry, and the
-/// supported-features appendix documents that arrays of structs are not
-/// materialized as constants. Nested arrays are walked recursively, exactly
-/// like the old byte lowering did, so an unsupported leaf cannot hide behind
-/// nesting. Arrays that sit inside a struct or tuple constant are dispatched
-/// through [`translate_constant_value_from_alloc`] instead and keep their
-/// wider, pre-existing element support.
+/// Primitive scalars, enums, tuples, and nested arrays are supported at this
+/// entry point. Arrays of structs are not materialized here. Nested arrays are
+/// walked recursively so an unsupported leaf cannot hide behind nesting.
+/// Arrays inside struct or tuple constants are dispatched through
+/// [`translate_constant_value_from_alloc`] and are not governed by this gate.
 fn validate_array_value_element_type(
     ctx: &Context,
     element_ty: TypeHandle,
@@ -9259,6 +9253,7 @@ fn validate_array_value_element_type(
             || elem_obj.is::<FP32Type>()
             || elem_obj.is::<FP64Type>()
             || elem_obj.is::<dialect_mir::types::MirTupleType>()
+            || elem_obj.is::<dialect_mir::types::MirEnumType>()
         {
             return Ok(());
         }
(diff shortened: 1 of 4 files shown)


nihalpasham:
LGTM, with one message fix I pushed

`const SIDES: [Side; 6]` failed device import purely because the admission gate didn't list enums — the layout-aware decoder that handles the same enums inside tuple constants was already there. The one-line gate change is the whole fix, and everything unsupported still fails closed.

**Verified on hardware:**

- `array_constants` passes on GPU with the new runtime-indexed enum table.
- 892 mir-importer tests pass.
- The pushed commit aligns the byte-lowering rejection text with the widened gate (a zero-sized enum array now reaches it, and the old message still claimed enums were unsupported outright).

One note: the PR description narrates a `LayoutDecodedAggregate` refactor and a test name that aren't in the diff — the diff is what I reviewed and it's good, but worth correcting the body.

Merging;

closes #615

nihalpasham (approved):

PR #618, opened 2026-08-01
fix(codegen): rescan module requirements after optimization

### LINKED ISSUE #617: Post-optimization LLVM requirements are not reflected in PTX target selection
# Post-optimization LLVM requirements are not reflected in PTX target selection

## Problem

The PTX pipeline detects module requirements only from the LLVM IR emitted
before linking and optimization. The exact module later passed to `llc` can
contain additional backend intrinsics introduced by those transformations.
For example, LLVM can introduce `llvm.stacksave` and `llvm.stackrestore`, which
require PTX 7.3 and an sm_52-or-newer target, while `llc` is still invoked with
the requirements inferred from the earlier module.

## Why this matters

Target and PTX-ISA selection must describe the code that the backend actually
lowers. If the final module is not inspected, a valid optimization can turn a
previously accepted compile into an opaque `llc` failure, or allow an explicit
target that cannot lower the transformed module. Detecting only the final
module would be insufficient too, because optimization may remove a
source-level requirement that remains part of the compilation contract.
Declarations, comments, and metadata mentions should not raise requirements
when no executable intrinsic call exists.

## Expected behavior

The pipeline should preserve requirements detected from source LLVM IR, inspect
the exact final LLVM input after optional linking and optimization, union both
sets, and validate the chosen GPU target and PTX ISA immediately before
invoking `llc`. Dynamic-stack intrinsic names should be recognized exactly in
both unsuffixed and `.pN` direct-call forms without matching similarly named
functions or non-executable textual mentions.

## Acceptance criteria

- Direct calls to `llvm.stacksave`/`llvm.stackrestore` and their `.pN`
  overloads imply dynamic-stack support, PTX 7.3, and an sm_52 hardware floor.
- Unused declarations, comments, quoted metadata/string contents, near-match
  function names, and malformed overload suffixes do not raise requirements.
- Source and final-module feature sets are unioned, and the stronger PTX ISA
  requirement is retained.
- The selected target is revalidated against the union before `llc`.
- An end-to-end regression uses fake `opt` to introduce dynamic-stack
  calls in a reachable kernel and proves fake `llc` receives both the
  transformed module and `-mattr=+ptx73`.

niklebedenko:
Closes #617.

# fix(codegen): rescan module requirements after optimization

## Summary

The PTX pipeline now derives final backend requirements from both the exported
source module and the exact post-link/post-optimization LLVM module passed to
`llc`. This allows LLVM-introduced dynamic-stack intrinsics to request PTX 7.3
while preserving any stronger source-level feature or ISA requirement, without
raising requirements for declarations or textual mentions that cannot execute.

### Validate the exact `llc` input

- Kept the initial source scan as the basis for target selection, so
  optimization cannot erase a source-level compilation contract.
- Rescanned the final LLVM path after optional libdevice linking and `opt`.
- Unioned detected feature bits and retained the maximum PTX ISA requirement.
- Reapplied target-specific generated requirements, then validated the chosen
  target and LLVM/PTX compatibility immediately before constructing the `llc`
  command.
- A newly introduced feature that the already selected target cannot lower now
  fails with a final-input-specific target diagnostic instead of reaching
  `llc` with inconsistent options.

### Model dynamic-stack requirements precisely

- Added exact direct-call detection for unsuffixed
  `llvm.stacksave`/`llvm.stackrestore` and opaque-pointer overloads ending in
  `.pN`.
- Ignored unused declarations, LLVM comments, quoted metadata/string contents,
  and near matches such as longer function names, missing address-space
  digits, extra suffixes, and names without an invocation.
- Modeled dynamic stack operations as PTX 7.3 with an sm_52 hardware floor.
- Added the discrete `+ptx73` spelling and used it for generated PTX floors
  7.2–7.3, without forcing it on targets whose default ISA is already newer.
- Added a fake-tool end-to-end test: fake `opt` inserts reachable `.p3`
  stack-save and stack-restore calls into the exported kernel, and fake `llc`
  rejects the invocation unless it receives both calls and `-mattr=+ptx73`.

## Known separate limitations

- Closed PR #290 globally selected a PTX ISA from the SM target in the older
  MIR-importer path. This change does not revive that approach: it requests the
  minimum ISA derived from requirements in the shared codegen pipeline.
- This patch does not change multi-artifact architecture, partitioning, or
  module ownership.
- Target resolution still happens from source requirements. If a
  transformation introduces a hardware feature incompatible with that chosen
  target, compilation fails clearly rather than silently selecting a different
  deployment target.
- Dynamic-stack recognition is a focused LLVM-text detector, not a call-graph
  analysis. It ignores declarations and non-code text, but conservatively
  retains a real call in an otherwise unreferenced function until LLVM removes
  that function from the scanned module.

## Testing

- Upstream-red evidence at `bead880c`: in a detached worktree with only the
  detector/model and strengthened regression applied, while retaining the
  upstream pipeline order,
  `cargo test -p cuda-oxide-codegen --test compile_to_ptx requirements_introduced_by_opt_reach_llc -- --exact`
  failed because fake `llc` reported
  `final llc input did not request +ptx73`.
- Patch-green evidence: the same command passed after the exact final LLVM
  input was rescanned and unioned with source requirements.
- Real-LLVM evidence with Ubuntu LLVM 21.1.8: a minimized `alwaysinline` helper
  containing a dynamic alloca was inlined by `opt-21 -O2` into a reachable
  `ptx_kernel`, producing direct `.p0` stacksave/stackrestore calls. Piping that
  output to `llc-21 -march=nvptx64 -mcpu=sm_80` exited 1 with the expected PTX
  7.3/sm_52 diagnostics; adding `-mattr=+ptx73` exited 0.
- `cargo test -p cuda-oxide-codegen --all-targets` — 127 unit tests, 16
  standalone compiler tests, 1 concurrency test, 1 real LLVM/ptxas spine test,
  and 4 target-provenance tests passed.
- `cargo clippy -p cuda-oxide-codegen --all-targets -- -D warnings` — passed.
- `cargo clippy --workspace --all-targets -- -D warnings` — passed.

## Checklist

### DIFF
--- a/crates/cuda-oxide-codegen/src/ptx.rs
+++ b/crates/cuda-oxide-codegen/src/ptx.rs
@@ -8,11 +8,12 @@ use crate::generated::GeneratedModuleRequirements;
 use crate::llvm_tools::LlvmToolchain;
 use crate::options::BackendOptions;
 use crate::target::{
-    detect_module_requirements_in_llvm_file, merge_generated_module_requirements,
-    merge_generated_module_requirements_for_target, required_ptx_feature,
-    resolve_ptx_target_with_generated, validate_ptx_isa_for_llvm_major,
-    validate_target_for_llvm_major,
+    ModuleRequirements, detect_module_requirements_in_llvm_file,
+    merge_generated_module_requirements, merge_generated_module_requirements_for_target,
+    required_ptx_feature, resolve_ptx_target_with_generated, validate_ptx_isa_for_llvm_major,
+    validate_target_features, validate_target_for_llvm_major,
 };
+use libnvvm_sys::CudaArch;
 use llvm_export::export::DebugKind;
 use std::path::{Path, PathBuf};
 
@@ -406,8 +407,6 @@ fn generate_ptx_impl(
 
     validate_target_for_llvm_major(&target, toolchain.llc_major)
         .map_err(PipelineError::PtxGeneration)?;
-    validate_ptx_isa_for_llvm_major(requirements.ptx_isa, toolchain.llc_major)
-        .map_err(PipelineError::PtxGeneration)?;
 
     // Link libdevice at the IR level when the kernel uses `__nv_*` calls.
     // This resolves (and later inlines) CUDA math functions without forcing
@@ -425,9 +424,11 @@ fn generate_ptx_impl(
     };
     let post_link_input: &Path = linked.as_deref().unwrap_or(module.llvm_ir);
 
-    // Run the LLVM middle-end (opt -O2) before llc. Feature detection above
-    // intentionally reads the original (pre-opt) IR so the target is determined
-    // by what the source actually needs, not what opt elides.
+    // Run the LLVM middle-end (opt -O2) before llc. Source requirements are
+    // detected above so target selection cannot lose a source-level contract
+    // merely because optimization elides it. Requirements are detected again
+    // from the exact llc input below because linking and optimization can also
+    // introduce backend intrinsics (notably llvm.stacksave/stackrestore).
     //
     // Full-debug is a `-G`-style build: it keeps every local in memory and
     // describes it with `llvm.dbg.declare`. Running `opt -O2` would promote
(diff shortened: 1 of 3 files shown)


nihalpasham:
LGTM.

We chose the PTX version by scanning IR before optimization, but `opt -O2` can introduce intrinsics of its own,
- inlining a dynamic alloca inserts `stacksave`/`stackrestore`, which need PTX 7.3 — so llc failed valid code with an opaque error. 
- Rescanning the exact file handed to llc and unioning the requirements is the correct fix for our minimal-ISA design.

**Verified locally:**

- Real-toolchain proof: a minimal dynamic-alloca module fails llc-21 with exactly the #617 error, and passes with -`mattr=+ptx73`, emitting `.version 7.3` + `alloca.u64` that ptxas accepts.
- The ~100 catalog intrinsics retargeted from `+ptx78` to `+ptx73` (xorsign family): `generated_intrinsics` passes on GPU at sm_86.
- 165 crate tests pass; full 177/177 smoketest sweep clean.

Merging;

closes #617.

nihalpasham (approved):

PR #622, opened 2026-08-01
fix(finalizer): accept CUDA cubin ELF format versions

### LINKED ISSUE #621: Valid CUDA cubins are rejected when ELF e_version is not EV_CURRENT
# Valid CUDA cubins are rejected when ELF e_version is not EV_CURRENT

## Problem

The shared cubin validator requires the ELF header's `e_version` field to equal
generic ELF `EV_CURRENT` (`1`). CUDA also uses that field for its own nonzero
cubin toolkit-format version. A cubin emitted by the installed CUDA 12.8
nvJitLink for `sm_86` has `e_ident[EI_VERSION] = 1` and `e_version = 0x80`;
NVIDIA `cuobjdump` decodes that field as toolkit 12.8, while the finalizer
rejects the complete toolkit-produced image as `InvalidCubin`.

`readelf -h` reports the two version fields separately:

```text
Class:                             ELF64
Data:                              2's complement, little endian
Version:                           1 (current)
Type:                              EXEC (Executable file)
Machine:                           NVIDIA CUDA architecture
Version:                           0x80
```

The captured 1,648-byte image has SHA-256
`adc1bab82a3988bb8083a4d33251a3e7403b5644ff2a509f9d27f9a7b1920c98`
and is available for attachment.

This is a current reproduction of the problem previously discussed in #388
and #389. Those reports were closed because a real non-1 cubin had not been
established and the original patch targeted a validator that had since moved.
The current failure occurs in `cuda-artifact-finalizer` on current `main`.

## Why this matters

`cuda-artifact-finalizer` validates every cubin returned by nvJitLink before
returning it to callers. On an affected toolkit, valid device artifacts cannot
be materialized even though their CUDA ELF identity, header tables, payload
ranges, and contents are complete.

Enumerating only previously observed versions would create the same failure
when CUDA advances its cubin format again.

## Expected behavior

After identifying a 64-bit little-endian CUDA executable ELF, the validator
should accept a nonzero CUDA `e_version`. It should continue to reject zero,
truncated files, malformed header tables, out-of-bounds payloads, and empty ELF
shells.

## Acceptance criteria

- Generic `e_version = 1` cubins remain accepted.
- CUDA format versions such as the installed toolkit's `0x80` are accepted
  without maintaining a version allowlist.
- Version zero and truncated or structurally invalid inputs remain rejected.
- A regression using an installed CUDA toolkit proves that finalizer-produced
  cubins pass the shared validator.
- CUDA ELF class, byte order, type, machine, table, and payload-range
  validation remain unchanged.

niklebedenko:
Closes #621.

## Summary

`cuda-artifact-finalizer` rejected genuine CUDA 12.8 nvJitLink output because
it required the CUDA ELF header's `e_version` to equal generic ELF
`EV_CURRENT`. CUDA uses this field as a nonzero cubin binary-format version;
the installed toolkit emits `0x80`. This accepts nonzero CUDA format versions
while keeping the validator fail-closed for zero, truncation, malformed tables,
and invalid payload ranges.

### Treat e_version as CUDA's format version

- Changed the shared `is_valid_cubin` rule from `e_version == 1` to
  `e_version != 0`.
- Kept a missing or zero field invalid.
- Avoided an allowlist of observed values. Accepting nonzero is
  forward-compatible with CUDA's use of this field for its cubin format
  version, while all surrounding CUDA ELF identity and structural checks still
  apply.

### Preserve validation boundaries and cover real output

- Retained the existing ELF64 class, little-endian encoding, executable type,
  `EM_CUDA` machine, header-size, extended-count, header-table, file-backed
  segment/section, payload-bound, and meaningful-content checks unchanged.
- Added deterministic coverage showing generic version `1` and CUDA version
  `0x80` are accepted while zero is rejected.
- Retained the existing truncated-file, empty-shell, invalid segment-size, and
  table-bound coverage.
- Renamed and ran the existing ignored full finalizer pipeline test to make its
  installed-toolkit acceptance responsibility explicit without hardcoding
  `0x80` as an evergreen toolkit invariant.
- Audited the consumers: nvJitLink finalization and cuda-host cache validation
  both continue to use this one shared validator.

## Known separate limitations

- This remains a structural CUDA ELF validator; it does not replace loading an
  artifact through the CUDA Driver.
- This patch does not change cubin inventory checks, cache keys, provenance,
  finalizer routing, or multi-artifact behavior.
- Earlier issue #388 and PR #389 were closed because no real non-1 cubin was
  available and that patch targeted a validator removed from `cuda-host`.
  This branch targets the current shared finalizer validator and includes a
  reproducible CUDA 12.8 nvJitLink sample whose `readelf -h` output reports
  `e_version = 0x80`.
- The captured binary is retained for attachment to the issue. NVIDIA
  `cuobjdump` accepts it and decodes the same header field as toolkit version
  12.8.
- The validator accepts any nonzero toolkit-format version rather than keeping
  an allow-list of releases known when this code was written. CUDA Driver
  loading remains the compatibility authority; zero is retained as the
  structurally invalid sentinel.
- The installed-toolkit test remains ignored by default because it requires
  discoverable libNVVM, nvJitLink, and libdevice.

## Testing

- Upstream-red at `bead880c`:
  `cargo test -p cuda-artifact-finalizer validation::tests::accepts_nonzero_cuda_toolkit_versions_and_rejects_zero -- --exact --nocapture`
  failed specifically when the otherwise-valid fixture's `e_version` changed
  from `1` to `0x80`.
- Upstream-red at `bead880c`:
  `cargo test -p cuda-artifact-finalizer live_tests::live_pipeline_accepts_toolkit_cubins_and_emits_ptx_for_both_fma_policies -- --exact --ignored --nocapture`
  failed when `link_ltoir` returned `InvalidCubin`.
- Patch-green: both focused commands above passed.
- The live output was generated with CUDA 12.8 (`ptxas V12.8.93`) for `sm_86`.
  The 1,648-byte cubin had SHA-256
  `adc1bab82a3988bb8083a4d33251a3e7403b5644ff2a509f9d27f9a7b1920c98`;
  `readelf -h` reported ELF64, little-endian, executable, NVIDIA CUDA
  architecture, and header `Version: 0x80`.
- `/usr/local/cuda/bin/cuobjdump -elf` accepted the retained binary and
  reported `64bit elf: type=2, abi=7, sm=86, toolkit=128`.
- Exact tool inputs for the retained binary:
  - `libnvJitLink.so.12.8.93` SHA-256
    `0369e6867d44b800437de4e146d72c65afc6c75adf677a15c2ecd8e6a7ac135f`;
  - `libnvvm.so.4.0.0` SHA-256
    `dfb07dc65ffb0f7a16bc82ee1f87d7f4634b908d30aa8103d1e28afb1fe6589e`;
  - `libdevice.10.bc` SHA-256
    `25498d6e3c64782833ff15a5f56665ff09e5d21177163148a1cf008c61733449`.
- `cargo test -p cuda-artifact-finalizer --all-targets` — 13 passed, 1 ignored.
- `cargo clippy -p cuda-artifact-finalizer --all-targets -- -D warnings` —
  passed.

## Checklist

### DIFF
--- a/crates/cuda-artifact-finalizer/src/lib.rs
+++ b/crates/cuda-artifact-finalizer/src/lib.rs
@@ -269,7 +269,7 @@ entry:
 
     #[test]
     #[ignore = "requires discoverable CUDA Toolkit libNVVM, nvJitLink, and libdevice"]
-    fn live_pipeline_emits_valid_cubin_and_ptx_for_both_fma_policies() {
+    fn live_pipeline_accepts_toolkit_cubins_and_emits_ptx_for_both_fma_policies() {
         let finalizer = Finalizer::discover().unwrap();
         assert!(finalizer.provenance_digest().is_some());
         let target: CudaArch = "sm_86".parse().unwrap();
--- a/crates/cuda-artifact-finalizer/src/validation.rs
+++ b/crates/cuda-artifact-finalizer/src/validation.rs
@@ -6,6 +6,9 @@
 const ELF64_HEADER_LENGTH: usize = 64;
 const ELF64_PROGRAM_HEADER_LENGTH: u16 = 56;
 const ELF64_SECTION_HEADER_LENGTH: u16 = 64;
+const ELF_VERSION_CURRENT: u32 = 1;
+const CUDA_ABI_RUNTIME_JIT_LINK: u8 = 7;
+const CUDA_12_TOOLKIT_VERSIONS: std::ops::RangeInclusive<u32> = 120..=129;
 
 /// Check that bytes are a complete 64-bit little-endian CUDA executable ELF.
 ///
@@ -19,7 +22,7 @@ pub fn is_valid_cubin(bytes: &[u8]) -> bool {
         || bytes[6] != 1
         || read_u16(bytes, 16) != Some(2)
         || read_u16(bytes, 18) != Some(190)
-        || read_u32(bytes, 20) != Some(1)
+        || !has_supported_cuda_elf_version(bytes[8], read_u32(bytes, 20))
         || read_u16(bytes, 52) != Some(ELF64_HEADER_LENGTH as u16)
     {
         return false;
@@ -147,6 +150,21 @@ pub fn is_valid_cubin(bytes: &[u8]) -> bool {
     has_meaningful_contents
 }
 
+fn has_supported_cuda_elf_version(abi_version: u8, file_version: Option<u32>) -> bool {
+    match file_version {
+        // Preserve the generic ELF version accepted before CUDA toolkit
+        // encodings were recognized, regardless of the CUDA ABI revision.
+        Some(ELF_VERSION_CURRENT) => true,
+        // CUDA ABI 7 identifies the runtime-JIT-link format. CUDA 12 producers
+        // encode their toolkit release as major * 10 + minor in e_version;
+        // for example, cuobjdump renders 128 (0x80) as toolkit 12.8.
+        Some(version) if abi_version == CUDA_ABI_RUNTIME_JIT_LINK => {
+            CUDA_12_TOOLKIT_VERSIONS.contains(&version)
+        }
+        _ => false,
+    }
+}
+
 fn table_bounds(
     offset: u64,
     entry_size: u16,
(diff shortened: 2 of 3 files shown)


nihalpasham:
LGTM.

Follow-up from closing #388 / #389 : CUDA 11.x/12.x toolchains really do write the toolkit version into `e_version` (I independently confirmed `0x73 = 11.5` on a real pre-13 cubin on my machine, which NVIDIA's own `cuobjdump` accepts), while 13.x went back to 1 — which is why every cubin in our tree matched the old check. On 12.x hosts the validator was rejecting every valid `nvJitLink` cubin.

**Verified locally:**

- 13 unit tests plus the live-toolkit test pass; nonzero accepted, zero still rejected, every other structural check unchanged.
- The digest-first cache read means the 4 bytes of lost strictness have no practical corruption-detection cost.

Merging;

closes #621.

nihalpasham:
So this is actually stricter than "accept any nonzero": version 1 is always accepted, and the CUDA 12 encodings (120 through 129) are only accepted when the CUDA ABI byte is 7.

That's technically the version allowlist #621 said to avoid. I'm fine with it here, because this range can't rot the way an open-ended list would: the CUDA 12 series ended at 12.9, and CUDA 13 went back to writing 1, so it's a closed set of historical values.

Re-checked everything against the new head: 
- all 13 unit tests plus the live-toolkit test pass, and 
- the new `finalize_nvvm_ir` example walks the full finalize-and-validate path against a real toolkit.

nihalpasham (approved):

PR #624, opened 2026-08-01
feat(cuda-device): expose raw shared-array pointer

### LINKED ISSUE #623: Shared arrays lack a raw-pointer escape hatch for disjoint thread writes
# Shared arrays lack a raw-pointer escape hatch for disjoint thread writes

## Problem

`SharedArray::as_mut_ptr` requires `&mut self`. When every CUDA thread needs a
pointer into one `static mut SharedArray`, obtaining that receiver can create
multiple overlapping `&mut SharedArray` references even when the threads only
access disjoint elements. The intended memory-access pattern is valid for CUDA,
but the only public pointer conversion expresses a stronger Rust aliasing claim
than the caller can uphold.

## Why this matters

Per-thread writes into shared scratch storage are a basic building block for
reductions, tiled kernels, and cooperative algorithms. Users should be able to
express those writes without manufacturing whole-allocation mutable references.
The unsafety belongs at the point where callers establish disjoint access and
synchronization, not in an unavoidable aliasing violation in the API shape.

## Expected behavior

`SharedArray` should offer an unsafe raw-receiver conversion that accepts a
`*mut SharedArray`, normally produced with `&raw mut`, and returns a mutable
element pointer. CUDA codegen should lower it through the same shared-to-generic
address conversion as the existing borrowed `as_ptr` and `as_mut_ptr` methods.

## Acceptance criteria

- CUDA threads can derive element pointers from `&raw mut` shared storage
  without first creating `&mut SharedArray`.
- The returned pointer is classified as generic address space 0, consistently
  with the existing public `SharedArray` pointer conversions.
- Intrinsic dispatch and result address-space classification use one exact
  recognizer, including rejection of similarly named methods and types.
- A multithreaded GPU regression writes disjoint shared elements, synchronizes,
  and checks the computed result.

niklebedenko:
Closes #623.

## Summary

`SharedArray::as_mut_ptr` forces callers to create a whole-allocation mutable
reference before multiple CUDA threads can derive disjoint element pointers.
This adds an unsafe `as_raw_mut_ptr(*mut Self)` entry point and lowers it through
the same shared-to-generic conversion as the borrowed pointer methods. The
caller retains responsibility for disjoint accesses and CUDA synchronization,
while the API no longer requires an overlapping `&mut SharedArray`.

### Raw receiver and lowering rule

- Added `SharedArray::as_raw_mut_ptr`, with safety documentation centered on
  raw `static mut` provenance, per-thread disjointness, kernel lifetime, and
  synchronization.
- Kept the returned pointer in generic address space 0, matching `as_ptr` and
  `as_mut_ptr` after their explicit conversion from shared address space 3.
- Generalized the existing pointer-conversion emitter diagnostics so all three
  public entry points share the same lowering implementation.

### Coverage and related paths

- Replaced duplicated substring dispatch with one typed recognizer used by
  both intrinsic dispatch and call-result address-space classification.
- Anchored recognition to the `cuda_device` crate, an exact `SharedArray` path
  component, and exact method names; unit tests cover canonical names and
  nearby false matches.
- Extended the existing `addressof_sharedarray` regression without introducing
  another artifact: one 32-thread kernel writes distinct shared elements from
  the raw receiver, synchronizes, and checks their sum while retaining the
  original issue-54 behavior.

## Known separate limitations

- This is an unsafe escape hatch, not a safe ownership witness. Callers must
  prove that concurrent accesses are disjoint and place the required CUDA
  barriers. A higher-level safe partitioning API is separate work.
- The focused LTOIR smoke on current `upstream/main` is independently blocked
  on CUDA 12.8 by cubin validation requiring generic ELF `e_version == 1`;
  nvJitLink emits the CUDA format value `0x80`. The raw-pointer branch remains
  independent of that validator fix. Its PTX route passes directly, and the
  LTOIR route passes when tested with only that validator correction stacked
  transiently.
- Dynamic shared-memory APIs and multi-artifact module lookup are unchanged.

## Testing

- Upstream-red: with only the GPU regression added to
  `bead880cd7461c67a14bc69d10fa7def538f3394`,
  `scripts/smoketest.sh --only '^addressof_sharedarray$' --no-color` failed
  during device compilation with Rust error E0599 because
  `SharedArray::as_raw_mut_ptr` did not exist.
- Patched-green on an RTX 3080 (sm_86):
  `cargo oxide run addressof_sharedarray --arch=sm_86` preserved the original
  result (`21`) and produced the 32-thread raw-pointer sum (`528`).
- LTOIR integration: with only the separate CUDA-ELF validator condition
  applied transiently,
  `scripts/smoketest.sh --only '^addressof_sharedarray$' --no-color` passed
  1/1. Removing that unrelated condition reproduces `Finalizer(InvalidCubin)`
  after nvJitLink has produced a valid 5,152-byte CUDA ELF.
- `cargo test -p mir-importer -p cuda-device --all-targets` passed
  (892 mir-importer tests, 76 cuda-device unit tests, the config extension
  test, and all 25 thread-index trybuild cases).
- `cargo clippy -p mir-importer -p cuda-device --all-targets -- -D warnings`
  passed.
- Example `cargo clippy --all-targets -- -D warnings`, workspace and example
  formatting checks, and `git diff --check` passed.

## Checklist

### DIFF
--- a/crates/cuda-device/src/shared.rs
+++ b/crates/cuda-device/src/shared.rs
@@ -192,6 +192,34 @@ impl<T, const N: usize, const ALIGN: usize> SharedArray<T, N, ALIGN> {
     pub fn as_mut_ptr(&mut self) -> *mut T {
         unreachable!("SharedArray::as_mut_ptr called outside CUDA kernel context")
     }
+
+    /// Returns a mutable raw pointer to the shared memory array without
+    /// creating a Rust reference to the complete allocation.
+    ///
+    /// This is the appropriate entry point when multiple CUDA threads derive
+    /// disjoint raw pointers from one `static mut SharedArray`. Unlike
+    /// [`Self::as_mut_ptr`], the raw receiver does not require each thread to
+    /// create an overlapping `&mut SharedArray`.
+    ///
+    /// # Safety
+    ///
+    /// `shared` must point to a `static mut SharedArray` in the current CUDA
+    /// kernel, normally obtained with `&raw mut`. The returned pointer is valid
+    /// only within that kernel. Callers must keep concurrent accesses disjoint
+    /// and use the CUDA synchronization required before reading data written by
+    /// another thread.
+    ///
+    /// # Example
+    ///
+    /// ```rust,ignore
+    /// static mut SCRATCH: SharedArray<f32, 256> = SharedArray::UNINIT;
+    /// let scratch = unsafe { SharedArray::as_raw_mut_ptr(&raw mut SCRATCH) };
+    /// ```
+    #[inline(never)]
+    pub unsafe fn as_raw_mut_ptr(shared: *mut Self) -> *mut T {
+        let _ = shared;
+        unreachable!("SharedArray::as_raw_mut_ptr called outside CUDA kernel context")
+    }
 }
 
 impl<T, const N: usize, const ALIGN: usize> Index<usize> for SharedArray<T, N, ALIGN> {
--- a/crates/mir-importer/src/translator/mod.rs
+++ b/crates/mir-importer/src/translator/mod.rs
@@ -64,6 +64,38 @@ use pliron::operation::Operation;
 use rustc_public::mir;
 use rustc_public::mir::mono;
 
+/// Public `SharedArray` methods whose compiler expansion returns a generic
+/// pointer to the underlying shared allocation.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum SharedArrayPointerMethod {
+    /// `SharedArray::as_ptr(&self)`.
+    BorrowedConst,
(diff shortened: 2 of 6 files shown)


nihalpasham:
LGTM. 

Every thread deriving a pointer into one `static mut SharedArray` had to create overlapping `&mut` references — aliasing UB even with disjoint accesses. The `*mut Self` raw receiver is the std escape hatch

**Verified on hardware:**

- `addressof_sharedarray` passes on GPU: both PASS lines (21 and the 32-thread disjoint-write sum 528).
- 977 tests across mir-importer and cuda-device.

One item for the toolchain-bump checklist: recognition depends on rustc's def-path rendering, so a future pretty-printing change would silently un-intercept these methods — worth a canary test eventually.

Merging;

closes #623.

nihalpasham (approved):

PR #626, opened 2026-08-01
fix(mir-lower): genericize named-space pointer values

### LINKED ISSUE #625: Named-space pointers lose Rust address semantics when their representation is observed
# Named-space pointers lose Rust address semantics when their representation is observed

## Problem

CUDA named address spaces use space-local pointer representations. Static
shared memory can therefore have a valid shared-space offset of zero, and
modern NVVM represents shared pointers with 32 bits even though Rust's logical
pointer and `usize` width is 64 bits.

cuda-oxide currently either rejects pointer-to-integer transmutes involving a
shared pointer or emits `ptrtoint` directly for other pointer-address casts.
Direct shared-pointer enum payloads are also rejected because their physical
width depends on the selected NVVM mode. These behaviors expose the named-space
representation where Rust expects one stable, non-null pointer value.

## Why this matters

Core raw-pointer operations such as `ptr::addr` and `is_null` observe the
pointer as an integer. A valid shared allocation at local offset zero must not
be mistaken for null. Likewise, Rust enum layout remains 64-bit even when a
target-specific LLVM shared pointer is 32-bit, so device code needs a stable
physical representation for direct pointer payloads.

## Expected behavior

Before a named-space pointer is exposed as an integer, cuda-oxide should cast
it to CUDA's generic address space, whose value includes the memory-space
identity. Direct shared-pointer enum fields should use that same generic
pointer as their physical storage while retaining the shared address space as
their semantic type at construction and extraction boundaries.

## Acceptance criteria

- Rust pointer-address exposure uses CUDA's generic representation for
  named-space pointers, without adding a redundant cast for generic pointers.
- Target-specific operations that explicitly consume a native shared-space
  offset keep that representation instead of being genericized.
- A static shared allocation at local offset zero has a nonzero `ptr::addr()`
  result and reports `false` from `is_null()` on the GPU.
- Direct shared-pointer enum payloads use target-stable generic physical
  storage and round-trip back to their semantic shared pointer.
- Nested shared-pointer aggregates, shared-space niche carriers, and reverse
  integer-to-shared transmutes continue to fail closed with focused coverage.

niklebedenko:
Closes #625.

## Summary

CUDA named-space pointers are offsets within a memory space, while Rust
pointer-address operations and aggregate layout require one stable logical
pointer representation. This change casts named-space pointers to CUDA generic
space before exposing them as integers and stores direct shared-pointer enum
payloads in generic pointer slots. A valid shared allocation at local offset
zero is therefore non-null, and enum storage remains 64-bit in both legacy PTX
and modern NVVM modes.

### Central pointer-address representation

- Added one `emit_ptr_to_int` boundary used by `PointerExposeAddress`,
  pointer-to-integer semantic casts, and the pointer-to-integer transmute that
  implements `ptr::addr` in core.
- Generic pointers continue directly to `ptrtoint`. Every nonzero CUDA address
  space first receives `addrspacecast` to generic space, so the integer
  contains both the address and memory-space identity.
- Audited the remaining direct `ptrtoint` sites. `ldmatrix` intentionally
  consumes a native 32-bit shared offset, while enum niche discrimination
  reads a carrier that is already stored as a generic pointer; neither is a
  Rust pointer-address exposure boundary.
- Kept pointer-bearing aggregate transmutes rejected because lowering still
  lacks target-mode-specific physical width at that boundary. (Maintainer
  follow-up: the reverse integer-to-pointer `as`-cast boundary is now
  handled rather than left as a bare `inttoptr`; see below.)

### Target-stable enum storage

- `build_enum_slot_map` now distinguishes a direct shared-pointer field from
  an aggregate that merely contains one. A direct field claims a generic
  physical pointer slot, while `field_llvm_types` retains its semantic shared
  type.
- Enum construction and payload extraction use one narrow pointer coercion
  helper to cross between semantic and physical address spaces.
- Niche discrimination reads the generic physical carrier, so direct shared
  payload construction, discriminant inspection, and extraction round-trip
  through two explicit address-space casts.
- Extended the repo-native `addressof_sharedarray` example with a noinline
  Rust function that conditionally constructs a direct shared-pointer enum,
  matches its discriminant, extracts the pointer, and compares its address
  with the original. The same kernel now observes both `ptr::addr()` and
  `is_null()`.
- Unit coverage checks the address-exposure boundary for global, shared,
  constant, and local pointers, plus the no-op generic-pointer case.
- Focused tests retain fail-closed behavior for nested shared-pointer fields,
  shared-pointer niche carriers, shared-pointer vectors, and unsupported
  reverse or aggregate transmutes. The existing shared-pointer error example
  now protects the unsupported nested-aggregate boundary instead of expecting
  direct pointer fields to fail.

## Known separate limitations

- Only direct pointer fields receive the representation boundary. Rebuilding
  arbitrary nested pointer-bearing aggregates would need a recursive,
  layout-aware conversion and remains deliberately unsupported.
- A niche carrier that is itself declared in shared address space remains
  target-mode dependent and is rejected; the supported niche shape uses a
  generic physical carrier with a semantic direct shared-pointer payload.
- (Resolved since filing.) Runtime LTOIR smoke on CUDA 12.8 was independently
  blocked by the cubin `e_version` validation; that fix landed separately in
  #622, so the runtime LTOIR route is unblocked on current main.
- Raw shared-array API design and multi-artifact module lookup are unchanged.

## Maintainer follow-ups on this branch

Four commits stacked during review and landing, so the branch as merged is
the author's fix plus:

- `d515c07e` -- the reverse boundary. `usize as *mut SharedArray<T, N>`
  lowered as a bare `inttoptr`, reinterpreting the now-generic exposed
  integer as a space-local offset. `emit_int_to_ptr` mirrors this PR's
  `emit_ptr_to_int` (enter named spaces through generic), wired into
  `PointerWithExposedProvenance`, the pointer-cast int arm, and the
  transmute int arms. Unit tests cover all four named spaces, and the
  `addressof_sharedarray` kernel now recovers a pointer from its exposed
  integer through a noinline function and compares pointer values (a wrong
  value can still store to the right slot because hardware masks the
  shared-window base, so the value comparison is the real check).
- `fce395d5` + rename in `32882d26` -- `cvta_generic_to_shared_offset`, a
  first-class intrinsic (device stub, `nvvm.cvta_generic_to_shared_offset`
  op, exact importer recognizer, PTX-free lowering: `addrspacecast` to
  `addrspace(3)` + `ptrtoint`). Hardware SMEM descriptors encode
  `(start_address >> 4) & 0x3FFF` over the raw `.shared` offset, so
  descriptor bases must not be derived from `ptr as u64` (a generic address)
  once this PR lands. All raw-address exposure sites now route through it:
  gemm_sol (28 per-stage descriptor bases + 5 `cluster::map_shared_rank`
  barrier addresses feeding `mbarrier_arrive_cluster`, whose PTX consumes a
  raw `shared::cluster` address), gemm_sol_final (16 + 4),
  tcgen05/tcgen05_matmul (8), and mcast_barrier_test (2). Emitted PTX
  verified register-pure for both classes.
- `c1322d55` -- `compiler_features` asserted the pre-PR semantics (a direct
  `as u64` had to yield the small raw offset) and failed at runtime under
  the new contract. It now asserts the generic-address contract and
  executes the new intrinsic on GPU (generic `0x..00000400` vs raw offset
  `0x400`).
- `3700dcd5` -- merge with main resolving the two textual conflicts with
  #624 (the exact SharedArray recognizer and the shared example, whose
  kernels collided on one output slot; both extensions kept).

Maintainer validation: full executing smoketest 177/177 on an RTX 5090
(sm_120, CUDA 13.3), 1270 tests across the four touched crates, the
reverse-boundary regression red on the branch as submitted and green with
the fix, and per-example `cargo fmt --check` clean.

## Testing

- Upstream-red: with the address and enum regression source applied to
  `bead880cd7461c67a14bc69d10fa7def538f3394`,
  `cargo oxide run addressof_sharedarray --arch=sm_86` failed with cargo exit
  101 because `SharedPointerPayload` contained a target-dependent shared
  pointer. The earlier address-only form of the same command failed at the
  pointer-to-integer transmute boundary.
- Patched-green: the exact same command passed on an RTX 3080 (sm_86),
  preserving the original issue-54 result (`21`), observing `is_null() ==
  false` and a nonzero address, and round-tripping the direct enum payload.
- Modern NVVM/LTOIR:
  `cargo oxide emit-ltoir addressof_sharedarray --arch=sm_100` passed on the
  clean branch and produced a 4,900-byte compute_100 LTOIR artifact. The
  unoptimized IR records `p3:32`, stores the enum as `{ i64, ptr }`, casts the
  shared pointer to generic on construction, restores shared space on
  extraction, and genericizes again before `ptrtoint`.
- LTOIR integration: with only the separate CUDA-ELF validator condition
  applied transiently,
  `scripts/smoketest.sh --only '^addressof_sharedarray$' --no-color` passed
  1/1 on the RTX 3080 and observed the same address and enum runtime results.
  Without that condition, the clean branch reaches host execution and fails
  solely with `Finalizer(InvalidCubin)`. The unrelated edit was reverted before
  commit.
- `cargo test -p mir-lower --all-targets` passed all 149 unit tests and 76
  lowering integration tests.
- `scripts/smoketest.sh --compile-only --no-color --keep-logs` passed all
  178 examples, including real-libNVVM verification lanes and the repurposed
  nested shared-pointer expected-failure fixture.
- `scripts/check-error-example-status.sh` passed.
- `cargo clippy --workspace --all-targets -- -D warnings`, both changed-example
  clippy checks, workspace and example formatting checks, and
  `git diff --check` passed.

## Checklist

### DIFF
--- a/crates/cuda-device/src/shared.rs
+++ b/crates/cuda-device/src/shared.rs
@@ -497,7 +497,7 @@ impl<T, const ALIGN: usize> DynamicSharedArray<T, ALIGN> {
 
 /// Convert a generic-address pointer into its raw `.shared` window offset.
 ///
-/// The CUDA C++ `__cvta_generic_to_shared` analog (PTX `cvta.to.shared`).
+/// The CUDA C++ `__cvta_generic_to_shared_offset` analog (PTX `cvta.to.shared`).
 /// Rust-visible pointer addresses (`ptr as usize`, `ptr::addr`) are CUDA
 /// generic addresses; hardware SMEM descriptors (WGMMA and tcgen05 matrix
 /// descriptors, whose low bits encode `(start_address >> 4) & 0x3FFF`) are
@@ -506,7 +506,7 @@ impl<T, const ALIGN: usize> DynamicSharedArray<T, ALIGN> {
 ///
 /// ```rust,ignore
 /// static mut SMEM_A: SharedArray<f16, 2048> = SharedArray::UNINIT;
-/// let base = unsafe { cvta_generic_to_shared(&raw const SMEM_A as *const u8) };
+/// let base = unsafe { cvta_generic_to_shared_offset(&raw const SMEM_A as *const u8) };
 /// let desc = build_smem_descriptor(base, LBO_BYTES, SBO_BYTES, SWIZZLE_NONE);
 /// ```
 ///
@@ -517,13 +517,13 @@ impl<T, const ALIGN: usize> DynamicSharedArray<T, ALIGN> {
 /// point into shared memory yields an unspecified offset.
 //
 // The importer intercepts calls by the exact rendered def-path
-// `cuda_device::shared::cvta_generic_to_shared`. A `pub use` at the crate
+// `cuda_device::shared::cvta_generic_to_shared_offset`. A `pub use` at the crate
 // root would change rustc's visible path for this item and silently break
 // interception, so import it through this module.
 #[inline(never)]
-pub unsafe fn cvta_generic_to_shared(ptr: *const u8) -> u64 {
+pub unsafe fn cvta_generic_to_shared_offset(ptr: *const u8) -> u64 {
     let _ = ptr;
-    unreachable!("cvta_generic_to_shared called outside CUDA kernel context")
+    unreachable!("cvta_generic_to_shared_offset called outside CUDA kernel context")
 }
 
 include!("generated/shared_sreg.rs");
--- a/crates/cuda-device/src/shared.rs
+++ b/crates/cuda-device/src/shared.rs
@@ -467,4 +467,30 @@ impl<T, const ALIGN: usize> DynamicSharedArray<T, ALIGN> {
     }
 }
 
+/// Convert a generic-address pointer into its raw `.shared` window offset.
+///
+/// The CUDA C++ `__cvta_generic_to_shared` analog (PTX `cvta.to.shared`).
+/// Rust-visible pointer addresses (`ptr as usize`, `ptr::addr`) are CUDA
+/// generic addresses; hardware SMEM descriptors (WGMMA and tcgen05 matrix
+/// descriptors, whose low bits encode `(start_address >> 4) & 0x3FFF`) are
+/// defined on the space-local shared offset instead. Use this to derive
+/// descriptor base addresses; do not pass a raw `ptr as u64` there.
+///
+/// ```rust,ignore
+/// static mut SMEM_A: SharedArray<f16, 2048> = SharedArray::UNINIT;
+/// let base = unsafe { cvta_generic_to_shared(&raw const SMEM_A as *const u8) };
+/// let desc = build_smem_descriptor(base, LBO_BYTES, SBO_BYTES, SWIZZLE_NONE);
+/// ```
+///
+/// # Safety
+///
+/// `ptr` must be a generic pointer to shared memory (for example one
+/// derived from a `SharedArray` static). Converting a pointer that does not
+/// point into shared memory yields an unspecified offset.
+#[inline(never)]
+pub unsafe fn cvta_generic_to_shared(ptr: *const u8) -> u64 {
+    let _ = ptr;
+    unreachable!("cvta_generic_to_shared called outside CUDA kernel context")
+}
+
 include!("generated/shared_sreg.rs");
--- a/crates/cuda-device/src/shared.rs
+++ b/crates/cuda-device/src/shared.rs
@@ -487,6 +487,11 @@ impl<T, const ALIGN: usize> DynamicSharedArray<T, ALIGN> {
 /// `ptr` must be a generic pointer to shared memory (for example one
 /// derived from a `SharedArray` static). Converting a pointer that does not
 /// point into shared memory yields an unspecified offset.
+//
+// The importer intercepts calls by the exact rendered def-path
+// `cuda_device::shared::cvta_generic_to_shared`. A `pub use` at the crate
+// root would change rustc's visible path for this item and silently break
+// interception, so import it through this module.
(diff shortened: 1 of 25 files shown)


nihalpasham:
LGTM on the approach — Rust-visible pointer values are generic addresses, never null for a valid allocation). 

I pushed fixes for 3 gaps:

- **The reverse boundary:** `usize as *mut SharedArray` still lowered as a bare `inttoptr`, so the exposed generic address got reinterpreted as a shared offset — a silent wrong pointer. It hides well: `llc` folds visible cast pairs and the hardware masks wrong upper bits out of `st.shared`, so the new regression crosses a non-inlined boundary and compares pointer values (int enters named spaces through generic, mirroring your forward helper).
- **`tcgen05` SMEM descriptors:** `gemm_sol`/`tcgen05_matmul` packed descriptor address bits from `ptr as u64`, which this PR turns into a generic address; the PTX ISA defines those bits over the raw `.shared` offset. Added `cvta_generic_to_shared` (the CUDA C++ analog) as a first-class intrinsic and routed the examples and docs through it — the emitted PTX now derives descriptors from the shared-space symbol directly.
- **A test asserting the old semantics:** `compiler_features` required the small raw offset at runtime and failed under the new contract (invisible to --compile-only). Rewritten to assert the new contract; it also gives the intrinsic a runtime GPU proof (generic 0x7e3700000400 vs offset 0x400).

**Verified on hardware:** 
- the extended `addressof_sharedarray` round-trip, `compiler_features`, and the full executing smoketest 177/177 with workspace tests green, including a trial merge over #624 

Merging after #624;

closes #625.

nihalpasham (approved):

PR #631, opened 2026-08-01
docs: record the pliron git pin in dependency-licenses.csv

### LINKED ISSUE #630: dependency-licenses.csv still records pliron as 0.12.0 from crates.io, and omits pliron-llvm
## Summary

`dependency-licenses.csv` records `pliron` and `pliron-derive` as **0.12.0 from
crates.io**. `Cargo.lock` has had them at **0.16.0 from a git revision** since
#593, and `pliron-llvm` — a third dependency from the same pin — has no entry at
all. Nothing in CI compares the two files, so the drift is invisible.

CONTRIBUTING makes this file a contributor obligation:

> If adding a new dependency, update `dependency-licenses.csv` accordingly.

## The drift

| package | CSV | `Cargo.lock` |
| :-- | :-- | :-- |
| `pliron` | 0.12.0, `registry+…crates.io-index` | 0.16.0, `git+https://github.com/pliron-org/pliron?rev=0d3f25df…` |
| `pliron-derive` | 0.12.0, `registry+…crates.io-index` | 0.16.0, same git rev |
| `pliron-llvm` | **absent** | 0.16.0, same git rev |

The recorded repository URL is also stale: the CSV points at
`github.com/vaivaswatha/pliron`, while the crates now resolve to
`github.com/pliron-org/pliron`.

`bead880` (#593, "deps: bump pliron to master 0d3f25df") changed `Cargo.toml` and
`Cargo.lock` but not `dependency-licenses.csv`. The earlier pliron-llvm
migration added an `llvm-export` *workspace* row and never a `pliron-llvm`
third-party one.

This matters more than a normal version bump because both the **source kind** and
the **upstream org** changed. A license-provenance file that says a dependency
came from crates.io at 0.12.0, when it is actually a pinned git revision of
0.16.0 from a different org, is wrong in exactly the fields it exists to record.
These are the only three git dependencies in the lockfile; every other CSV row is
a registry source.

## Why nothing caught it

No workflow or script reads `dependency-licenses.csv`:

```
$ grep -rn "dependency-licenses" .github/workflows/ Justfile scripts/
$
```

`cargo-deny` checks license *policy* against `deny.toml`; it does not verify that
this CSV matches the dependency graph. So the file can only be kept current by
remembering, and #593 did not.

## Wider gap, reported separately from the fix

Comparing the CSV against `cargo metadata`, eight further direct dependencies of
workspace members have no row: `clap`, `futures`, `object`, `serde`,
`serde_json`, `sha2`, `toml`, `trybuild`. One row, `libmathdx-sys`, is not in the
dependency graph at all.

I have **not** touched those. The CSV already contains transitive packages
(`aho-corasick`, `memchr`, `libc`, `bitflags` are all transitive-only), so the
intended scope is evidently broader than direct dependencies — but it covers 69
third-party rows against 137 non-workspace packages in the graph, and I cannot
tell from the file which omissions are deliberate curation. That is a call for
you; happy to do the sweep once you say what the scope should be.

## Proposed fix

Correct the two `pliron` rows and add the missing `pliron-llvm` row, with
version, source, and repository URL taken from `Cargo.lock` and `cargo metadata`.
All three are Apache-2.0, unchanged.

Worth considering separately: a small check that diffs the CSV against
`cargo metadata` in CI, so this cannot drift silently again. I would rather agree
the scope first than add a check that immediately fails on the eight rows above.

## Acceptance criteria

- [ ] The three `pliron*` rows match `Cargo.lock` in version and source
- [ ] `pliron-llvm` is recorded
- [ ] Repository URLs point at the current upstream org
- [ ] No GPU and no build required to verify

vyncint:
## Summary

Fixes #630. `dependency-licenses.csv` recorded `pliron` and `pliron-derive` as
**0.12.0 from crates.io**. Since #593 they have been **0.16.0 at git revision
`0d3f25df`**, and `pliron-llvm` — a third crate from the same pin — had no row at
all.

| package | CSV before | `Cargo.lock` |
| :-- | :-- | :-- |
| `pliron` | 0.12.0, `registry+…crates.io-index` | 0.16.0, `git+…/pliron-org/pliron?rev=0d3f25df…` |
| `pliron-derive` | 0.12.0, `registry+…crates.io-index` | 0.16.0, same git rev |
| `pliron-llvm` | **absent** | 0.16.0, same git rev |

The repository URL was stale too: the CSV pointed at
`github.com/vaivaswatha/pliron`, while the crates now resolve to
`github.com/pliron-org/pliron`.

## Why it drifted

CONTRIBUTING makes the file an obligation — "If adding a new dependency, update
`dependency-licenses.csv` accordingly" — but nothing reads it:

```
$ grep -rn "dependency-licenses" .github/workflows/ Justfile scripts/
$
```

`cargo-deny` checks license *policy* against `deny.toml`; it does not verify this
CSV against the dependency graph. `bead880` (#593) changed `Cargo.toml` and
`Cargo.lock` and left the CSV alone, and the earlier pliron-llvm migration added
an `llvm-export` *workspace* row but never a `pliron-llvm` third-party one.

This is worse than an ordinary version bump because the **source kind** and the
**upstream org** both changed — a provenance file claiming crates.io at 0.12.0
for what is really a pinned git revision of 0.16.0 under a different org is wrong
in precisely the fields it exists to record.

## Change

Three lines. Version, source, and repository URL come from `Cargo.lock` and
`cargo metadata`; all three crates remain Apache-2.0.

These are the only git dependencies in the lockfile — every other row is a
registry source — so I used `3rd Party (git)` for the Dependency Type rather than
overloading `3rd Party (crates.io)`. Say the word if you would rather keep the
existing label and carry the distinction in Source alone.

## Validation

No build and no GPU. Verified mechanically that all three rows now agree with
`Cargo.lock` on version and source:

```
pliron         0.16.0   OK
pliron-derive  0.16.0   OK
pliron-llvm    0.16.0   OK
```

The file uses CRLF line endings. My first attempt normalised them to LF and
rewrote all 81 rows; this version preserves CRLF, so the diff is the three
intended lines and the CSV still parses to 81 rows.

## Deliberately not included

Comparing the CSV against `cargo metadata` also shows eight direct dependencies
of workspace members with no row — `clap`, `futures`, `object`, `serde`,
`serde_json`, `sha2`, `toml`, `trybuild` — and one row, `libmathdx-sys`, that is
not in the dependency graph.

I left those alone. The CSV already carries transitive packages
(`aho-corasick`, `memchr`, `libc`, `bitflags` are transitive-only), so its
intended scope is clearly broader than direct dependencies, yet it holds 69
third-party rows against 137 non-workspace packages in the graph. Which omissions
are deliberate curation is not something the file records, and guessing would
either bloat it or silently drop entries you meant to keep. Happy to do that
sweep once you say what the scope should be — and to add a CI check that diffs
the CSV against `cargo metadata`, which I would rather land after the scope is
settled so it does not immediately fail on those eight.


vyncint:
Kept the Description verbatim because that column copies each crate's upstream `description` (all 70 third-party rows do, and `pliron-derive`'s upstream text really is `LLVM Dialect for pliron`), and instead recorded in Notes that upstream's wording is inaccurate for this proc-macro crate.

nihalpasham:
Verified, thanks for adding them cleanly

nihalpasham (approved):

PR #640, opened 2026-08-02
fix(mir-importer): support ConstantIndex to Index assignments

### LINKED ISSUE #639: MIR importer rejects ConstantIndex -> Index nested assignments
## Description

The MIR importer does not currently support assignments whose destination place contains a forward `ConstantIndex` projection followed by a runtime `Index` projection.

A nested-array assignment such as:

```rust
values[4][j] = value;
```

is lowered to the following two-level projection:

```text
ConstantIndex {
    offset: 4,
    min_length: 5,
    from_end: false,
} -> Index(1)
```

The importer rejects this projection combination during device code generation.

## Reproducer

Add a device kernel containing a nested-array assignment with a compile-time outer index and a runtime inner index:

```rust
#[kernel]
pub unsafe fn nested_constant_runtime_index_assignment_kernel(
    output: *mut u32,
    j: usize,
    value: u32,
) {
    let mut values = [[0u32; 4]; 5];

    values[4][j] = value;

    *output = values[4][j];
}
```

Then build or run the example through `cargo oxide`.

## Actual behavior

Device code generation fails with:

```text
Unsupported construct: 2-level projection
ConstantIndex { offset: 4, min_length: 5, from_end: false }
-> Index(1)
not yet implemented for assignment
```

The example compilation exits with status 101.

## Expected behavior

Assignments with a forward `ConstantIndex` followed by a runtime `Index` should be translated using the existing projected-place address traversal, just as other supported nested assignment projection combinations are.

The kernel should compile and the assigned value should be written and read back correctly.

## Additional context

Other two-level assignment projection combinations are already handled, including:

* `Deref -> Index | ConstantIndex`
* `Index -> Index`
* `Field -> Index | ConstantIndex`
* `Index | ConstantIndex -> Field`

The missing case is:

```text
ConstantIndex { from_end: false, .. } -> Index(_)
```

uurl:
Fixes #639 

## Summary

Support nested assignments whose destination place contains a forward `ConstantIndex` projection followed by a runtime `Index` projection.

This fixes device code generation for assignments such as:

```rust
values[4][j] = value;
```

## Problem

The MIR generated for the assignment above contains this two-level destination projection:

```text
ConstantIndex {
    offset: 4,
    min_length: 5,
    from_end: false,
} -> Index(1)
```

The assignment projection dispatcher did not contain an arm for this combination, causing the MIR importer to report:

```text
Unsupported construct: 2-level projection
ConstantIndex { offset: 4, min_length: 5, from_end: false }
-> Index(1)
not yet implemented for assignment
```

## Implementation

Add an assignment projection arm matching:

```rust
(
    mir::ProjectionElem::ConstantIndex {
        from_end: false, ..
    },
    mir::ProjectionElem::Index(_),
)
```

The arm delegates the complete destination place to `store_through_place_address`.

The shared projected-place address walker already supports a forward constant index followed by a runtime index, so this avoids duplicating address-calculation logic and keeps assignment handling consistent with the existing projection cases.

## Regression coverage

Extend the `nested_index_assignment` example with a kernel that:

1. Creates a nested local array.
2. Assigns through `values[4][j]`.
3. Reads the same location back.
4. Writes the result to host-visible output.

The example now validates both:

* `Index -> Index`
* `ConstantIndex -> Index`

Without the importer change, the new kernel fails during device code generation with exit status 101.

With the importer change, the example compiles and runs successfully on the GPU.

## Validation

The following checks pass:

```text
cargo fmt --all -- --check
cargo test -p mir-importer
cargo oxide run nested_index_assignment
```

Results:

```text
mir-importer unit tests: 893 passed, 0 failed
nested_index_assignment: PASS
smoketest: 178 passed, 0 failed
```

### DIFF
--- a/crates/mir-importer/src/translator/statement.rs
+++ b/crates/mir-importer/src/translator/statement.rs
@@ -670,40 +670,24 @@ pub fn translate_statement(
                         )
                     }
                     (
-                        mir::ProjectionElem::Index(_outer_index_local),
-                        mir::ProjectionElem::Index(_inner_index_local),
+                        mir::ProjectionElem::Index(_) | mir::ProjectionElem::ConstantIndex { .. },
+                        mir::ProjectionElem::Index(_) | mir::ProjectionElem::ConstantIndex { .. },
                     ) => {
-                        // `_local[i][j] = value` for nested arrays. The shared
-                        // walk-and-store path already handles chained runtime
-                        // indexes, so delegate to it instead of re-deriving the
-                        // address here. That keeps this 2-level arm from drifting
-                        // from the (Deref, Index) arm above and the N-projection
-                        // fallback below, which use the same helper. The store
-                        // target of an assignment is always a mutable place, so
-                        // the helper's mutable-address request is correct here.
-                        store_through_place_address(
-                            ctx,
-                            body,
-                            value_map,
-                            place,
-                            result_value,
-                            rvalue_op_opt,
-                            last_inserted,
-                            prev_op,
-                            block_ptr,
-                            loc,
-                        )
-                    }
-                    (
-                        mir::ProjectionElem::ConstantIndex {
-                            from_end: false, ..
-                        },
-                        mir::ProjectionElem::Index(_),
-                    ) => {
-                        // `_local[CONST][j] = value` for nested arrays. The shared
-                        // address walker already handles a forward constant index
-                        // followed by a runtime index, so delegate the complete
-                        // destination place instead of duplicating address logic.
+                        // Nested array element assignment with any mix of
+                        // runtime and constant indexes: `_local[i][j]`,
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Confirmed the gap:** 
- the walker already handled constant indexes, 
- only the assignment arm list in front of it was incomplete. 
 
- that folds your arm and the (`Index`, `Index`) arm into a single arm covering all four constant/runtime combinations (`values[i][j]`, `values[4][j]`, `values[i][2]`, `values[1][2]`), since the bodies were identical, plus two regression kernels for the newly covered shapes. 
 
**Validated on GPU:** 
- all five kernels in `nested_index_assignment` pass, 
- mir-importer suite green. 
 
Thanks for the clean fix and for following the sibling-arm pattern.

nihalpasham (approved):

PR #642, opened 2026-08-02
fix(mir-lower): support nested shared-pointer enum payloads

### LINKED ISSUE #641: Support nested shared-memory pointers in enum payload storage
## Problem

`mir-lower` cannot currently lower enum payloads whose Rust layout contains a shared-memory pointer nested inside a struct or tuple.

A direct shared-memory pointer payload can be represented by converting the pointer from CUDA shared address space (`addrspace(3)`) to generic address space for physical enum storage. However, this conversion is not applied recursively when the pointer is contained inside an aggregate payload.

For example, an enum payload with this shape fails during lowering:

```rust
struct Inner {
    pointer: *mut u32,
    cookie: u32,
}

struct Outer {
    inner: Inner,
    guard: u32,
}

enum Value {
    None,
    Some(Outer),
}
```

When `pointer` refers to shared memory, the semantic LLVM type contains a nested `ptr addrspace(3)`, while the physical enum storage requires a target-stable representation.

## Current behavior

Enum slot-map construction or payload lowering rejects aggregate payloads containing nested shared-memory pointers because the semantic aggregate type cannot be stored directly in the generic enum representation.

The existing behavior is stricter than necessary: structs and tuples have explicit, recursively traversable layouts, so their shared-pointer leaves can be converted safely.

## Expected behavior

Enum payload storage should:

1. Recursively traverse struct and tuple payloads.
2. Replace each `ptr addrspace(3)` leaf with a generic pointer in the physical storage type.
3. Emit `addrspacecast` operations when:

   * constructing the enum payload, from shared to generic;
   * extracting the enum payload, from generic back to shared.
4. Preserve the original aggregate shape, field ordering, padding, size, and alignment.
5. Continue rejecting aggregate shapes whose storage cannot yet be converted safely, including:

   * arrays containing shared-memory pointers;
   * vectors containing shared-memory pointers.

## Scope

The initial implementation should support:

* direct shared-memory pointer payloads;
* shared-memory pointers nested inside structs;
* shared-memory pointers nested inside tuples;
* arbitrary struct/tuple nesting composed from supported leaves;
* symmetric construction and extraction;
* direct-tag and niche enum layouts where applicable.

Arrays and vectors containing shared-memory pointers should remain fail-closed with an explicit diagnostic.

## Acceptance criteria

* A nested struct/tuple enum payload containing a shared-memory pointer lowers successfully.
* Enum construction converts shared-pointer leaves to generic pointers.
* Enum extraction restores generic-pointer leaves to shared address space.
* The resulting IR uses `addrspacecast`, not integer pointer round trips.
* Enum size, alignment, offsets, and padding remain identical to the rustc-recorded layout.
* Existing enum layout tests continue to pass.
* A GPU runtime regression verifies the nested pointer round trip.
* A negative example verifies that arrays containing shared-memory pointers remain unsupported.
* The implementation works through the PTX and NVVM code-generation paths.

uurl:
Closes #641

## What this fixes

Wrapping a shared-memory pointer inside a struct or tuple before putting it in an enum made the kernel fail device codegen:

```rust
struct Cursor {
    data: SharedPtr<f32>,   // pointer into CUDA shared memory
    offset: u32,
}

let direct: Option<SharedPtr<f32>> = Some(ptr);              // worked before
let nested: Option<Cursor> = Some(Cursor { data: ptr, offset: 4 });  // failed before, works now
```

This is an ordinary pattern: any handle struct that groups a shared pointer with a length, index, or flag, made optional or put in a state enum, hit this.

## Why it happens

- A shared-memory pointer does not have one fixed physical width across our output modes:
  - PTX and legacy NVVM: 64-bit
  - modern NVVM: 32-bit (`ptr addrspace(3)`)
- Enum payload storage must have one stable layout regardless of output mode, so shared pointers cannot sit in enum storage as themselves.
- The existing code substituted a target-stable storage type, but only when the pointer **was** the payload directly. One level of nesting fell into the unsupported path:

```text
Before this PR: the substitution only looked at the payload's top level.

  Option<SharedPtr<f32>>              Option<Cursor>
  =====================              ==============
  payload = SharedPtr<f32>           payload = struct Cursor
      |                                  |
      | top level IS a shared            | top level is a struct, and the
      | pointer, special case            | walker did not look inside it
      v fires                            v
  stored as a generic ptr            no storage type found
      |                                  |
      v                                  v
     OK                              kernel REJECTED

After this PR: the walk recurses through struct/tuple nesting.

  Option<Cursor>
  ==============
  payload = struct Cursor
      |-- data:   SharedPtr<f32>  -> stored as generic ptr
      |                              (addrspacecast on construct/extract)
      |-- offset: u32             -> unchanged
      v
  same size, alignment, offsets, padding as rustc's recorded layout
      |
      v
     OK in all three output modes
```

## The fix, in plain terms

- When computing an enum payload's storage type, walk the payload recursively:
  - shared pointer leaf: store it as a generic pointer (64-bit in every mode)
  - struct field or tuple element: recurse
  - bool leaf: keep the existing byte-faithful `i8` handling
  - other scalars: unchanged
- When a payload is **constructed**, convert each shared-pointer leaf to its storage form; when it is **extracted**, convert back. Both directions are plain `addrspacecast` ops, no `ptrtoint`/`inttoptr` round trip.
- This is layout-safe for structs and tuples because rustc's recorded layout already gives every pointer field a 64-bit slot: field offsets, padding, size, and alignment are unchanged by the substitution.

## The one shape still rejected, and why

Arrays (and vectors) of shared pointers inside a payload remain unsupported. For example:

```rust
struct Table {
    // an ARRAY of shared pointers, not a single one
    pointers: [&'static SharedArray<u32, 128>; 2],
}

let slot: Option<Table> = Some(Table { pointers: [a, b] });  // rejected
```

This now fails with a precise diagnostic instead of a generic failure:

```text
enum payload storage: arrays containing shared-memory pointers are not
supported; recursive target-stable storage currently supports direct
pointers and struct/tuple nesting
```

The reason it fails closed rather than recursing:

- A struct has a fixed, named list of fields, so the construct/extract coercions are a bounded set of per-field casts at known offsets.
- An array is indexed dynamically at runtime, and swapping the element type changes the element stride in the mode where the two pointer widths differ. Converting values would require per-element work at every dynamic access site, not a bounded set of casts.
- Until that is designed properly, a loud, specific error is safer than a layout that silently disagrees between output modes.

## Tests

- Unit coverage: recursive storage-type conversion, construction and extraction through nested structs, symmetric shared-to-generic and generic-to-shared casts, slot maps with nested shared-pointer leaves, rejection of arrays and of vectors.
- `addressof_sharedarray` example: runtime round-trip of the nested shape on GPU.
- New error-category example locks in the array diagnostic, wired into the smoketest.

## Validation

```text
cargo fmt --all -- --check
cargo test -p mir-lower
cargo test -p cuda-oxide-codegen
cargo clippy -p mir-lower -p cuda-oxide-codegen --all-targets -- -D warnings
```

### DIFF
--- a/crates/mir-lower/src/convert/enum_payload_storage.rs
+++ b/crates/mir-lower/src/convert/enum_payload_storage.rs
@@ -1,0 +1,309 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Target-stable physical storage for enum payload values.
+//!
+//! Rust treats CUDA pointers as one logical pointer-sized value, while modern
+//! NVVM uses a 32-bit physical representation for address-space-3 pointers.
+//! Enum storage therefore cannot retain a semantic shared pointer directly.
+//! Direct and struct/tuple-nested shared pointers are represented as CUDA
+//! generic pointers in the enum and converted at construction/extraction
+//! boundaries. Rust `bool` leaves are represented by canonical `i8` bytes at
+//! the same boundary.
+//!
+//! Arrays and vectors containing shared pointers remain fail-closed. Rebuilding
+//! an array creates one operation sequence per element, while pointer vectors
+//! have separate ABI and address-space-cast semantics. Those shapes need their
+//! own bounded/code-shape design rather than being accepted implicitly here.
+
+use llvm_export::op_interfaces::{CastOpInterface, CastOpWithNNegInterface};
+use llvm_export::ops as llvm;
+use llvm_export::types as llvm_types;
+use llvm_export::types::PointerTypeExt;
+use pliron::builtin::types::{IntegerType, Signedness};
+use pliron::context::Context;
+use pliron::irbuild::dialect_conversion::DialectConversionRewriter;
+use pliron::irbuild::inserter::Inserter;
+use pliron::op::Op;
+use pliron::result::Result;
+use pliron::r#type::{TypeHandle, Typed};
+use pliron::value::Value;
+
+#[derive(Clone)]
+struct StorageRewrite {
+    ty: TypeHandle,
+    rewrote_shared_pointer: bool,
+}
+
+enum TypeShape {
+    Bool,
+    Pointer(u32),
+    Array(TypeHandle, u64),
+    Vector(TypeHandle),
+    Struct(Vec<TypeHandle>),
+    Identity,
+}
+
+/// Return the physical LLVM type used to store one semantic enum payload.
+///
+/// The transformation is recursive through LLVM structs, which cover MIR
+/// structs and tuples after ordinary type conversion:
+///
+/// - `ptr addrspace(3)` becomes a CUDA generic pointer;
+/// - `i1` becomes the canonical one-byte `i8` memory representation;
+/// - structs are rebuilt field by field;
+/// - bool-only arrays are rebuilt recursively;
+/// - arrays/vectors containing shared pointers are rejected.
+pub(crate) fn enum_payload_storage_type(
+    ctx: &mut Context,
+    semantic_ty: TypeHandle,
+) -> std::result::Result<TypeHandle, anyhow::Error> {
+    Ok(rewrite_storage_type(ctx, semantic_ty)?.ty)
+}
+
+fn rewrite_storage_type(
+    ctx: &mut Context,
+    semantic_ty: TypeHandle,
+) -> std::result::Result<StorageRewrite, anyhow::Error> {
+    let shape = {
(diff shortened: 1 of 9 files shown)


nihalpasham:
Ran the full battery locally: 
- `addressof_sharedarray` passes on GPU including the new nested round-trip, 
- all four enum ABI/stride/provenance examples pass, 
- NVVM IR builds succeed for both the legacy (`sm_90`) and modern (`sm_120`) dialects, 
- the error example fails with the new precise diagnostic, and mir-lower tests are green. 

The recursive storage/coercion layer is the correct generalization of the old direct-pointer special case, and rejecting shared-pointer arrays with a named diagnostic instead of a generic failure is good.

nihalpasham (approved):

PR #649, opened 2026-08-03
feat: add scalar f16 and bf16 min/max intrinsics

honeyspoon:
## Overlap with the `f16` primitive

Stating this up front: 8 of the 24 are reachable from safe Rust today on nightly
with `#![feature(f16)]`, because LLVM lowers the generic min/max intrinsics
straight to these instructions:

| Rust | LLVM | PTX |
|---|---|---|
| `a.min(b)` / `a.max(b)` | `llvm.minnum.{f16,bf16}` | `min.f16` / `max.f16` |
| `a.minimum(b)` / `a.maximum(b)` | `llvm.minimum.{f16,bf16}` | `min.NaN.f16` / `max.NaN.f16` |

Verified on the pinned `llc` at `sm_80 +ptx70`. So the 4 plain and 4 `NaN` forms
duplicate instructions the language already emits; the remaining **16 have no
Rust spelling** - every `ftz` form and every `xorsign.abs` form.

Two reasons to keep the overlapping 8 rather than trim to 16:

- The intrinsic pins the exact instruction. `a.min(b)` routes through
  `llvm.minnum`, which the optimizer is free to fold, reassociate under
  fast-math, or select differently; the generated stub is a guarantee.
- It matches how the family already works. `min_f16x2` / `max_f16x2` live in
  `packed_alu` even though `llvm.minnum.v2f16` reaches `min.f16x2` generically,
  and `float::` exposes explicit-rounding arithmetic alongside plain `+` and `*`.

Happy to drop the 8 if you would rather the surface stay minimal.

Completes the `extended_minmax` family with the scalar 16-bit forms. The family already covered `f32`, `f16x2`, and `bf16x2`; LLVM also declares the one-element `f16` and `bf16` variants, and every form admitted here lowers on the pinned backend.

## What this adds

24 intrinsics:

| Format | Forms | Floor |
|---|---|---|
| `f16` | all 8 modifier combinations of `min`/`max` over `ftz`, `NaN`, `xorsign.abs` | `sm_80`/PTX 7.0, or `sm_86`/PTX 7.2 with `xorsign.abs` |
| `bf16` | the 4 combinations without `ftz` | same |

## What it deliberately excludes

The four `bf16` `ftz` declarations. LLVM declares `llvm.nvvm.fmin.ftz.bf16` and its relatives, but the NVPTX backend has no selection pattern for them:

```
LLVM ERROR: Cannot select: intrinsic %llvm.nvvm.fmin.ftz.bf16
```

They also carry no tablegen selection record in `intrinsics/imported.json`, which is the same signal from the other direction. The exclusion is recorded in `canonical_extended_minmax_variants` and in the `cuda_device::bf16` module docs so it is not mistaken for an oversight.

## Carrier

A scalar half rides as its `u16` bit pattern (`i16` in the dialect) - the convention `packed_conversion` already uses for `e4m3x2` and `e5m2x2` operands - and inline PTX uses the 16-bit `h` register class.

The lowering helper's `packed: bool` becomes a three-state `MinMaxCarrier`, so `f32`, scalar halves, and packed pairs each select their own register class instead of two of them sharing one boolean.

## ABI ids

The new variants continue at **i0830**, not i0578: the ids directly after this family's original block were reserved by other families before the family grew. `extended_minmax_abi_id` makes the discontinuity explicit, and the admission test pins both block boundaries (i0577 -> i0830).

## Naming

`min.f16` and `min.bf16` carry no modifier at all, so the identifier, source record, LLVM symbol, and selection record are now built by joining the parts that are *present* rather than by a format string with fixed separators. The construction is unchanged for all 28 pre-existing variants - the admission test still pins `min_ftz_f16x2`/i0550 first and `max_xorsign_abs_f16x2`/i0577 at index 27.

## Evidence

Measured, not asserted. Two new fixtures per backend profile:

- **Pinned LLVM 22.1.2**: all 24 typed declarations canonicalize under `llvm-as`/`llvm-dis`, lower to their exact instruction heads, and assemble under CUDA 13.3 `ptxas`. The inline-PTX route produces a **byte-identical cubin** to the typed route in each target tier.
- **CUDA 13.3 libNVVM**: the same forms compile to PTX 9.3, assemble to a cubin byte-identical to the pinned-LLVM one, and survive `-gen-lto` plus nvJitLink.

## Hardware verification

The `generated_intrinsics` example gained two things:

1. All 24 intrinsics in the existing `compile_extended_minmax` coverage kernel, so they are driven through device codegen and libNVVM verification. This is the check that catches a stale `generated_intrinsic_targets.rs`.
2. A launched `scalar_half_minmax` kernel with bit-exact host assertions, run on an **sm_86 RTX A2000**. Inputs `+1.0` and `-2.0` are asymmetric and opposite-signed so the `xorsign.abs` forms cannot pass by accident; a second launch with a NaN operand separates the plain forms from the `NaN` forms.

That test **corrected an assumption**: PTX's canonical NaN is `0x7FFF` - all-ones exponent and mantissa with a clear sign, the 16-bit analogue of the `0x7fffffff` the ISA specifies for `f32` - not the IEEE default quiet NaN (`0x7E00` for f16, `0x7FC0` for bf16). The measured value is pinned in the test and documented in both modules.

## Verification

Rebased on `a77daf85`; 2 commits, both DCO signed off.

- `cargo run -p cuda-intrinsics-gen -- check` - clean
- `probe --all --skip-terminal` - 853/853 routes; each of the 24 new ones lowers to its expected instruction
- `check-abi-history --base-ref upstream/main` - 829 base entries preserved, 853 total
- `cargo fmt --all -- --check` - clean in all three workspaces
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `RUSTDOCFLAGS='-D warnings' cargo doc` - clean
- Tests: cuda-intrinsics-gen 288, cuda-intrinsics 853, mir-importer 918, mir-lower 245, dialect-nvvm 50
- `scripts/smoketest.sh -o '^generated_intrinsics$'` - passes `--compile-only` (verified by libNVVM) and launched on hardware

Full CI is green on the fork PR (50/50 checks, including `compile all examples`): https://github.com/honeyspoon/cuda-oxide/pull/115

## Note on scope

There is no tracking-issue box for these forms - #274 covered the packed `min/max.f16x2` variants and is fully checked off. Happy to file an issue first if you would rather the design be agreed before review.

https://claude.ai/code/session_01LXYptgNnAcDSbkgyWGYQzk


nihalpasham:
Ran the full battery locally:
- `cuda-intrinsics-gen check` reports generated outputs current, and `check-abi-history` confirms all 829 base entries preserved, 853 total, so the second ABI block is append-only as documented.
- `cargo test --workspace` green.
- smoketest executed `generated_intrinsics` on sm_120: every bit-exact assertion passes, including the measured `0x7FFF` canonical NaN, and the Blackwell libNVVM variant passes too.
- checked the bf16-ftz exclusion against the PTX ISA grammar and our pinned `imported.json`: those forms have no NVPTX selection pattern, so leaving them out is right (clang actually declares those builtins and they die at instruction selection).
- on trimming the overlap: keeping all 24 is the right call. Rust has no bf16 primitive, so the real overlap is only the 4 plain/NaN f16 forms, and the intrinsic guarantees the exact instruction where the primitive goes through the optimizer.

The three-state `MinMaxCarrier` is a strictly better model than the packed boolean. Thanks for the thorough work, hardware-measured NaN detail included.

nihalpasham (approved):

Issue #131, opened 2026-06-07
match xs[i] over an array of an enum fails to import (Downcast on non-ADT type: Array)


hbtweb:
I was matching on an element of an array of enums inside a kernel — `match xs[i] { E::A(x) => ..., ... }` — and it wouldn't import:

```
Downcast on non-ADT type: Array
```

Happens with both a runtime index and a constant index.

### Reproduce

```rust
#[derive(Clone, Copy)]
pub enum E { A(u32), B(u32), C }

#[kernel]
pub fn match_runtime_index(index: u32, mut out: DisjointSlice<u32>) {
    let idx = thread::index_1d();
    if let Some(out_elem) = out.get_mut(idx) {
        let xs: [E; 4] = [E::A(7), E::B(8), E::C, E::A(100)];
        // matching an element of the array -> projection chain [Index, Downcast, Field]
        *out_elem = match xs[index as usize] {
            E::A(x) => x,
            E::B(y) => y + 1000,
            E::C   => 9999,
        };
    }
}
```

Launch with `index = 0..3`; expected `[7, 1008, 9999, 100]`. Pre-fix the kernel **fails to import** with `Downcast on non-ADT type: Array`. (A literal index — `match xs[0] { … }` — hits the `ConstantIndex` arm and fails the same way.)

### What I found digging in

<details><summary>Root cause (mir-importer / rvalue.rs)</summary>

`match xs[i]` lowers to a projection chain `[Index, Downcast, Field]`. In `translate_place_iterative`, the `Index` / `ConstantIndex` arms advanced the projected SSA *value* but never narrowed the running `current_rust_ty`, so it stayed at the outer `Array(Adt(E), N)`. The following `Downcast`/`Field` then handed that array type to `apply_enum_field_projection`, which rejected it with `Downcast on non-ADT type: Array`.

Fix that worked: narrow `current_rust_ty` to the element type on `Index`/`ConstantIndex` (a small `element_rust_ty` helper), mirroring how the `Field` arm already advances the running type.

While I was in there: `translate_place_iterative` has no `Subslice` arm, so something like `match xs[a..][i]` would hit the same stale-type class — probably worth a follow-up.
</details>

### Related
[...]

PR #652, opened 2026-08-04
Address an enum payload in memory, so a payload can be mutated in place

### LINKED ISSUE #651: Enum payloads cannot be mutated in place: no address for a Downcast projection
## Summary

Mutating an enum payload in place is not expressible. Both spellings need the address of the payload, and the importer's address walker has no arm for `Downcast`, so both fail:

```rust
enum Slot { Occupied(f32), Empty }

let mut slot = Slot::Occupied(x);

// (1) mutable borrow of a payload
if let Slot::Occupied(value) = &mut slot {
    *value *= 2.0;
}

// (2) direct assignment, which is how (1) often arrives after inlining
// (slot as Occupied).0 = x * 2.0;
```

(1) fails with

```
Unsupported construct: Rvalue::Ref: cannot compute an in-memory address for the
mutable borrow of place (((*_1) as variant#0).0) (projection [Deref,
Downcast(VariantIdx(0)), Field(0, ...)]); the value-copy fallback would silently
discard writes through the borrow
```

and (2) with

```
Unsupported construct: 2-level projection Downcast(VariantIdx(0)) -> Field(0, ...)
not yet implemented for assignment
```

## Why the refusal is right and still a gap

The first message is a deliberate refusal, not an oversight: `Rvalue::Ref` will not fall back to a value copy for a mutable borrow, because writes through the borrow would land in the copy and the original place would never change. Refusing loudly is the correct behaviour given no address can be computed. What is missing is the address.

`translate_place_addr_from_slot` marks the spot:

```rust
// Enum-variant downcast (`(x as Variant).field`). Addressing an
// enum payload in memory needs variant/niche layout machinery
// [...] This arm is the designed extension point for the
// enum-layout work tracked in issues #131/#146;
// until that lands, punt [...]
mir::ProjectionElem::Downcast(_) => return Ok(None),
```

Both #131 and #146 are closed, so the extension point is unclaimed.

## What the workaround costs

Today the payload has to be rebuilt from a matched copy:

```rust
let slot = match slot {
    Slot::Occupied(value) => Slot::Occupied(value * 2.0),
    Slot::Empty => Slot::Empty,
};
```

That is expressible, so this is a gap in what compiles rather than in what runs. It bites hardest where the borrowing form is the natural one: a helper taking `&mut T`, or any `if let ... = &mut e` over an enum a kernel is updating.

## What a solution needs

1. **An address for the payload.** A downcast moves no address of its own, since a payload shares the enum's storage; the variant only decides which field the following `Field` names.
2. **The shared-byte case.** Variants share bytes, so a payload field either has a slot of its own or its bytes are already held by a differently typed field of another variant. Both must resolve to the same storage, or a write lands somewhere the reader will not look.
3. **A loud failure for anything it cannot resolve.** Falling back to a value copy for a mutable borrow is exactly the silent miscompile the current code refuses to emit.

## Open design questions

- Should the payload address be a new operation, or does `MirFieldAddrOp` extend to an enum pointee? `all_field_types` already runs variant by variant, so one flattened index names a (variant, field) pair without a second attribute.
- Should the assignment path get its own `(Downcast, Field)` arm, or delegate to the shared walk-and-store helper that the other 2-level pairs already use?
- Is a payload behind a niche-encoded variant in scope, or should the first pass cover Direct layouts and reject the rest?

alejandro-soto-franco:
Mutating an enum payload in place needs the address of that payload, and the importer's address walker had no arm for `Downcast`, so a mutable borrow of a payload was refused outright and `(x as Variant).field = v` failed as an unimplemented projection pair. This adds the addressing and opens both paths.

Closes #651.

## The one gap behind two errors

`Rvalue::Ref` refuses a mutable borrow it cannot address rather than falling back to a value copy, because writes through the borrow would land in the copy and the original place would never change. That refusal is correct; the missing address is the defect. The assignment path failed separately with "2-level projection Downcast -> Field not yet implemented for assignment". One address fixes both.

## Changes

`all_field_types` already concatenates the variants in order, so one flattened index names a (variant, field) pair and `MirFieldAddrOp` extends to a third pointee kind beside struct and union:

```rust
/// Position of variant `variant`'s field `field` in `all_field_types`.
pub fn flat_field_index(&self, variant: usize, field: usize) -> Option<usize>
```

The walker records the variant at `Downcast` and consumes it at the `Field` that follows, since the downcast moves no address of its own. Lowering resolves the position through `build_enum_slot_map`, which already answers the part that makes enums awkward: a payload field either has an LLVM slot of its own, or its bytes are held by a differently typed field of another variant. The first is a slot GEP, the second a byte-offset GEP, and both name the same storage.

The assignment path becomes a delegation to the same walk-and-store helper the other 2-level pairs use.

This was the extension point marked in the walker for #131 and #146, both since closed.

## Design decisions

**Chosen:** extend `MirFieldAddrOp` to an enum pointee, keyed by the flattened field index. **Alternative rejected:** a dedicated payload-address operation mirroring `mir.enum_payload` in value space. It would need a second attribute for the variant and its own verifier and lowering, to reach storage the flattened index already names.

**Chosen:** the walker still returns no address when the pointee does not resolve to an enum payload position. **Alternative rejected:** treating a downcast as a zero-offset address unconditionally, which would silently address the wrong bytes for any shape the walker misreads.

## Validation

New `enum_payload_addr` example, 2^20 elements on an RTX 5060 (sm_120), CUDA 13.3. Each kernel reads its value back after mutating, from a sentinel-filled buffer, so a write into a copy shows as an unchanged element rather than passing quietly.

| kernel | path exercised | result |
|---|---|---|
| `assign_payload` | `(x as Variant).field = v` | exact match |
| `borrow_payload` | `&mut` into a `#[device]` helper, so the borrow survives a call | exact match |
| `shared_bytes` | two payload variants of different types at one offset, so the no-slot byte-offset path runs | exact match |
| `rebuild_payload` | the match-and-rebuild workaround, as a baseline | exact match |

`flat_field_index` also has a unit test covering a field past its variant's count, a variant with no fields, and an out-of-range variant.

## The caveat

This buys expressiveness, not speed. Against the rebuild-from-a-copy workaround, in-place mutation measures 0.0113 ms to 0.0105 ms over 20 runs, a ratio of 1.08 on a kernel that is bandwidth-bound either way. The value is that the borrowing form compiles at all, not that it runs faster.

## Gates

3287 workspace tests pass, `cargo clippy --workspace --all-targets -- -D warnings` clean, four device kernels verified against a host reference.


nihalpasham:
Verified the mechanism at every layer (the importer extension point, the flattened field index, the slot-map lowering) and on hardware:
- `enum_payload_addr` on RTX 5090: all six kernels exact match; full smoketest green, since the op-verifier change touches every example.
- `cargo test --workspace` green.

Pushed on top:
- a fail-closed gate for the two payload kinds enum storage keeps in converted form (bools stored as a full byte, shared-memory pointers stored generic). A write through a raw payload address stored the unconverted form; for a niche enum like `Flag { On(bool), Off }` that is a `store i1` into the very byte the tag compare reads back, which we captured in the generated .ll before gating. Both shapes now reject with a clear error, and an error-category fixture locks the diagnostic in.
- read-only borrows of those payloads keep compiling exactly as on main: the importer falls back to the sound value copy for shared borrows, and only mutation takes the gated address path.
- the byte-offset arm no longer truncates the offset with `as u32` and no longer defaults a missing slot-map entry to offset 0, which is the tag byte; plus a new kernel that genuinely exercises the byte-offset path on GPU (the existing one took the named-slot path).
- the address walker keeps the strict Downcast/Field pairing, so a raw per-variant index can never slip through as a flattened one now that the op verifier accepts enum pointees.

Closes #651. #306 (SetDiscriminant) stays open and is untouched here by design.

nihalpasham (approved):

PR #653, opened 2026-08-04
Bind the runtime row width into the disjoint slice's index space

### LINKED ISSUE #516: `index_2d_runtime` still carries an unsafe launch-uniformity obligation
`thread::index_2d_runtime` takes a runtime row stride and is `unsafe`. The
caller is responsible for that stride being identical in every thread of the
launch, which is the obligation `DisjointSlice::tile_2d32_rt` carried before
`Uniform<T>` discharged it.

The witness applies unchanged. A `Uniform<u32>` parameter proves launch
uniformity through the kernel ABI, so `index_2d_runtime` can take one and
become safe, with the same argument: under a uniform stride the coordinate to
offset mapping is injective, so distinct threads resolve distinct elements.

The change is larger than the `tile_2d32_rt` one. `index_2d_runtime` is the
indexing path that predates the proof-carrying views, so the signature change
reaches intrinsic dispatch and the macro-side coordinate routing rather than a
single method.

Until this lands, the two runtime-stride APIs discharge the same premise by
different means: one through the type system, one through an `unsafe` caller
obligation.

The witness lands in #515.

alejandro-soto-franco:
Tile disjointness needs one row width per slice for the slice's whole lifetime. A width supplied per call cannot give that, which is why #515 was closed, so this moves the row width to where device code cannot choose it.

Closes #516. Reworks the design closed in #515.

## Why no per-call witness can carry this

Uniformity is closed under arithmetic whose operands are all uniform. It is not closed under selection. Safe code can pick between two `Uniform<u32>` values on a thread-varying condition, or pass different ones at two call sites on one slice, and the selected row width then differs between threads:

```rust
let chosen = if thread_varying { wide } else { narrow };
c.tile_2d32_rt(coord, chosen)   // both values launch-uniform, the choice is not
```

Two threads that disagree about the row width describe two grids over one buffer. With 1x1 tiles, thread `(1, 0)` at width 5 and thread `(0, 5)` at width 100 both resolve flat index 5. Binding at the host boundary removes the choice: the host writes one word into the launch packet and device code has no step at which it could substitute another.

## Changes

A sealed trait gives each index space the runtime layout it needs:

```rust
pub trait SpaceLayout: sealed::Sealed {
    /// Runtime layout data marshalled with the slice, or `()` when the index
    /// space is fully described by its type.
    type Data: Copy;
}
```

`()` for the five spaces fixed in their types, `u32` for `RuntimeRowMajorTiles` and `Runtime2DIndex`. The host supplies it through `RowWidth`, or `RowWidthOwned` for the owned async launches. `tile_2d32_rt` therefore takes only a thread coordinate, and `index_2d_runtime` takes the slice it will address and is now safe, which is what closes #516.

`mir.disjoint_slice` was a dialect type of exactly two fields, so the third word could not reach the device. It now carries the index space's layout fields after `{ ptr, len }`: the verifier accepts them, the importer reads them from rustc's layout rather than from the index space's name, signature conversion flattens them into kernel parameters, and the entry prologue repacks them.

`()` is zero-sized, so every existing slice keeps its two-word layout and kernel ABI. Const assertions check that.

## Design decisions

**Chosen:** the payload lives in the slice, marshalled from the host. **Alternative rejected:** a consuming `with_row_width` on the device, binding once per slice. It reads as sound because `DisjointSlice` is neither `Copy` nor `Clone`, so it can be called at most once. The argument to that one call can still be selected on a thread-varying condition, which is the same defect one call up.

**Chosen:** read the payload from rustc's own layout of the struct. **Alternative rejected:** matching on the index space's name in the importer, which would silently miss any space added later.

## Validation

RTX 5060 (sm_120), CUDA 13.3.

| example | result |
|---|---|
| `gemm` | 538 GFLOPS, max error 1.5e-5 |
| `tiled_gemm` | 1447 GFLOPS, max error 1.5e-5 |
| `gemm_views` naive, safe against raw | 1117 against 1090 GFLOPS, bitwise equal |
| `gemm_views` tiled, safe against raw | 1350 against 1360 GFLOPS, bitwise equal |
| `async_mlp` | `RowWidthOwned` across four concurrent batches |

A new compile-fail case rejects a per-call width, so reverting this change fails a test rather than passing quietly. The eight existing cases keep their expectations, re-derived against the new signatures.

## The caveat

Leaving the entry prologue unfixed compiled and ran, and gave every thread a zero row width. Hardware caught that; the host-side suite did not. Anything else extending `mir.disjoint_slice` needs the same end-to-end check.

## Gates

3289 workspace tests pass, `cargo clippy --workspace --all-targets -- -D warnings` clean, four device examples verified against host references.

## Maintainer edits during review

Changes layered on top of the contribution during review, validated end to end on an RTX 5090 (sm_120):

- **`(row, col)` witness with per-slice resolution.** A `Runtime2DIndex` witness no longer stores a flat index. It carries the thread's raw `(row, col)` coordinates, and whichever slice it is handed to resolves them against that slice's own host-bound row width (`row * width + col`, requiring `col < width`). Witnesses minted from two slices with different widths can therefore never put two threads on one element, even when safe code selects between them under a thread-varying condition.
- **Sealed launch-packet backstop.** The `#[cuda_module]` macro picks the two-word or three-word packet from the index space's spelling, which a type alias can hide or fake. A hidden sealed trait, `__LaunchContractDisjointSliceAbi`, now bounds every generated launch method by the packet shape the macro chose, so a spelling/semantics mismatch is a compile error instead of the driver reading past the argument array. Compile-fail tests cover both directions.
- **Call-site flattening for by-value slices.** Passing a runtime-width `DisjointSlice` by value to a non-inlined device function used to drop the width word at the internal call boundary. Call lowering now flattens the index-space fields through the call ABI, symmetric with the callee signature.
- **Fallback arm and tests.** The no-history aggregate fallback in mir-lower accepts the lowered `{ ptr, i64, u32 }` slice shape and still fails closed for anything else, with a direct unit test. A new GPU example, `row_width_slice`, holds the regressions: nonzero width readback, two-width witness mixing, and the by-value call ABI.
- **Book updates.** The safety model, bounds checks, API quick reference, glossary, and async-MLP project pages now teach the safe slice-bound form everywhere the old unsafe per-call form survived.
- **Vocabulary: row width, not pitch.** The API says what the value is: `RowWidth` / `RowWidthOwned` on the host, `DisjointSlice::row_width()` on the device, and matching prose throughout. Raw CUDA driver names are unaffected.


nihalpasham:
Lands the architecture the #515 close asked for, hardened during review; everything below ran on an RTX 5090 (sm_120):
- gemm 2949 GFLOPS / tiled_gemm 9059, max error 1.5e-5 against host references; `gemm_views` safe vs raw bitwise-equal on both kernel pairs; `async_mlp` green.
- ABI stability confirmed: a fixed-width example's PTX is byte-identical to main, and runtime-width kernels carry exactly three parameters.

Pushed on top (full list in the PR description's maintainer-edits section):
- witness redesign: a `Runtime2DIndex` witness now carries the thread's `(row, col)` and every slice resolves it against its own host-bound row width. The submitted design stored a flat index baked with the minting slice's width; with two slices of widths 5 and 100, threads (1,0) and (0,5) both resolved flat index 5, and we reproduced that as real memory corruption on hardware before fixing it. Unit tests and the new `row_width_slice` GPU example replay the scenario.
- sealed launch-packet backstop, so a type alias can no longer produce a two-word packet for a three-parameter kernel; call-site flattening for by-value slices through non-inlined helpers; and the no-history fallback arm with a direct unit test.
- fmt gate fixed, the book now teaches the safe slice-bound form everywhere the old unsafe form survived, and the vocabulary moved from "pitch" to row width across the API (`RowWidth` / `RowWidthOwned`, `row_width()`).

Closes #516. Thanks for taking the #515 redesign all the way through the pipeline.

nihalpasham (approved):

PR #654, opened 2026-08-04
Give a thread a contiguous run, whole or clipped, and walk runs by grid stride

### LINKED ISSUE #583: DisjointSlice cannot express a thread owning N contiguous elements

alejandro-soto-franco:
Two of this issue's three requirements are already met on main, so this adds only the tail and the grid stride.

Closes #583.

## What already works

`LinearTiles<N>` plus `tile_thread32` gives thread `t` the elements `t * N .. t * N + N` behind a proof derived from a `ThreadIndex32`, so a thread can already own `N` contiguous elements without `get_unchecked_mut`. The issue's worked example predates that index space. Requirement 1 is covered.

## The tail

`tile_thread32` accepts only a whole tile, so the one thread whose run straddles the end of the buffer gets `None` and drops back to raw indexing. Launches round the grid up, so that thread almost always exists, and the fallback lands exactly where the arithmetic is most awkward to get right.

```rust
pub enum ThreadRunMut32<'a, T, const N: usize> {
    /// The thread owns all `N` elements.
    Full(StaticViewMut32<'a, T, N>),
    /// The parent ended part-way through the run.
    Clipped(RuntimeViewMut32<'a, T>),
}

pub fn thread_run32<'kernel>(
    &mut self,
    thread: ThreadIndex32<'kernel>,
) -> Option<ThreadRunMut32<'_, T, N>>
```

`None` now means only that a thread owns nothing at all.

## The grid stride

`grid_stride_runs32` walks every run a thread owns with the period read from `gridDim.x * blockDim.x` inside the call. A passed stride is the defect the issue describes: one that disagrees with the launch makes threads overlap silently and nothing in the signature catches it. With the period taken from the launch, every tile index a thread takes is congruent to its own index modulo the period, and `t < period` holds for a 1D launch, so distinct threads never take the same tile.

## Design decisions

**Chosen:** an enum for the tail, answering the issue's second design question. **Alternative rejected:** a single sub-view whose extent is always a runtime value. That would put a bounds branch on every thread to serve the one whose run is clipped, where the enum keeps `at_const` and its absent branch for the whole-run case.

**Chosen:** extend the `view.rs` vocabulary. `RuntimeViewMut32` is the 1D counterpart of the existing `RuntimeTileMut32`, and both follow the established shape of a view parameterised by extents, indexed by a bounded-index proof, returning an `InBounds` wrapper. **Alternative rejected:** new `Disjoint*` types on the slice, which the issue's own retrospective on #571 identifies as the wrong shape.

## Validation

RTX 5060 (sm_120), CUDA 13.3, via the new `thread_runs` example over a 4093-element buffer with a run width of 4, so the tail runs in every kernel.

| check | result |
|---|---|
| one run per thread, tail included | all 4093 elements doubled |
| grid-stride coverage | 256 threads covered all 1024 runs exactly once each |
| clipped runs | exactly one, of the expected 1 element |

Coverage is checked by value rather than by inspection: a thread that took a run twice, or none, changes the result.

## The caveat

The issue notes its earlier attempt carried no performance benefit. This one carries a cost. Against a raw twin that computes the base by hand and stores through `get_unchecked_mut`, over 2^22 elements and 20 timed runs:

| kernel | time |
|---|---|
| safe runs | 0.0633 ms |
| raw `get_unchecked_mut` | 0.0625 ms |

A ratio of 1.012. The proof and the bounds check together cost about one percent on a kernel that is bandwidth-bound, and the tail stops needing raw indexing.

`ThreadRunMut32::at` copies the base pointer and length out of the variant rather than borrowing through it, because a `&mut` through an enum downcast has no in-memory address the backend can name. That is #651, filed separately.

## Gates

3286 workspace tests pass, `cargo clippy --workspace --all-targets -- -D warnings` clean, three device kernels verified against host references.


nihalpasham:
Verified the soundness core and the hardware behavior:
- `ThreadIndex32` is only constructible from a launch context proving 1D and `grid.x * block.x <= 2^32`, so the flat id always sits below the grid-stride period; clipping only ever shortens the last run; `next_run` terminates (strictly increasing tile, `period == 0` guarded).
- `thread_runs` on RTX 5090: PASS, covering Full, Clipped, and None in one launch.
- `cargo test --workspace` and clippy clean.

I pushed two commits on top:
- the start/length arithmetic was duplicated in `thread_run32` and `next_run` with no host tests; it now lives in one helper (`checked_run_bounds`) with unit tests on every boundary. A differential harness confirmed the helper is exactly equivalent to both original copies: zero mismatches over the boundary grid plus an exhaustive small-domain sweep.
- parity and docs: `N > u32::MAX` is now rejected like the sibling APIs (it used to truncate `len()`), and `grid_stride_runs32` documents the `u32::MAX` start-offset cutoff for buffers past 2^32 elements. The added tests kill both mutants we tried (a `wrapping_mul` swap and a dropped boundary guard).

Closes #583. The `raw_parts` copy-out at the enum boundary can be revisited once #652 lands.

nihalpasham (approved):

PR #655, opened 2026-08-04
Make the warp an index space, so a warp reduction writes safely

### LINKED ISSUE #584: No safe path for writes whose uniqueness is guaranteed by the algorithm, not a ThreadIndex
## Summary

`DisjointSlice` has two access paths with nothing between them:

| | proof required | bounds checked |
|---|---|---|
| `get_mut(ThreadIndex)` | a `ThreadIndex` | yes |
| `get_unchecked_mut(usize)` | none | **no** |

An algorithm whose write uniqueness comes from somewhere *other* than a `ThreadIndex` has no index to offer, so it must drop to `get_unchecked_mut` and lose the bounds check as well as the proof — even though the bounds are often trivially checkable.

## The gap, concretely

**Warp reduction.** Only lane 0 writes, and warp indices are unique by construction. There is no `ThreadIndex` for "the warp I belong to":

```rust
#[kernel]
pub fn warp_sums(input: DisjointSlice<f32>, mut out: DisjointSlice<f32>, n: usize) {
    let gid = thread::index_1d().get();
    let mut v = if gid < n { unsafe { *input.get_unchecked(gid) } } else { 0.0 };
    v = warp::reduce_sum_f32(u32::MAX, v);

    if warp::lane_id() == 0 {
        let warp_idx = gid / 32;
        // Uniqueness holds: one write per warp, warp_idx distinct per warp.
        // But there is no ThreadIndex for warp_idx, so this is fully unchecked --
        // an out-of-range warp_idx is UB rather than a rejected write.
        unsafe { *out.get_unchecked_mut(warp_idx) = v; }
    }
}
```

**Scatter with known-unique destinations.** A permutation guarantees uniqueness, but the destination index comes from data, not from the thread index:

```rust
// `perm` is a permutation, so each dst is written exactly once -- but that fact
// lives in the caller's head, and the index is unchecked either way.
let dst = unsafe { *perm.get_unchecked(gid) };
unsafe { *out.get_unchecked_mut(dst) = value; }
```

In both cases the *disjointness* obligation is real and unavoidable. What is avoidable is throwing away the bounds check at the same time.

## Why an `unsafe` convenience method is not the answer

I proposed exactly that in #572 (now closed): `unsafe fn write(&mut self, idx, value) -> bool` and `unsafe fn read(&self, idx) -> Option<T>`, both bounds-checked, both `T: Copy`.

The argument for it was that its obligations are a strict subset of `get_unchecked_mut`'s (bounds discharged, disjointness still on the caller), so it is monotonically safer than the path callers use today. That still seems true — but it is an **escape hatch, not a fix**. It makes the unsound path slightly less unsound rather than making the sound path expressible, and it grows the API surface with a third access method whose safety contract users must read carefully to distinguish from the other two.

Notably, the returned `bool`/`Option` invites exactly the wrong reading: it looks like the safety signal and is not.

## What a real solution looks like

A **witness for uniqueness established by other means**, so these writes can be safe rather than merely checked. The warp-reduction case would mint one from "I am lane 0 of warp `w`", and the API would accept it where it accepts a `ThreadIndex` today.

The hard part, and the reason I am filing this rather than proposing it: **#515 shows where such a witness must not live.** That PR added a per-call `Uniform<T>` witness and was closed because

> safe code can select between two `Uniform<u32>` values under a thread-varying condition, or use different strides at two call sites on one slice, and two threads then hold `&mut` to the same element (stride 5 at (1,0) and stride 10 at (0,5) both resolve element 5).

So a uniqueness witness cannot be a per-call parameter either — safe code could vary it between calls on the same slice and reintroduce the aliasing. The transferable rule from that rejection is that **a witness must be bound to the domain over which the invariant holds**, and for write-disjointness that domain is the slice's whole lifetime, not one call.

Which suggests the witness belongs in the *slice type* — a `DisjointSlice` variant whose construction records how uniqueness is established (thread index, warp index, permutation) and which then hands out safe checked accessors for that scheme.

## Open design questions

- Can the warp-index case be expressed as an ordinary `ThreadIndex` over a coarser index space, making this a special case of #583 rather than a new mechanism?
- For the scatter case, is a data-dependent permutation ever provable, or is that irreducibly `unsafe`? If irreducible, the honest fix may be documentation pointing users at `get_unchecked_mut` with a worked disjointness argument, rather than a new method.
- Should `get_unchecked_mut`'s docs state the bounds obligation more prominently? Today the escape hatch is the *only* option for these patterns, and nothing signposts that.

alejandro-soto-franco:
This issue asks first whether the warp case can be an ordinary `ThreadIndex` over a coarser index space rather than a new mechanism. It can, so no uniqueness witness of a new kind is introduced.

Closes #584.

## The index space

`WarpIndex` has one element per warp. The witness is minted only for the lane that holds the completed reduction:

```rust
pub fn warp_index<'kernel>() -> Option<ThreadIndex<'kernel, WarpIndex>>
```

`Some` for lane 0 of each warp in a 1D launch, `None` for every other lane. A slice declared over the space accepts it through the ordinary `get_mut`, so the store is bounds-checked and needs no `unsafe`:

```rust
let total = warp::reduce_sum_f32(contribution);
if let Some(warp_slot) = thread::warp_index()
    && let Some(slot) = sums.get_mut(warp_slot)
{
    *slot = total;
}
```

Distinct warps get distinct indices and only one lane per warp holds a witness, so it is unique across the launch on the same terms as `index_1d`.

This also satisfies the rule #515 established. The witness is bound to the domain over which the invariant holds, which here is the slice type, not a per-call parameter that safe code could vary between calls.

## Why not divide the global thread index

The index is built from the block and the lane. `index_1d() / 32` differs whenever `blockDim.x` is not a multiple of the warp size, because the last warp of each block is then partial. At `blockDim.x = 48`, global indices 32 and 48 are both a lane 0 and both divide to 1, so two different warps would share a slot. The example below launches exactly that shape.

## Design decisions

**Chosen:** a coarser index space reusing `ThreadIndex` and `get_mut`. **Alternative rejected:** a distinct uniqueness witness accepted wherever a thread index is accepted, which the issue sketches. It would need its own type, its own minting rules and a second accessor on the slice, to express what an index space already expresses.

**Chosen:** the scatter case stays `unsafe`, and `get_unchecked_mut` now says so. **Alternative rejected:** the bounds-checked `unsafe write` of #572. That a permutation is a bijection is a fact about buffer contents, so no device-side type can carry it, and a returned `bool` reads as a safety signal it is not.

## Validation

RTX 5060 (sm_120), CUDA 13.3, via the new `warp_sums` example. 14 warps over 7 blocks of 48 threads, so every block has a partial second warp.

| check | result |
|---|---|
| warp sums against a host reference | 0 absolute error |
| slots written | every one, checked by value |
| safe against a raw twin | equal bitwise across every warp |

The raw twin derives the warp index by hand and stores through `get_unchecked_mut`, which is what such kernels write today.

## The caveat

The cost, over 16384 warps and 50 timed runs:

| kernel | time |
|---|---|
| safe warp index | 0.0095 ms |
| raw `get_unchecked_mut` | 0.0094 ms |

A ratio of 1.006. The proof and the bounds check together are within a percent of the unchecked store.

Five trybuild expectations are re-derived here. Adding a launch-contract impl changes the candidate list rustc prints in those errors, with no change to what they reject.

## Gates

3286 workspace tests pass, `cargo clippy --workspace --all-targets -- -D warnings` clean, two device kernels verified against a host reference and against each other.


nihalpasham:
The WarpIndex mechanism checks out end to end:
- re-derived the uniqueness argument independently: `div_ceil` warps-per-block (`blockDim=48`: gids 32 and 48 land in different slots, where `/32` would alias them), lane-0 gating, checked arithmetic degrading to None, and no laundering path (`WarpIndex` implements no `IndexFormula`, so `get_mut_indexed` cannot take it).
- `cargo test --workspace` green on the pinned toolchain, including the five re-derived trybuild snapshots.
- `warp_sums` on RTX 5090: exact match against the host reference.

I pushed two commits on top:
- the 48-thread blocks leave a 16-lane tail warp, but both kernels called `reduce_sum_f32`, whose contract needs all 32 lanes launched; `shfl.sync` sourcing never-launched lanes is undefined per the PTX ISA. Both twins now reduce over exactly the live lanes (member mask from the live-lane count, butterfly from `live/2`), keeping the partial-warp geometry that proves the index math. Host reference still matches exactly.
- doc fixes: the `warp_index` rustdoc called a two-argument `reduce_sum_f32` that does not exist (now the real one-arg call with the full-warp caveat inline), and `warp_index` joined the reserved-names list since the macro now claims that name inside kernel bodies.

Closes #584.

nihalpasham (approved):

Issue #656, opened 2026-08-04
smoketest: `standard` examples are not gated by compute capability (2 false PASS, 2 false FAIL, 8 failures on sm_75)


vyncint:
**Description**

`scripts/smoketest.sh` classifies an example as `standard` by default, and the `standard` contract is *execution must succeed* (`smoketest.sh:321`). Nothing in the script gates that on compute capability — each example is expected to self-declare a skip by printing `skipping:`. That self-declaration is incomplete and inconsistent, so on a pre-Ampere GPU the suite reports two examples as passing when they executed nothing, two as failing when they did exactly the right thing, and eight as hard failures.

I ran the full suite on a Tesla T4 (sm_75): **178 examples, 1515s, 11 failures**. Upstream CI compiles examples without executing them, so none of this is reachable from `ubuntu-latest`.

---

## 1. Two examples report a full `PASS` without executing

`cp_async_small` and `cp_async_zfill` guard correctly in host code, but print a capitalised marker:

```
     Running `target/oxide-examples/release/cp_async_small`
=== cp.async small-copy example (4-byte and 8-byte) ===

Skipping: cp.async requires sm_80+, device is sm_75 -- PASS (skipped)
```

smoketest verdict: **`PASS`** — not `PASS (skipped)`.

Two things go wrong in sequence:

- the skip check at `smoketest.sh:333` is `grep -qE '^[[:space:]]*skipping:'`, which is case-sensitive and does not match `Skipping:`;
- the example is then rescued by the success check at `:337`, `grep -qE 'SUCCESS|PASS|Complete'`, because its message contains the substring `PASS`.

So it is counted as a genuine execution pass, indistinguishable in the summary from an example that ran and validated results. The comment at `:329-332` says examples may gate themselves "without having to fake a success marker" — these two do exactly that, and pass only by accident. Reword the message without the word `PASS` and both would report `FAIL (no success marker)`.

## 2. Two examples report `FAIL` despite behaving correctly

**`tma_copy`** checks capability, falls back to `verify_ptx_only`, and exits 0:

```
⚠️  WARNING: TMA requires sm_90+ (Hopper or newer)
ℹ️  PTX load failed (expected on pre-Hopper): DriverError(218, "a PTX JIT compilation failed")
```

[...]

PR #658, opened 2026-08-04
fix(examples): gate arch-dependent examples on compute capability

vyncint:
Fixes the reporting half of #656.

Eleven examples in the `standard` smoketest category assume the device can run instructions above its architecture. On a Turing GPU (sm_75) that produced three distinct wrong verdicts, none of them reachable from CI, which compiles examples without executing them.

## What was wrong

**Two reported a full `PASS` having executed nothing.** `cp_async_small` and `cp_async_zfill` print a capitalised `Skipping:`, which the case-sensitive check at `smoketest.sh:333` does not match; they then satisfy the success check at `:337` through the substring `PASS` inside their own message:

```
=== cp.async small-copy example (4-byte and 8-byte) ===

Skipping: cp.async requires sm_80+, device is sm_75 -- PASS (skipped)
```

Verdict: `PASS`, not `PASS (skipped)` — indistinguishable in the summary from an example that ran and validated results.

**Two reported `FAIL` while behaving correctly.** `tma_copy` and `tma_multicast` handle an unsupported device and exit 0, but printed no recognised marker. `tma_multicast` already carried the `skipping:` marker in its driver-error arm (`main.rs:181`), directly under a comment explaining it is there so "smoketest's failure-marker scan doesn't flag this as a regression" — but the capability arm at `:154`, which is the one that actually runs on pre-Blackwell hardware, omitted it.

**Seven failed outright** with `DriverError(218, "a PTX JIT compilation failed")`: `barrier`, `future_apis`, `cvt_f16x2`, `generated_intrinsics`, `mcast_barrier_test` and `coop_groups_demo` had no capability guard, and `clc` had one that only warned before continuing into the same panic.

## What changed

Each example now takes the shape the thirteen already-correct examples use — check `compute_capability()`, print the documented lowercase `skipping:` marker, return. No harness changes; `smoketest.sh` is untouched.

Floors used: sm_80 for mbarrier (`barrier`, `future_apis`), packed f16 conversion (`cvt_f16x2`, matching the floor `cvt_packed` already gates on) and bf16 (`generated_intrinsics`); sm_90 for clusters (`coop_groups_demo`, `mcast_barrier_test`); sm_100 for CLC (`clc`).

## Verification

Tesla T4 (sm_75), driver 595.71.05, CUDA 13.2, `cargo oxide fmt --check` clean:

```
FMT_EXIT=0
[ 1/13] barrier                          ... PASS (skipped) [99s]
[ 2/13] clc                              ... PASS (skipped) [3s]
[ 3/13] cluster                          ... PASS (skipped) [2s]
[ 4/13] coop_groups_demo                 ... PASS (skipped) [5s]
[ 5/13] cp_async_small                   ... PASS (skipped) [71s]
[ 6/13] cp_async_zfill                   ... PASS (skipped) [2s]
[ 7/13] cvt_f16x2                        ... PASS (skipped) [25s]
[ 8/13] future_apis                      ... PASS (skipped) [2s]
[ 9/13] generated_intrinsics             ... PASS (skipped) [22s]
[10/13] mcast_barrier_test               ... PASS (skipped) [3s]
[11/13] tma_copy                         ... PASS (skipped) [4s]
[12/13] tma_multicast                    ... PASS (skipped) [3s]
[13/13] vecadd                           ... PASS [2s]
Fail:    0 / 13
```

`vecadd` (executes) and `cluster` (already guarded) are unchanged controls. Before this change, the same selection on the same hardware gave nine failures and the two false passes.

The before-state came from a full 178-example run on sm_75: 11 failures, listed in #656.

## Deliberately not included

Two items from #656 are left out rather than guessed at:

- **`wgmma_mma_bf16`** — it is compile-only but categorised `standard`. Moving it into `WGMMA_EXAMPLES` would not fix it: its output matches none of the non-Hopper markers at `smoketest.sh:475`, so it would fall through to `:488` and report `PASS (wgmma, executed)`, which is equally untrue. That needs a decision about whether a compile-only wgmma bucket should exist.
- **`cutile_inter_kernel`** — fails on sm_75 inside the third-party toolchain (`tileiras: Cannot find option named 'sm_75'`). I do not know cuTile's actual floor, and an invented one is worse than none.

Also unaddressed here: the `standard` success check accepts `SUCCESS|PASS|Complete` anywhere in the output, which is what let the two false passes through in the first place. Tightening it is a harness change with blast radius across every example, so I would rather do it separately if you want it.


nihalpasham:
Validated the half that needed newer hardware:
- this machine is CC 12.0, where no guard should fire: all 11 touched examples ran their real paths (zero `skipping:` lines in any log) and every success marker printed. A guard that misfired here would have silently turned a real test into a skip, so this was the important half to confirm.
- `--compile-only` over the same set mirrors the examples-compile CI job: green.
- all 12 guard strings match the harness skip pattern, and the examples stay in the standard smoketest category.
- one stale claim found for a separate follow-up, not this PR: the sm_100a `tma_multicast` PTX JIT-loads and passes on an RTX 5090, so the "consumer Blackwell does not support TMA multicast" fallback text is outdated.

Merging. 

Partially addresses #656 (the examples half; the harness-side tightening stays open).

nihalpasham (approved):

PR #659, opened 2026-08-04
feat: add packed f16x2 fma ftz and saturating forms

honeyspoon:
Completes the `packed_alu` fma modifier coverage for `f16x2`. The family already had `fma.rn.f16x2` and `fma.rn.relu.f16x2`; LLVM also declares the `ftz` and `sat` combinations, each with a single tablegen selection record, and all four lower on the pinned backend.

## What this adds

| Intrinsic | PTX | Native floor |
|---|---|---|
| `fma_ftz_f16x2` | `fma.rn.ftz.f16x2` | PTX 4.2 / sm_53 |
| `fma_sat_f16x2` | `fma.rn.sat.f16x2` | PTX 4.2 / sm_53 |
| `fma_ftz_sat_f16x2` | `fma.rn.ftz.sat.f16x2` | PTX 4.2 / sm_53 |
| `fma_ftz_relu_f16x2` | `fma.rn.ftz.relu.f16x2` | PTX 7.0 / sm_80 |

Floors are **not** collapsed across the group. The first three mirror the existing `fma_f16x2` per-backend floors exactly - PTX 6.0 / sm_70 for the pinned LLVM route, PTX 4.2 / sm_75 for libNVVM.

None of the four is reachable any other way: generic `llvm.fma.v2f16` lowers only to the plain `fma.rn.f16x2`, which already existed, and `ftz`/`sat`/`relu` have no Rust or generic-LLVM spelling. Verified on the pinned `llc`.

## Why the bf16x2 counterparts are excluded

They **do not exist in the PTX ISA**. ptxas rejects hand-written `fma.rn.ftz.bf16x2`:

```
error : Illegal modifier '.ftz' for instruction 'fma'
```

while accepting `fma.rn.relu.bf16x2` beside it, so the assembler is the authority. LLVM declares the intrinsics anyway, which is why instruction selection is merely what fails first (`Cannot select: intrinsic %llvm.nvvm.fma.rn.ftz.bf16x2`); they also carry no tablegen selection record in `intrinsics/imported.json`. The recipe comment and the `bf16x2` module docs state this, so the six-vs-two asymmetry reads as an ISA fact rather than an omission.

## Fail-closed recipe lookup

`packed_alu_recipe` returned `PackedAluRecipe` unconditionally, so an operation one format supports and another does not had nowhere to go but a panic. It now returns `Option` and `packed_bf16x2_alu_recipe` rejects the four pairs by name, so the family errors with `is outside the closed packed-ALU recipe` instead of admitting a variant that cannot lower.

`packed_alu_backend_floor` also takes the caller's already-resolved recipe rather than re-deriving it, so the new `Option` is discharged where it is created rather than behind an `expect` justified by a caller invariant.

## Evidence

Measured for both backend profiles:

- **Pinned LLVM 22.1.2** - the inline-PTX bundles lower with every exact instruction head present and assemble under CUDA 13.3 ptxas. Independent typed calls to the `llvm.nvvm` symbols emit the same heads.
- **CUDA 13.3 libNVVM** - the same forms compile to PTX 9.3, assemble, and survive `-gen-lto` plus nvJitLink.

## Hardware verification

The launched `f16x2_arith` example goes from 9 to 14 checked operations on an **sm_86 RTX A2000**, with cases chosen to discriminate rather than merely execute:

- A subnormal input pair (`0x0002_0001`) is **preserved** by `fma_f16x2` and **flushed to zero** by `fma_ftz_f16x2` and `fma_ftz_relu_f16x2` - direct proof the `ftz` modifier acts, not just that the stub compiles.
- `2*3+7 = 13` and `4*5+11 = 31` both clamp to `1.0` under the saturating forms.

All 14 pass bit-exactly.

## What I could not verify

Two gaps, stated rather than left to be found:

- **Assembly is checked at sm_75, not at the admitted sm_70 floor.** CUDA 13.3 ptxas no longer accepts sm_70, so the evidence records assembly at sm_75, as the existing `fma_f16x2` record already does for the same reason.
- **Only sm_86 was exercised on hardware.** No Blackwell or Hopper run.

## Verification

Rebased on `e28248c1`; 3 commits, all DCO signed off.

- `cargo run -p cuda-intrinsics-gen -- check` - clean
- `probe --all --skip-terminal` - 857/857 routes; all four new ones lower to their expected instruction
- `check-abi-history --base-ref upstream/main` - 853 base entries preserved, 857 total (this PR uses i0854-i0857)
- `cargo fmt --all -- --check` - clean in all three workspaces
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `RUSTDOCFLAGS='-D warnings' cargo doc` - clean
- Tests: cuda-intrinsics-gen 288, cuda-intrinsics 857, dialect-nvvm 50, mir-lower 245
- `scripts/smoketest.sh -o '^f16x2_arith$'` - launched on hardware

Full CI is green on the fork PR (50/50 including `compile all examples`): https://github.com/honeyspoon/cuda-oxide/pull/116

Follows #649. Independent of it - different family and contract, no shared code, non-overlapping ABI ids.

https://claude.ai/code/session_01LXYptgNnAcDSbkgyWGYQzk


nihalpasham:
Ran the full battery locally (sm_120):
- `cargo test --workspace`: 3294 passed. `check-abi-history` vs main: 853 base entries preserved, 857 total, append-only holds.
- normalized compare of the catalog and both evidence JSONs: 0 existing records modified, exactly the 4 new ones. The 866 modified probe files change only the catalog SHA header (scripted over all of them, not sampled).
- all 4 new forms probed through llc and the real ptxas 13.3 (binary SHA matches your recorded evidence): emitted heads `fma.rn.{ftz,sat,ftz.sat,ftz.relu}.f16x2` confirmed at sm_70/sm_80.
- `f16x2_arith` on GPU: all 6 fma variants exact against the host oracle, including the subnormal-flush and saturation checks.
- checked the bf16x2 exclusion against real ptxas: `fma.rn.ftz.bf16x2` and `fma.rn.sat.bf16x2` both rejected with `Illegal modifier`, while the `fma.rn.relu.bf16x2` control assembles. Your final commit documenting that at the exclusion site is exactly right; the pinned LLVM declares those intrinsics but has no selection patterns, so they would die at instruction selection.

Merging.

nihalpasham (approved):

PR #662, opened 2026-08-05
ci: verify dependency-licenses.csv covers what the workspace declares

### LINKED ISSUE #661: dependency-licenses.csv is unverified and missing 8 direct dependencies and 13 workspace members
## Summary

`dependency-licenses.csv` is the "tracked license inventory" the README points at, and CONTRIBUTING asks contributors to update it by hand when adding a dependency. Nothing verifies it, and it has fallen behind: **8 of 28 directly declared third-party dependencies and 13 of 23 workspace members have no row.**

Missing third-party direct dependencies:

| Crate | Kind | Declared by |
|---|---|---|
| `clap` | normal | cargo-oxide |
| `futures` | normal | cuda-async |
| `object` | normal | oxide-artifacts |
| `serde` | normal | cuda-intrinsics-gen |
| `serde_json` | normal | cargo-oxide, cuda-intrinsics-gen |
| `sha2` | normal | cargo-oxide, cuda-artifact-finalizer, cuda-host, cuda-intrinsics-gen |
| `toml` | normal | cargo-oxide, cuda-intrinsics-gen |
| `trybuild` | dev | cuda-core, cuda-device, cuda-macros |

Missing workspace members: `cargo-oxide`, `cuda-artifact-finalizer`, `cuda-async`, `cuda-intrinsics`, `cuda-intrinsics-gen`, `cuda-oxide-codegen`, `fuzzer`, `libnvvm-sys`, `mir-transforms`, `nvjitlink-sys`, `nvvm-transforms`, `oxide-artifacts`, `reserved-oxide-symbols`.

Dev-only dependencies are in scope by the file's own precedent: `expect-test` is dev-only and *is* recorded, so `trybuild`'s absence is an omission by the repo's existing standard rather than a scope question. Likewise `tempfile` was added along with its row in #417, so the convention works when remembered — these 21 are the cases where it wasn't.

## Proposed fix

Add `scripts/check-dependency-licenses.sh`, wire it into `cargo-deny.yml` as a second job, and add the 21 missing rows.

The two halves then line up: `cargo deny check` enforces the license *policy* over the whole resolved graph from `deny.toml`, and this guard enforces that the human-readable inventory still lists what the workspace declares.

## Where I deliberately drew the scope lines

**Presence, not versions.** 24 of the 81 existing rows already disagree with the resolved graph (`anyhow` 1.0.100 vs 1.0.103, `syn` 2.0.114 vs 3.0.3, and so on). That is ordinary `Cargo.lock` drift. Enforcing versions would fail on every routine bump while saying nothing about licensing, and would need a CSV edit for each one. I did not "fix" those 24 numbers either — they would go stale again immediately.

**Direct dependencies, not the full transitive graph.** The resolve graph has 160 packages; the CSV records 81, of which 50 are transitive-only. Requiring all 109 transitive crates would be a large mechanical change for no licensing gain, because `deny.toml`'s `[licenses]` allow-list already covers every transitive crate. The guard therefore requires workspace members plus direct dependencies, and treats existing transitive rows as welcome but optional.

**No stale-row check.** One row, `libmathdx-sys`, is not in the root workspace graph at all — it is the first-party sys crate under `examples/mathdx_ffi_test/`. Since the CSV legitimately covers first-party crates from outside the root workspace, a "no extra rows" rule would flag it wrongly. The guard only requires presence.

## Two observations I am not fixing here

1. **Vendored `rustlantis` has no automated license coverage.** `crates/fuzzer/rustlantis/` is tracked third-party source, dual MIT/Apache-2.0, carrying its own `LICENSE-APACHE` and `LICENSE-MIT` (so CONTRIBUTING's vendored-file rule is satisfied). But `crates/fuzzer/Cargo.toml` sets `exclude = ["rustlantis/"]` and it is a separate workspace, so cargo-deny never sees it and the CSV does not record it. It is the one piece of in-tree third-party code with no automated license coverage. Recording it would need a new `Dependency Type` value such as `3rd Party (vendored)`, which is a schema decision for you rather than something I should invent. It is also not linked into any artifact, so its license does not propagate into distributed binaries.
2. **The 24 stale versions** above, if you would rather they were refreshed as a one-off.

## Verification

CPU-only; no GPU, no CUDA toolkit, no LLVM — `cargo metadata` reads manifests without running build scripts.

|---|---|---|---|
| 1 | Guard against the CSV as it is on `main` | fail | exit 1, lists exactly the 21 crates above |
| 2 | Guard after adding the 21 rows | pass | exit 0, "records all 51 declared crates" |
| 3 | Delete one third-party row (`sha2`) | fail | exit 1, names `sha2` |
| 4 | Delete one workspace-member row (`oxide-artifacts`) | fail | exit 1, names `oxide-artifacts` |
| 5 | `python3` absent from `PATH` | fail | exit 1, "refusing to report success from a check that cannot run" |
| 6 | `cargo` absent from `PATH` | fail | exit 1, same |
| 7 | CSV reduced to its header (layout change) | fail | exit 1 via the parse self-test |
| 8 | All tools present, unmodified CSV | pass | exit 0 |

The fail-closed behaviour in 5–7 is deliberate, following #643: that guard reported success for three months because `rg ... \|\| true` turned a missing binary into an empty result set.

The CSV edit is insert-only — `git diff --numstat` is `21 0`, no existing row touched — and CRLF line endings are preserved (103/103 lines still carry `\r`), which is the trap that produced a whole-file diff in #631. All 102 rows still parse as 9-field CSV, and every new row's license intersects `deny.toml`'s allow-list.

`scripts/check-error-example-status.sh` and the reserved-prefix guard both still pass. The change touches four files and overlaps none of the seven open PRs.

vyncint:
Fixes #661. Branched from `main`; touches four files and overlaps none of the seven open PRs (checked file-by-file against #650, #652, #653, #654, #655, #658, #659).

## What was wrong

`dependency-licenses.csv` is the "tracked license inventory" the README points at, and CONTRIBUTING asks contributors to update it by hand. Nothing verified it, and it had fallen behind by **8 of 28 directly declared third-party dependencies** and **13 of 23 workspace members**.

| Crate | Kind | Declared by |
|---|---|---|
| `clap` | normal | cargo-oxide |
| `futures` | normal | cuda-async |
| `object` | normal | oxide-artifacts |
| `serde` | normal | cuda-intrinsics-gen |
| `serde_json` | normal | cargo-oxide, cuda-intrinsics-gen |
| `sha2` | normal | cargo-oxide, cuda-artifact-finalizer, cuda-host, cuda-intrinsics-gen |
| `toml` | normal | cargo-oxide, cuda-intrinsics-gen |
| `trybuild` | dev | cuda-core, cuda-device, cuda-macros |

Plus `cargo-oxide`, `cuda-artifact-finalizer`, `cuda-async`, `cuda-intrinsics`, `cuda-intrinsics-gen`, `cuda-oxide-codegen`, `fuzzer`, `libnvvm-sys`, `mir-transforms`, `nvjitlink-sys`, `nvvm-transforms`, `oxide-artifacts`, `reserved-oxide-symbols`.

Dev-only deps are in scope by the file's own precedent — `expect-test` is dev-only and *is* recorded — so `trybuild`'s absence is an omission by the existing standard, not a scope question. And `tempfile` arrived with its row in #417, so the convention works when remembered; these 21 are where it wasn't.

## What this adds

`scripts/check-dependency-licenses.sh`, run from `cargo-deny.yml` as a second job, plus the 21 rows. The two jobs then cover the two halves of one concern: `cargo deny check` enforces the license *policy* over the whole resolved graph from `deny.toml`; this guard enforces that the human-readable inventory still lists what the workspace declares.

## Where I drew the scope lines, and why

**Presence, not versions.** 24 of the 81 existing rows already disagree with the resolved graph (`anyhow` 1.0.100 vs 1.0.103, `syn` 2.0.114 vs 3.0.3, …). That is ordinary `Cargo.lock` drift. A version check would fail on every routine bump while saying nothing about licensing. I also deliberately did **not** refresh those 24 numbers — they would go stale again immediately, and that churn is the reason not to enforce them.

**Direct dependencies, not the full graph.** The resolve graph has 160 packages; the CSV has 81, of which 50 are transitive-only. Requiring all 109 transitive crates would be a large mechanical change for no licensing gain, since `deny.toml`'s `[licenses]` allow-list already covers every transitive crate. Existing transitive rows stay welcome but optional.

**No stale-row rule.** `libmathdx-sys` is not in the root workspace graph — it is the first-party sys crate under `examples/mathdx_ffi_test/`. Because the CSV legitimately covers first-party crates from outside the root workspace, a "no extra rows" rule would flag it wrongly.

## Verification

CPU-only. No GPU, no CUDA toolkit, no LLVM — `cargo metadata` resolves the graph without running build scripts, which is why the new job needs neither.

|---|---|---|---|
| 1 | Guard against the CSV as it is on `main` | fail | exit 1, lists exactly those 21 crates |
| 2 | Guard after adding the rows | pass | exit 0, "records all 51 declared crates" |
| 3 | Delete a third-party row (`sha2`) | fail | exit 1, names `sha2` |
| 4 | Delete a workspace-member row (`oxide-artifacts`) | fail | exit 1, names `oxide-artifacts` |
| 5 | `python3` absent from `PATH` | fail | exit 1, "refusing to report success from a check that cannot run" |
| 6 | `cargo` absent from `PATH` | fail | exit 1, same |
| 7 | CSV reduced to its header | fail | exit 1 via the parse self-test |
| 8 | All tools present, unmodified CSV | pass | exit 0 |

5–7 are deliberate, following #643: that guard reported success for three months because `rg ... || true` turned a missing binary into an empty result set.

Two details worth calling out because they are easy to get wrong here:

- The CSV edit is **insert-only** — `git diff --numstat` is `21 0`, no existing row modified — and rows land in the file's established `LC_ALL=C` order without reordering anything.
- **CRLF preserved** (103/103 lines still carry `\r`). Writing this file in text mode is what produced a whole-file 81-row diff in #631, so the insertion used binary I/O.

All 102 rows still parse as 9-field CSV, every new row's license intersects `deny.toml`'s allow-list, `bash -n` is clean, and both sibling guards (`check-error-example-status.sh`, reserved-prefix) still pass.

## Two things I found but did not fix

1. **Vendored `rustlantis` has no automated license coverage.** `crates/fuzzer/rustlantis/` is tracked third-party source, dual MIT/Apache-2.0, carrying its own `LICENSE-APACHE` and `LICENSE-MIT` — so CONTRIBUTING's vendored-file rule is satisfied. But `crates/fuzzer/Cargo.toml` sets `exclude = ["rustlantis/"]` and it is a separate workspace, so cargo-deny never sees it and the CSV does not record it. It is the only in-tree third-party code with no automated license coverage. Recording it needs a new `Dependency Type` value such as `3rd Party (vendored)` — a schema decision for you, not one I should invent. It is also not linked into any artifact, so its license does not propagate into distributed binaries.
2. **The 24 stale versions**, if you would prefer a one-off refresh.

Happy to move the job out of `cargo-deny.yml` into its own guard workflow (alongside `naming-guard` / `status-guard`) if you would rather keep that file strictly about cargo-deny — I put it there because the concern is dependency licensing and it gets the weekly schedule for free.


nihalpasham:
Verified this end to end:
- replayed the guard against main: it reports exactly the 21 missing crates from #661; against this branch it reports none, with the required set at the expected 51.
- row data checked: all 13 workspace-member rows match their crates' `Cargo.toml`, all 8 third-party versions match `Cargo.lock`, and every license intersects `deny.toml`'s allow-list.
- reproduced a locale failure: on stock en_US.UTF-8, `comm` rejects the C-sorted lists (`file 2 is not in sorted order`) and the script aborts even with a complete CSV. CI's C.UTF-8 was unaffected, but CONTRIBUTING points people at running this locally.

I pushed one commit on top:
- `export LC_ALL=C` right after `set -euo pipefail`, so every command in the script sorts and compares by the same byte-wise rules on any machine. 
- Re-verified: passes under both locales with the complete CSV, and still fails loudly naming the crate when a row is deleted.

Follow-ups that can be filed separately: `--locked` on the `cargo metadata` call, and a second pass over `crates/rustc-codegen-cuda` (its own workspace, invisible to both cargo-deny and this guard today).

Thanks for closing the loop on the inventory. Closes #661.

nihalpasham (approved):

Issue #663, opened 2026-08-05
cargo-deny and the license inventory both stop at the root workspace: 68 third-party crates in example workspaces are unenforced


vyncint:
**Description**

`cargo deny check` resolves the **root workspace**, and the guard added in #662 checks the same graph against `dependency-licenses.csv`. Every example under `crates/rustc-codegen-cuda/examples/` is its own `[workspace]`, so neither sees any of them. Six examples declare third-party dependencies, pulling **68 distinct crates that are in neither the root dependency graph nor the CSV** — and nothing enforces the license policy over any of them.

This is not a violation today, and I want to be precise about that: `deny.toml`'s allow-list (`MIT`, `Apache-2.0`, `Apache-2.0 WITH LLVM-exception`, `BSD-2-Clause`, `BSD-3-Clause`, `ISC`, `Unicode-3.0`, `Zlib`) plausibly covers all 68, though I have not resolved each crate's license to confirm it. The problem is that **if it did not, no gate would say so**.

**Where the 68 come from**

| Example | Uncovered crates | Direct declarations |
|---|---|---|
| `cutile_inter_kernel` | 62 | `cutile`, `cutile-compiler`, `cutile-cuda-core` (git), `tokio` |
| `hashmap_v2` | 5 | `rayon` |
| `hashmap_v3` | 5 | `rayon` |
| `async_mlp` | 2 | `tokio` |
| `async_vecadd` | 2 | `tokio` |
| `libm_math` | 1 | `libm` |

`cutile_inter_kernel` dominates because the `cutile-rs` git dependency pulls a large transitive graph — `wasm-bindgen`, `wit-bindgen`, `wasmparser`, `windows-targets`, `parking_lot`, `rand`, `phf`, and so on. The others are small: `rayon` brings `crossbeam-*`, `tokio` brings `mio`/`socket2`.

`half` and `hashbrown` are *not* in this list — they already appear in the root graph, so they are covered.

**Why this is in scope**

#662 established that the CSV legitimately records crates from outside the root workspace: `libmathdx-sys` lives under `examples/mathdx_ffi_test/` and has a row. So "examples are out of scope" is not the existing convention.

These crates are also not hypothetical — they are compiled by a documented workflow. `cargo oxide run cutile_inter_kernel` builds all 62, and `scripts/smoketest.sh` builds them on every full run.

**Expected behavior**

[...]

PR #664, opened 2026-08-05
ci: extend the license inventory guard to the example workspaces

vyncint:
Addresses #663.

`cargo deny check` resolves the root workspace, and the guard from #662 checks that same graph. Every example under `crates/rustc-codegen-cuda/examples/` sets its own `[workspace]`, so neither sees any of them — six examples declare third-party dependencies, pulling **68 distinct crates** recorded in neither place, with no license gate over any of them.

Not a violation today: `deny.toml`'s allow-list plausibly covers all 68. The problem is that nothing would say so if it did not.

## What the guard now does

For each example lock file, every package must be covered — in the root graph, in the CSV, or carrying no `source` field.

That last case is a path dependency, first-party by construction. **Name matching is avoided deliberately:** `cutile_inter_kernel` pulls `cuda-core`, `cuda-async` and `cuda-bindings` by *git* rather than by path, so a heuristic over crate names would misfile first-party crates as third-party. I only found that by checking, having assumed the opposite.

Lock files are parsed directly rather than via `cargo metadata` — resolving 178 workspaces would be slow and would need the network. The existing `cargo-deny / license-manifest` job already runs this script and stays CPU-only, so there is no workflow change.

It also skips the 172 examples that declare only path dependencies, so in practice this checks 6 workspaces.

## Rows added

Eight crates, for the five small examples:

| Crate | Version | License | Pulled by |
|---|---|---|---|
| `tokio`, `tokio-macros` | 1.52.3, 2.7.0 | MIT | `async_mlp`, `async_vecadd` |
| `rayon`, `rayon-core` | 1.12.0, 1.13.0 | MIT OR Apache-2.0 | `hashmap_v2`, `hashmap_v3` |
| `crossbeam-deque`, `-epoch`, `-utils` | 0.8.6, 0.9.18, 0.8.21 | MIT OR Apache-2.0 | via `rayon` |
| `libm` | 0.2.16 | MIT | `libm_math` |

Every license, repository and description is read from that crate's own manifest rather than transcribed from memory — `libm` is plain `MIT`, not the dual license the others use, which is exactly the sort of thing worth not guessing in a license file.

Following #631 and #662: the CSV edit is **insert-only** (`git diff --numstat` is `8 0`, no existing row modified), rows land in the file's existing `LC_ALL=C` order, and CRLF endings are preserved via binary I/O.

## The one thing left open

`cutile_inter_kernel` sits in `INVENTORY_EXEMPT_EXAMPLES`. Its git dependency on `cutile-rs` resolves ~60 further crates — `wasm-bindgen`, `wit-bindgen`, `wasmparser`, `windows-targets` — that exist in this tree only to build one interop example. Whether those belong in the inventory is the question in #663, and I would rather leave a documented, single-line lever than add 60 rows on a guess. **Delete that one entry and the guard requires the rows instead.**

Names in that list are validated against the examples on disk, so a typo or a rename fails the run rather than quietly exempting nothing — or everything. Given that this PR exists because a guard's coverage stopped short of what it appeared to check, an exemption that could silently widen seemed worth closing off.

## Verification

CPU-only; no GPU, CUDA toolkit or LLVM.

|---|---|---|---|
| 1 | Guard against the CSV as it is on `main` | fail | exit 1, lists the 6 examples and 68 crates |
| 2 | After adding the rows | pass | exit 0, both halves report OK |
| 3 | Delete a new row (`rayon`) | fail | exit 1, names `rayon` |
| 4 | Drop the exemption | fail | exit 1, names `cutile_inter_kernel` |
| 5 | Typo the exemption | fail | exit 1, "names no such example: cutile_inter_kernl" |
| 6 | Lock glob matches nothing | fail | exit 1, "parse self-test failed: found 0 example lock files" |

3–6 follow the #643 lesson, and #662's locale bug: a guard that cannot fail loudly is the defect, not the coverage gap.

`bash -n` clean, the root-workspace half still reports its 51 crates unchanged, and `CONTRIBUTING.md` gains four lines so the rule is discoverable before CI enforces it.

## Not included

`crates/rustc-codegen-cuda` is its own workspace too, and was the follow-up suggested on #662 — but I measured it and its 93-package graph is a **strict subset** of the root's 160, so `cargo deny check` already covers every third-party crate the shipped backend uses. Only the first-party `rustc_codegen_cuda` crate itself lacks a row. Nothing to enforce there; noted in #663 rather than changed here.


vyncint:
Pushed `a10bb84d`: `--locked` on the `cargo metadata` call, which you suggested reviewing #662. It seemed better folded in here than sent as a follow-up, since it touches the same line this PR already moves — a separate PR would just conflict.

Verified `cargo metadata --locked --format-version 1` exits 0 against the committed lockfile, and the guard still reports both halves OK.

Your other #662 follow-up — a second pass over `crates/rustc-codegen-cuda` — I measured and did **not** act on: its 93-package graph is a strict subset of the root's 160, so `cargo deny check` already covers every third-party crate the shipped backend uses. Only the first-party `rustc_codegen_cuda` crate itself lacks a CSV row. Noted in #663 rather than changed here, since there is nothing to enforce.

nihalpasham:
The guard architecture is consistent with the #662 first half (presence-only, no network, same CI job) and the design choices held up under review — first-party by absence of a lock source field, exemption list validated against disk, parse self-tests against the silent-blindness failure mode. I pushed one commit on top:

- the lockfile search is now recursive with each lock attributed to its top-level example: `cutile_inter_kernel/simt/Cargo.lock` (59 packages) already exists below the old depth-1 glob, so a future nested sub-workspace would have escaped silently; also dropped the stale "178 workspaces" count.
- both halves re-run clean locally.
- The `cutile_inter_kernel` exemption stays a single documented line; whether those ~60 rows belong in the inventory remains #663's question.

nihalpasham (approved):

PR #666, opened 2026-08-05
fix(examples): future_apis needs sm_90, not sm_80

vyncint:
Follow-up to #658, which I got wrong for this one example.

## What was wrong

#658 gated `future_apis` at sm_80, reasoning from its `mbarrier_{init,arrive,test_wait,inval}` calls. Those really are sm_80 instructions and they are the only arch-sensitive ones the emitted PTX contains — but the kernels also construct `ManagedBarrier<_, TmaBarrier>`, and the backend detects that as a TMA feature. Asking for anything lower is refused outright:

```
CUDA target sm_86 cannot lower detected feature Tma + Sm80; cuda-oxide
requires a target compatible with sm_100 for this module
```

Under an auto-detected target the build instead succeeds at `.target sm_100`, and the module then fails to load. So the too-low floor did not present as a clean skip — it presented as `DriverError(218, "a PTX JIT compilation failed")` on **every pre-Blackwell device**, Ampere and Hopper alike, not merely the pre-Ampere hardware #658 was aimed at.

Both compiler behaviours are the designed ones: fail closed on an explicit low `--arch`, raise the target when it is auto-detected. Only the floor in the example was wrong.

## The change

`major < 8` → `major < 10`, matching `clc` and `tma_multicast`, with the compiler's own diagnostic quoted in the comment so the floor is not re-derived from the visible instructions next time. That was exactly my mistake: the instructions in the PTX say sm_80, and they are not what sets the requirement.

## Verification

NVIDIA A10G (`compute_cap 8.6`):

```
[ 1/ 2] future_apis                      ... PASS (skipped)
[ 2/ 2] vecadd                           ... PASS
Pass:    2 / 2
```

```
=== Future APIs Test (Unified) ===

skipping: TMA-typed barriers require sm_100+ (device is sm_86)
```

Before this change, the same example on the same device failed with `DriverError(218)` — including after `touch crates/*/src/lib.rs` to force a full backend rebuild, so it was not a stale-artifact effect. `cargo fmt --check` clean.

## How this went unnoticed in #658

@nihalpasham validated #658 on CC 12.0, where sm_100 ≤ the device and the example runs its real path correctly — so nothing there could have shown it. I validated on sm_75, where the guard fires and the example skips, which also hides it. The failure only appears in the band between the two, on Ampere and Hopper, and that is where I found it: running the full suite on an A10G, which sits inside the documented `sm_80+` support floor.

Worth noting for the target-selection discussion in #656: `Tma + Sm80` resolving to a requirement of sm_100 rather than sm_90 is the part I do not understand, since TMA itself is Hopper. I have not investigated whether that mapping is intentional and am not proposing a change to it here — this PR only corrects the example's floor to match whatever the backend requires today.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/future_apis/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/future_apis/src/main.rs
@@ -244,10 +244,17 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let ctx = CudaContext::new(0)?;
 
-    // The mbarrier APIs exercised below require sm_80 (Ampere) or later.
+    // sm_100, not sm_80. The mbarrier calls here would run on Ampere, but the
+    // kernels also construct `ManagedBarrier<_, TmaBarrier>`, and the backend
+    // detects that as a TMA feature: building for anything lower is refused
+    // with "CUDA target sm_86 cannot lower detected feature Tma + Sm80;
+    // cuda-oxide requires a target compatible with sm_100 for this module".
+    // Under an auto-detected target the build instead succeeds at sm_100 and
+    // the module then fails to load, so a lower floor here reads as
+    // DriverError(218) on every pre-Blackwell device.
     let (major, minor) = ctx.compute_capability()?;
-    if major < 8 {
-        println!("skipping: mbarrier requires sm_80+ (device is sm_{major}{minor})");
+    if major < 10 {
+        println!("skipping: TMA-typed barriers require sm_100+ (device is sm_{major}{minor})");
         return Ok(());
     }
 
--- a/crates/rustc-codegen-cuda/examples/future_apis/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/future_apis/src/main.rs
@@ -244,17 +244,21 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let ctx = CudaContext::new(0)?;
 
-    // sm_100, not sm_80. The mbarrier calls here would run on Ampere, but the
-    // kernels also construct `ManagedBarrier<_, TmaBarrier>`, and the backend
-    // detects that as a TMA feature: building for anything lower is refused
-    // with "CUDA target sm_86 cannot lower detected feature Tma + Sm80;
-    // cuda-oxide requires a target compatible with sm_100 for this module".
-    // Under an auto-detected target the build instead succeeds at sm_100 and
-    // the module then fails to load, so a lower floor here reads as
-    // DriverError(218) on every pre-Blackwell device.
+    // sm_90, not sm_80. `ManagedBarrier::init_by` emits `fence.proxy.async`
+    // for every barrier kind (the TmaBarrier typestate parameter is inert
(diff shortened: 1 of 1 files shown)


vyncint:
Heads-up for review: **the description above documents the superseded first revision.** The head of this branch is `8615368f`, which corrects the floor to **sm_90**, so the title (now fixed) and the `sm_100` reasoning and skip-message output quoted in the body no longer match the diff. The code comment on the branch is the accurate account.

I confirmed the sm_90 floor independently today, arriving at it from the other direction, and it can be pinned with no GPU by rewriting only the `.target` line of the generated PTX:

```bash
cd crates/rustc-codegen-cuda/examples/future_apis
for arch in sm_86 sm_90; do
  sed "s/^\.target .*/.target $arch/" future_apis.ptx > /tmp/fa_$arch.ptx
  printf '%-8s : ' "$arch"; ptxas -arch=$arch /tmp/fa_$arch.ptx -o /dev/null 2>&1 | head -1; echo
done
```

```text
sm_86    : ptxas fatal   : Modifier '.async' requires .target sm_90 or higher
sm_90    :
```

The module contains no actual TMA — `cp.async.bulk.`, `expect_tx`, `.sync_restrict`, `tcgen05` and `clusterlaunchcontrol.` are all absent — so the proxy fence emitted next to every `mbarrier.init` is the sole trigger for both the sm_90 requirement and the `Tma` classification.

I also verified on an A10G (sm_86, driver 595.71.05) that the `major < 9` gate takes the skip path and exits 0, so the example reports `PASS (skipped)`. That was via #677, which made the identical change before I realised this PR already existed; I have closed it as a duplicate. The skip message string differs between the two, so I am not quoting output as if it came from this branch.

**One question**, which I raised on #676 before closing that as a duplicate too: `ManagedBarrier::init_by` emits `fence.proxy.async` for *every* `Kind`. Its purpose is to make generic-proxy writes visible to the async proxy, so a `ManagedBarrier<_, GeneralBarrier>` that never feeds an async copy plausibly does not need it. If the fence can be predicated on `Kind`, plain barriers would keep working on Ampere exactly as the original sm_80 comment intended, instead of the whole API being Hopper+. I have not attempted it because I do not want to guess at the ordering requirement. Worth a separate issue, or is the unconditional fence deliberate?

nihalpasham:
Good catch on the symptom; the floor needed one correction, which I pushed on top:

- the guard is `major < 9`, not `< 10`: the requirement is `fence.proxy.async` (sm_90; ptxas rejects it below that), emitted by `ManagedBarrier::init_by` for every barrier kind — the `TmaBarrier typestate` is inert `PhantomData` and detects nothing. 
- On a Hopper box the device hint resolves the target to sm_90 and the example genuinely runs, so skipping below sm_100 would have silently dropped coverage on exactly the generation these APIs were built for.

**Verified on RTX 5090 (sm_120):**
- all five tests execute and pass; on pre-Hopper devices the guard now exits with a clean skip instead of DriverError(218).

Closes #676. 

The compiler-side question it leaves open (a bare sm_90 fence classified as Tma with an sm_100 no-hint default) is split into its own issue.

nihalpasham (approved):

Issue #657, opened 2026-08-04
vectorization: f32x4_view_copy accesses lower to generic space, not .global


vyncint:
**Description**

The `vectorization` example inspects its own generated PTX at runtime and asserts that `f32x4_view_copy` performs 128-bit `ld.global`/`st.global` transactions. It does not, and the example fails.

The copy *is* fused to 128 bits — but in the **generic** address space rather than global, so the example's `ld.global` search finds nothing. There is also a local-memory round trip in constructing the view.

**Reproducer**

```
cargo oxide run vectorization      # fails the assertion
cargo oxide build vectorization --arch sm_75   # PTX alone is enough to see it
```

**Actual behavior**

Runtime report:

```
f32x4       float4           16    16  ld.global.v2.b64       yes
f64x4       double4          32    16  ld.global.v2.b64       yes
  !! f32x4_view_copy has no 128-bit ld.global vector op
  !! f32x4_view_copy has no 128-bit st.global vector op

✗ FAILED: 2 problem(s)
```

The emitted memory operations inside `f32x4_view_copy`:

```
cvta.global.u64  %rd58, %rd25;
st.local.v2.b32  [%rd34+8], {%r33, %r29};
ld.local.b64     %rd5,  [%rd34+8];
cvta.global.u64  %rd60, %rd40;
st.local.v2.b32  [%rd48+8], {%r55, %r51};
ld.local.b64     %rd49, [%rd48+8];
ld.v2.b64        {%rd54, %rd55}, [%rd53];
st.v2.b64        [%rd56], {%rd54, %rd55};
```

Compare the direct `f32x4` kernel in the same module, which is what the example expects:

```
ld.global.v2.b64  {%rd21, %rd22}, [%rd20];
st.global.v2.b64  [%rd3], {%rd21, %rd22};
```

So the width is right and the address space is not. The example's message is literally true but describes the wrong defect — worth fixing in the example's wording whatever the outcome here.

**What I ruled out**

- **Not architecture-specific.** Building for `sm_75` and `sm_90` produces byte-identical output for this kernel, down to register numbers. (This also corrects an earlier version of this issue, which claimed newer hardware was needed to tell — the assertion reads generated PTX, so a compile-only build settles it with no GPU.)
[...]

PR #669, opened 2026-08-05
docs(examples): vectorization misdescribes what f32x4_view_copy emits

vyncint:
Split out of #657 as the part that needs no decision from you.

The failure message reads as though the transaction had been scalarized:

```
!! f32x4_view_copy has no 128-bit ld.global vector op
```

SASS on sm_86 says otherwise:

| | `f32x4` (direct) | `f32x4_view_copy` |
|---|---|---|
| global | `LDG.E.128` x1, `STG.E.128` x1 | — |
| generic | — | `LD.E.128` x1, `ST.E.128` x1 |
| local | — | `LDL.64` x2, `STL.64` x2 |
| **total** | **2** | **6** |

The 128-bit access **is** there. What is missing is the *global window*: the view path addresses through the generic one and pays four local-memory accesses, plus 37 `PRMT` byte-permutes, on the way.

## What this changes

The message and its comment only. **The check is untouched and still fails** — the pessimization it detects is real, and this PR does not paper over it. #657 remains open for the codegen question.

I am sending it separately because it is correct either way: if the codegen is fixed the message never prints, and if it is not, anyone who hits it currently gets pointed at the wrong thing. I spent a while chasing a scalarization that was never happening because I believed this string.

`cargo fmt --check` clean; no behaviour change.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/vectorization/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/vectorization/src/main.rs
@@ -272,7 +272,16 @@ fn main() {
             Some(line) => println!("f32x4_view_copy: {line}"),
             None => {
                 errors += 1;
-                println!("  !! f32x4_view_copy has no 128-bit {dir}.global vector op");
+                // Deliberately not "has no 128-bit vector op": measured on
+                // sm_86, this path does reach 128 bits, as a generic `LD.E.128`
+                // / `ST.E.128` rather than the `LDG.E.128` / `STG.E.128` the
+                // direct path gets, and pays four extra `LDL.64`/`STL.64`
+                // local accesses on the way. What is missing is the *global*
+                // window, not the width. See #657.
+                println!(
+                    "  !! f32x4_view_copy: no 128-bit {dir}.global vector op \
+                     (the 128-bit access is there, but in the generic window)"
+                );
             }
         }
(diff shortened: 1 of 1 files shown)


vyncint:
Pushed `d08ddb43`, prompted by #675.

This PR reworded the failure message for a check that #675 makes **pass** — it fixes the generic-window fallback at the root, in the enum filler lowering. So on a tree with #675 the `None` arm here is never taken, and the comment I added, which described the fallback in the present tense, would be asserting in-tree a pessimization that no longer happens.

The comment is now framed as what a *future* failure most likely means rather than what the compiler currently does, so it reads correctly whichever order the two land in. The printed message is unchanged — naming the address space rather than the width is still the right thing to say when the check fires.

If you would rather this PR just went away now that the root cause is fixed, that is fine too: the message only matters on a regression. I have kept it because a regression here is easy to misread as scalarization, which is exactly the trap #657 set for me.

nihalpasham:
makes sense on the wording — the SASS evidence in #657 shows the 128-bit access exists but sits in the generic window, and the old message sent readers hunting for a scalarization that never happened. 

With #675 merged the check passes and neither message prints, but the corrected failure text is what we want on record if this ever regresses.

nihalpasham (approved):

PR #670, opened 2026-08-05
Construct a disjoint slice when a kernel builds one

### LINKED ISSUE #667: device codegen error for DisjointSlice::from_raw_parts
**Description**

Using `DisjointSlice::from_raw_parts` gives a device codgen error. This could be related to using the DisjointSlice from a separate `#[device]` method, as this was required for my repro.

**Minimal reproducer**

```rust
use cuda_device::{DisjointSlice, device, kernel, thread};
use cuda_host::cuda_module;

#[cuda_module]
mod kernels {
    use super::*;

    #[device]
    fn increment_device<'a>(vector: &mut DisjointSlice<'a, f32>) {
        if let Some(value) = vector.get_mut(thread::index_1d()) {
            *value += 1.0;
        }
    }

    #[kernel]
    pub fn disjoint_from_raw_parts_repro(ptr: *mut f32, len: usize) {
        let mut view = unsafe { DisjointSlice::from_raw_parts(ptr, len) };
        increment_device(&mut view);
    }
}

    let context = CudaContext::new(0)?;
    let stream = context.default_stream();
    let output = DeviceBuffer::<f32>::zeroed(&stream, 1)?;
    let module = kernels::load(&context)?;
    unsafe {
        module.disjoint_from_raw_parts_repro(
            &stream,
            LaunchConfig::for_num_elems(1),
            output.cu_deviceptr() as *mut f32,
            output.len(),
        )?;
    }

    Ok(())
}
```

**Expected behavior**

compiles ok

**Actual behavior**

```
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Translation failed: disjoint_from_raw_parts_repro: ["/home/mrobins/.cargo/git/checkouts/cuda-oxide-6d394bb007f5e114/e28248c/crates/cuda-device/src/disjoint.rs": line: 179, column: 9] Compilation error: invalid input program.
       Unsupported construct: Scalar-lowered ADT expected exactly one runtime field, found 0
```

**Environment**
- GPU: A40
- CUDA driver version: 580.173.02
```
binary: rustc
commit-date: 2026-04-02
release: 1.96.0-nightly
LLVM version: 22.1.2
```
- `llc --version` (or `llc-22 --version`):

```
❯ llc-21 --version
Ubuntu LLVM version 21.1.8
  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: znver3

  Registered Targets:
    aarch64     - AArch64 (little endian)
    aarch64_32  - AArch64 (little endian ILP32)
    aarch64_be  - AArch64 (big endian)
    amdgcn      - AMD GCN GPUs
    arm         - ARM
    arm64       - ARM64 (little endian)
    arm64_32    - ARM64 (little endian ILP32)
    armeb       - ARM (big endian)
    avr         - Atmel AVR Microcontroller
    bpf         - BPF (host endian)
    bpfeb       - BPF (big endian)
    bpfel       - BPF (little endian)
    hexagon     - Hexagon
    lanai       - Lanai
    loongarch32 - 32-bit LoongArch
    loongarch64 - 64-bit LoongArch
    m68k        - Motorola 68000 family
    mips        - MIPS (32-bit big endian)
    mips64      - MIPS (64-bit big endian)
    mips64el    - MIPS (64-bit little endian)
    mipsel      - MIPS (32-bit little endian)
    msp430      - MSP430 [experimental]
    nvptx       - NVIDIA PTX 32-bit
    nvptx64     - NVIDIA PTX 64-bit
    ppc32       - PowerPC 32
    ppc32le     - PowerPC 32 LE
    ppc64       - PowerPC 64
    ppc64le     - PowerPC 64 LE
    r600        - AMD GPUs HD2XXX-HD6XXX
    riscv32     - 32-bit RISC-V
    riscv64     - 64-bit RISC-V
    sparc       - Sparc
    sparcel     - Sparc LE
    sparcv9     - Sparc V9
    spirv       - SPIR-V Logical
    spirv32     - SPIR-V 32-bit
    spirv64     - SPIR-V 64-bit
    systemz     - SystemZ
    thumb       - Thumb
    thumbeb     - Thumb (big endian)
    ve          - VE
    wasm32      - WebAssembly 32-bit
    wasm64      - WebAssembly 64-bit
    x86         - 32-bit X86: Pentium-Pro and above
    x86-64      - 64-bit X86: EM64T and AMD64
    xcore       - XCore
    xtensa      - Xtensa 32
```

alejandro-soto-franco:
Building a `DisjointSlice` inside a kernel fails device codegen whenever the construction survives optimisation, which it does as soon as the slice crosses a call.

## What

`DisjointSlice::from_raw_parts` fails device codegen when the slice crosses a `#[device]` boundary:

```text
Unsupported construct: Scalar-lowered ADT expected exactly one runtime field, found 0
```

The constructor writes a struct literal. That literal usually folds into its use before import, and it survives whenever the slice crosses a call the optimiser keeps. The type translator gives the ADT its own slice type rather than a struct, so the aggregate arm took it for a scalar-lowered newtype, looked for the one field carrying the collapsed scalar type, and found none.

## How

`mir.construct_disjoint_slice` takes the data pointer, the element count and the runtime layout words the index space carries. `mir.construct_slice` cannot serve: its result is a `MirSliceType` and its arity is fixed at two, while a runtime row width adds a word after `{ ptr, len }`.

Lowering is the same insert chain one field longer, built through `make_disjoint_slice_struct` so the field order cannot drift from the layout `resolve_aggregate_slots` indexes identically.

Field selection in the importer is positional, and the op's verifier checks it against the result type: the data pointer must point to the element type, the length must be an integer, and each remaining operand must match the index space's layout types in order. A reordered or retyped field fails verification rather than writing a row width into the length slot.

## Testing

On an RTX 5060 (sm_120):

- `disjoint_from_raw_parts` covers both shapes. The two-word form increments 4096 elements, and the three-word form scales 888 elements at a row width of 37, which is neither a power of two nor a multiple of the block width. Every element's expected value depends on its own index, so a constant write or a misaddressed element reports a mismatch. Both exact.
- Lowering unit tests pin the slot layout for both widths. Writing the row width into any slot but the third fails them.
- Verifier unit tests cover a missing index-space operand, a surplus one, a mismatched width type and a plain slice result.

Closes #667.


nihalpasham:
Verified the new op and the fix on hardware:

- `disjoint_from_raw_parts` on RTX 5090: both kernels exact match (4096 elements incremented; 888 elements scaled at row width 37), covering the two-word and three-word slice shapes.
- the `mir.construct_disjoint_slice` verifier + positional ZST filtering is the right fix per the compiler-gaps-are-bugs rule — the importer arm, not the kernel, was what needed to change.
- cargo test across dialect-mir / mir-importer / mir-lower green.

Closes #667.

nihalpasham (approved):

PR #672, opened 2026-08-05
Reduce over the live lanes of a partial warp

alejandro-soto-franco:
A block whose width is not a multiple of 32 leaves its last warp short, and the library offers no reduction that such a warp may call.

## What

The warp reductions shuffle with the full 32-lane member mask, so every lane must be launched and converged. Blocks of 48 threads give a second warp of 16 live lanes, and the PTX ISA makes `shfl.sync` undefined when a thread sources a lane that is inactive or outside the member mask.

#655 met this in the `warp_sums` example and carried a live-lane butterfly private to it, correct only for a power-of-two tail by its own admission. Any block width whose tail is not a power of two, 45 threads for instance, has nothing to call.

## How

```rust
pub fn live_lanes_1d() -> u32;
pub fn reduce_sum_f32_partial(val: f32, live_lanes: u32) -> f32;
pub fn reduce_max_f32_partial(val: f32, live_lanes: u32) -> f32;
pub fn reduce_min_f32_partial(val: f32, live_lanes: u32) -> f32;
pub fn reduce_sum_f64_partial(val: f64, live_lanes: u32) -> f64;
pub fn reduce_max_f64_partial(val: f64, live_lanes: u32) -> f64;
pub fn reduce_min_f64_partial(val: f64, live_lanes: u32) -> f64;
```

A power-of-two count takes the same butterfly as the full-warp form with the mask and the first offset cut to the live lanes, which keeps `lane ^ offset` inside the mask at every step.

Any other count folds the upper part of the span into the lower half and halves the span, reaching an arbitrary count in `ceil(log2(live))` steps. Every shuffle names a source below the live count, clamped to the last live lane where the arithmetic would run past it, so no thread sources a lane that was never launched. The clamped result is discarded wherever the source lies outside the current span. One broadcast then leaves the total in every lane, as the full-warp form does.

`warp_sums` now calls the library form and drops its private helper.

## Testing

On an RTX 5060 (sm_120):

- A host model of the fold totals correctly for every live count from 1 to 32 and never names a source at or above the live count. The second property is the one the device code cannot show by inspection.
- `partial_warp_reduce` covers a 16-lane tail from 48-thread blocks and a 13-lane tail from 45-thread blocks, for sum and for maximum, exact against the host reference.
- `warp_sums` passes unchanged, safe over raw 1.021.

One measurement belongs in the record. Substituting the full-warp form into those kernels also matches the reference exactly on sm_120, so this hardware returns something harmless for a shuffle sourcing a never-launched lane. The ISA promises nothing about that value on any architecture, which is the reason to name the live lanes rather than rely on it. The example's README says so, so a passing run is not read as evidence that the full-warp form is safe there.

Follow-up to #655.


nihalpasham:
The lane math checks out end to end:

- re-derived the two reduction paths independently: pow2 counts get the standard butterfly with the mask cut to the live lanes; arbitrary counts fold the upper half down with sources clamped, so no shuffle ever names an unlaunched lane (undefined per the PTX ISA).
- `partial_warp_reduce` and `warp_sums` on RTX 5090: max error 0 across pow2 and odd tails, SUCCESS both.
- host-model tests prove totals and never-out-of-mask sourcing for every live count 1..=32; clippy clean after your iterator fix.

This gives the library the `partial-warp` form #655's review had to hand-roll inside `warp_sums`.

nihalpasham (approved):

PR #673, opened 2026-08-05
Write a canonical-storage enum payload by rebuilding the enum

alejandro-soto-franco:
Mutating a `bool` enum payload has no lowering, even when the write never leaves the function it is written in.

## What

Enum storage keeps two payload kinds in a converted form: a `bool` occupies a full `i8` byte and a shared-memory pointer is stored generic. A raw payload address cannot carry that conversion, because the address escapes and the loads and stores made through it elsewhere are typed with the semantic type. #652 therefore refused to hand one out, which left this with no lowering at all:

```rust
if let Flag::On(value) = &mut flag {
    *value = input[i] > 0.0;
}
```

## How

The borrow in that shape does not survive to the importer. Optimisation folds it into a store to the payload place:

```text
((_9 as On).0: bool) = Gt(move _12, const 0f32)
```

It is that store which asks for the address, so the write needs none. The enum is rebuilt around the new payload and stored whole, which is what a whole-enum assignment already does, and `construct_enum` coerces each payload to its canonical storage on the way in.

```text
(_flag as On).0 = v   ->   _e  = load _flag
                           _e' = construct_enum On(v)
                           store _e' -> _flag
```

Valid MIR assigns to a variant's field only when the enum already holds that variant, so rebuilding with the same variant leaves the discriminant where it was. A variant of several fields keeps the others as they stand: each is read back out of the loaded value and passed through unchanged.

A borrow that escapes, into a `#[device]` call for instance, reaches no store site and keeps the refusal. The diagnostic now names which of the two cases it is.

## Testing

On an RTX 5060 (sm_120):

- `enum_payload_addr` gains `mutate_bool_payload`, whose match arms differ so a write landing in a copy reports a mismatch rather than passing, and `mutate_multi_field_payload`, where the sibling `f32` of a two-field variant has to survive the rebuild. Both exact over 1048576 elements.
- Removing the write from the first kernel fails it at element 1, so the check discriminates.
- `error_enum_bool_payload_addr` still fails closed on the escaping borrow, and the smoketest verdict matches the updated diagnostic.
- The other six kernels of `enum_payload_addr` are unchanged, in-place over rebuild 1.070.

Follow-up to #652.

### DIFF
--- a/crates/mir-importer/src/translator/mod.rs
+++ b/crates/mir-importer/src/translator/mod.rs
@@ -47,6 +47,7 @@ pub mod block;
 pub mod body;
 pub(crate) mod layout;
 pub(crate) mod location;
+pub(crate) mod payload_store;
 pub mod rvalue;
 pub mod statement;
 pub mod terminator;
--- a/crates/mir-importer/src/translator/payload_store.rs
+++ b/crates/mir-importer/src/translator/payload_store.rs
@@ -1,0 +1,288 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Writing an enum payload whose bytes use canonical storage.
+//!
+//! Enum storage keeps two payload kinds in a converted form: a `bool` occupies
+//! a full `i8` byte, and a shared-memory pointer is stored generic. The value
+//! paths coerce at that boundary. A raw payload address cannot, because the
+//! address escapes and the loads and stores made through it elsewhere are
+//! typed with the SEMANTIC type, so mir-lower refuses to hand one out.
+//!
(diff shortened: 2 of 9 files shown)


nihalpasham:
Verified the rewrite at both layers and on hardware:

- `enum_payload_addr` on RTX 5090: SUCCESS, in-place/rebuild ratio 1.005, and the error example still refuses escaping borrows with the sharpened diagnostic.
- `cargo test -p mir-importer` green (923 tests); clippy clean after your follow-up commit.
- the rebuild path reuses `construct_enum`'s existing storage coercion, so the `bool-as-i8` conversion happens in one place and sibling fields of multi-field variants carry through unchanged.

This unblocks the if `let ... = &mut` pattern #652 made loud. Nested canonical payloads (a bool inside a struct payload) stay refused; a follow-up issue for those is welcome. #306 (SetDiscriminant) stays open by design.

nihalpasham (approved):

PR #674, opened 2026-08-05
feat: Mir pass registry

drbh:
this pr adds an opt-in staged MIR pass registry to cuda-oxide.

the goal is to give MIR optimizations one owned, validated entry point instead of adding ad-hoc environment checks and pipeline hooks for each pass.

- adds `CUDA_OXIDE_MIR_PASSES` plus a typed `CompileOptions::with_mir_pass_pipeline(...)` API
- parses and validates the complete comma-separated selection before any transformation runs, so an invalid later pass cannot leave a module partially transformed
- runs selected passes at explicit compiler stages: `PrePreparation`, `PostMem2Reg`, and `PostPreparation`
- preserves the requested order for passes in the same stage while keeping stage execution in compiler order
- reports invalid selections as input errors
- adds registry and preparation tests for ordering, empty/invalid pipelines, and debug-mode rejection

this intentionally adds the registry infrastructure without enabling an in-tree optimization yet. follow-up optimization branches can register a pass at the stage whose IR contract they need, without changing the public option or orchestration plumbing again.

this pr is based off the discussions in #398 and #636. both were adding opt-in compiler optimizations, and the question kept coming up how each pass should be selected and where it should run. instead of having every follow-up add another env option and more pipeline plumbing, this puts that in a small registry and lets each pass pick its stage.

### running tests

```bash
cargo test -p cuda-oxide-codegen --lib mir_pass_registry::
```

outputs

```
running 5 tests
test mir_pass_registry::tests::empty_registry_accepts_only_the_empty_pipeline ... ok
test mir_pass_registry::tests::empty_pipeline_runs_nothing ... ok
test mir_pass_registry::tests::invalid_pipeline_does_not_run_a_prefix ... ok
test mir_pass_registry::tests::selection_runs_only_entries_declared_for_each_stage ... ok
test mir_pass_registry::tests::selected_passes_run_in_order ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 146 filtered out; finished in 0.01s
```

and

```bash
cargo test -p cuda-oxide-codegen --lib prep::
```

outputs

```
running 2 tests
test prep::tests::debug_mode_rejects_requested_mir_passes ... ok
test prep::tests::invalid_staged_pipeline_is_rejected_before_preparation ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 149 filtered out; finished in 0.00s
```

### implementation notes

this pr does not add any new optimization passes so the registry can be reviewed independently of any specific optimization. the first follow up is https://github.com/drbh/cuda-oxide/pull/1, which adds `warp-aggregate-constant-fp-atomics`: an explicitly opt-in `PostMem2Reg` pass for coalescing eligible contended f16/f32 atomic adds within a warp. on a RTX 4090 the `atomic_f16` benchmark, improves the highest-contention f32 case from 5.68 ms to 0.18 ms (31.4x). if/when the pass registry lands, i'll re open the pr to this repo to serve as the first example of an optional workload specific pass.

please let me know if the approach should be updated, split differently, or if there is a better direction for the registry and stage plumbing - happy to adjust anything accordingly, thanks!

### DIFF
--- a/crates/cuda-oxide-codegen/src/api.rs
+++ b/crates/cuda-oxide-codegen/src/api.rs
@@ -121,6 +121,7 @@ pub struct CompileOptions {
     debug_info: DebugInfo,
     verbose: bool,
     linking: Linking,
+    mir_pass_pipeline: Option<String>,
 }
 
 impl CompileOptions {
@@ -133,6 +134,7 @@ impl CompileOptions {
             debug_info: DebugInfo::None,
             verbose: false,
             linking: Linking::SelfContained,
+            mir_pass_pipeline: None,
         }
     }
 
@@ -165,6 +167,12 @@ impl CompileOptions {
     /// perform that link.
     pub fn with_linking(mut self, linking: Linking) -> Self {
         self.linking = linking;
+        self
+    }
+
+    /// Select an optional post-preparation MIR pass pipeline.
+    pub fn with_mir_pass_pipeline(mut self, pipeline: impl Into<String>) -> Self {
+        self.mir_pass_pipeline = Some(pipeline.into());
         self
     }
 
@@ -202,6 +210,11 @@ impl CompileOptions {
     /// Requested libdevice linking policy.
     pub fn linking(&self) -> Linking {
         self.linking
+    }
+
+    /// Requested optional post-preparation MIR pass pipeline.
+    pub fn mir_pass_pipeline(&self) -> Option<&str> {
+        self.mir_pass_pipeline.as_deref()
     }
 
     /// Whether progress and tool-selection diagnostics were requested.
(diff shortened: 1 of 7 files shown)


nihalpasham:
The registry is a good start: 

**Gist:**
- pliron `Passes` through `PassManager::run_pass`, 
- the `env` var read once at the options boundary, 
- whole-spec validation before any transformation, and 
- fail-loud on debug builds. 

This should avoid the plumbing #398/#636 kept re-inventing.

I pushed three commits on top:

- dropped the speculative `PostPtx` stage — a textual post-PTX rewrite hook has no in-tree precedent and is the mechanism we've declined before; `OptEntry.build` becomes a plain constructor.
- empty stages now cost nothing: no pass-manager run and no extra module verification when nothing is selected for a stage, so default builds are unaffected (the two now-redundant follow-up verifies in `prepare_mir_module` are removed rather than gated).
- duplicate names in the spec are rejected loudly (`Duplicate` error variant) — running the same pass twice is never what the user meant.

Registry + preparation tests green (9 total).

nihalpasham (approved):

Issue #605, opened 2026-07-31
Value-threaded WGMMA accumulator lowering for pipelined use (wait_group<N>, multiple groups in flight)


nihalpasham:
Title: Value-threaded WGMMA accumulator lowering for pipelined use (wait_group<N>, multiple groups in flight)
Labels: enhancement, cuda-feature, IR-lowering, perf

Follow-up to #598, which landed the first working WGMMA MMA lowering. That lowering fuses the whole fence/MMA/commit/wait sequence into one PTX asm block and round-trips the 32 per-thread accumulator values through memory on either side:

```text
asm {
  ld.f32  x32          acc -> regs
  wgmma.fence
  wgmma.mma_async
  wgmma.commit_group
  wgmma.wait_group 0
  st.f32  x32          regs -> acc
}
```

That shape is sound by construction, but it forecloses everything that makes WGMMA fast in practice:

- only `wait_group<0>` (no partial waits),
- one group in flight at a time,
- a 32-load + 32-store memory round-trip per fused region, so a K-loop pays it every iteration.

CUTLASS and Triton instead thread the accumulator through registers as a value across separate asm statements, which enables register-resident K-loops, multiple groups in flight, and `wait_group<N>`. We should get there too, with two prerequisites to solve:

1. **Aggregate SSA.** Rust MIR keeps the address-taken `[[f32; 8]; 4]` accumulator in memory; we have no machinery to promote it to 32 SSA scalars across the region. Related local-aggregate work: #397, #399, #400.
2. **Spill soundness.** Naive value threading across separate asm blocks is unsound at the LLVM layer: nothing stops LLVM from spilling the in-flight accumulator vregs between the `mma_async` asm and the `wait_group` asm, which the PTX ISA forbids. The design must either fuse larger regions (fence through wait including intervening descriptor arithmetic) or find a representation LLVM provably cannot spill mid-flight.

Acceptance sketch:

- a K-loop with descriptor offset arithmetic between MMAs keeps the accumulator in registers across iterations (no per-iteration ld/st of the accumulator),
- `wait_group<N>` with N > 0 lowers, allowing multiple groups in flight,
- the existing deferred-accumulator path remains the fallback for shapes the value-threaded path cannot prove.

nihalpasham:
Assigned to @uurl

uurl:
## Tentative implementation plan

I plan to implement this issue as a sequence of six reviewable PRs.
The boundaries are provisional and may change based on review feedback or
[...]

PR #679, opened 2026-08-06
test(codegen): validate spill-free 32-value WGMMA carrier

uurl:
This PR is the first of six planned PRs that incrementally implement #605.

## Summary

Add feasibility coverage for carrying the 32 per-thread f32 accumulator
values of BF16 m64n64k16 WGMMA through one convergent inline-PTX region.

The tests verify that:

- `nvvm.inline_ptx` accepts 32 tied f32 inputs and 32 f32 results;
- MIR lowering represents the outputs as an LLVM aggregate followed by
  32 `extractvalue` operations;
- a real `wgmma.mma_async.sync.aligned.m64n64k16.f32.bf16.bf16`
  sequence compiles for `sm_90a`;
- all 32 results remain observable after `wait_group 0`;
- PTX contains no local-memory materialization;
- `ptxas -v` reports zero stack frame, zero spill stores, and zero spill
  loads.

Observed locally:

    Used 58 registers
    0 bytes stack frame
    0 bytes spill stores
    0 bytes spill loads

## Motivation

This isolates the primary backend feasibility question for #605 before
introducing a production value-form WGMMA operation or changing the
existing pointer-based deferred fallback.

## Scope

This PR adds tests only. It does not:

- change the public WGMMA API;
- change the existing deferred accumulator lowering;
- implement K-loop fusion;
- implement partial waits;
- claim runtime numerical validation on Hopper hardware.

### DIFF
--- a/crates/cuda-oxide-codegen/tests/wgmma_value_carrier_ptx.rs
+++ b/crates/cuda-oxide-codegen/tests/wgmma_value_carrier_ptx.rs
@@ -1,0 +1,354 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+ * All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Proves that the codegen pipeline can carry thirty-two tied `f32` values
+//! through one inline-PTX operation and produce spill-free `sm_90a` code.
+
+#![cfg(unix)]
+
+use cuda_oxide_codegen::experimental::{CodegenModule, CompileOptions, Compiler, Target};
+use dialect_mir::{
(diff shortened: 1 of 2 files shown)


nihalpasham:
Step 1 of the #605 plan lands what it promised: 
- proof, before any production code, that the backend can carry 32 tied f32 accumulator values through one convergent asm region without spilling.
- Reproduced the ptxas evidence independently here: 58 registers, 0 bytes stack frame, 0 spill stores/loads for sm_90a (ptxas 13.3, llc-21).
- The lowering test rides the same `struct-return` + `extractvalue` machinery #598's fused path already uses; the 32-output constraint accounting checks out against the `InlinePtxOp` verifier.
- Scope read correctly: this proves the single-region carrier, not cross-statement liveness — that stays with steps 4-5, as the plan says.

Merging.

nihalpasham (approved):

PR #681, opened 2026-08-06
ci: enforce deny.toml over the example workspaces

vyncint:
Closes the policy half of #663. #664 closed the inventory half.

## Problem

`cargo deny check` resolves the **root workspace**. Every example under
`crates/rustc-codegen-cuda/examples/` sets its own `[workspace]`, so the root run
stops at that boundary and the license, source and advisory policies never see a
crate an example pulls on its own.

The gap is **131 distinct crates**: the root workspace resolves 139 third-party
crates, and the example locks add 131 beyond it. All 187 examples add at least
one.

### I was wrong that this was harmless

#663 says "This is not a violation today". Running the policy over the example
workspaces disproves that:

```
error[vulnerability]: Invalid pointer dereference in `fmt::Pointer` impl for `Atomic` and `Shared`
crossbeam-epoch 0.9.18 — RUSTSEC-2026-0204
  crossbeam-epoch v0.9.18
  └── crossbeam-deque v0.8.6
      └── rayon-core v1.13.0
          └── rayon v1.12.0
              └── hashmap_v2 v0.1.0
Solution: Upgrade to >=0.9.20
```

It affects `hashmap_v2` and `hashmap_v3`, and nothing reports it. My original
count of 68 was also low — I had counted only the six examples with *declared*
third-party dependencies, missing transitive build-dep graphs (`bindgen`,
`clang-sys`, `libloading`, `zerocopy`).

## Change

- `scripts/check-example-license-policy.sh` runs `cargo deny check` over the
  example workspaces, and a new `example-policy` job in `cargo-deny.yml` calls
  it. `ci.yml` already `uses:` that whole workflow, so no change is needed there.
- `crossbeam-epoch` 0.9.18 → 0.9.20 in the two rayon examples. Without it the
  guard would fail on the tree it is added to. Both examples still build
  (`cargo oxide build hashmap_v2` / `hashmap_v3`).

### One run per dependency set, not per example

Each example depends on `cuda-core` and friends by path, so every lock re-lists
the root workspace's own transitive crates — which is why all 187 look like they
pull third-party code. Grouping examples by their exact set of third-party
`(name, version)` pairs collapses 187 workspaces to **27** runs.

That is an equivalence, not a sample: licenses, sources and advisories are
per-crate properties, so two workspaces resolving an identical crate set get an
identical verdict. `bans.multiple-versions` is a graph property, but `deny.toml`
sets it to `warn`, so it cannot change the exit status. The script asserts the
representatives cover every crate.

### cutile_inter_kernel is exempt

It cannot satisfy the policy as written:

```
error[source-not-allowed]: detected 'git' source not explicitly allowed   (x7)
error[unlicensed]: cuda-bindings = 0.1.0 is unlicensed
bans ok, licenses FAILED, sources FAILED
```

The first would need cutile-rs in `[sources] allow-git`; the second a license
field on a crate this repository does not own. Both are policy calls for you, so
the example is listed rather than the policy loosened — the same reason it is
already exempt from the inventory guard. As there, every exempt name is checked
against the examples on disk, so a typo or a rename fails the run instead of
quietly exempting nothing, or everything.

## Verification

All local, no GPU needed:

```
bash scripts/check-example-license-policy.sh
  Checking deny.toml over 27 representative example workspaces.
  ... advisories ok, bans ok, licenses ok, sources ok   (x27)
  OK: deny.toml holds over every example workspace outside POLICY_EXEMPT_EXAMPLES.
```

- **186 of 187** example workspaces satisfy `deny.toml` unchanged once the
  advisory is fixed; the 187th is the exempt one above.
- `cargo deny check` at the root — still `advisories ok, bans ok, licenses ok, sources ok`
- `bash scripts/check-dependency-licenses.sh` — still passes with the new
  lockfiles, so the bump needs no new CSV rows
- `cargo oxide build hashmap_v2` and `hashmap_v3` — both produce PTX

The `example-policy` job gets `timeout-minutes: 15` rather than 10: it installs
cargo-deny and then resolves 27 workspaces, so it does more work than the two
existing jobs.

Refs #663


nihalpasham:
**Gist:** This closes the policy half of #663 the same way the merged #664 closed the inventory half: same script shape, same CI job, same single documented exemption. Running it over the tree immediately surfaced a live advisory (`RUSTSEC-2026-0204` via crossbeam-epoch in the rayon examples), so the gap pretty helpful.

- dropped the top-level `--config` flag (0.20-only; every 0.19 install failed all 26 runs) in favor of cargo-deny's manifest-dir walk-up, verified on 0.19.4 and 0.20; added `--locked` so a stale lock fails loudly;
- checking is per lock file, so a nested sub-workspace's own graph is genuinely policy-checked rather than only folded into its parent's grouping key;
- the source string joins the grouping key (the same name+version from crates.io and from a git fork are different policy questions);
- parse self-test floors mirroring #664's guard; crossbeam-epoch CSV row refreshed to 0.9.20.

Guard re-run clean end to end (26 representatives); hashmap_v2/v3 build and pass the smoketest on sm_120. 

Closes #663.

nihalpasham (approved):

PR #684, opened 2026-08-07
perf(mir-importer): materialize a bare array constant as one immutable global

### LINKED ISSUE #397: Small runtime-indexed local arrays spill to PTX local memory

vyncint:
Refs #397.

## Summary

`const TABLE: [T; N]` indexed by a runtime value was built element by element and
stored into the local's stack slot, so **every thread wrote its own copy of data
the module image already carries**: one `st.local` per element, then `ld.local`
reads of thread-private memory. The identical data written `&[T; N]` has always
become a single immutable device global read with one load, so two spellings of
one table cost very different amounts.

This assigns the constant with `mir.memcpy` from that same promoted global.

Three parts:

- **`mir-importer`** — a new arm in the existing "array initializers assigned to
  an addressable local" fast-path block emits `mir.global_alloc` + `mir.memcpy`
  instead of building the array as a value. The promoted-global emission is
  factored out of `translate_ptr_to_array_constant`, so both spellings now share
  one code path and one content-keyed dedup.
- **`dialect-mir` / `mir-lower` / `llvm-export`** — `MirGlobalAllocOp` carries a
  `global_immutable` marker through to `llvm::GlobalOp`, and the exporter writes
  LLVM `constant` rather than `global` for it.
- **a write-once guard**, without which this would regress the mutable case.

## Why a copy, and not the global's address

The local is ordinary mutable storage — `let mut t = TABLE; t[0] = x;` is legal —
so handing out the global's address would be wrong in general, and proving the
local is never written is exactly the analysis I wanted to avoid writing.

A `memcpy` from `constant` storage is unconditionally correct, and **`opt` already
supplies that proof**: `isOnlyCopiedFromConstantMemory` deletes the copy and
rewrites the reads to the global wherever the local is never written, and keeps
the copy wherever it is.

That is also why the `constant` keyword is load-bearing rather than cosmetic. It
is what makes reads of the storage invariant, which both enables the elimination
and makes `llc` select `ld.global.nc` (the read-only data cache).

A whole-array **load** cannot reach the global however it is spelled. I probed
four shapes through `opt -passes='default<O2>'` at LLVM 22.1.2 on the pinned
toolchain, NVPTX triple, 256 x f32:

| shape | after `opt -O2` |
|---|---|
| `alloca` + `insertvalue` chain + aggregate `store` (today) | 256 `st.local` survive |
| `constant` global + whole-array `load` + `store` to alloca | **fails** — SROA splits the FCA load (`.fca.N.gep`) and folds it back into per-element stores |
| `constant` global + `llvm.memcpy` into alloca | **collapses to `gep @global; load`** |
| same, but the alloca is also written | copy **retained** |

So only a memcpy survives to where the proof runs. Plain external linkage is
enough — no `private`/`unnamed_addr` needed — which is why the exporter change is
one keyword.

## The marker is a property of storage, not of a pointer

`global_immutable` is set **only** at the promoted-constant sites, never for a
Rust `static`, whose bytes the host can still overwrite by symbol
(`ensure_static_global_alloc` is untouched). `MirPtrType::is_mutable` is
deliberately not consulted: a shared reference to a mutable static is an
immutable pointer to mutable storage, and #413 records that `is_mutable` must not
be read as a promise about the pointee. There is a test asserting lowering does
not *infer* immutability.

## The write-once guard

Without it this trades one cost for another. When the local **is** written the
copy survives, and NVPTX expands a surviving `memcpy` into a **byte loop** —
measurably worse than the element-wise stores it replaced. So
`constant_local_is_written_once` rejects a local that is assigned more than once,
borrowed at all (either mutability), or written by a call.

The guard only chooses between two *correct* lowerings — both give the local a
private copy — so an unrecognised write form costs performance, never
correctness. That is why its statement match ends in a catch-all rather than an
error.

Checked directly: a kernel doing `let mut t = TABLE; t[0] = input[i] as f32;`
still emits 64 `st.local.b32` with no memcpy loop, while the read-only kernel
beside it emits no depot and one `ld.global.nc.b32`.

## Measured

A10G, from the evidence log's `nvidia-smi` block:

```
name, compute_cap, memory.total [MiB], driver_version
NVIDIA A10G, 8.6, 23028 MiB, 595.71.05
```

8388608 threads x 64 divergent lookups per thread (536870912 lookups per launch),
50 timed launches after 10 warmup. Same kernels and same workload in both runs;
only the report's labels differ. Every kernel bit-correct against a host
reference in both.

**Before** (`.gpu-evidence/20260807T001402Z.log`):

```
kernel            entries    space   us/launch   Glookups/s   correct
----------------------------------------------------------------------
lut16_value            16   .local      1282.3       418.67       yes
lut16_ref              16  .global       374.3      1434.36       yes
lut16_constant         16   .const      1283.3       418.35       yes
lut256_value          256   .local     16044.1        33.46       yes
lut256_ref            256  .global       321.8      1668.22       yes
lut256_constant       256   .const     16045.2        33.46       yes
```

**After** (`.gpu-evidence/20260807T012016Z.log`):

```
kernel            entries     spelling   us/launch   Glookups/s   correct
--------------------------------------------------------------------------
lut16_value            16     [f32; N]       374.3      1434.52       yes
lut16_ref              16    &[f32; N]       374.2      1434.60       yes
lut16_constant         16  ConstantMem      1294.6       414.70       yes
lut256_value          256     [f32; N]       324.1      1656.52       yes
lut256_ref            256    &[f32; N]       324.1      1656.52       yes
lut256_constant       256  ConstantMem     16053.5        33.44       yes

time relative to the `&[f32; N]` spelling (1.00x = same cost):
--------------------------------------------------------------------------
   16 entries:  [f32; N]   1.00x   ConstantMem    3.46x
  256 entries:  [f32; N]   1.00x   ConstantMem   49.53x
```

| entries | spelling | before | after | speedup |
|---|---|---|---|---|
| 16 | `[f32; N]` | 1282.3 us | **374.3 us** | **3.43x** |
| 256 | `[f32; N]` | 16044.1 us | **324.1 us** | **49.50x** |
| 16 | `&[f32; N]` | 374.3 us | 374.2 us | 1.00x (control) |
| 256 | `&[f32; N]` | 321.8 us | 324.1 us | 0.99x (control) |

`[f32; N]` now costs what `&[f32; N]` costs, to within 1%, and the controls did
not move. PTX for the 256-entry kernel goes from **309 instructions with a 1 KiB
local depot and 256 `st.local`** to **70 instructions with no depot**, and the
table read becomes `ld.global.nc`.

Two entry counts are measured because `ptxas` can rescue a small table on its
own. Worth flagging: **`ptxas -v` reported "0 bytes stack frame" for every one of
these kernels, before and after** — it is not a usable signal for this, and only
the runtime measurement exposed it.

## How this differs from #398 and #636

Both were closed for good reasons that this does not reproduce.

- They attacked the **read**, replacing a dynamic load with a select chain. #636
  measured **16-19% slower** on the workload it targeted. The dominant cost here
  is the **initialization** — N stores per thread against K reads — which select
  chains do not touch. That is also why the small-table case only moves 3.4x
  while the 256-entry case moves 49.5x: the cost is linear in N.
- #398 was rejected partly because `gemm_sol`'s PTX grew ~64%. This **shrinks**
  PTX (309 -> 70 on the affected kernel) and adds no pipeline stage: no extra
  `opt` run, no unroller, no env plumbing. Kernels that never materialize an
  array constant are untouched.
- No textual IR rewriting. The transformation is in the typed MIR the importer
  already builds, and the elimination is `opt`'s own existing pass.

I have not tried to close #397 with this — the iterator-adapter half (#399) and
mutable local scratch arrays are untouched, and the guard deliberately declines
anything it cannot see is write-once.

## Also here, deliberately unfixed

`ConstantMemory::get()` returns `T` **by value**, so a `ConstantMemory<[f32; N]>`
reads all N entries out of `.const` and writes them into a local depot. It still
measures 3.46x / 49.53x off the other two spellings, so `.const` cannot back a
lookup table today. This is a separate defect in `cuda-device`'s API shape rather
than in lowering, so I left it alone and kept it in the example as the contrast.
Happy to file it separately if useful.

## New example

`const_table_lookup` measures the three spellings of one table side by side and
keeps their agreement checked. The index comes from a per-thread LCG so it is
unpredictable and lane-divergent, and the kernel is lookup-bound rather than
bandwidth-bound (4 bytes in, 4 bytes out per thread). All six kernels verify
bit-exactly against a host reference computed with the same add order.

## Validation

Local (`fmt`, `clippy`, unit tests, docs, doctests all clean):

```
cargo fmt --check
cargo test -p mir-lower -p llvm-export -p dialect-mir -p dialect-nvvm \
           -p mir-importer -p cuda-oxide-codegen --lib --tests     # 17 suites ok
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace
cargo test --doc --workspace --exclude cuda-bindings
bash scripts/check-dependency-licenses.sh                          # OK
bash scripts/check-error-example-status.sh                          # OK
```

On the A10G:

- **full smoketest 188/188, 0 fail** (`.gpu-evidence/20260807T005028Z.log`)
- **scoped re-run 16/16** with the guard in place
  (`.gpu-evidence/20260807T012016Z.log`), covering `array_constants`,
  `array_index`, `array_init`, `enum_array_match`, `cast_tests`,
  `tuple_array_provenance`, `tuple_constant_provenance`, `uninit_const`,
  `device_global`, `constant_memory_coeffs`, `place_read_fallbacks`,
  `ref_index_projections`, plus `vecadd` / `vectorization` as passing controls
  and `tma_copy` as a skip control

`array_constants` is the existing regression test for array-constant *values*, so
it is the one that matters most here; it passes in both runs.

New test: `immutable_marking_survives_lowering_and_is_not_assumed` asserts the
marker survives MIR -> LLVM lowering **and** that a plain static does not acquire
it.

instruction-level evidence above is PTX only.


nihalpasham:
**Gist:** This PR uses the established lowering for this problem: 
- emit the constant as a private read-only global plus one memcpy, and let opt's `isOnlyCopiedFromConstantMemory` delete the copy. clang/nvcc and rustc's own backend do the same thing. The rejected #398/#636 instead attacked the read with select chains, and measured slower.

- llvm-export test for the load-bearing keyword: a marked global exports constant, an unmarked twin stays global, on both NVVM dialects;
- importer test that promoted globals are marked immutable and dedup by the full (type, bytes) image;
- a `lut256_mut_copy` kernel + bit-exact host reference covering the write-once guard's fallback end to end: the written copy stays per-thread and element-materialized (262 `.local` lines vs 0 for `lut256_value`).

- all 7 kernels bit-correct, `lut256_value` at 1.00x of `lut256_ref` with `ld.global.nc` and 
- zero `.local`; full smoketest sweep 187/187. 
 
Partially addresses #397 (the read-only by-value subcase; written scratch arrays and #399's iterator shapes stay open).

nihalpasham (approved):

PR #685, opened 2026-08-07
feat(wgmma): add 32-value SSA accumulator carrier

uurl:
This PR is stacked on #679 and is the second of six planned PRs for #605.

## Summary

Add an internal value-form BF16 WGMMA group that carries the 32 per-thread
`f32` accumulator values explicitly through SSA operands and results.

The new `nvvm.wgmma_mma_group_values_m64n64k16_f32_bf16` operation:

- takes 32 `f32` accumulator values followed by one or more `u64`
  descriptor pairs;
- returns 32 `f32` accumulator values;
- lowers the complete fence/MMA+/commit/`wait_group<0>` sequence into one
  convergent inline-PTX region;
- ties each accumulator input to its corresponding output register;
- does not materialize the accumulator through `ld.f32` / `st.f32` inside
  the WGMMA lifetime.

The existing pointer-form deferred group remains unchanged and available as
the fallback path.

## Motivation

#679 established that a real BF16 `m64n64k16` WGMMA can carry all 32
per-thread accumulator values through tied inline-asm operands/results and
reach `ptxas` for `sm_90a` without spills.

This PR turns that feasibility result into an internal dialect operation and
production lowering that later transformations can use to keep accumulators
register-resident across WGMMA regions.

## Implementation

- Add `WgmmaMmaGroupValuesM64N64K16F32Bf16Op`.
- Verify:
  - exactly 32 `f32` accumulator operands;
  - exactly 32 `f32` results;
  - one or more descriptor pairs;
  - all descriptors are `u64`.
- Register the new handwritten NVVM operation.
- Add its `MirToLlvmConversion` implementation.
- Lower it to one convergent inline-PTX statement with:
  - 32 `=f` outputs;
  - tied input constraints `0..31`;
  - `l` constraints for descriptors;
  - a memory clobber.
- Extract the 32 LLVM aggregate results back into individual SSA values.
- Add dialect and lowering coverage for the new operation.

## Validation

```text
dialect-nvvm:
  50 passed

mir-lower:
  166 unit tests passed
  90 integration tests passed

cuda-oxide-codegen:
  137 unit tests passed
  integration tests passed
````

The WGMMA backend feasibility test from #679 was also re-run successfully:

```text
ptxas info : Compiling entry function 'wgmma_value_carrier_kernel' for 'sm_90a'
0 bytes stack frame
0 bytes spill stores
0 bytes spill loads
Used 58 registers
```

`git diff --check` is clean.

## Scope

This PR does not yet:

* adapt the existing accumulator pointer path to the value-form operation;
* change selection or fallback behavior;
* recognize or fuse a counted K-loop;
* support partial `wait_group<N>` values;
* keep multiple committed groups in flight.

Those are left to subsequent PRs.

### DIFF
--- a/crates/cuda-oxide-codegen/tests/wgmma_value_carrier_ptx.rs
+++ b/crates/cuda-oxide-codegen/tests/wgmma_value_carrier_ptx.rs
@@ -6,6 +6,10 @@
 
 //! Proves that the codegen pipeline can carry thirty-two tied `f32` values
 //! through one inline-PTX operation and produce spill-free `sm_90a` code.
+//!
+//! The probe covers both group shapes the value-form lowering emits: a
+//! single `wgmma.mma_async` and a chain of several under one
+//! fence/commit/wait sequence in the same asm region.
 
 #![cfg(unix)]
 
@@ -33,7 +37,7 @@ use std::num::NonZeroUsize;
 
 const ACCUMULATOR_LEN: usize = 32;
 
-fn build_wgmma_value_carrier_kernel(module: &mut CodegenModule) {
+fn build_wgmma_value_carrier_kernel(module: &mut CodegenModule, mma_count: usize) {
     module.edit(|ctx, module| {
         let module_region = module.get_operation().deref(ctx).get_region(0);
         let module_block = module_region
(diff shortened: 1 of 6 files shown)


nihalpasham:
**Gist:** Step 2 of the #605 plan: an internal operation that carries all 32 accumulator values as plain values instead of behind a pointer, so later steps can keep them in registers across a WGMMA region. Nothing user-facing routes onto it yet, so merging it is risk-free by construction.

- the #679 feasibility probe had only ever assembled one MMA instruction per region, and 
- this operation chains several; the probe now also assembles the two-MMA shape. 
- Assembler result: zero spills, all 32 values stay in registers.

Merging (start of the stack; #686-#689 follow in order).

nihalpasham (approved):

PR #686, opened 2026-08-07
feat(wgmma): adapt linear pointer accumulators to SSA values

uurl:
This PR is the third of six planned PRs for #605 and builds on the 32-value SSA carrier introduced by #685.

## Summary

Adapt canonical pointer-form BF16 WGMMA accumulator sequences to the
32-value SSA carrier introduced in the previous PR.

For a canonical `*mut [[f32; 8]; 4]` accumulator, the lowering now:

- computes typed pointers for the 32 per-thread `f32` accumulator elements;
- loads each accumulator element exactly once before the WGMMA region;
- feeds the 32 scalar values into
  `WgmmaMmaGroupValuesM64N64K16F32Bf16Op`;
- recovers the 32 SSA results after the convergent WGMMA region;
- stores each accumulator element exactly once after `wait_group<0>`.

Unsupported accumulator pointer shapes continue to use the existing deferred
pointer-form WGMMA group.

## Motivation

The previous PR introduced a value-form WGMMA group capable of carrying all
32 accumulator values explicitly through SSA and lowering them to tied PTX
register operands/results.

The existing public WGMMA API still exposes the accumulator through a pointer,
however. Without an adapter, straight-line pointer-form sequences continue to
use the deferred pointer group and materialize accumulator memory inside that
inline-PTX region.

This PR connects the public pointer-form sequence to the new SSA carrier while
preserving the existing deferred path as a fallback.

## Implementation

The WGMMA deferred-accumulator preparation pass now recognizes the canonical
public accumulator shape:

```text
*mut [[f32; 8]; 4]
```

and rewrites a closed linear sequence:

```text
fence
mma+
commit
wait_group<0>
```

into:

```text
32 typed accumulator loads
        ↓
value-form WGMMA group
        ↓
32 accumulator stores
```

The 32 accumulator addresses are produced with typed
`MirArrayElementAddrOp` operations across the `4 x 8` nested array layout.

For non-canonical accumulator shapes, the pass retains the existing
`WgmmaMmaGroupM64N64K16F32Bf16Op` deferred pointer-form fallback.

## Register-lifetime property

For a linear sequence containing multiple MMA instructions, accumulator
memory traffic occurs once at the boundaries of the complete WGMMA region.

For example, a sequence containing two MMAs lowers conceptually as:

```text
load x32
    ↓
[fence
 mma
 mma
 commit
 wait_group 0]
    ↓
store x32
```

rather than performing 32 loads and 32 stores for every individual MMA.

The convergent inline-PTX region itself contains no accumulator:

```text
ld.f32
st.f32
.reg .f32
```

materialization.

## Fallback behavior

The existing deferred pointer-form lowering remains intact.

A supported pointer-form WGMMA sequence whose accumulator does not match the
canonical `[[f32; 8]; 4]` shape continues to lower through the deferred
pointer group rather than being rejected or silently changed.

## Tests

The new coverage verifies that:

* a canonical `[[f32; 8]; 4]` accumulator selects the SSA value adapter;
* a two-MMA sequence performs exactly 32 LLVM loads and 32 LLVM stores;
* all 32 loads occur before the WGMMA inline-asm region;
* all 32 stores occur after the WGMMA inline-asm region;
* the WGMMA region contains both MMA instructions inside one convergent asm;
* all 32 accumulator values use tied register constraints;
* all 32 aggregate results are recovered with `extractvalue`;
* the WGMMA asm contains no accumulator `ld.f32`, `st.f32`, or `.reg .f32`;
* a non-canonical pointer accumulator preserves the deferred pointer fallback.

## Validation

```text
dialect-nvvm:
  50 passed

mir-lower:
  166 unit tests passed
  91 integration tests passed

cuda-oxide-codegen:
  137 unit tests passed
  all integration tests passed
```

The spill-free WGMMA carrier backend test also remains green:

```text
bf16_wgmma_uses_thirty_two_tied_f32_values_without_spills ... ok
```

WGMMA smoketests:

```text
wgmma           PASS
wgmma_mma_bf16  PASS
```

`git diff --check` is clean.

## Scope

This PR does not yet:

* carry accumulator SSA values across a counted K-loop;
* fuse loop control or descriptor recurrence into the WGMMA inline-PTX region;
* support partial `wait_group<N>`;
* keep multiple committed WGMMA groups in flight;
* change the public WGMMA API.

Those are left to subsequent PRs.


nihalpasham:
**Gist:** Step 3: kernels using the standard accumulator shape now load its 32 values once, run the whole WGMMA region on values, and store once at the end, instead of bouncing the accumulator through memory inside the region. Anything nonstandard keeps the old fallback path. 

- three clippy failures CI caught in the new pass (two mechanical rewrites and the house-style allowance for a long argument list).

Verified here: 
- `wgmma` smoketests compile green on this box (execution needs sm_90a hardware, gated as usual); 
- the generated PTX shows the memory round trip is gone.

Merging after #685.

nihalpasham (approved):

PR #687, opened 2026-08-07
feat(wgmma): keep counted K-loop accumulators in SSA

uurl:
This PR is the fourth of six planned PRs for #605 and builds on the
linear SSA accumulator adapter introduced by #686.

## Summary

Extend the value-threaded BF16 WGMMA path across a canonical counted K-loop.

For supported loops, the lowering now keeps all 32 per-thread `f32`
accumulator values in one convergent inline-PTX lifetime across every K
iteration, rather than materializing the accumulator through memory on each
iteration.

The resulting shape is conceptually:

```text
32 accumulator loads
        ↓
one convergent inline-PTX region
{
    fence

    counted K-loop:
        wgmma.mma_async
        update descriptor A
        update descriptor B
        update loop counter
        branch

    commit_group
    wait_group<0>
}
        ↓
32 accumulator stores
```

Unsupported or non-canonical control-flow shapes continue through the existing
linear/deferred paths.

## Motivation

#686 moved straight-line pointer-form WGMMA regions onto the 32-value SSA
carrier, eliminating accumulator memory materialization from inside the
convergent WGMMA asm scope.

A counted K-loop still presents a stronger lifetime requirement: if each loop
iteration becomes a separate LLVM inline-asm statement, LLVM regains visibility
between asynchronous WGMMA operations and may spill accumulator values while a
group is still in flight.

This PR closes that boundary for the canonical counted-loop case by placing the
entire asynchronous accumulator lifetime, including loop control and descriptor
recurrence, inside one convergent inline-PTX statement.

### Counted-loop recognition

The WGMMA accumulator preparation pass now uses the reusable
`mir-transforms` loop and induction analyses to recognize a restricted,
provable counted-loop shape.

The optimization requires a canonical loop with:

- a single recognizable natural loop;
- a single preheader;
- a single latch/back-edge;
- a compile-time-known positive trip count;
- a closed fence / MMA / commit / `wait_group<0>` lifetime;
- one WGMMA operation per loop iteration;
- affine descriptor recurrences;
- the canonical `[[f32; 8]; 4]` accumulator layout.

If those conditions cannot be proved, the pass leaves the loop untouched and
preserves the existing lowering behavior.

Loop analysis is also skipped for regions containing unreachable blocks. This
keeps malformed/negative-test MIR on the established diagnostic path instead of
querying dominance information that does not exist for unreachable CFG nodes.

### Internal loop carrier

A new internal NVVM operation represents a counted BF16 WGMMA loop with:

```text
32 x f32 accumulator inputs
descriptor A base
descriptor B base
descriptor A delta
descriptor B delta
trip count

→

32 x f32 accumulator results
```

The verifier closes the operation over the expected accumulator, descriptor,
delta, and trip-count types.

### Lowering

The counted-loop operation lowers to one convergent, side-effecting LLVM
inline-asm statement.

The 32 accumulator inputs are tied to 32 `=f` outputs, as in the existing
value-form carrier.

The PTX region:

```text
fence
guard zero trip count
loop:
    wgmma.mma_async
    desc_a += desc_a_delta
    desc_b += desc_b_delta
    trips -= 1
    branch if trips != 0
commit_group
wait_group 0
```

remains entirely inside that single LLVM-visible asm boundary.

The 32 aggregate results are then recovered with `extractvalue` and stored back
to the canonical accumulator only after the final `wait_group<0>`.

## Register-lifetime property

For a K-loop with `N` iterations, accumulator memory traffic is now:

```text
32 loads total
N WGMMA instructions
32 stores total
```

rather than:

```text
N × 32 loads
N × 32 stores
```

There is no accumulator `ld.f32`, `st.f32`, or local `.reg .f32` materialization
inside the counted-loop inline-PTX template.

Descriptor recurrence and loop control are also kept inside the same convergent
asm scope, so LLVM does not observe an intermediate accumulator lifetime between
iterations.

## Fallback behavior

This remains a conservative optimization.

Loops that do not satisfy the canonical counted-loop contract are not partially
rewritten. Existing straight-line SSA adaptation and deferred pointer-form
fallback behavior remain available.

Existing malformed WGMMA control-flow cases also continue to produce their
previous diagnostics rather than being consumed by the new loop analysis.

## Tests

Coverage includes:

- recognition and lowering of a canonical counted WGMMA K-loop;
- exactly 32 accumulator loads before the fused loop region;
- exactly 32 accumulator stores after the fused loop region;
- one convergent inline-asm statement for the complete counted loop;
- tied register constraints for all 32 accumulator values;
- descriptor base and delta operands carried into the PTX loop;
- loop trip-count control inside the PTX region;
- one fence, final commit, and final `wait_group<0>`;
- no accumulator `ld.f32`, `st.f32`, or `.reg .f32` materialization inside the
  counted-loop template;
- preservation of existing malformed join/control-flow diagnostics.

## Validation

```text
dialect-nvvm:
  50 tests passed

mir-lower:
  167 unit tests passed
  92 integration tests passed

cuda-oxide-codegen:
  137 unit tests passed
  all integration tests passed
```

The existing backend carrier feasibility probe also remains green:

```text
0 bytes stack frame
0 bytes spill stores
0 bytes spill loads
Used 58 registers
```

That probe validates the 32-value tied-register carrier independently; it is
not a direct `ptxas` spill measurement of the new counted-loop production path.

WGMMA smoketests also remain green:

```text
wgmma           PASS
wgmma_mma_bf16  PASS
```

`git diff --check` is clean.

## Scope

This PR does not yet:

- lower partial `wait_group<N>` values;
- keep multiple committed WGMMA groups in flight;
- pipeline asynchronous groups across the K-loop;
- change the public WGMMA API;
- generalize fusion to arbitrary loop CFGs or descriptor expressions.

Those remain for subsequent PRs in #605.


nihalpasham:
**Gist:** Step 4: for one narrow, provable loop shape, the whole K-loop now lives inside a single asm region, which is what keeps the 32 accumulators in registers across iterations. The recognizer being strict is a feature, not a limitation: anything it cannot prove falls back to the old path.

- this is the first asm in the codebase containing labels and branches, and nothing had ever assembled it. 
- added a probe that builds the real loop through the real compiler pass and requires llc and the assembler (sm_90a) to accept it with zero spills. They do.

Merging after #686 (the rebase collapses the stack to this PR's own commit).

nihalpasham (approved):

PR #688, opened 2026-08-07
feat(wgmma): support partial waits with multiple groups in flight

uurl:
This PR is the fifth of six planned PRs for #605 and builds on the register-resident counted K-loop path introduced by #687.

## Summary

Add value-threaded BF16 WGMMA support for statically known partial
`wait_group<N>` values and multiple committed groups in flight.

The new path recognizes a restricted, provably safe pipelined sequence and
fuses the complete asynchronous lifetime into one convergent inline-PTX
statement.

For `wait_group<1>`, the supported shape is conceptually:

```text
fence

mma acc0
commit

mma acc1
commit
wait_group<1>

mma acc0
commit
wait_group<1>

mma acc1
commit
wait_group<1>

wait_group<0>
```

The lowering keeps two independent 32-value accumulator slots alive across the
pipeline and only exposes them back to LLVM after the final full wait.

## Motivation

The previous PRs established:

- a spill-free 32-value SSA carrier;
- a value-form WGMMA accumulator representation;
- pointer-to-value adaptation for straight-line regions;
- a register-resident canonical counted K-loop.

Those paths still fully drain the asynchronous WGMMA lifetime with
`wait_group<0>` before accumulator values escape.

To allow multiple groups to remain in flight, the lowering must support partial
waits while preserving the PTX accumulator lifetime rules. In particular, an
accumulator belonging to a still-pending WGMMA group must not be reused before
the corresponding wait makes that slot safe again.

This PR adds that legality model for a narrow static pipeline.

## Accumulator slot model

A partial-wait pipeline uses:

```text
max_pending_groups + 1
```

independent accumulator slots.

For `wait_group<1>`, that means two 32-value accumulator sets:

```text
slot 0: acc0_0 ... acc0_31
slot 1: acc1_0 ... acc1_31
```

Groups are assigned to slots round-robin.

A slot may only be reused after the preceding `wait_group<1>` has retired enough
older groups to make that accumulator set safe again.

This avoids reusing the same accumulator registers while a group referencing
them may still be pending.

## Internal representation

Add a value-form WGMMA pipeline operation carrying:

- one or more 32-value accumulator slots;
- one descriptor pair per MMA;
- a statically known maximum pending-group count;
- 32 results for each accumulator slot.

The verifier checks the exact accumulator/result shape, descriptor typing, and
supported pending-group depth.

The initial implementation admits statically known values in the PTX-supported
range and requires the pipeline topology to match the accumulator-slot count.

## Pipeline recognition

The deferred-accumulator preparation pass now recognizes a conservative
straight-line partial-wait pipeline.

The supported region requires:

- one initial WGMMA fence;
- one or more BF16 `m64n64k16` MMA operations;
- explicit `commit_group` boundaries;
- statically known `wait_group<N>` operations;
- `N > 0` for partial waits inside the pipeline;
- enough distinct canonical accumulator objects to avoid unsafe slot reuse;
- a final `wait_group<0>` before any accumulator value escapes;
- canonical mutable generic `[[f32; 8]; 4]` accumulator storage;
- no unsupported control flow or intervening operations.

Unsupported shapes remain fail-closed and continue through the existing
fallback/diagnostic paths.

The existing straight-line `wait_group<0>` path and counted K-loop path remain
unchanged.

## Lowering

The pipeline lowers to one convergent, side-effecting LLVM inline-asm statement.

For a two-slot `wait_group<1>` pipeline, the resulting structure is:

```text
load slot 0: 32 f32
load slot 1: 32 f32

one convergent inline-PTX region
{
    fence

    mma slot 0
    commit

    mma slot 1
    commit
    wait_group 1

    mma slot 0
    commit
    wait_group 1

    ...

    wait_group 0
}

store slot 0: 32 f32
store slot 1: 32 f32
```

Each accumulator input is tied to its corresponding `=f` output.

The entire asynchronous lifetime remains inside the same LLVM-visible asm
boundary, so LLVM cannot introduce a spill boundary between committed WGMMA
groups and their waits.

Accumulator element addresses are recomputed after the final `wait_group<0>`
instead of being kept live across the asm region, reducing unnecessary register
pressure.

## Legality checks

The new path rejects unsafe pipeline shapes, including:

- a partial wait without a final `wait_group<0>`;
- accumulator-slot reuse before it is safe;
- unsupported `wait_group<N>` values;
- insufficient accumulator slots for the requested pending-group depth;
- malformed descriptor or accumulator types;
- unsupported control flow inside the asynchronous lifetime.

These checks preserve the existing fail-closed behavior for regions that cannot
be proven safe.

## Backend feasibility

A new `sm_90a` backend probe exercises:

```text
64 tied f32 accumulator values
2 accumulator slots
multiple committed WGMMA groups
wait_group<1>
final wait_group<0>
```

`ptxas` accepts the generated PTX with:

```text
0 bytes stack frame
0 bytes spill stores
0 bytes spill loads
Used 90 registers
```

This directly validates the two-slot carrier required by the initial
multiple-groups-in-flight path.

The existing 32-value carrier probe also remains green.

## Tests

Coverage includes:

- positive lowering of a two-slot partial-wait pipeline;
- multiple committed WGMMA groups in one convergent asm region;
- statically lowered `wait_group<1>`;
- mandatory final `wait_group<0>`;
- round-robin accumulator-slot reuse;
- rejection of unsafe accumulator reuse;
- rejection of a partial wait without a final full wait;
- rejection of unsupported `wait_group<8>`;
- verifier coverage for the new pipeline carrier;
- backend compilation of the 64-value two-slot carrier with zero spills.

## Validation

```text
cuda-oxide-codegen:
  137 unit tests passed
  all integration tests passed

dialect-nvvm:
  50 tests passed

mir-lower:
  168 unit tests passed
  95 integration tests passed
```

Backend pipeline probe:

```text
0 bytes stack frame
0 bytes spill stores
0 bytes spill loads
Used 90 registers
```

WGMMA smoketests:

```text
wgmma           PASS
wgmma_mma_bf16  PASS
```

`git diff --check` is clean.

## Scope

This PR intentionally keeps pipeline selection conservative.

It does not yet:

- generalize partial-wait fusion to arbitrary CFGs;
- infer arbitrary accumulator scheduling;
- pipeline the counted K-loop introduced by #687;
- expose new public WGMMA APIs;
- add broader end-to-end examples or user-facing documentation.

Those integration and hardening tasks remain for the final PR in #605.


nihalpasham:
**Gist:** Step 5: partial waits (`wait_group<N>`) with several groups in flight, restricted to a straight-line shape where accumulator slots rotate in a fixed pattern, all inside one asm statement so the compiler can never spill a slot the hardware still owns.

- the wait depth was stored as a bare integer attribute; it is now a proper typed attribute that enforces its own `1..=7` range, with the exact rejection messages tested.

Merging after #687.

nihalpasham (approved):

PR #689, opened 2026-08-07
docs(wgmma): finalize value-threaded pipeline integration

### LINKED ISSUE #605: Value-threaded WGMMA accumulator lowering for pipelined use (wait_group<N>, multiple groups in flight)

uurl:
Fixes #605

This PR is the sixth and final PR in the planned series for #605 and builds on
the partial-wait pipeline support introduced by #688.

## Summary

Finalize the public integration, end-to-end coverage, and documentation for the
value-threaded BF16 WGMMA lowering path.

The compiler-side implementation from the preceding PRs now supports three
conservative shapes:

- linear full-drain regions ending in `wait_group<0>`;
- a canonical register-resident counted K-loop;
- straight-line static partial-wait pipelines with multiple committed groups
  in flight.

This PR updates the public WGMMA contract and examples to match those
capabilities and adds a source-level compile-only example that exercises
`wait_group<1>` through the complete Rust-to-PTX pipeline.

No new WGMMA lowering representation or selection algorithm is introduced
here.

## Motivation

The previous PRs established the value-threaded implementation incrementally:

1. #679 validated the spill-free 32-value backend carrier.
2. #685 introduced the internal 32-value SSA accumulator representation.
3. #686 adapted canonical pointer-form accumulators to the value path.
4. #687 kept a canonical counted K-loop register-resident across iterations.
5. #688 added static partial waits and multiple committed groups in flight.

After those changes, the public documentation and compile-only BF16 example
still described the original deferred implementation:

```text
fence
mma
commit
wait_group<0>
```

and still stated that partial waits, multiple accumulators, and loop-spanning
lifetimes were unsupported.

This PR brings those surfaces in sync with the implemented compiler behavior.

## Public lowering contract

The `cuda_device::wgmma` documentation now describes the three supported BF16
`m64n64k16.f32.bf16.bf16` lowering shapes.

### Linear full drain

A canonical `[[f32; 8]; 4]` accumulator is adapted to 32 SSA `f32` values,
processed inside one convergent WGMMA lifetime, and stored only after the final
`wait_group<0>`.

Unsupported full-drain pointer shapes retain the existing deferred pointer-form
fallback when that sequence can still be proven safe.

### Canonical counted K-loop

A restricted compile-time counted loop with one MMA per iteration and affine
`u64` descriptor recurrences is fused into one convergent inline-PTX region.

The accumulator therefore crosses the complete K-loop as register-carried
values instead of paying an accumulator load/store round trip on every
iteration.

### Static partial-wait pipeline

A static:

```text
wait_group<N>
```

with `1 <= N <= 7` uses:

```text
N + 1
```

independent canonical accumulator slots.

Committed groups may remain in flight, accumulator slots are reused only after
the corresponding partial wait makes reuse safe, and every accepted pipeline
ends with a mandatory:

```text
wait_group<0>
```

before accumulator values escape.

Unsupported or unprovable shapes remain fail-closed.

## End-to-end example

Extend the existing `wgmma_mma_bf16` compile-only example with a second kernel
that exercises two independent accumulator slots:

```text
wgmma_fence

mma acc0
commit_group

mma acc1
commit_group

wait_group<1>
wait_group<0>

observe acc0 and acc1
```

This exercises the public API through:

```text
Rust source
    ↓
cuda_device::wgmma
    ↓
MIR importer
    ↓
WGMMA region selection
    ↓
MIR → LLVM lowering
    ↓
PTX generation
```

The example intentionally uses zero-valued descriptors and remains compile-only;
it must not be launched.

The longer round-robin accumulator reuse schedule remains covered by the
`mir-lower` integration tests added in #688.

## Documentation

Update:

- `cuda_device::wgmma` module documentation and safety/lowering contract;
- the MIR importer documentation for the BF16 pointer-form operation;
- the `wgmma_mma_bf16` example and README;
- the WGMMA section of the matrix-accelerator book chapter.

The documentation now explicitly describes value-threaded full drains,
register-resident counted loops, static partial waits, the deferred fallback,
and fail-closed behavior.

## Validation

The compiler regression gate remains green:

```text
cuda-oxide-codegen:
  137 unit tests passed
  16 compile_to_ptx tests passed
  all remaining integration tests passed

dialect-nvvm:
  50 tests passed

mir-lower:
  168 unit tests passed
  95 integration tests passed
```

Both backend carrier probes remain green:

```text
32-value WGMMA carrier:
  PASS

64-value / two-slot wait_group<1> carrier:
  PASS
```

The WGMMA smoketest also passes both selected examples:

```text
wgmma           PASS
wgmma_mma_bf16  PASS

Pass: 2 / 2
Fail: 0 / 2
```

The updated public-API example reports:

```text
SUCCESS: BF16 WGMMA value-threaded and partial-wait lowering compiled.
```

`git diff --check` is clean.

## Scope

This PR intentionally does not add another WGMMA compiler transformation.

In particular, it does not:

- generalize the counted-loop recognizer to arbitrary CFGs;
- support dynamic `wait_group` values;
- infer arbitrary accumulator scheduling;
- add F16 or TF32 WGMMA lowering;
- provide Hopper runtime/numerical validation for the compile-only examples.

Those remain separate extensions beyond the scope of #605.


nihalpasham:
**Gist:** Step 6 closes the series: the public WGMMA docs now describe what the compiler actually does after steps 2 through 5, and a new compile-only kernel drives the partial-wait path through the entire Rust-to-PTX pipeline. #605's plan is delivered end to end.

Verified after the stack rebase: 
- the diff collapses to the five doc/example files; wgmma smoketest lane green.

Closes #605 (and the docs half of #399's ask). Merging last.

nihalpasham (approved):

PR #691, opened 2026-08-07
feat(cuda-device): let a ConstantMemory table be read at a runtime index

### LINKED ISSUE #690: ConstantMemory<[T; N]> cannot be indexed at a runtime index: get() returns T by value, so the table is copied into the local depot
## Problem

`ConstantMemory<T>` exposes exactly one accessor, [`get`], which returns `T`
**by value**. That is right for a scalar, but it leaves no way to index a
`ConstantMemory<[T; N]>` at a runtime index without materializing the whole
array first. A value has no address, so the copy is spilled to the thread's
local depot — one `st.local` per element in *every thread* — and the lookup then
reads thread-private memory.

The canonical CUDA constant-memory use case is exactly this:

```cpp
__constant__ float coeffs[256];
__global__ void k(...) { acc += coeffs[i & 255]; }   // runtime index
```

so today `#[constant]` cannot express the shape it exists for.

The existing examples do not show it because they index at *compile-time*
constants (`constant_memory_coeffs` does `c[0]`, `c[1]`; `constant_memory`
destructures `let [c0, c1, c2, c3] = COEFFS.get();`). SROA cleans those up
completely. Only a runtime index falls off the cliff.

## The documented workaround makes it worse

`get`'s doc currently advises:

> For large `T` (e.g. `ConstantMemory<[f32; 1024]>`) call `.get()` once before
> the loop and reuse the local

"Reuse the local" *is* the pathology — a local reused across a runtime-indexed
read is precisely what forces the array into the depot. There is no fast path
today, so the advice cannot be followed to a good outcome.

## Reproducer

```rust
#[constant]
static TABLE: ConstantMemory<[f32; 256]> = ConstantMemory::UNINIT;

#[kernel]
pub fn lut(input: &[u32], mut output: DisjointSlice<f32>) {
    let idx = thread::index_1d();
    let i = idx.get();
    let v = TABLE.get()[(input[i] >> 24) as usize & 255];
    if let Some(o) = output.get_mut(idx) { *o = v; }
}
```

PTX on `sm_86`, one kernel, one lookup:

```
555 instructions, .local depot 1024 bytes
  ld.const.b32  x256      <- reads the whole table
  st.local.b32  x256      <- writes it into the thread's own depot
  ld.local.b32  x1        <- the actual lookup
```

The same kernel reading through a reference to the storage is **42
instructions, no depot, one `ld.const.b32`**.

## The lowering is already correct

This is an API gap, not a codegen defect. A reference into a `#[constant]`
static already keeps address space 4 all the way down — the symbol stays
`addrspace(4) global` in LLVM IR (so a host `set_<name>` remains visible) and a
runtime-indexed read emits a single `ld.const.b32`. Nothing in `mir-lower` or
`llvm-export` needs to change; there is simply no safe way to obtain that
reference.

## Measured cost

A10G (sm_86, driver 595.71.05), 256-entry `f32` table, 64 dependent lookups per
thread over 8388608 threads, 50 timed launches after 10 warmup. All variants
bit-identical against a host reference:

| index pattern | `get()` copy | reference to storage | `const [f32; N]` (global) |
|---|---:|---:|---:|
| divergent | 16048.3 us | 7616.3 us | **321.7 us** |
| warp-uniform | 16065.4 us | **258.5 us** | 322.4 us |

So a reference is **2.1x** the throughput of the `get()` copy on a divergent
index and **62x** on a warp-uniform one.

Note the `get()` row barely moves between the two index patterns (16048 vs
16065): the per-thread depot copy dominates so completely that the access
pattern stops mattering.

## A second finding worth documenting

The table above also shows constant memory is **not** simply "faster memory",
which the current docs do not say anywhere:

- **warp-uniform index** — `.const` wins, 258.5 us vs 322.4 us (1.25x). One
  broadcast per warp. This is what the constant cache is for.
- **divergent index** — `.const` loses badly, 7616.3 us vs 321.7 us
  (**23.7x slower**). The constant cache serves distinct addresses in sequence,
  so a warp asking for up to 32 different entries serializes, while the same
  table in global memory coalesces and stays resident in L1.

A divergent lookup table should therefore be a plain `const TABLE: [f32; N]`,
which since #684 is materialized as one immutable device global read with
`ld.global.nc` — no host upload and no `#[constant]` needed. Choosing `.const`
by the qualifier rather than by the access pattern is a 23.7x mistake.

## Proposed fix

Add a borrowing accessor beside `get`:

```rust
pub fn get_ref(&self) -> &T
```

It is safe under the same contract `get` already relies on: device code only
reads this storage, so no `&mut` exists to alias, and the host writes only
between launches via `cuMemcpyHtoD`. Borrowing through the `UnsafeCell` keeps
the storage mutable as far as the compiler is concerned, so a `set_<name>`
between launches stays visible — the property that makes the wrapper necessary
in the first place.

Plus documentation: correct `get`'s "reuse the local" advice, and record which
address space suits which access pattern, with the numbers above.

I have this working and measured, and will open a PR.

## Not proposed here

`impl Index for ConstantMemory<[T; N]>` would be nicer at the call site than
`get_ref()[i]`, but it only helps arrays, while `get_ref` also covers structs.
Happy to add it on top if preferred.

vyncint:
Closes #690.

## Summary

`ConstantMemory<T>` had one accessor, `get`, returning `T` **by value**. Right for
a scalar, but it left no way to index a `ConstantMemory<[T; N]>` at a runtime
index: a value has no address, so the copy is spilled to the thread's local depot
— one `st.local` per element in *every thread* — and the lookup then reads
thread-private memory. `#[constant]` could not express the shape constant memory
exists for:

```cpp
__constant__ float coeffs[256];
acc += coeffs[i & 255];      // runtime index
```

This adds `get_ref(&self) -> &T`, which borrows the storage so the index stays in
constant space.

**No codegen change.** A reference into a `#[constant]` static already keeps
address space 4 all the way down: the symbol stays `addrspace(4) global` in LLVM
IR (so a host `set_<name>` remains visible) and the read emits one
`ld.const.b32`. Only the safe way to obtain that reference was missing. For the
reproducer kernel, **555 PTX instructions with a 1 KiB depot become 42 with
none**.

The existing examples never showed this because they index at compile-time
constants — `constant_memory_coeffs` does `c[0]`, `c[1]`; `constant_memory`
destructures `let [c0, c1, c2, c3] = COEFFS.get();`. SROA cleans those up
entirely. Only a runtime index falls off the cliff, so both keep using `get` here
and both still pass.

## Safety

Safe under the contract `get` already relies on: device code only reads this
storage, so no `&mut` exists to alias it, and the host writes only between kernel
launches via `cuMemcpyHtoD`, synchronized out-of-band relative to device
execution. Holding a shared reference for the duration of a launch observes no
write that `get` would not.

Borrowing *through* the `UnsafeCell` is what keeps this honest: the storage stays
mutable as far as the compiler is concerned, so reads are not folded to the
all-zero placeholder and a `set_<name>` between launches stays visible. That is
the property the wrapper exists for, and the emitted `ld.const.b32` confirms it
survives.

## Measured

A10G, from the evidence log's `nvidia-smi` block:

```
name, compute_cap, memory.total [MiB], driver_version
NVIDIA A10G, 8.6, 23028 MiB, 595.71.05
```

256-entry `f32` table, 64 dependent lookups per thread over 8388608 threads, 50
timed launches after 10 warmup. Every variant bit-identical against a host
reference computed with the same add order:

```
kernel                   reached by          index   us/launch   Glookups/s   correct
--------------------------------------------------------------------------------------
const_getval_div              get()      divergent     16048.3        33.45       yes
const_ref_div             get_ref()      divergent      7616.3        70.49       yes
global_div           const [f32; N]      divergent       321.7      1668.98       yes
const_getval_uni              get()   warp-uniform     16065.4        33.42       yes
const_ref_uni             get_ref()   warp-uniform       258.5      2076.72       yes
global_uni           const [f32; N]   warp-uniform       322.4      1665.15       yes
```

| index pattern | `get()` | `get_ref()` | speedup |
|---|---:|---:|---:|
| divergent | 16048.3 us | 7616.3 us | **2.11x** |
| warp-uniform | 16065.4 us | **258.5 us** | **62.14x** |

Note the `get()` row barely moves between the two index patterns (16048 vs
16065): the per-thread depot copy dominates so completely that the access pattern
stops mattering. That is the signature of the defect.

## The second half: `.const` is not simply faster memory

The same run settles which address space a table belongs in, which the docs did
not say anywhere. Constant memory is served by a broadcast-oriented cache:

| index pattern | `.const` (`get_ref`) | `.global` (`const [f32; N]`) | winner |
|---|---:|---:|---|
| warp-uniform | **258.5 us** | 322.4 us | `.const`, 1.25x |
| divergent | 7616.3 us | **321.7 us** | `.global`, **23.7x** |

On a warp-uniform index every lane wants the same entry and it is one broadcast —
what the constant cache is for. On a divergent index the cache serves distinct
addresses in sequence, so a warp asking for up to 32 different entries
serializes, while the same table in global memory coalesces and stays resident in
L1.

So a **divergent** lookup table should be a plain `const TABLE: [f32; N]`, which
since #684 is materialized as one immutable device global read with
`ld.global.nc` — no host upload and no `#[constant]` at all. Choosing `.const` by
the qualifier rather than by the access pattern is a 23.7x mistake, and both the
guidance and these numbers are now in `get_ref`'s docs.

## Docs corrected

`get`'s own advice was actively pointing at the pathology:

> For large `T` (e.g. `ConstantMemory<[f32; 1024]>`) call `.get()` once before the
> loop and reuse the local

A local reused across a runtime-indexed read is precisely what forces the array
into the depot, so the advice could not be followed to a good outcome. It now
points at `get_ref` instead. The module-level docs gained the two-accessor
distinction, and the module example uses `get_ref`.

## New example

`constant_memory_table` measures all six combinations of {`get`, `get_ref`,
`const [f32; N]`} x {divergent, warp-uniform} and checks each against its own
host reference. The two index patterns are made comparable rather than merely
different: both read the input buffer once and write the output once, and the
`_uni` kernels seed from `input[i & !(WARP - 1)]` — one warp-uniform load — so
only the index's divergence varies, not the memory traffic or the arithmetic
count.

## Validation

Local, all clean:

```
cargo fmt --check
cargo test -p cuda-device --lib --tests
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace
cargo test --doc -p cuda-device
bash scripts/check-dependency-licenses.sh            # OK, no new third-party crates
bash scripts/check-error-example-status.sh           # OK
```

On the A10G — scoped smoketest **9/9, 0 fail** (`.gpu-evidence/20260807T134040Z.log`),
covering every constant-memory example plus controls:

```
[ 1/ 9] array_constants         ... PASS      [ 6/ 9] tma_copy       ... PASS (skipped)
[ 2/ 9] const_table_lookup      ... PASS      [ 7/ 9] uninit_const   ... PASS
[ 3/ 9] constant_memory_coeffs  ... PASS      [ 8/ 9] vecadd         ... PASS
[ 4/ 9] constant_memory_table   ... PASS      [ 9/ 9] vectorization  ... PASS
[ 5/ 9] device_global           ... PASS
```

I scoped rather than swept: this adds an inherent method and touches no existing
code path, so the constant-memory examples plus controls are the relevant
surface. Say the word if you would rather see the full 188.

instruction-level evidence here is PTX only.

## Not in this PR

`impl Index for ConstantMemory<[T; N]>` would read better at the call site than
`get_ref()[i]`, but it only helps arrays while `get_ref` also covers structs. Glad
to add it on top if you prefer that spelling.


vyncint:
Note on coverage, so the evidence in this PR is not read as broader than it is.

I ran a full **190/190** sweep on the A10G for the array-constant stack
(#694 + #715 + #717) — `.gpu-evidence/20260808T080039Z.log`. **That sweep does not
cover this PR**: `get_ref` is on a separate branch and was not part of it.

This PR's execution evidence remains its own scoped run, **9/9**
(`.gpu-evidence/20260807T134040Z.log`), covering every constant-memory example
plus controls. Happy to run the full sweep for this branch too if you would rather
not merge on the scoped one — it is one box run.

nihalpasham:
**Gist:** Not a compiler change, just the missing safe accessor. `get()` returns the table by value, and a value has no address, so indexing it at a runtime index forced a copy of the whole table into each thread's private memory first. 
- references into `#[constant]` statics already stay in constant memory all the way through the compiler; `get_ref()` simply gives you a safe way to obtain one. 
- the doc fix matters as much as the API: `get()`'s old advice actively recommended the slow pattern.

- all 6 kernels bit-correct, SUCCESS marker; 
- the new accessor compiles to a single constant-memory load with no per-thread copy, and 
- the doc's advice about when constant memory loses to global (divergent indices) reproduces on this card.

Closes #690.

Merging.

nihalpasham (approved):

PR #694, opened 2026-08-07
perf(mir-importer): promote a field-less enum array constant like an integer one

### LINKED ISSUE #692: Field-less enum array constants are still materialized per thread, and &[Op; N] is rejected outright
## Problem

#684 stopped bare array constants being built per thread, but its element gate
admits only primitive scalars and nested arrays of those. A **field-less enum**
array is left out, even though such an array holds nothing but discriminants and
so is byte-identical to the `[u32; N]` of its values.

`const OPS: [Op; 256]` where `Op` is a field-less `#[repr(u32)]` enum is still
materialized into the per-thread local depot — one `st.local` per element in
*every thread*, for data the module image already carries — and the lookup then
reads thread-private memory.

There is also no way to spell around it. The reference form that works for scalar
tables is rejected outright:

```rust
const R: &[Op; 256] = &OPS;
// error: Compilation error: invalid input program.
```

`validate_ptr_to_array_constant_type` keeps enums out of the pointer-to-array
path too, so a field-less enum table has no fast spelling at all.

## Reproducer

```rust
#[derive(Clone, Copy)]
#[repr(u32)]
pub enum Op { Add = 3, Sub = 5, Mul = 7, Div = 11 }

const OPS: [Op; 256] = /* ... */;

#[kernel]
pub fn lut(input: &[u32], mut output: DisjointSlice<u32>) {
    let idx = thread::index_1d();
    let i = idx.get();
    let v = OPS[(input[i] >> 24) as usize & 255] as u32;
    if let Some(o) = output.get_mut(idx) { *o = v; }
}
```

## Measured

A10G (sm_86, driver 595.71.05), 64 dependent lookups per thread over 8388608
threads, 50 timed launches after 10 warmup, divergent index. Each enum kernel is
paired with a `[u32; N]` kernel over the same discriminants, so the pair isolates
the element type; all four are bit-identical against a host reference.

| entries | `[Op; N]` | `[u32; N]` | enum / u32 |
|---|---:|---:|---:|
| 16 | 1309.4 us | 361.4 us | 3.62x |
| 256 | 16073.3 us | 327.3 us | **49.11x** |

PTX for the 256-entry pair: **309 instructions with a 1 KiB local depot and 256
`st.local`** for the enum against **70 instructions with no depot** for the
integer table.

## Proposed fix

Widen the gate to admit field-less enums and nested arrays of them. The byte
image is rustc's evaluated allocation, so the discriminants are already exact;
what the gate has to establish is that the element is cheap to read back, which a
field-less enum is — the whole element *is* its discriminant.

Payload-carrying enums, tuples and structs should stay out, and not because the
byte image would be wrong — see the separate report I am filing about per-field
array copies, which dominate those shapes and would leave a promoted global as
dead weight beside an unchanged depot.

I have this working and measured, and will open a PR.

## Note on `array_constants`

`array_constants`' `verify-code-shape.sh` pins the enum array's current
`insertvalue` chain, so it needs updating to the new shape. Its runtime
assertions pass unchanged; the values were never in question.

vyncint:
Closes #692. Follow-up to #684.

## Problem

Issue #692 names two defects around field-less enum tables:

- A `const OPS: [Op; 256]` (field-less `#[repr(u32)]` enum) was rebuilt element by element in every thread's local depot, even though it is byte-identical to the `[u32; 256]` of its discriminants. Measured 49x slower than the equivalent u32 table at 256 entries.
- The reference spelling `const R: &[Op; 256] = &OPS;` did not even compile: "invalid input program".

```text
#[repr(u32)] enum Op { Add = 3, Sub = 5, Mul = 7, /* ... */ }
const OPS: [Op; 256] = [ /* ... */ ];
acc += OPS[(h >> 24) as usize & 255] as u32;   // runtime index

Before: 256 x st.local per thread, 1 KiB depot, 309 PTX instructions
        16073 us/launch (A10G), 49x the u32 table
After:  table lives once in the module image, one ld.global.nc per
        lookup, no depot, 70 instructions, 330 us/launch, 1.00x
```

## Fix

- The bare-array promotion gate becomes a predicate owned by this path (`promotable_array_element`): primitive scalars, field-less enums, and nested arrays of either, plus a byte-size agreement guard between the dialect's storage layout and rustc's evaluated allocation.
- The discriminant bytes come from rustc's own constant evaluation, so the values were never in question; the guard establishes that reading an element back is cheap.
- `enum_table_lookup` example with paired enum/u32 kernels so the integer row doubles as a control; bit-correct against a host reference.

## Maintainer commit on top (12429315)

- The pointer-to-array constant path now delegates its admission decision to the same `promotable_array_element` predicate, so `const R: &[Op; N] = &OPS` compiles and the two paths cannot drift apart. This closes the second defect of #692, making the "Closes" above accurate.
- New `enum256_ref_table` kernel reading through `&[Op; 256]` at a runtime index, verified bit-exact; on an RTX 5090 it runs at the bare enum table's exact cost, reading the same deduplicated immutable global.


vyncint:

```
===== SMOKETEST SUMMARY =====
Pass:    190 / 190
Fail:    0 / 190
Elapsed: 1218s
```

```
```

nihalpasham:
**Gist:** A table of C-style enums is stored as the same bytes as a table of the integers, so it deserves the same treatment #684 gave integer tables: keep it in the module image once instead of having every thread rebuild it in private memory. The PR widens the acceptance check the right way, with its own predicate plus a guard that the compiler's byte layout agrees with rustc's.

- the issue's second part: adding the same table behind a reference (`const R: &[Op; N] = &OPS`) did not just run slow, it refused to compile. 
- the reference path now uses the same acceptance check as the plain path (one check, both paths, cannot drift), with a unit test and a kernel that reads through the reference at a runtime index.

- all 5 kernels bit-correct; 
- the 49x perf problem is gone (1.00x at 16 entries, 1.19x at 256 on this card), and 
- the reference runs at the plain table's exact cost; 
- 927 mir-importer tests green; `array_constants` code-shape guard PASS; full sweep 188/188.

Closes #692.

Merging.

nihalpasham (approved):

PR #695, opened 2026-08-07
fix(mir-lower): lower scoped atomic load/store and fences to inline PTX

### LINKED ISSUE #696: DeviceAtomic load/store and AcqRel/SeqCst orderings fail libNVVM verification under --materialize-cubin
**Description**

`DeviceAtomic*::load` and `DeviceAtomic*::store` cannot be called: every call fails inside the libNVVM verifier under `--materialize-cubin`, because they lower to LLVM `load atomic` / `store atomic`, which libNVVM rejects outright. Separately, any `AcqRel` or `SeqCst` ordering emits an LLVM `fence`, which libNVVM also rejects. The result is that a documented part of the device atomics API is unusable in the only configuration that produces a real kernel, and the failure surfaces as a backend verifier message naming an LLVM basic block rather than an error at the Rust call site.

The `atomics` example in this repository is affected and does not currently build with `--materialize-cubin`, which is how I concluded the path is uncovered rather than that I was holding it wrong.

**Minimal reproducer**

The shipped example reproduces it with no new code:

```console
$ cargo oxide build atomics --arch sm_120 --materialize-cubin
error: [rustc_codegen_cuda] Failed to embed device artifact: libnvvm: libnvvm error in nvvmVerifyProgram: 6
       --- libNVVM log ---
       error: Error: :  Function `atomic_block_scope_acqrel_test' Basic Block `bb1':
         context:   fence syncscope("block") release
         Illegal instruction: fence
```

With the fence issue resolved, the same example then fails on `atomic_load_store_test` for the load/store reason. A minimal kernel showing that half on its own:

```rust
#[kernel]
pub fn publish(flags: &[u32]) {
    let tid = thread::threadIdx_x() as usize;
    // SAFETY: one thread per element.
    let slot = unsafe { DeviceAtomicU32::from_ptr((flags.as_ptr() as *mut u32).add(tid)) };
    slot.store(tid as u32 + 1, AtomicOrdering::Release);
    thread::sync_threads();
    let _seen = slot.load(AtomicOrdering::Acquire);
}
```

```console
error: Error: :  Function `publish' Basic Block `bb3':
  context:   store atomic i32 %v45, ptr %v34 syncscope("device") release, align 4
  Atomic loads/stores are not supported
```

**Expected behavior**

`DeviceAtomicU32::store(.., Release)` and `::load(.., Acquire)` compile and emit `st.release.gpu.b32` / `ld.acquire.gpu.b32`, and an `AcqRel` atomic emits `fence.acq_rel.gpu`. All three PTX instructions have existed since sm_70.

**Actual behavior**

The build fails in `nvvmVerifyProgram` with the messages above. Because `load` and `store` are unavailable, the only remaining way to re-read a location another thread is publishing to is `read_volatile`, and Rust's volatile semantics lower to a system-scope strongly-ordered access (`LDG.E.64.STRONG.SYS`) even when the data never leaves one device. In a hash-table kernel of mine that was 6 system-scope loads where the equivalent CUDA C++ had 2.

**Environment**

- GPU: RTX 5070 Ti (sm_120), also reproduced on RTX 5060
- CUDA driver version: 580.159.04, CUDA toolkit 13.0.88
- `rustc --version --verbose`: `rustc 1.96.0-nightly (55e86c996 2026-04-02)`, LLVM version 22.1.2
- `llc --version`: LLVM 21.1.8
- cuda-oxide at `20a5616`, and reproduced on current `main`

**Additional context**

PTX ISA has the scoped forms these should lower to:
[`ld`](https://docs.nvidia.com/cuda/parallel-thread-execution/#data-movement-and-conversion-instructions-ld),
[`st`](https://docs.nvidia.com/cuda/parallel-thread-execution/#data-movement-and-conversion-instructions-st),
[`fence`](https://docs.nvidia.com/cuda/parallel-thread-execution/#parallel-synchronization-and-communication-instructions-membar).

I have a fix in #695 that lowers all three to inline PTX, in the same style `convert_packed_atom_add` already uses in that file, plus lowering tests and a regression example that fails to compile on `main` and passes with the change. Happy to take it in whatever direction maintainers prefer, including splitting the fence change out.

One design question I could not settle from the outside: on plain load/store I reject `SeqCst` with an explanatory error rather than approximating it as a relaxed access plus a fence. Erroring seemed safer than quietly weakening someone's ordering, but the opposite choice is defensible and is yours to make.

realitymatrix:
Fixes #696.

## Problem

- libNVVM rejects three LLVM IR constructs outright: `load atomic` ("Atomic loads/stores are not supported"), `store atomic`, and `fence` ("Illegal instruction: fence"). Any kernel using `DeviceAtomic*::load/store` or a fence failed to build under `--materialize-cubin`, with the error naming an LLVM basic block instead of the Rust call site.
- On the normal llc path the requested scope was ignored: a `cta`- or `gpu`-scope atomic load/store came out as slow system-scope SASS.

```text
Before:
  $ cargo oxide build atomics --materialize-cubin --arch sm_120
  error: libnvvm: nvvmVerifyProgram: 6
    context: store atomic i32 %v45, ptr %v34 syncscope("device") release
    Atomic loads/stores are not supported

After:
  build succeeds; the PTX carries the exact scoped instructions:
    st.release.gpu.b32 [%rd1], %r2;
    ld.acquire.gpu.b32 %r3, [%rd1];
```

## Fix

- Lower atomic load/store to inline PTX `ld.{sem}.{scope}.{ty}` / `st.{sem}.{scope}.{ty}`, and fences to `fence.acq_rel.{scope}` (SeqCst fence via the `llvm.nvvm.membar` intrinsic), following the pattern `convert_packed_atom_add` already uses in the same file.
- This is the established shape: libcu++ (`cuda::atomic` under nvcc) emits per-op inline PTX at the header layer for exactly this reason. An inline-asm call passes through libNVVM untouched and llc handles it too, so one lowering serves both backends, and the user's scope is now honored on both.
- Scope comes from the op's `AtomicScope` attribute; PTX type and register class from operand width.
- Exact-template lowering tests plus a runnable `scoped_atomic_load_store` example registered in the NVVM-verify smoketest lane.

## Maintainer commit on top (1c4261a4)

- `~{memory}` clobber kept unconditionally, Relaxed included: without it LLVM may move plain same-address accesses across the asm, breaking the single-thread coherence Rust still guarantees for Relaxed. libcu++ keeps the clobber on relaxed too; the slowdown previously blamed on it was misattributed.
- SeqCst load/store now compile instead of erroring, using libcu++'s mapping fused into one template: `fence.sc.{scope}; ld.acquire.{scope}.{ty}` and `fence.sc.{scope}; st.release.{scope}.{ty}`.
- Float atomics (f16/f32/f64) are staged through the matching integer register class (`b16/h`, `b32/r`, `b64/l`) with LLVM-dialect bitcasts on each side. This restores `DeviceAtomicF16` load/store and makes the f32/f64 constraints type-correct.
- The example's AcqRel probe switched from `compare_exchange` (whose lowering emits no fence, contrary to its old comment) to `fetch_add(0, AcqRel)`, which genuinely exercises the fence-splitting RMW path.
- Template tests extended: SeqCst per scope, the f32 bitcast round trip, the f16 arm, and the Relaxed clobber.

## Scope

RMW atomics (`fetch_add`, CAS) are untouched: they still lower to native `atomicrmw`/`cmpxchg`, and the legacy-NVVM fail-closed guard from #101 still applies to them. This PR is the load/store/fence family only.


realitymatrix:
Quick orientation, since the diff touches three lowering paths at once.

The short version: DeviceAtomic*::load and ::store cannot be called today. They lower to LLVM load atomic / store atomic, and libNVVM rejects that form outright, so the build dies in the verifier with a message pointing at an LLVM basic block rather than at the user's code. The same applies to any AcqRel or SeqCst ordering, which emits an LLVM fence.

How I ran into it: I was chasing a performance gap between a Rust kernel and its CUDA C++ equivalent, and the SASS showed the Rust build using system scope loads where CUDA used GPU scope. The fix looked obvious, use the scoped atomic load the API already provides. It would not compile.

The change itself is small in principle. The PTX instructions all exist (ld.acquire.gpu, st.release.gpu, fence.acq_rel.cta, all since sm_70), so these three ops now emit inline PTX, in the same style convert_packed_atom_add already uses a few lines below in the same file.

Worth flagging separately: the atomics example in this repo does not currently build with --materialize-cubin. That is how I confirmed the path was uncovered, and it is why this PR adds a new example rather than only unit tests. The new one fails to compile on main and passes with the change.

Left as draft because two decisions deserve your opinion rather than mine:

Acquire, Release and AcqRel all map to fence.acq_rel, since PTX has no separate acquire or release fence.
I reject SeqCst on plain load/store with an explanatory error instead of approximating it as a relaxed access plus a fence. Erroring felt safer than quietly weakening someone's ordering, but you may prefer the approximation.

Happy to split the fence change into its own PR if you would rather review them separately.

realitymatrix:
Quick orientation, since the diff touches three lowering paths at once. Filed the bug separately as #696.

The short version: `DeviceAtomic*::load` and `::store` cannot be called today. They lower to LLVM `load atomic` / `store atomic`, and libNVVM rejects that form outright, so the build dies in the verifier with a message pointing at an LLVM basic block rather than at the user's code. The same applies to any `AcqRel` or `SeqCst` ordering, which emits an LLVM `fence`.

How I ran into it: I was chasing a performance gap between a Rust kernel and its CUDA C++ equivalent, and the SASS showed the Rust build using system scope loads where CUDA used GPU scope. The fix looked obvious, use the scoped atomic load the API already provides. It would not compile.

The change itself is small in principle. The PTX instructions all exist (`ld.acquire.gpu`, `st.release.gpu`, `fence.acq_rel.cta`, all since sm_70), so these three ops now emit inline PTX, in the same style `convert_packed_atom_add` already uses a few lines below in the same file.

Worth flagging separately: the `atomics` example in this repo does not currently build with `--materialize-cubin`. That is how I confirmed the path was uncovered, and it is why this PR adds a new example rather than only unit tests. The new one fails to compile on `main` and passes with the change, and it is registered in `NVVM_VERIFY_EXAMPLES` so the compile-only lane exercises it without a GPU.

Two decisions deserve a maintainer's opinion rather than mine, and are the main things I would like reviewed:

1. Acquire, Release and AcqRel all map to `fence.acq_rel`, since PTX has no separate acquire or release fence.
2. I reject `SeqCst` on plain load/store with an explanatory error instead of approximating it as a relaxed access plus a fence. Erroring felt safer than quietly weakening someone's ordering, but you may prefer the approximation.

Happy to split the fence change into its own PR if you would rather review them separately.

realitymatrix:
Pushed one more commit, `0aa4e86`. It came out of using this feature in anger, so the context is worth more than the diff.

## What I was doing

I maintain a benchmark comparing the same hash-blocked TSDF kernel written in CUDA C++, Rust via cuda-oxide, and Triton. The Rust arm was 1.5x slower than the CUDA one on its insert path and I could not explain why. Eleven hypotheses died: register pressure (Rust uses *fewer*), `f32::clamp`, `read_volatile`, static instruction count (1.02x parity, 520 against 528), dynamic compare-exchange count (Rust issues 12% *fewer*), memory-ordering scope, the shared counter, the fence, the publication store, the publication spin, and occupancy (identical, 84.3% both).

Every count favoured Rust and it was still slower. Nsight Compute found it in one run: `long_scoreboard` was the only warp stall that worsened, and it worsened by more than the whole gap. L1 sector hit rate was 28.9% against the CUDA arm's 55.9%, with identical request counts and 1.70x the sectors pushed to L2.

The cause was mine, not yours. I was reading a hash key and a published block index with `DeviceAtomic*::load(Relaxed)` because those locations are shared. A scoped atomic load is coherent at its scope, an L1 is not coherent across SMs, so it bypasses L1 on every call however weak the ordering. The CUDA version reads both plainly and relies on the algorithm: a stale key costs one extra probe, and a stale `-1` just enters the wait loop where the atomic load *does* observe the publication.

Reading plainly first and falling back to `load` only while genuinely waiting restored the hit rate to 54.1%, L2 sectors to 552,775 against CUDA's 568,128, and `long_scoreboard` to 11.06 against 11.54. Across a 12-cell, 3-pass, 2-GPU sweep the Rust arm went from 1.05-1.27x of hand-written CUDA C++ on the full integrate path to **1.01-1.11x**, and 1.01-1.03x on real depth data.

## What is in the commit

**A doc note on `cuda_device::atomic`**, which is the part I would actually argue for. Now that these methods can be called, people will reach for them, and the reflex to use an atomic whenever a location is shared is exactly what cost me 24%. The note states the characteristic plainly and gives the probe-loop shape as the worked example. I could not find this written down anywhere and it is not inferable from the type signatures.

**A lowering change: emit `~{memory}` only for ordering-carrying atomics.** A memory clobber forces every value LLVM was holding in a register to spill and reload around the access. That is right for an access whose purpose is to order other memory operations, and over-conservative for a relaxed one, which orders nothing, in exactly the construct most likely to sit in a hot loop.

I want to be straight about this second one: I measured it in isolation and it moved my benchmark by about 1%, which is noise. It is justified on principle, not on a demonstrated win. If you would rather keep the clobber unconditionally for safety, say so and I will drop it; the doc note stands on its own. The lowering test now asserts the constraint string per ordering rather than assuming a clobber everywhere.

## Verification unchanged

`cargo test --workspace` passes (3358 tests). The `atomics` example still passes 20/20, `scoped_atomic_load_store` still fails to compile on `main` and passes here, and `check-example-smoketest-contract.sh` and `check-dependency-licenses.sh` both pass.

The two open questions from before still stand and are the main things I would like reviewed: mapping Acquire, Release and AcqRel all to `fence.acq_rel`, and rejecting `SeqCst` on plain load/store rather than approximating it.

nihalpasham:
**Gist:** libNVVM rejects LLVM load atomic/store atomic/fence outright, so they have to stop being IR instructions; libcu++ emits per-op inline PTX at the header layer for exactly this reason. As a bonus the `llc` path now honors the requested scope instead of defaulting to system scope.

**I pushed two commits on top:**

- the `~{memory}` clobber is unconditional again, Relaxed included: without it LLVM may move plain same-address accesses across the asm, breaking single-thread coherence Rust still guarantees (the slowdown blamed on the clobber was misattributed);
- `SeqCst` load/store now compile instead of erroring, per libcu++: `fence.sc.{scope}` fused with an acquire load / release store in one template;
- f16/f32/f64 stage through integer register classes with bitcasts on each side, restoring `DeviceAtomicF16` and making the float constraints type-correct; template tests extended for all of the above;
- the example's fence probe now uses the fused `SeqCst` store/load pair. The original compare_exchange probe's comment was wrong (`cmpxchg` emits no fence), a fetch_add replacement trips the legacy-NVVM RMW guard in the GPU-less examples-compile lane, and `core::sync::atomic::fence` does not translate yet (importer gap, filing separately). The SeqCst pair keeps runtime `fence.sc.gpu` coverage while staying in the load/store family, so the example compiles on BOTH NVVM generations; legacy libNVVM verified it and wrote LTOIR for compute_90 here.

- atomics 20/20 normally, and the issue's headline repro now compiles and 
- runs green under `--materialize-cubin` (it does not compile at all without this PR); 
- `scoped_atomic_load_store` green in both modes; full sweep 188/188; 168 + 94 mir-lower tests green.

RMW atomics through legacy NVVM stay behind the #101 guard; that per-op validation is separate work. Fixes #696.

Merging.

nihalpasham (approved):

PR #699, opened 2026-08-07
feat(cuda-core): expose per-kernel register/local/const usage and the stack-size limit

### LINKED ISSUE #697: Per-kernel register/local/const usage and the stack-size limit are unreachable from cuda-core

`CudaFunction::attribute` already wraps `cuFuncGetAttribute`, but it is private and `cu_function` is `pub(crate)`, so only four attributes are reachable. Register count, local (stack+spill) size, and constant size — the numbers `ptxas -v` prints — cannot be read at runtime. There is also no wrapper for `cuCtxSetLimit(CU_LIMIT_STACK_SIZE, ..)`.

**Proposed API / behavior**

```rust
let f = module.load_function("my_kernel")?;
println!("{} regs, {} B local, {} B cmem",
         f.num_registers()?, f.local_size_bytes()?, f.const_size_bytes()?);

ctx.set_stack_size(32 * 1024)?;
assert_eq!(ctx.stack_size()?, 32 * 1024);
```

**Alternatives considered**

Running `ptxas -v` at build time. That needs a toolkit installed, reports on the PTX rather than the loaded module, and cannot see what the driver's JIT actually did. The stack limit is worse: its reservation is `stack_bytes × sm_count × max_threads_per_sm` and is invisible to every allocation API — it only shows up as less memory available to the next `cuMemAlloc`. On a 40-SM card at the 128 KB default that is ~5.4 GB, and it scales with SM count, so on newer cards the invisible reservation grows. Today you can neither read the per-kernel frame that drives it nor set the limit.

**Additional context**

CUDA driver API: `cuFuncGetAttribute`, `cuCtxGetLimit`/`cuCtxSetLimit`.

nh13:
Closes #697.

## Summary

`CudaFunction::attribute` already wraps `cuFuncGetAttribute`, but it is private and `cu_function` is `pub(crate)`, so callers can reach only `max_threads_per_block`, `static_shared_memory_bytes`, `max_dynamic_shared_memory_bytes` and `required_cluster_dimensions`. This adds the three remaining resource figures — the ones `ptxas -v` prints — plus a wrapper for the device stack-size limit, so a caller can right-size it instead of guessing. The stack limit reserves `stack_bytes × sm_count × max_threads_per_sm` of device memory, and that reservation is invisible to every allocation API: it only shows up as less memory available to the next allocation.

## Changes

- `CudaFunction::num_registers`, `local_size_bytes`, `const_size_bytes`, each a one-line call to the existing private `attribute` helper, in the style of the four accessors above them.
- `CudaContext::stack_size` / `set_stack_size` wrapping `cuCtxGetLimit`/`cuCtxSetLimit` for `CU_LIMIT_STACK_SIZE`, matching that file's `bind_to_thread` + `MaybeUninit` idiom.
- `CudaContext::max_threads_per_multiprocessor`, so the reservation formula is computable from the crate. Happy to drop this if you'd rather keep the diff to the stack limit itself — though see the device results below, where it is what makes the reservation figure come out right.

Additive only; no existing signature changes and no new dependencies.

## Testing

- [x] `cargo check -p cuda-core` → 0
- [x] `cargo fmt -p cuda-core -- --check` → 0
- [x] `cargo clippy -p cuda-core --all-targets` → 0
- [x] `cargo doc -p cuda-core --no-deps` → 0
- [x] **Exercised against a real device** (Tesla T4, `sm_75`, CUDA 13.2). Loaded a 22-kernel PTX module and cross-checked every new accessor against `ptxas -v` on the same PTX: `num_registers` and `local_size_bytes` agree exactly on **22/22 kernels, 0 mismatches** (e.g. `spine_chain` 218 regs / 2048 B, `spine_anchor` 128 / 2208, `seed_intv_kernel` 126 / 144). Two independent producers — the driver's JIT and the offline assembler — reporting the same numbers.
- [x] `stack_size()` reads CUDA's documented 1024 B default; `set_stack_size(32768)` and `set_stack_size(16384)` each read back exactly, unclamped.
- [x] `multiprocessor_count() × max_threads_per_multiprocessor() × 32 KB` = 1.34 GB on that T4, matching an independent instrument in the calling application that computes the same reservation. This is the check that justifies including `max_threads_per_multiprocessor`.
- [ ] `const_size_bytes` returns 0 for all 22 kernels. That is *correct* for this module — it declares no user `__constant__` memory — but it therefore does not distinguish a correct 0 from an unimplemented one. The other two accessors go through the identical one-line path, so I believe it is fine, but I would not claim it is covered.
- [ ] `cargo test --workspace` — not run: developed on macOS with headers only, where linking any cuda-core test binary fails with `ld: library 'cuda' not found`, identically on a clean tree.
- [ ] New example added — none; say the word if you'd like one printing a loaded kernel's resource usage.

## Checklist


nihalpasham:
**Gist:** The per-kernel numbers `ptxas -v` prints (registers, local bytes, constant bytes) plus the device stack-size limit, each a one-line wrapper over a helper that already existed privately, in the file's exact house style. The 22-kernel cross-check against ptxas on a T4 is the evidence that makes review easy.

**I pushed one commit on top (docs only):**

- `const_size_bytes` reports user-declared constant memory, not the `cmem[0]` parameter bank the doc claimed (your own T4 zeros for kernels that all have parameters show the difference);
- `local_size_bytes` is the per-thread local memory size; dropped the "plus spill" wording (spill counts are traffic through that memory, not extra bytes on top);
- noted the stack limit is device-wide state shared by the whole process, next to the reservation formula that motivates `max_threads_per_multiprocessor`.

**Verified live on an RTX 5090:** 
- sensible figures for a simple kernel (12 registers, zero local, zero constant), and 
- the stack limit sets, reads back, and restores cleanly.

Closes #697. (#707/#708, the spill diagnostics, stay open; they are compiler-side.)

Merging.

nihalpasham (approved):

PR #700, opened 2026-08-07
fix(cuda-bindings): select targets/ by full build target, and reach Tegra

### LINKED ISSUE #698: toolkit_target_dir() matches only the triple's architecture, so non-Linux targets probe Linux paths and Tegra is unreachable
**Description**

`toolkit_target_dir` in `crates/cuda-bindings/build.rs` matches only the first segment of the `TARGET` triple, so the architecture alone selects the CUDA `targets/<dir>` layout and the OS is never consulted. Every non-Linux target claims a Linux directory, and CUDA for Tegra is unreachable entirely.

**Minimal reproducer**

Build `cuda-bindings` for any non-Linux target:

```
aarch64-apple-darwin      -> probes {toolkit}/targets/sbsa-linux/include
x86_64-pc-windows-msvc    -> probes {toolkit}/targets/x86_64-linux/include
```

**Expected behavior**

Non-Linux targets should probe no `targets/` tree. aarch64 Linux should be able to resolve a Tegra install, which lives at `targets/aarch64-linux/`.

**Actual behavior**

Both `find_cuda_include_dir` and `collect_lib_paths` are handed a path that cannot exist on those platforms. Separately, `aarch64` hardcodes `sbsa-linux` — the ARM *server* tree — so CUDA for Tegra (Jetson, Drive) under `targets/aarch64-linux/` is never probed. Both build natively as `aarch64-unknown-linux-gnu`, so the triple alone cannot tell them apart.

**Environment**

- GPU: n/a (build-time only)
- CUDA driver version: n/a
- `rustc --version --verbose`: nightly-2026-04-03

**Additional context**

Selecting a single `targets/` directory is deliberate — see the doc comment on `find_cuda_include_dir` and #172; globbing `targets/*` was proposed in #88 and rejected because `sbsa-linux` sorts before `x86_64-linux` on multi-target installs.

nh13:
Closes #698.

## Problem

`toolkit_target_dir()` matched only the first segment of the target triple, so architecture alone decided the CUDA `targets/<dir>` layout and the OS was never consulted:

- `aarch64-apple-darwin` probed `targets/sbsa-linux/` (a Linux-only tree) and `x86_64-pc-windows-msvc` probed `targets/x86_64-linux/`.
- CUDA for Tegra was unreachable: Jetson and Drive install under `targets/aarch64-linux/`, while aarch64 was hard-wired to the ARM server tree (`sbsa-linux`).

```text
# A CUDA-for-Tegra install ships only:
#   /usr/local/cuda/targets/aarch64-linux/include/cuda.h

Before: probes include/, then targets/sbsa-linux/  -> not found, error
After:  probes include/, targets/sbsa-linux/, targets/aarch64-linux/
        -> found; matching lib dirs join the link search path
```

## Fix

- Selection keys off `CARGO_CFG_TARGET_ARCH` + `CARGO_CFG_TARGET_OS` (robust against 3- vs 4-field triples; correctly excludes Android), returns nothing for non-Linux, and yields ordered candidates for aarch64 Linux: `["sbsa-linux", "aarch64-linux"]`, walked by the existing `cuda.h` existence probe. `sbsa-linux` stays first, so installs carrying both trees resolve exactly as today.
- The table lives in one shared file (`toolkit_target.rs`) used by both `build.rs` consumers and a synthetic-tree integration suite. Preserves the single-architecture anti-glob decision from #172, with a test asserting it.
- The shape matches CMake's `FindCUDAToolkit` selection almost exactly.

## Maintainer commit on top (b65208b6)

- `CUDA_TOOLKIT_TARGET_DIR` env override (the nvcc `-target-dir` / CMake `CUDAToolkit_TARGET_DIR` equivalent): if set, it is the single `targets/` candidate, still existence-probed so typos fail with the existing clear error; `cargo:rerun-if-env-changed` wired; two override tests added. This settles the dual-tree ambiguity (a Jetson cross sysroot carrying both trees can now pin `aarch64-linux`).
- The two hand-lockstep copies of the old logic are mirrored to the new selection: cargo-oxide's doctor (`cuda_header_candidates`) and cuda-core's `build.rs`. Without this, a Tegra user would build fine and then `cargo oxide doctor` would exit 1 claiming cuda.h is missing. Their lockstep comments now name `toolkit_target.rs`.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -3931,7 +3931,13 @@ pub fn doctor(ctx: &Context) {
     // cuda-bindings' build script (issue #87).
     print!("CUDA headers (cuda.h)... ");
     let toolkit = cuda_toolkit_root(|var| std::env::var(var).ok());
-    let header_candidates = cuda_header_candidates(&toolkit, std::env::consts::ARCH);
+    let target_dir_override = std::env::var("CUDA_TOOLKIT_TARGET_DIR").ok();
+    let header_candidates = cuda_header_candidates(
+        &toolkit,
+        target_dir_override.as_deref(),
+        std::env::consts::ARCH,
+        std::env::consts::OS,
+    );
     match header_candidates.iter().find(|path| path.is_file()) {
         Some(found) => println!("✓ {}", found.display()),
         None => {
@@ -4232,9 +4238,9 @@ pub fn doctor(ctx: &Context) {
 /// variable among `CUDA_TOOLKIT_PATH`, `CUDA_HOME`, else `/usr/local/cuda`.
 ///
 /// Kept in lockstep BY HAND with `crates/cuda-bindings/build.rs`
-/// (`cuda_toolkit_dir` / `find_cuda_include_dir` / `toolkit_target_dir`):
-/// doctor cannot import that probe because build.rs logic is not a library.
-/// If the build.rs discovery changes, mirror it here.
+/// (`cuda_toolkit_dir` / `find_cuda_include_dir`): doctor cannot import that
+/// probe because build.rs logic is not a library. If the build.rs discovery
+/// changes, mirror it here.
 fn cuda_toolkit_root(mut get_env: impl FnMut(&str) -> Option<String>) -> String {
     ["CUDA_TOOLKIT_PATH", "CUDA_HOME"]
         .iter()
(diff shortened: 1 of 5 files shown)


nihalpasham:
**Gist:** The toolkit directory is now chosen from both the target's CPU architecture and its operating system, instead of the first word of the triple. Concretely: Mac and Windows builds stop probing Linux-only directories, and Jetson/Drive installs (which live under `targets/aarch64-linux/`) finally resolve, with the ARM-server tree still preferred when both exist. Matches CMake's `FindCUDAToolkit`, with tests against synthetic toolkit trees.

- The `CUDA_TOOLKIT_TARGET_DIR` override you offered (same idea as nvcc's `-target-dir`), so a machine carrying both ARM trees can pin one; wired for rebuilds, two tests.
- The same old logic existed hand-copied in two more places: `cargo oxide doctor` and cuda-core's build script. Mirrored the fix into both, so a Tegra user cannot end up with a working build and a doctor that says cuda.h is missing. The keep-in-lockstep comments now point at the shared file.

Verified here: 10/10 toolkit tests linking against the real toolkit on this box; both mirrors compile clean.

Closes #698. #139 (Windows lib paths) stays open, as you noted.

Merging.

nihalpasham (approved):

Issue #701, opened 2026-08-07
#[cuda_module] and #[kernel] emit host-only code, so a device-only kernel crate still needs cuda.h


nh13:
**What problem does this solve?**

`#[cuda_module]` and `#[kernel]` emit host-only code into the consumer crate — `LoadedModule`, `load`/`load_named`/`from_module`, the launch/prepare-launch/constant methods, and the `CudaKernel` marker impl. All of it names `::cuda_host` / `::cuda_core`, so a crate that only *compiles kernels* and never loads or launches anything still has to depend on cuda-host → cuda-core → cuda-bindings → `cuda.h`. That makes a device-only kernel crate unbuildable anywhere without a CUDA toolkit, including on a developer machine doing a type-check.

**Proposed API / behavior**

```toml
[dependencies]
cuda-device = "0.2"
cuda-macros = { version = "0.2", default-features = false }  # no `host`
```

```rust
#[kernel]
pub fn scale(out: *mut f32, n: usize) { /* ... */ }   // compiles with no cuda-host in the tree
```

**Alternatives considered**

Adding cuda-host purely to satisfy the expansion, which is what we do today; it pulls bindgen and a `cuda.h` requirement into a crate that emits no host code. Splitting the kernels into a crate that avoids `#[cuda_module]` doesn't help, because bare `#[kernel]` emits `impl cuda_host::CudaKernel` too.

**Additional context**

The `async` feature in `cuda-macros/src/lib.rs` already gates emitted host items the same way, so there is an in-file precedent for the shape.

vyncint:
This appears to be resolved by #702, which put the generated host surface behind a default-on `host` feature — the exact shape proposed here.

Verified on `15fa9d9d` with a throwaway crate outside the repo, taking the dependency lines from this issue verbatim:

```toml
[dependencies]
cuda-device = { path = ".../crates/cuda-device" }
cuda-macros = { path = ".../crates/cuda-macros", default-features = false }
```

```rust
#![no_std]
use cuda_device::{DisjointSlice, kernel, thread};

#[kernel]
pub fn scale(input: &[f32], mut out: DisjointSlice<f32>) {
    let idx = thread::index_1d();
    let i = idx.get();
    if let Some(o) = out.get_mut(idx) { *o = input[i] * 2.0; }
}
```

Both halves of the ask hold:

[...]

PR #702, opened 2026-08-07
feat(cuda-macros): put the generated host surface behind a default-on `host` feature

nh13:
Partially addresses #701 (the non-generic surface; generic `#[kernel]`s remain host-coupled and are tracked in the issue).

## Problem

`#[cuda_module]` and `#[kernel]` emit host-only code into the consumer crate: `LoadedModule`, `load`/`load_named`/`from_module`, the launch and constant methods, the nested-module loader, and the `CudaKernel` marker impl. All of it names `::cuda_host` or `::cuda_core`, so a crate that only compiles kernels still drags in cuda-host -> cuda-core -> cuda-bindings -> `cuda.h` at build time.

```text
        #[cuda_module] mod kernels { #[kernel] fn scale(...) {...} }
                            | expands to
   TODAY (always)           |    WITH default-features = false
   kernels        -> PTX    |    kernels        -> PTX
   PTX-merge markers        |    PTX-merge markers
   LoadedModule + load()    |
   launch()/prepare_..()    |         (nothing else)
   impl CudaKernel ...      |
        |                   |
   needs cuda.h + toolkit   |    needs nothing host-side
```

## Fix

- New `host` feature on cuda-macros, **in `default`**, following the `async` feature precedent in the same file. Existing consumers expand byte-identically.
- Gates exactly the host surface; kernel bodies and the PTX-merge markers the codegen collector consumes are never gated.
- The decision is threaded as an `emit_host: bool` parameter rather than a `cfg!` read, because the dev-dep cycle on cuda-host re-unifies the feature under `cargo test --no-default-features`; the parameter makes both settings testable in one build (five new unit tests).
- Workspace: the `cuda-macros` entry takes `default-features = false`; `cuda-host` names `features = ["host"]` explicitly.

## Maintainer commit on top (67c1c129)

- Documented the mixed-graph pattern in the Cargo.toml feature comment and a new crate-level doc: a device-only kernel crate consumed by a host application must forward the feature itself (`host = ["dep:cuda-host", "cuda-macros/host"]`), because proc-macro features unify globally per build graph.
- One-line note beside the generic-kernel emitters: generics stay host-coupled by design for now (TypeId machinery) and are deliberately not gated.
- Verified with a scratch crate outside the repo: `cargo check` with `default-features = false` passes with zero cuda-host/cuda-core/cuda-bindings in `cargo tree`; adding `--features cuda-macros/host` fails with the documented E0433.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,11 @@ nvvm-transforms = { path = "crates/nvvm-transforms" }
 cuda-device = { path = "crates/cuda-device" }
 cuda-artifact-finalizer = { path = "crates/cuda-artifact-finalizer" }
 cuda-host = { path = "crates/cuda-host" }
-cuda-macros = { path = "crates/cuda-macros" }
+# No default features: `host` is opt-in per member, so that depending on
+# cuda-device does not unify the host surface back on for kernel-only crates.
+# `cuda-host` names `features = ["host"]`, so anything that loads or launches a
+# module still gets it.
+cuda-macros = { path = "crates/cuda-macros", default-features = false }
 cuda-bindings = { path = "crates/cuda-bindings" }
 cuda-core = { path = "crates/cuda-core" }
 cuda-async = { path = "crates/cuda-async" }
--- a/crates/cuda-device/Cargo.toml
+++ b/crates/cuda-device/Cargo.toml
@@ -8,6 +8,11 @@ repository.workspace = true
 description = "Device-side intrinsics and types for CUDA kernels in Rust"
 
 [dependencies]
+# Inherits the workspace entry, which sets `default-features = false` and so
+# leaves the `host` feature off. That matters here specifically: cuda-device is
+# in every kernel crate's graph, so requesting `host` at this edge would unify
+# it back on for all of them and defeat the feature. `cuda-host` requests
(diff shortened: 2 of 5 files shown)


nihalpasham:
**Gist:** The right split at the right boundary: everything the macros generate that mentions the host crates now sits behind a `host` feature that is on by default, so existing users see zero change, while a crate that only defines kernels can turn it off and build with no CUDA installation at all. Kernel bodies and the markers the compiler collects are never gated. Threading the on/off decision as a parameter instead of reading the feature flag inline is what makes both settings testable in one build.

- documented the one sharp edge where users will look: cargo features are global per build, so a kernel-only crate used inside a host application must forward the feature itself (`host = ["dep:cuda-host", "cuda-macros/host"]`) or it hits a confusing build error.
- a one-line note in code that generic kernels intentionally stay coupled to the host crates for now.

**Verified with a scratch crate outside the repo:** 
- a kernel-only build has zero host crates in its dependency tree, and turning the feature on fails as documented.

On your two questions: 
- keeping the feature approach (matches the existing `async` feature in the same file; the attribute variant can be revisited if unification bites in practice), and 
- holding off on a new std helper crate for generics. 
- Partially addresses #701, which stays open for the generic-kernel half.

Merging.

nihalpasham (approved):

PR #704, opened 2026-08-07
feat(mir-lower): carry the originating Rust static onto shared-memory globals

### LINKED ISSUE #703: Static shared-memory globals are anonymous, so a kernel's shared footprint cannot be attributed back to source

`create_shared_global` (`mir-lower/src/convert/ops/memory.rs`) mints every static shared allocation's symbol from a process-global `AtomicUsize`, so the emitted `__shared_mem_N` carries no hint of its origin. A consumer attributing a kernel's shared-memory footprint back to source sees only anonymous blocks and a total, and cannot tell which Rust `SharedArray` or `Barrier` accounts for which one. That makes occupancy tuning guesswork: you can see the kernel's shared bytes per block but not which declaration to shrink.

**Proposed API / behavior**

Keep the symbol anonymous, but record the originating Rust path alongside it, e.g. in the exported LLVM IR:

```llvm
; shared source: my_kernel::TILE
@__shared_mem_0 = addrspace(3) global [64 x i32] zeroinitializer, align 4
```

**Alternatives considered**

Renaming the symbol after the static. A Rust path is not a legal LLVM identifier in general, and `__shared_mem_N` is matched literally by existing export tests, so a rename is a much larger change for the same information. Counting `SharedArray` declarations by hand against a PTX total, which is what we do now and which stops working as soon as two arrays are the same size.

**Additional context**

The name is reachable already: `GlobalAlloc::from(alloc_id)` -> `GlobalAlloc::Static(static_def)` -> `static_def.name()`, used by `static_target_from_constant` in `mir-importer`.

nh13:
Closes #703.

## Summary

`create_shared_global` mints every static shared allocation's symbol from a process-global counter, so `__shared_mem_N` carries no hint of its origin and a consumer attributing a kernel's shared-memory footprint back to source sees only anonymous blocks. This plumbs the Rust path through as an optional, purely descriptive `source_name` alongside the existing `alloc_key`, and renders it as a comment above the definition in the exported IR. The emitted symbol is deliberately unchanged.

```llvm
; shared source: my_kernel::TILE
@__shared_mem_0 = addrspace(3) global [64 x i32] zeroinitializer, align 4
```

## Changes

- **mir-importer** — `shared_static_source_name` resolves the constant to `GlobalAlloc::Static(..).name()` at both shared alloc sites (`SharedArray`, `Barrier`). Deliberately infallible and separate from `static_target_from_constant`: a diagnostic label must never turn a translation into a hard error.
- **dialect-mir** — optional `source_name: StringAttr` on `MirSharedAllocOp`.
- **mir-lower** — read it and stamp it on the `llvm::GlobalOp`. `create_shared_global` gained a `SharedAllocSpec` parameter struct because the extra field tripped `clippy::too_many_arguments`; it mirrors the `DeviceGlobalSpec` already in that file.
- **llvm-export** — `GlobalOpExt::{set_,}shared_source_name` and the comment line. The attribute key follows the `cuda_oxide_global_source` precedent but is never indexed and has no uniqueness requirement; nothing in code generation reads it.

No rename: a Rust path is not a legal LLVM identifier in general, and `__shared_mem_N` is matched literally by existing export tests. Not covered: dynamic shared memory (`MirExternSharedOp` has no single originating static to name).

**Please push back on the delivery vehicle.** A comment in the `.ll` is grep-parseable only and does not reach the PTX. I chose it because the `.ll` is already written on every build and it needs no new artifact-format version, but if the intent is for tooling to consume this from PTX, post-processing in `cuda-oxide-codegen/src/ptx.rs` is the honest answer and I'd rather be told now.

**One question on the rendered name.** The path is the true post-macro-rewrite Rust path, which means it embeds the generated entry symbol:

```
addressof_sharedarray::kernels::cuda_oxide_codegen_v1_cuda_oxide_kernel_246e25db_sharedarray_late_use::SCRATCH
```

Accurate, but noisy, and it carries the codegen hash. I can strip that segment down to the user-facing kernel name if you prefer; I left it verbatim rather than guess at a normalization you might not want.

## Testing

- [x] `cargo fmt --all -- --check` → 0
- [x] `cargo check --workspace --all-targets` → 0
- [x] `cargo clippy -p dialect-mir -p llvm-export -p mir-lower -p mir-importer --all-targets -- -D warnings` → 0
- [x] `cargo test -p dialect-mir -p llvm-export -p mir-lower -p mir-importer` → 0, **1319 passed, 0 failed** (7 new tests)
- [x] **End-to-end on Linux + CUDA 13.2 (T4, `sm_75`)**, which is what closes the gap this change would otherwise carry. Built `examples/addressof_sharedarray` — it declares two distinct named shared statics in one kernel — with a backend built from this branch, and read the emitted `.ll`:

  ```llvm
  ; shared source: …::sharedarray_late_use::SCRATCH
  @__shared_mem_1 = addrspace(3) global [32 x i32] zeroinitializer, align 4
  ; shared source: …::sharedarray_late_use::OUTPUT_NORM
  @__shared_mem_0 = addrspace(3) global [1 x float] zeroinitializer, align 4
  ```

  Both statics resolve, each onto its own global, symbols unchanged. The array types corroborate the mapping independently: `SCRATCH` is `SharedArray<u32, THREADS>` with `THREADS = 32`, and `OUTPUT_NORM` is `SharedArray<f32, 1>`. This exercises the `mir-importer` half, which no workspace test can reach — it needs real rustc MIR constants, i.e. the `rustc-codegen-cuda` driver.
- [ ] `cargo test --workspace` — dies at cuda-bindings with `ld: library 'cuda' not found` before any test binary runs, identically on a clean tree (macOS, headers only).

Mutation-tested and gated on the exit code, not on grepping output: suppressing the mir-lower stamp fails 3 of the 4 new lowering tests (101); suppressing the exporter's write fails 2 of the 3 new export tests (101). Both reverted and re-verified at 0.

## Checklist


nihalpasham:
**Gist:** A label, not a behavior change: each shared-memory block in the compiler output now carries a comment naming the Rust static it came from, so nobody has to match anonymous blocks back to source by array size. Plumbed as an optional attribute through the existing extension points; symbols and generated code untouched, and a missing label can never fail a build. I also brought the branch up to date with main here (one side-by-side conflict with #684's additions, both kept).

- the shared-memory example emits both `; 
- shared source:` comments in the .ll with unchanged PTX and passing execution; llvm-export/mir-lower/dialect-mir suites green.

Closes #703.

Merging.

nihalpasham (approved):

PR #709, opened 2026-08-07
Address a tuple field read instead of copying the whole array

### LINKED ISSUE #693: Projecting a field of a tuple array element copies the whole array, once per field
## Problem

Reading **one field** of a tuple element out of a local array copies the *whole
array* to a fresh stack slot first — once per field projected.

#231 fixed the nested-array shape (`scratch[row][col]`). This is the
array-of-tuple shape, where the projection is a field rather than a second index,
and it still materializes the entire array per projection.

## Reproducer

```rust
// (u8, u32): 8 bytes with three bytes of interior padding.
const PAIRS: [(u8, u32); 256] = /* ... */;

#[kernel]
pub fn tuple_table(input: &[u32], mut output: DisjointSlice<u32>) {
    let idx = thread::index_1d();
    let i = idx.get();
    let (a, b) = PAIRS[(input[i] >> 24) as usize & 255];
    if let Some(o) = output.get_mut(idx) { *o = a as u32 + b; }
}
```

Emitted LLVM IR contains **three** whole-array allocas — the table's own storage
plus one per destructured field:

```llvm
%v11 = alloca [256 x { i8, [3 x i8], i32 }], align 4   ; the table
%v29 = alloca [256 x { i8, [3 x i8], i32 }], align 4   ; copy for field `a`
%v30 = getelementptr inbounds [256 x { i8, [3 x i8], i32 }], ptr %v29, i32 0, i64 %v26
%v34 = alloca [256 x { i8, [3 x i8], i32 }], align 4   ; copy for field `b`
%v35 = getelementptr inbounds [256 x { i8, [3 x i8], i32 }], ptr %v34, i32 0, i64 %v26
```

`%v29` and `%v34` survive `opt -O2`. On `sm_86` this kernel is **925 PTX
instructions with a 4096-byte local depot and 878 `st.local`** — two 2 KiB
per-thread copies of a 2 KiB table, for a single element read.

## Why it matters beyond the instruction count

The copies dominate, so they mask improvements to the table's own storage. While
extending #684 to more element types (#692) I confirmed that promoting such a
table to an immutable device global changes nothing measurable: the depot and the
878 stores stay exactly as they are, and the module merely gains a 2 KiB global.
That is why field-less enums are worth promoting and tuples are not — until these
copies are gone.

Measured on A10G (sm_86, driver 595.71.05), with and without the promotion, the
kernel is identical at 925 instructions / 4096-byte depot / 878 `st.local`.

## Expected

A field projection of an array element should stay address-based to the final
scalar load, as #231 established for the nested-index case:

```llvm
%elem = getelementptr inbounds [256 x { i8, [3 x i8], i32 }], ptr %table, i32 0, i64 %index
%a_ptr = getelementptr inbounds { i8, [3 x i8], i32 }, ptr %elem, i32 0, i32 0
%a = load i8, ptr %a_ptr
```

with no additional whole-array alloca, and one per *field*, not one per array.

## Notes

- Both fields projected from the same element produce two separate copies, so the
  cost scales with the number of fields read, not with how many elements are
  touched.
- I am not proposing a fix here; I ran into this while measuring #692 and it
  seemed worth recording with the evidence rather than leaving it implicit in a
  code comment.

alejandro-soto-franco:
Closes #693.

## Problem

Reading one field of a tuple stored in an array copied the whole array first, once per field, because `mir.field_addr`'s verifier accepted struct, union and enum pointees but not tuples. The read then fell back to the value path, which materializes the entire array as a first-class aggregate.

```text
const PAIRS: [(u8, u32); 256] = /* 2 KB table */;
let (a, b) = PAIRS[i & 255];    // read ONE entry

Before: copy all 256 pairs -> local stack (for `a`)
        copy all 256 pairs -> local stack (for `b`)
        878 st.local, 4 KiB local depot per thread
After:  gep to entry i, one small load per field
        512 st.local, 2 KiB depot (the table's own storage,
        a separate known limitation)
```

## Fix

The missing tuple arm at all three layers, no new MIR op:

- `dialect-mir`: `MirFieldAddrOp::verify` accepts a `MirTupleType` pointee (field types via declaration order).
- `mir-importer`: the read classifier routes tuple field projections onto the address path.
- `mir-lower`: `convert_field_addr` resolves the GEP slot through `StructLayoutInfo::of_tuple`, the same memory-order map the write path already uses. This is the load-bearing part: rustc reorders `(u8, u32)` in memory, and the GEP must name the memory slot, not the declaration index. Unit tests at both IR layers lock that distinction.

## Maintainer commits on top

- d548317b: the verifier change also unlocks tuple-field writes (`arr[i].1 = x`) and `&mut` tuple-field borrows, which previously failed dialect verification loudly. Added the missing coverage: a dialect-level store-through-field_addr test on the reordered `(u8, u32)` tuple, and a `tuple_field_store` kernel in the example (runtime-index field write plus a `&mut` field borrow, bit-exact against a host replica).
- 08bef988: the full smoketest caught `array_constants`' code-shape guard rejecting the over-aligned-ZST tuple kernel on this branch: the address path's final load states natural alignment only, so routing a `repr(align(N))`-carrying tuple onto it emits an under-aligned (though still correct) load shape. Over-aligned-ZST tuples now stay on the value path. Gate shape from PR #715 (vyncint), with its byte-size gap closed (`[Align32; N]` array fields are measured in bytes, not element count); unit-tested for the nested-array cases, and ordinary tuples verified to stay on the address path.


nihalpasham:
**Gist:** The missing tuple arm in `mir.field_addr`, fixed at all three layers. 
- the issue: Rust stores tuple fields in a different order than you write them ((u8, u32) puts the u32 first in memory to avoid padding), so the generated address calculation has to use the field's memory position, not its source position, or every read of `.1` returns garbage. 
- the PR gets this right by reusing the same written-order-to-memory-order table the compiler already uses when building tuples, so reads and writes can never disagree, and both affected compiler stages carry a reordered-tuple test that fails if the wrong index is ever used.

**I pushed two commits on top:**

- the same change that fixes reads also quietly makes tuple-field writes (`arr[i].1 = x`) and `&mut` field borrows compile, where before they were hard errors. Nothing tested them, so I added a compiler-level test plus an example kernel that writes fields both ways and checks every value against the CPU.
- the full smoketest caught one exotic case: a tuple carrying a zero-sized `repr(align(N))` field produced a code shape our alignment guard rejects, so tuples like that now stay on the old copy path instead. The guard idea comes from #715 ; I closed a hole in its detection and added a test for it.

- all three example kernels exact-match; 
- `array_constants` + its shape guard PASS; 925/169/89 crate tests green; full sweep 188/188.

Closes #693. (#715 is the same fix filed independently four days later; closing it as already-fixed.)

Merging.

nihalpasham (approved):

PR #710, opened 2026-08-07
feat(cuda-core): add a context scheduling-policy wrapper

### LINKED ISSUE #705: No way to set the context synchronization policy (cuCtxSetFlags / cuDevicePrimaryCtxSetFlags have no wrapper)

`CudaContext` exposes no way to choose the context's synchronization scheduling policy. The default is `CU_CTX_SCHED_AUTO`, under which the driver may spin the calling thread while waiting on a sync. On a host with plenty of cores that is nearly free; on a small one it is not, and today there is no way to change it from cuda-core without reaching into `cuda_core::sys`.

Concretely, on a 2-physical-core host (g4dn.xlarge, 4 vCPU, T4) running a sync-heavy pipeline — ~1550 kernel batches, each ending in a stream synchronize — we measured **~20 CPU-seconds of the process's host CPU, about 19% of the total**, disappearing when the context is switched to `CU_CTX_SCHED_BLOCKING_SYNC`:

| arm | host CPU-seconds (whole process) | mapping-loop wall |
|---|---|---|
| default (`CU_CTX_SCHED_AUTO`) | 105.5 / 105.6 / 105.8 | 61.21 / 61.26 / 61.60 s |
| `CU_CTX_SCHED_BLOCKING_SYNC` | 85.3 / 85.9 / 85.8 | 62.35 / 62.55 / 64.44 s |

3 reps per arm, arms interleaved in one boot, ranges disjoint on both columns. Both arms produce byte-identical output, so the workload is identical and the difference is sync-wait accounting rather than work. (CPU-seconds are `CPUPCT × total process wall`; the wall column is the mapping loop only, which excludes index load — hence the two columns do not share a denominator.)

For that workload blocking sync turned out to be a net **loss** in wall time: the per-sync wake-up latency exceeded the value of the returned cores. That is a perfectly good answer, and I am not claiming the default is wrong. The point is narrower: the default spends a measurable fraction of host CPU on spinning, and whether that trade is worth it is workload- and host-specific, so it should be reachable from the safe API.

**Proposed API / behavior**

```rust
ctx.set_sync_policy(SyncPolicy::BlockingSync)?;   // or a thin flags wrapper
```

One wrinkle worth your judgement: `CudaContext` retains the device's *primary* context (`cuDevicePrimaryCtxRetain`), so the flags cannot be chosen at creation the way `cuCtxCreate` allows. Both `cuCtxSetFlags` (on the bound current context) and `cuDevicePrimaryCtxSetFlags` (on the device's primary context by ordinal) would work, and CUDA's own documentation for the former cross-references the latter. Neither is wrapped in cuda-core today. I do not have a strong opinion on which fits the crate's ownership model better.

**Alternatives considered**

Calling `cuda_core::sys::cuCtxSetFlags` directly, which is what we do today. It works, but it is the same class of gap as the stack-size limit in #697 — a one-line driver call with no wrapper — and reaching into `sys` means every consumer re-derives the primary-context caveat above.

**Additional context**

CUDA driver API: `cuCtxSetFlags`, `cuDevicePrimaryCtxSetFlags`, `CUctx_flags::CU_CTX_SCHED_BLOCKING_SYNC`. `CudaContext` currently has 17 public methods, none of which touches context flags.

alejandro-soto-franco:
Closes #705.

`CudaContext` exposed no way to choose the primary context's `CU_CTX_SCHED_*` policy. The default (`CU_CTX_SCHED_AUTO`) may spin the calling thread while waiting on a sync, which is close to free on a host with spare cores and a real cost on a CPU-constrained one. Reaching `CU_CTX_SCHED_BLOCKING_SYNC` today means going through `cuda_core::sys` directly and re-deriving the primary-context caveat below at every call site.

## Changes

- `SyncPolicy`: `Auto` / `Spin` / `Yield` / `BlockingSync`, mapped to and from `CU_CTX_SCHED_*`.
- `CudaContext::set_sync_policy`: read-modify-write over `cuDevicePrimaryCtxGetState`/`cuDevicePrimaryCtxSetFlags_v2`, replacing only the 3 scheduling bits (`CU_CTX_SCHED_MASK`) so any other flag bit (`CU_CTX_MAP_HOST`, etc.) set independently survives.
- `CudaContext::sync_policy`: reads the current policy back.

## Rationale

Chosen: `cuDevicePrimaryCtxSetFlags_v2`/`cuDevicePrimaryCtxGetState`, scoped to the device `CudaContext` retains. Alternative: `cuCtxSetFlags`/ `cuCtxGetFlags`, scoped to whatever context is current on the calling thread. Rejected: `CudaContext` retains and owns one device's primary context specifically, so the device-scoped pair matches what the type actually holds rather than depending on thread-local current-context state a caller could change out from under it.

CUDA's own docs note `cuDevicePrimaryCtxSetFlags` can return `CUDA_ERROR_INVALID_CONTEXT` once the primary context is active, which `CudaContext::new` always makes it (retain, then `cuCtxSetCurrent` inside `bind_to_thread`). Measured on an RTX 5060 (driver 595.71.05, CUDA 13.3): `set_sync_policy` right after `CudaContext::new` succeeds. The wrapper still returns `Result`, so a driver or usage pattern where the documented restriction does apply surfaces as an `Err`, not a panic.

## Validation

`crates/cuda-core/tests/context_sync_policy.rs`, its own test binary (the scheduling flags are shared, process-wide state on the primary context, so sharing a binary with concurrently-run tests would race): round-trips `set_sync_policy` -> `sync_policy` through every named value in sequence, landing on `BlockingSync` last. Run on an RTX 5060 (sm_120a):

```text
test sync_policy_round_trips_through_every_named_value ... ok
```

Full `cargo test -p cuda-core` still green (26 + 1 + 13 + 3 + ... = every existing suite, 0 failed).

Gates: `cargo fmt --all -- --check`, `cargo clippy --workspace -- -D warnings`, `cargo check --workspace --all-targets`, `cargo doc --no-deps -p cuda-core` (RUSTDOCFLAGS=-D warnings), `cargo deny check`, all green.

### DIFF
--- a/crates/cuda-core/src/context.rs
+++ b/crates/cuda-core/src/context.rs
@@ -406,14 +406,17 @@ impl CudaContext {
     /// (`CU_CTX_SCHED_MASK`) are replaced, any other flag bit the process has
     /// set independently (e.g. `CU_CTX_MAP_HOST`) is preserved.
     ///
-    /// CUDA's own docs note `cuDevicePrimaryCtxSetFlags` can return
-    /// `CUDA_ERROR_INVALID_CONTEXT` once the primary context is active.
-    /// Measured on an RTX 5060 (driver 595.71.05, CUDA 13.3): calling this
-    /// immediately after [`CudaContext::new`] -- i.e. after the context has
-    /// already been retained and made current, which `new` always does --
-    /// succeeds. Treat a `CUDA_ERROR_INVALID_CONTEXT` result as that
-    /// documented restriction surfacing on a driver or usage pattern where
-    /// it does apply, not as this wrapper being wrong.
+    /// The policy is process-wide state on the device's primary context:
+    /// it affects every [`CudaContext`] clone of this device, and any
+    /// runtime-API user of the same device in this process, not just this
+    /// handle. The read-modify-write over `CU_CTX_SCHED_MASK` is also not
+    /// atomic: a concurrent flag writer on the same device can interleave
+    /// between the get and the set, and one side's update is then lost.
+    ///
+    /// Historical note: before CUDA 11, `cuDevicePrimaryCtxSetFlags` failed
+    /// with `CUDA_ERROR_PRIMARY_CONTEXT_ACTIVE` once the primary context was
+    /// active. That restriction was lifted; the flags now apply to the
+    /// already-active context.
     pub fn set_sync_policy(&self, policy: SyncPolicy) -> Result<(), DriverError> {
         self.bind_to_thread()?;
         unsafe {
--- a/crates/cuda-core/src/context.rs
+++ b/crates/cuda-core/src/context.rs
@@ -88,6 +88,58 @@ impl PartialEq for CudaContext {
 }
 impl Eq for CudaContext {}
 
+/// Context scheduling policy (`CU_CTX_SCHED_*`), governing how the driver
+/// waits when the calling host thread blocks on GPU work (a stream or
+/// context synchronize).
+///
+/// Set via [`CudaContext::set_sync_policy`], read via
+/// [`CudaContext::sync_policy`]. The trade-off is host CPU against wake
+/// latency: [`Spin`](Self::Spin) burns a full core for the lowest latency,
+/// [`BlockingSync`](Self::BlockingSync) frees the core but wakes later. On a
+/// host with spare cores this trade rarely matters; on a CPU-constrained one
+/// (few physical cores, a sync-heavy pipeline) it can cost double-digit
+/// percentages of host CPU for no wall-clock benefit, which is the case
+/// [`BlockingSync`](Self::BlockingSync) exists to fix.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


nihalpasham:
**Gist:** The wrapper is correct: it targets the device's primary context, which is what `CudaContext` actually owns, rather than whatever context happens to be current on the calling thread; and it changes only the three scheduling bits while preserving any other flags already set, which is the part most wrappers get wrong. The policy is a proper enum, with an `"unknown"` for the encodings the driver never assigns.

**I pushed one commit on top (docs only):**

- removed a claimed driver error that does not exist in CUDA 13.3 (the old restriction was a different error and was lifted years ago);
- documented that the setting is shared by every handle to that device in the process, and that two threads changing flags at once are not protected against each other.

**Verified on an RTX 5090:** 
- the round-trip test passes on real hardware. 
- fyi: CI never executes this test (it runs library tests only), so the hardware run here was the first.

Closes #705.

Merging.

nihalpasham (approved):

PR #711, opened 2026-08-07
fix(mir-lower): scope the __shared_mem_N counter to one module

### LINKED ISSUE #706: __shared_mem_N names come from a process-global counter, so codegen is not reproducible
**Description**

`create_shared_global` in `mir-lower/src/convert/ops/memory.rs` names each static shared allocation from a process-global counter:

```rust
static SHARED_COUNTER: std::sync::atomic::AtomicUsize = AtomicUsize::new(0);
let counter = SHARED_COUNTER.fetch_add(1, Ordering::SeqCst);
let name = format!("__shared_mem_{counter}");
```

The index therefore depends on the order in which allocations reach lowering, not on anything about the program. Whenever more than one codegen unit is lowered concurrently in a process, the mapping from Rust `static` to `__shared_mem_N` is scheduling-dependent, so two builds of identical source can emit the same globals under different names.

**Expected behavior**

Emitted symbol names are a function of the input, so a rebuild of unchanged source produces byte-identical artifacts.

**Actual behavior**

Names are assigned in completion order from shared mutable state. Beyond reproducible builds, this makes any tooling that pins a generated name fragile — including cuda-oxide's own tests: `crates/llvm-export/tests/export_test.rs` asserts on the literal `__shared_mem_20` in six places, which holds only as long as the counter reaches that value in the same order.

**Environment**

- GPU: n/a (codegen-time only)
- CUDA driver version: n/a
- `rustc --version --verbose`: nightly-2026-04-03

**Additional context**

A per-module or per-function counter, or an index derived from the allocation's `alloc_key`, would make the name a function of the input. I have not proposed a patch because the right scope depends on what guarantees you want to make about symbol stability across versions — happy to write one if you have a preference.

Noticed while adding source attribution for these globals (see the companion issue); that change deliberately uses a sidecar attribute rather than a rename, so it neither depends on nor fixes this.

alejandro-soto-franco:
Closes #706.

## Problem

Static shared-memory globals were named from a process-global `AtomicUsize`, so the `__shared_mem_N` index depended on how many other modules the process had lowered first and in what thread order. Two builds of identical source could emit the same globals under different names.

```text
Before: lower module A, then module B, in one process:
    A's tile -> @__shared_mem_0
    B's tile -> @__shared_mem_1
  next build lowers B first (parallel scheduling):
    B's tile -> @__shared_mem_0    <- same source, new name every build

After: the counter restarts at 0 for every module:
    B's tile -> @__shared_mem_0    <- stable on every rebuild
```

## Fix

- `MirToLlvmConversionDriver` gains `next_shared_mem_index: usize`, initialised where `shared_globals`/`device_globals` already are, threaded into `create_shared_global` exactly like the existing dedup maps. No new mechanism: the driver is already instantiated fresh once per module for the same reason.
- Regression test lowers two fresh modules in one process and asserts both name their first shared block `__shared_mem_0`, which the old code could never satisfy.

## Maintainer commit on top (b78a338b)

The identical process-global counter behind `__device_global_N` sat 200 lines below in the same file. Extended the same fix: `next_device_global_index` on the driver, threaded the same way, and the regression test now also asserts both modules name their first device global `__device_global_0`. Device codegen naming is now reproducible end to end.

### DIFF
--- a/crates/mir-lower/src/convert/ops/memory.rs
+++ b/crates/mir-lower/src/convert/ops/memory.rs
@@ -665,6 +665,7 @@ pub fn convert_global_alloc_dc(
     op: Ptr<Operation>,
     _operands_info: &OperandsInfo,
     device_globals: &mut DeviceGlobalsMap,
+    next_device_global_index: &mut usize,
 ) -> Result<()> {
     use pliron::builtin::attributes::{StringAttr, TypeAttr};
 
@@ -748,6 +749,7 @@ pub fn convert_global_alloc_dc(
             ctx,
             op,
             device_globals,
+            next_device_global_index,
             DeviceGlobalSpec {
                 key: &global_key,
                 mir_type: mir_global_type,
@@ -779,10 +781,15 @@ struct DeviceGlobalSpec<'a> {
     immutable: bool,
 }
 
+/// `next_device_global_index` is scoped to one `MirToLlvmConversionDriver`
+/// instance (one module), not a process-global counter: `N` is a function of
+/// this module's own MIR walk order, not of how many other modules have
+/// lowered a device global earlier in the process (#706).
 fn create_device_global(
     ctx: &mut Context,
     op: Ptr<Operation>,
     device_globals: &mut DeviceGlobalsMap,
+    next_device_global_index: &mut usize,
     spec: DeviceGlobalSpec<'_>,
 ) -> Result<pliron::identifier::Identifier> {
     // An explicit initializer is already the evaluated Rust allocation image.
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Gist:** The right fix using scoping that already existed: the naming counter moves onto the per-module lowering driver, where the other per-module state already lives for the same reason. A shared-memory block's name now depends only on its own module, so identical source produces identical names on every build, regardless of what else the process compiled first.

- The identical process-wide counter also named `__device_global_N`, 200 lines below in the same file. Gave it the same fix, and the regression test now checks both counters restart at zero for a fresh module. Codegen naming is reproducible end to end.

**Verified on an RTX 5090 (sm_120):** barrier and LTOIR examples green; mir-lower suites green.

Closes #706.

Merging

nihalpasham (approved):

Issue #712, opened 2026-08-07
verify.rs: unbounded recursion, a wrong doc comment, and zero test coverage


alejandro-soto-franco:
## Problem

`crates/cuda-oxide-codegen/src/verify.rs` (72 lines, from #314, untouched since) wraps `Operation::verify` for the standalone pipeline's 7 verification gates (`prep.rs`, `pipeline.rs`). Three separate, real issues in it:

**1. `find_inner_verification_error` recurses without a depth bound.** It walks every region/block/op below the failing operation to locate the innermost failure for a richer diagnostic. pliron's own `Operation::verify` (`region.rs`/`basic_block.rs`/`operation.rs`) already recurses the same way internally via `try_for_each`, so this is a second, independent recursive walk, triggered only on the error path. Recursion depth today is bounded by region nesting, not by op count: `MirFuncOp` is the only region-bearing op across `dialect-mir`, `dialect-nvvm` and the LLVM dialect, so a module of any size stays at roughly 3 stack frames (module, function, block), which is not a live crash risk today. The function has no structural reason to assume that stays true, and nothing stops a future region-bearing op (or a much deeper lowering pipeline) from making it one. An iterative rewrite removes the assumption for free rather than leaving it undocumented and unenforced.

**2. The doc comment on `verify_operation` is wrong.** It reads "mir-importer pipeline plumbing; not part of the frontend contract". Every call site (`prep.rs:38,73,85,103,116,160`, `pipeline.rs:374`) is inside `cuda-oxide-codegen`'s own standalone pipeline; `mir-importer` never calls it. Looks like leftover text from wherever this was adapted from.

**3. Zero test coverage.** No test in the crate exercises `verify_operation` or `find_inner_verification_error`. Missing coverage spans three cases: pass-through on success, a single-level failure, and the case that is the whole reason the function exists, where the innermost failure sits several levels below the operation `verify_operation` was called on.

## Scope note

[...]

PR #713, opened 2026-08-07
fix(cuda-oxide-codegen): make find_inner_verification_error iterative

alejandro-soto-franco:
Part of #712.

`find_inner_verification_error` re-walks the operation tree recursively to locate the specific operation that failed verification, for a richer diagnostic than `Operation::verify`'s own error carries. Recursion depth here tracks region nesting, not module size: `MirFuncOp` is the only region-bearing op across `dialect-mir`, `dialect-nvvm` and the LLVM dialect, so depth stays around 3 (module, function, block) regardless of how large a module gets, which keeps this from being a live crash risk today. Nothing in the function enforces that stays true, and a compile error should not depend on an unstated depth bound to avoid becoming a stack overflow instead.

## Changes

Rewrite the recursive descent as two passes over an explicit stack: a natural-order walk (`to_visit`) collected into `post_order`, then drained in reverse to recover the children-before-parent order the original recursive contract required (every descendant of an operation checked before the operation itself).

## Rationale

Chosen: two passes over one stack. Alternative: a single-pass iterative walk that tracks each node's remaining unvisited children explicitly (the usual way to do post-order in one pass). Rejected for this function specifically: the two-pass version reads as a direct transliteration of the original recursive contract, which matters here because getting the order wrong would silently change WHICH operation gets reported on a tree with more than one failure, and that is exactly the property the new tests below check.

## Validation

First tests for this file (previously zero): `verify_operation_passes_ through_a_fully_valid_tree`, `verify_operation_reports_a_direct_child_ failure`, and `verify_operation_finds_the_innermost_failure_among_several_ siblings` (three siblings, only the middle one malformed; asserts the returned operation is that one specifically, not the first or last sibling checked and not the module).

Full `cargo test -p cuda-oxide-codegen` green (182 tests across 8 binaries, 0 failed).

Gates: `cargo fmt --all -- --check`, `cargo clippy --workspace -- -D warnings`, `cargo check --workspace --all-targets`, `cargo doc --no-deps -p cuda-oxide-codegen` (RUSTDOCFLAGS=-D warnings), `cargo deny check`, all green.

### DIFF
--- a/crates/cuda-oxide-codegen/src/verify.rs
+++ b/crates/cuda-oxide-codegen/src/verify.rs
@@ -41,32 +41,198 @@ pub fn verify_operation(
     Ok(())
 }
 
-/// Recursively finds the innermost operation that failed verification.
+/// Finds the innermost operation that failed verification.
 ///
 /// Helps produce better error messages by pointing to the specific failing
 /// operation rather than just the containing module/function.
+///
+/// Iterative, not recursive: recursion depth here tracks REGION nesting
+/// (`MirFuncOp` is the only region-bearing op across every dialect this
+/// pipeline lowers through, so today that depth stays around 3 -- module,
+/// function, block -- regardless of how large a module gets). Nothing about
+/// this function enforces that stays true, and there is no reason a compile
+/// error, of all things, should depend on an unstated depth bound to avoid
+/// becoming a stack overflow instead.
+///
+/// Two passes over an explicit stack, not one: `to_visit` drains in the
+/// tree's natural (root-to-leaf, left-to-right) order, which is exactly
+/// backwards from what finding an INNERMOST failure needs. Collecting into
+/// `post_order` first and draining THAT reverses it back to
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** The recursion is gone and the answer cannot change: the iterative walk visits operations in exactly the order the recursive one did (children before parents, left to right), so the error that gets reported stays the same one. A compile error should never depend on an unstated recursion-depth budget to avoid becoming a stack overflow, and now it does not. The file also gains its first tests.

- fixed two wrong comments: another op type also nests regions, so the doc no longer claims only one does (it now says the walk assumes nothing about depth); and a test comment claimed a type-identity property pliron does not have.
- added the two tests the issue actually asked for: a failure buried several levels deep must be the one reported, and with two bad siblings the first one must win.

153 `cuda-oxide-codegen` tests green. Partially addresses #712 (the doc-comment item lands with #714).

Merging.

nihalpasham (approved):

PR #714, opened 2026-08-07
docs(cuda-oxide-codegen): fix verify_operation's stale mir-importer comment

alejandro-soto-franco:
Part of #712.

## What this PR now does

`verify_operation`'s doc comment described it as "mir-importer pipeline plumbing; not part of the frontend contract", which understated its in-crate role. The first version of this PR swapped that for a claim that mir-importer never calls it; that premise turned out to be wrong (easy to miss: both crates have a `pipeline.rs`), and the corrected comment now names the full consumer set:

- this crate's own standalone pipeline stages (`prep.rs`, `pipeline.rs`);
- `mir-importer`, via the `cuda_oxide_codegen::__private` re-export (`lib.rs:126`), which calls it at `mir-importer/src/pipeline.rs:319` as the per-function post-translation verification gate on the live `#[kernel]` path.

That second consumer is why the function is `pub` + `#[doc(hidden)]`, and why demoting it to `pub(crate)` or deleting the re-export would break every normal kernel build. The comment now says so, which is the maintenance-safety point of the fix.

Gates: `cargo fmt --all -- --check`, `cargo clippy -p cuda-oxide-codegen -- -D warnings`, `cargo doc --no-deps -p cuda-oxide-codegen` (RUSTDOCFLAGS=-D warnings), all green.

Note for #712: item 2 of the issue carries the same "mir-importer never calls it" claim and should be corrected the same way.

### DIFF
--- a/crates/cuda-oxide-codegen/src/verify.rs
+++ b/crates/cuda-oxide-codegen/src/verify.rs
@@ -14,7 +14,8 @@ use pliron::printable::Printable;
 ///
 /// On failure, attempts to find the innermost failing operation for better
 /// error messages.
-// mir-importer pipeline plumbing; not part of the frontend contract.
+// cuda-oxide-codegen's own standalone pipeline plumbing (prep.rs,
+// pipeline.rs); not part of the frontend contract.
 #[doc(hidden)]
 pub fn verify_operation(
     ctx: &Context,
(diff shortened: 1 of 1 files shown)


nihalpasham:
Thanks for the PR

A couple of points:
mir-importer does call `verify_operation`, through a re-export, as the verification gate on the normal `#[kernel]` build path. Easy to miss, since both crates have a file named `pipeline.rs`. Merging the original text would have told a future refactorer the re-export was safely deletable, which would have broken every kernel build's verification.

- the comment now names both consumers (this crate's own pipeline stages, and mir-importer through the re-export) and says why the function is public but hidden from docs. 
- Doc and lint gates green.

Note: issue #712's item 2 carries the same claim and needs the same one-line correction.

Merging.

nihalpasham (approved):

PR #715, opened 2026-08-08
perf(mir-importer): address a tuple field in place instead of copying its array

### LINKED ISSUE #693: Projecting a field of a tuple array element copies the whole array, once per field

vyncint:
Closes #693.

## Summary

`PAIRS[i].1` — one field of a tuple array element — loaded the **whole array** as
a value and projected into it, so every such read first spilled the entire table
into a fresh per-thread stack slot. A second field read spilled it again.

The reason was narrow, and `place_read_fallbacks` recorded it:

> tuple field reads, because `mir.field_addr` currently verifies struct pointees
> only while value-space `mir.extract_field` supports tuples.

A tuple carries the same layout facts a struct does — declaration order, memory
order, per-field offsets, total size — and `StructLayoutInfo::of_tuple` already
existed to hand them over. So:

- `mir.field_addr` accepts a tuple pointee;
- `convert_field_addr` builds its slot map through that same `StructLayoutInfo`,
  so the field's LLVM slot comes out of the walk that built the type — which is
  what keeps a reordered tuple like `(u8, u32, u64)` addressing correctly;
- the read classifier lets a tuple field read walk to its address.

## Measured

A10G, from the evidence logs' `nvidia-smi` block:

```
name, compute_cap, driver_version
NVIDIA A10G, 8.6, 595.71.05
```

1048576 threads x 64 divergent lookups into a 256-entry `(u8, u32)` table, 10
timed launches after 3 warmup. Every kernel bit-correct against a host reference
in both runs.

**Before** (`.gpu-evidence/20260808T000635Z.log`) / **after**
(`.gpu-evidence/20260808T000907Z.log`):

| kernel | reads | before | after | speedup |
|---|---|---:|---:|---:|
| `tuple_one_field` | `PAIRS[i].1` | 1 127 170 us | 479 670 us | **2.35x** |
| `tuple_two_fields` | `let (a,b)=PAIRS[i]` | 2 139 199 us | 485 414 us | **4.41x** |
| `scalar_control` | `CODES[i]` | 46.3 us | 46.3 us | 1.00x |

`scalar_control` reads a plain `[u32; N]` table over the same index stream and
came back **bit-identical at 46.3 us in both runs**, so this is a controlled
comparison rather than two unrelated timings.

PTX:

| kernel | before | after |
|---|---|---|
| `tuple_one_field` | 567 instrs, 2048 B depot, 512 `st.local` | 565, 2048, 512 |
| `tuple_two_fields` | 1084 instrs, 4096 B depot, 1024 `st.local` | 567, 2048, 512 |
| `scalar_control` | 70, 0, 0 | 70, 0, 0 |

Two things improve, which is why `tuple_one_field` moves at an essentially
unchanged instruction count: `tuple_two_fields` drops from two whole-array spills
to one, and **both** kernels now read only the field they want per loop iteration
instead of loading the whole element and extracting.

## What is deliberately left

The table's own materialization: the constant is still built element by element
into the one remaining slot. That is the other half, and it is not independent —
#694 excludes tuple elements from immutable-global promotion **because this copy
dominated**, which I measured there (promoting alone changed nothing).

With both in place the depot goes entirely: the same two kernels compile to **42
and 44 PTX instructions with no local memory at all**, reading through
`ld.global.nc`. I verified that composition locally before deciding to split it.
It belongs on top of #694 rather than here, since it edits that PR's gate and
would need several `array_constants` code-shape assertions rewritten in the same
change.

## Over-aligned tuples stay on the value path

A zero-sized field can carry `repr(align(N))`, raising the ABI alignment above
anything the LLVM storage type expresses — `(Align32, u8)` lowers to
`{ i8, [31 x i8] }`, naturally align-1. The address path states no alignment on
its final load, so lowering falls back to the field's natural alignment and
quietly under-states the ABI.

`array_constants`' own guard caught exactly this, which is how the case was found:
it rejects any `align 1` operation in the over-aligned kernel, and my first cut
introduced `load i8, ptr %v15, align 1`. That load is *sound* on its own — a `u8`
needs no alignment — but the guard is right that the address path cannot restate
what the ABI recorded, so such tuples keep the value path, which moves the whole
aggregate at its recorded alignment. Making the address path carry alignment
would be the better fix and is a separate piece of work.

## Validation

Local, all clean:

```
cargo fmt --check
cargo test -p mir-lower -p llvm-export -p dialect-mir -p dialect-nvvm \
           -p mir-importer -p cuda-oxide-codegen --lib --tests    # 18 suites ok
bash crates/rustc-codegen-cuda/examples/array_constants/verify-code-shape.sh
```

On the A10G — scoped smoketest **15/15, 0 fail** (`.gpu-evidence/20260808T001138Z.log`),
covering every example that reads through tuple/array projections plus controls:

```
[ 1/15] array_constants     PASS   [ 6/15] index_field_assign        PASS   [11/15] tuple_constant_provenance PASS
[ 2/15] array_index         PASS   [ 7/15] place_read_fallbacks      PASS   [12/15] tuple_field_lookup        PASS
[ 3/15] cast_tests          PASS   [ 8/15] ref_index_projections     PASS   [13/15] uninit_const              PASS
[ 4/15] const_aggregate     PASS   [ 9/15] tma_copy       PASS (skipped)    [14/15] vecadd                    PASS
[ 5/15] enum_array_match    PASS   [10/15] tuple_array_provenance    PASS   [15/15] vectorization             PASS
```

I scoped rather than swept, to the examples that project through tuples and
arrays plus controls. Happy to run the full 188 if you would rather see it.

instruction-level evidence here is PTX only.

## Notes

`place_read_fallbacks` documented tuple field reads as a fallback, which they no
longer are. Its entry now names over-alignment as what keeps a tuple on the value
path; the kernel and its runtime check are unchanged, and it still covers the
cases that genuinely do fall back (enum payload reads, ZSTs).

`tuple_field_lookup` is sized deliberately small — 1 Mi threads, 10 iterations.
Before this change a single launch of the tuple kernels took over a second, so a
larger benchmark would have made the smoketest sit on it.


vyncint:

```
===== SMOKETEST SUMMARY =====
Pass:    190 / 190
Fail:    0 / 190
Elapsed: 1218s
```

```
```

nihalpasham:
**Gist:** This is the right fix, but #709 fixed the same gap at the same three compiler layers, was filed first, and came with tests at both stages, so it is the one merging. Closing this as already-fixed 

**Two notes that matter:**

- Your over-alignment fallback turned out to be necessary in practice: our full test sweep on #709 hit the failure it prevents. That idea landed inside #709 , with one detection hole closed (array fields were being measured in element count rather than bytes, so a zero-byte over-aligned array slipped through).
- Still worth a small follow-up on top of merged #709 if you want it: your benchmark example with the measured 2.35x / 4.41x.

PR #716, opened 2026-08-08
feat(iket): add semantic in-kernel event tracing integration

lucifer1004:
## Summary

Add initial CUDA In-Kernel Event Tracing (IKET) support to cuda-oxide.

This change introduces:

- a semantic `iket` compiler dialect;
- device-side point-event and range annotations;
- whole-program NativeDump/ExtendedNativeDump selection and lowering;
- IKET-compatible placeholder instructions, metadata retained through final
  CUBIN materialization, typed payloads, and long-name globals; and
- an end-to-end example captured with `run-iket`.

## Design

The compiler pipeline keeps source intent separate from the physical IKET
encoding:

```text
cuda_device::iket annotations
  -> MIR importer
  -> semantic iket.* operations
  -> whole-program IKET planning
  -> NativeDump or ExtendedNativeDump placeholders and metadata
  -> run-iket discovery, patching, collection, and export
```

The semantic dialect contains no cursor, buffer-address, elected-lane, or
capture-session policy. Those details belong to IKET lowering and profiling.

The annotation surface provides:

- `iket::mark!`;
- token-paired `iket::range_start!` / `iket::range_end!`; and
- stack-style `iket::range_push!` / `iket::range_pop!`.

Payload kind and width remain explicit through the semantic pipeline. Typed
range tokens prevent pairing an end operation with a different start site.

## Instrumentation method selection

`CUDA_OXIDE_IKET=auto` is used when IKET operations are present:

- up to 30 unique user event names: `NativeDump`;
- 31 through 4,031 unique user event names: `ExtendedNativeDump`;
- above 4,031: compilation error.

`native`, `extended`, and `off` are available as explicit controls. A forced
method fails rather than silently selecting another method.

The count is over unique names in the compiler root, not dynamic event count
or static call-site count. Range-pop uses its reserved IKET event ID.

Names are measured in UTF-8 bytes. Names that do not fit the 31-byte inline
field use the IKET-compatible hash representation and a device-global copy of
the complete string.

## Example and profiler dependency

`iket_trace` exercises a point event, a typed range, a `u32` payload, and a
name longer than the inline field:

```bash
cargo oxide build iket_trace --materialize-cubin --arch sm_120

uv venv --python 3.12 .venv
uv pip install --python .venv/bin/python nvidia-cutlass-dsl==4.7.0

.venv/bin/run-iket --output-dir /tmp/iket-trace --clobber \
  profile --postprocess all -- \
  "$PWD/crates/rustc-codegen-cuda/examples/iket_trace/target/release/iket_trace"
```

The current public profiling workflow depends on `nvidia-cutlass-dsl`, which
provides the `run-iket` command. This is a profiler-time dependency: cuda-oxide
does not link it into the generated application or vendor profiling binaries.

The default `run-iket` flow performs a tracker pass, chooses the timestamp and
context-buffer budgets, then captures the application in a second pass. The
example does not enable cluster launch or cluster filtering.

## Validation

- `cargo test --workspace`
- `cargo oxide build iket_trace --materialize-cubin --arch sm_120`
- final CUBIN checks for `__iket_meta_info` and the complete long event name
- end-to-end `run-iket profile --postprocess all` on an SM120 GPU

The capture contains one launch, 64 markers, 32 ranges, 32 warp lifetimes,
32 locations, and the complete long event name. It produces JSON, Perfetto,
compressed Perfetto, and HTML outputs.


nihalpasham:
Thanks for the PR. Nice work.

I probed an unpatched-placeholder question on hardware and pushed a couple of (CI + compiler) related  fixes; here's where things stand.

**What I verified:** on sm_120, an annotated kernel sharing its block with live `SharedArray` data runs bit-correct without the profiler. The placeholder store does execute on every run (confirmed in SASS: all threads, raw shared address 0x20), but it lands in the driver-reserved first 1KB of the shared window; user allocations base at 0x400, so user data can't be touched. There's now a comment at `placeholder_config` recording that this safety comes from where the store lands, not from construction.

**What I couldn't verify here:** I don't have sm_90 or sm_100/110 hardware. sm_90 uses offset 0x3f0 and the sm_100/110 variants address by cluster rank; both rely on the same reserved-region argument but haven't been probed. If you (or anyone reading) can run the equivalent check there, that closes the last safety question; until then the code comment marks them as pending.

**The fixes pushed:** 
- an iket smoketest category (pinned `--arch` for GPU-less CI, execution on SM90+ hosts) and 
- a recognized SUCCESS marker in the example; 
- placeholder encoding keyed off the finally-resolved compile target instead of the early hint; and `CUDA_OXIDE_IKET=off` now strips annotations under full-debug builds instead of erroring.

**About the default:** 
- keeping `CUDA_OXIDE_IKET=auto` as you designed it. 
- Profiling shipped binaries without a rebuild is the point of the feature under ahead-of-time compilation, and the sm_120 evidence supports it. 
- The env-var mechanism (rather than a cargo feature) is also the right call: no feature-unification hazards.

**Reviewer note:** This is an experimental feature tracking an experimental external format, so (as discussed) it'll need to be iterated on.

nihalpasham (approved):

PR #717, opened 2026-08-08
perf(mir-importer): promote a tuple array constant now that its fields address

vyncint:
**Rebased onto `main`, now a single commit** (`369de33d`). This PR was previously
stacked on #694 and #715; #694 has merged, and #715 was closed as already-fixed by
#709. The commit that duplicated #709 has been dropped, so what remains is only
the half that #709 did not do.

## Why this is the other half

#694 kept tuple elements out of immutable-global promotion and said why: while a
tuple field read still copied the whole array to a stack slot, that copy dominated
so completely that promoting the table changed nothing measurable and only added a
global to the module image. I measured it both ways there and it was neutral.

#709 removed that copy. With the read addressing in place, the promotion is what
takes the depot away. Neither alone looked like much; together they remove the
per-thread table entirely.

`promotable_array_element` therefore now admits a tuple whose every field is itself
promotable. Structs stay out because the element-wise path does not build arrays of
them either, so there would be nothing to fall back to. A payload-carrying enum
stays out because reading one back still round-trips the payload through memory:
the address walker resolves enum payload fields for writes, not for reads.

## It widens the reference form too, deliberately

Since #694 that predicate is also the gate for the pointer-to-array form
(`const R: &[T; N] = &TABLE`, via `validate_ptr_to_array_constant_type`), and its
doc says `TABLE[i]` and `(&TABLE)[i]` must not drift apart. Widening it widens that
form as well. That is safe rather than incidental: nothing on that path enumerates
fields — the initializer is rustc's evaluated byte image, and the size-agreement
check rejects any layout the dialect reproduces differently — so a tuple element
travels it exactly as a scalar does.

Its boundary test asserted the old exclusion (`"bare tuple-array support must not
widen pointer-to-array constants"`). It now asserts the admission, plus a
struct-in-tuple case proving the recursion holds on that path too. The size guard
from #694 is unchanged and still backs both forms.

## Measured, against this commit's actual parent

Re-measured after the rebase rather than carried over, because the baseline moved:
with #709 merged, the old "before" no longer exists. Both runs are the same commit
of `tuple_field_lookup` — a 256-entry `(u8, u32)` table read at a divergent runtime
index, 1048576 threads x 64 lookups, 10 timed launches after 3 warmup — built by
the two compilers and nothing else changed.

```
name, memory.total [MiB], driver_version, compute_cap
NVIDIA A10G, 23028 MiB, 595.71.05, 8.6
Detected GPU arch: sm_86 (via nvidia-smi)
```

| kernel | parent (`d82a1506`) | **+ this** | speedup |
|---|---:|---:|---:|
| `tuple_one_field` (`PAIRS[i].1`) | 479 684.6 us | **68.1 us** | **7044x** |
| `tuple_two_fields` (`let (a,b)=PAIRS[i]`) | 485 388.7 us | **129.4 us** | **3751x** |
| `scalar_control` (`CODES[i]`) | 46.5 us | 46.6 us | 1.00x |

Every kernel bit-correct against a host reference in both builds.
`scalar_control` reads a plain `[u32; N]` over the same index stream and moved by
0.2%, so this is a controlled comparison rather than two unrelated runs.

Evidence: parent `.gpu-evidence/20260808T131239Z.log`, this commit
`20260808T130553Z.log`.

The PTX explains the size of it:

| kernel | depot | `st.local` | `ld.global.nc` | instrs |
|---|---:|---:|---:|---:|
| `tuple_one_field` on parent | 2048 B | 512 | 0 | 562 |
| `tuple_one_field` **+ this** | **0** | **0** | **4** | **67** |
| `tuple_two_fields` on parent | 2048 B | 512 | 0 | 564 |
| `tuple_two_fields` **+ this** | **0** | **0** | **4** | **59** |
| `scalar_control` | 0 | 0 | 4 | 67 → 67 |

Each thread's 2 KiB private copy of the table is gone and the reads come from the
read-only data cache instead — the shape the scalar table has always had.
`tuple_one_field` now matches `scalar_control` instruction for instruction.

### On the numbers this PR used to quote

The previous body reported 16 625x / 16 557x. That figure was the composition of
*two* fixes measured against a baseline that had neither, and its middle column
("#715 only") is essentially what `main` is today. Quoting it as this PR's delta
would overstate it, so it is withdrawn in favour of the 7044x / 3751x above.

Re-measuring was not merely pedantic: parent `tuple_two_fields` is 564 instructions
here where the old table recorded 1084, because #709 already removes the second
whole-array copy. Runtime-wise the two implementations of that fix land in the same
place — 479 684.6 us against my old 479 670 us — so #709 fully substitutes for the
commit that was dropped.

## Gates

Scoped smoketest on this branch, A10G, **12/12 pass**
(`.gpu-evidence/20260808T131526Z.log`) — every example touching tuple or const
array constants, plus `vecadd` as a pass control and `wgmma` as a
compile-only control:

```
array_constants  const_table_lookup  const_tuple_table_lookup  constant_memory_table
enum_constant_provenance  enum_table_lookup  place_read_fallbacks
tuple_array_provenance  tuple_constant_provenance  tuple_field_lookup
vecadd  wgmma (PTX compiled)
Pass: 12 / 12   Fail: 0 / 12
```

`const_tuple_table_lookup` is #709's own example and passes unchanged.

Locally: `cargo fmt --check` (workspace and the new example crate), `clippy
--all-targets --lib --tests -D warnings`, the unit-test set including
`-p mir-importer` (1394 tests), and `array_constants/verify-code-shape.sh` against
a freshly built backend.

## The code-shape guard

`array_constants` pinned four tuple tables as `insertvalue` chains, which this
replaces with byte images. Each is now asserted as the initializer bytes plus the
lowered element type, which is strictly stronger: the reordered `(u8, u32, u64)`
table went from four separate slot searches to one pattern that fixes all three
fields at their byte offsets, the padding between them, and both elements' stride
at once. I checked that a swapped field and a padding byte carrying data are both
rejected by it.


vyncint:

```
===== SMOKETEST SUMMARY =====
Pass:    190 / 190
Fail:    0 / 190
Elapsed: 1218s
```

```
```

vyncint:
Rebased onto `main` as a single commit and re-measured, now that #709 has merged.

The commit duplicating #709 is dropped. What is left is the promotion half, which
#709 does not do — it never touches `promotable_array_element`.

**The headline number changed, downward.** This PR used to quote 16 625x, which was
two fixes measured against a baseline that had neither. With #709 in, its marginal
delta on the A10G (sm_86, driver 595.71.05) is:

| kernel | parent `d82a1506` | + this | speedup |
|---|---:|---:|---:|
| `tuple_one_field` | 479 684.6 us | **68.1 us** | **7044x** |
| `tuple_two_fields` | 485 388.7 us | **129.4 us** | **3751x** |
| `scalar_control` | 46.5 us | 46.6 us | 1.00x |

Both bit-correct; the control moved 0.2%. Parent `.gpu-evidence/20260808T131239Z.log`,
this commit `20260808T130553Z.log`. Scoped smoketest 12/12, `20260808T131526Z.log`.

Worth flagging for review: the rebase is not a pure replay. #694 made
`promotable_array_element` the shared gate for the pointer-to-array form as well, so
widening it now widens `(&TABLE)[i]` too. I believe that is right per its own doc
contract — and safe, since that path is byte-image driven with a size-agreement check
rather than field enumeration — but it is a real behaviour change beyond the original
diff, and the boundary test that asserted the old exclusion is inverted here. Happy to
split that into its own commit if you would rather review it separately.

Also: apologies for #715 duplicating #709. I have changed how I scan before filing —
by file rather than by keyword, since GitHub PR search does not index paths or diffs,
which is precisely why I missed it.

nihalpasham:
**Reviewed and GPU-verified on an RTX 5090:** 
- all three kernels bit-correct and the promotion shape confirmed in the artifacts. 
- clean recursive extension of the existing gate; 
- the padding/size/relocation machinery needed no changes. 

Merging. 

Related to #745 but doesn't close it; that one's root cause lives in mir-lower.

nihalpasham (approved):

PR #718, opened 2026-08-08
fix(finalizer): compile accepted NVVM IR directly

lucifer1004:
## Summary

Treat `nvvmCompileProgram` as the authoritative acceptance boundary for NVVM IR. `nvvmVerifyProgram` rejects some modern atomic IR that the same libNVVM installation compiles successfully, causing valid device code to fail before compilation.

## Changes

- Remove the separate `nvvmVerifyProgram` preflight from the NVVM IR compilation path.
- Preserve the existing frontend compatibility checks, deterministic module ordering, compile options, diagnostics, and output validation.
- Remove the unused verification-option contribution from the artifact digest.
- Bump the finalizer recipe version so cached artifacts and fingerprints reflect the changed acceptance semantics.
- Add a live regression test demonstrating that:
  - an atomic load accepted by libNVVM compiles successfully; and
  - malformed NVVM IR is still rejected by `nvvmCompileProgram`.

## Testing

- [x] `cargo clippy -p cuda-artifact-finalizer --all-targets -- -D warnings`
- [x] `cargo test -p cuda-artifact-finalizer` — 13 passed, 2 live tests ignored
- [x] `cargo test -p cuda-artifact-finalizer nvvm::tests::live_compile_accepts_atomic_load_and_rejects_malformed_ir -- --exact --ignored` — passed with a discoverable CUDA Toolkit
- `cargo oxide run`: not applicable; this change is isolated to the artifact finalizer.
- `cargo test --workspace`: not run; the affected crate and live libNVVM path were tested directly.
- New example: not applicable.

## Checklist

- [x] SPDX headers on new source files — no new source files

### DIFF
--- a/crates/cuda-artifact-finalizer/src/nvvm.rs
+++ b/crates/cuda-artifact-finalizer/src/nvvm.rs
@@ -94,9 +94,9 @@ impl NvvmCompiler {
                 program.add_module(&self.libdevice, "libdevice.10.bc")?;
                 program.add_module(nvvm_ir, module_name)?;
 
-                let verify = options.nvvm_verify_options();
-                let verify_refs = verify.iter().map(String::as_str).collect::<Vec<_>>();
-                program.verify(&verify_refs)?;
+                // Compilation is the authoritative acceptance boundary.
+                // nvvmVerifyProgram rejects atomic loads and stores that the
+                // same libNVVM installation can compile successfully.
                 let compile = options.nvvm_compile_options();
                 let compile_refs = compile.iter().map(String::as_str).collect::<Vec<_>>();
                 Ok(program.compile(&compile_refs)?)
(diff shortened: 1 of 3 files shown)


nihalpasham:
**Gist:** the verify-then-compile double gate was kind of our own invention: nvcc and NVRTC both drive libNVVM through the compile call alone, and the standalone verifier demonstrably lags the compiler's real acceptance set. This converges us on the standard shape, and it does it completely: the option plumbing and digest field go with the call, the real frontend guards stay, and the recipe bump invalidates every cached artifact from the old semantics, per the provenance contract.

- the finalizer suite green, 
- both live tests pass against the real toolkit (modern atomic IR accepted, malformed IR still rejected by the compile call with the same log), and 
- the LTOIR smoketest lanes 4/4 including execution on the modern path.

**One small non-blocker for a follow-up:** the live test hardcodes sm_120a, which needs CUDA 12.8+; deriving the arch would spare maintainers on older toolkits a confusing failure.

Merging.

nihalpasham (approved):

PR #719, opened 2026-08-08
feat(finalizer): link PTX inputs to cubin

lucifer1004:
## Summary

Add a driver-independent PTX-to-cubin path backed by nvJitLink. This allows callers with an existing PTX artifact to produce a validated target-specific cubin without loading libNVVM or compiling NVVM IR first.

## Changes

- Add `LtoLinker::link_ptx_to_cubin` for compiling one named PTX input into a cubin.
- Keep the PTX path on `LtoLinker`, so discovering and using it does not require libNVVM or libdevice.
- Share tool-identity revalidation, diagnostics, output handling, and cubin validation between the existing LTOIR path and the new PTX path.
- Use PTX-specific nvJitLink options without enabling LTO mode.
- Accept PTX with or without one trailing NUL terminator, while rejecting empty inputs and interior NUL bytes.
- Add a PTX artifact digest covering the logical PTX bytes, diagnostic name, target and code-generation policy, and exact nvJitLink identity.
- Add unit coverage for PTX normalization, validation, options, and artifact identity.
- Add a live nvJitLink regression test that compiles an acquire-load PTX kernel into a valid cubin.

## Testing

- [x] `cargo clippy -p cuda-artifact-finalizer --all-targets -- -D warnings`
- [x] `cargo test -p cuda-artifact-finalizer` — 15 passed, 2 live tests ignored
- [x] `cargo test -p cuda-artifact-finalizer link::live_tests::live_ptx_pipeline_compiles_acquire_load_to_cubin -- --exact --ignored` — passed with a discoverable CUDA Toolkit
- `cargo oxide run`: not applicable; this PR exposes the finalizer capability without changing a cargo-oxide workflow.
- `cargo test --workspace`: not run; the affected crate and live nvJitLink path were tested directly.
- New example: not applicable.

## Checklist

- [x] SPDX headers on new source files — no new source files


nihalpasham:
**Gist:** 
- primitive that fills a gap; main's materializer still refuses PTX input outright and this is the missing route, 
- on the same already-loaded nvJitLink the finalizer uses for LTOIR, so one tool identity and one policy owner.  
 
**I pushed three small commits:** 
- the PTX C-string rule is now owned once by nvjitlink-sys (both the digest and the link route go through it), 
- the `NoLinkInputs` message covers both routes, and one real toolkit constraint is documented.

**Verified here on an RTX 5090 with CUDA 13.3 / nvJitLink 13.3.33:**
- all crate live tests pass, plus an end-to-end check the tests don't cover: a real cuda-oxide `vecadd.ptx` (sm_80) linked at sm_120, loaded via `cuModuleLoadData`, launched, 1024/1024 correct,
- `vprintf` stays an undefined symbol in the cubin and resolves at load; a `gpu_printf` kernel prints correctly on device,
- `-lineinfo` and `-g` are honored without `-lto`; `-fma` is accepted but provably inert on this route (byte-identical cubins either way, while standalone `ptxas --fmad=false` splits the same probe), so FMA policy for PTX input belongs to the producer. Now documented where the option is emitted.

Merging. 

Related to #722 but does not close it.

nihalpasham (approved):

PR #720, opened 2026-08-08
feat(cargo-oxide): export metadata interop artifacts

lucifer1004:
## Summary

Extend the metadata-declared interop device-crate workflow with configurable PTX or cubin output, named binary-target selection, device-only Cargo features, and optional source-identity sidecars. PTX remains the default behavior.

## Changes

- Extend `[[package.metadata.cuda-oxide.device-crates]]` with:
  - `artifact-kind = "ptx" | "cubin"`
  - `artifact-dir`, while continuing to accept the existing `ptx-dir` spelling
  - `bin` for selecting a named device binary target
  - `source-identity` for requesting an artifact identity sidecar
- Add `--device-features` to `cargo oxide run` and `cargo oxide build`, forwarding those features only to metadata-declared device crates.
- Validate named binary targets through Cargo metadata, build only the selected target, touch its actual source file to trigger device code generation, and use its normalized target name as the default artifact name.
- Resolve source-identity depfiles from the selected Cargo target rather than the enclosing package name.
- Export cubins by emitting NVVM IR and compile-option metadata, then using the shared CUDA artifact finalizer.
- Require a concrete CUDA target for cubin output and install the resulting cubin atomically.
- Add an optional v1 identity sidecar containing:
  - the target architecture;
  - the normalized `--device-features` selection;
  - the artifact SHA-256;
  - an aggregate source digest; and
  - Cargo depfile-derived source paths and SHA-256 values.
- Include artifact mode and device-only features in the interop codegen fingerprint.
- Keep PTX inspection limited to metadata entries that actually emit PTX.
- Clean generated cubin, NVVM IR, option, and identity files alongside existing PTX artifacts.
- Document the extended metadata contract and CLI behavior.

## Testing

- [x] `cargo clippy -p cargo-oxide --all-targets -- -D warnings`
- [x] `cargo test -p cargo-oxide` — 164 passed
- `cargo oxide run`: not run; the metadata parsing, target selection, artifact paths, fingerprints, identity generation, and CLI surface are covered by focused tests.
- `cargo test --workspace`: not run.
- New example: not applicable; the existing metadata interop workflow remains the integration point.

## Checklist


nihalpasham:
**Gist:**
- does for the interop path what `nvcc -cubin` does: an opt-in ahead-of-time native binary, PTX stays the default,
- the `.identity` file gives deployment systems the checksums Triton's kernel cache computes internally, so they can detect stale artifacts.

**Review confirmed two real bugs; I pushed five commits:**
- the depfile lookup replaced hyphens with underscores, but cargo keeps them, so the README's own `bin = "simt-device"` example aborted with "did not produce dependency file"; the name is now taken verbatim from cargo, with a test that builds a real hyphenated crate,
- the cubin was compiled for the requested arch, not the arch the backend actually resolved and recorded in its `.target` file (a detected sm_120a becomes sm_90a for WGMMA kernels); the finalizer and the identity record now read that file, and for PTX the identity records the PTX's own `.target` line,
- identity sources under any directory named `target` were silently dropped; now only the real cargo target directory is excluded,
- one `cargo metadata` call per device crate instead of two; `.gitignore` learned `*.identity`,
- new end-to-end example `interop_cubin_identity` (hyphenated bin, cubin, source identity, a decoy default bin so `--bin` regressions fail loudly), wired into smoketest and the compile-only CI lane.

- 170/170 cargo-oxide tests,
- the example runs on GPU and checks the contract against reality: identity digest matches the cubin bytes, recorded target matches the backend's `.target` file, every recorded source re-hashes to its digest,
- a source edit changes the source digest while the artifact digest stays put; toggling cubin to ptx re-triggers device codegen.

Merging.

Related to #722 but does not close it (that repro is the non-interop path).

Follow-up we'll file separately: `artifact-name` only works when it matches the backend-derived crate name (pre-existing, fails loudly).

nihalpasham (approved):

PR #727, opened 2026-08-08
build: pin rustfmt in rust-toolchain.toml like clippy

### LINKED ISSUE #726: rust-toolchain.toml omits rustfmt, so a fresh checkout cannot run the fmt gate and CI works around it twice
`rust-toolchain.toml` pins the components a contributor needs:

```toml
components = ["rust-src", "rustc-dev", "rust-analyzer", "clippy", "llvm-tools"]
```

`clippy` is there. `rustfmt` is not — even though `cargo fmt --check` is a
required CI gate exactly like clippy. On a machine that has never installed
rustfmt for this nightly, `cargo fmt` fails, and nothing in the toolchain file
says why.

CI compensates for the gap rather than closing it, in two separate workflows:

* `.github/workflows/fmt.yml:24` — with a comment naming the cause directly:
  > *"rust-toolchain.toml pins nightly-2026-04-03 but does not list rustfmt as a
  > component. Add it explicitly so `cargo fmt` resolves against the same
  > nightly the rest of CI uses."*
* `.github/workflows/unit-tests.yml:183` — the `generated-intrinsics` job needs
  it too, because `cuda-intrinsics-gen` shells out to rustfmt when it writes
  generated sources.

So the workaround has already had to be copied once. A third job needing rustfmt
would need it again.

This also bites outside CI: provisioning a fresh build machine from the repo
gives you a toolchain that cannot run `just fmt-check`, which is a confusing
first experience given the file exists precisely to make the toolchain
reproducible.

### Suggested fix

Add `rustfmt` to the component list and drop the two workaround steps, making
the toolchain file the single source of truth for both required formatting and
linting components.

vyncint:
Closes #726.

`clippy` was in the component list; `rustfmt` was not — though `cargo fmt
--check` is a required gate exactly like clippy. A fresh checkout produced a
toolchain that could not run the fmt gate.

CI compensated instead of closing the gap, and had **already copied the
workaround once**:

* `fmt.yml` — its comment named the cause outright: *"rust-toolchain.toml pins
  nightly-2026-04-03 but does not list rustfmt as a component."*
* `unit-tests.yml` — the `generated-intrinsics` job needs it too, because
  `cuda-intrinsics-gen` shells out to rustfmt when writing generated sources.

Both steps are removed; the toolchain file now carries the component, so a third
job needing rustfmt gets it for free.

## Verified

```
$ rustup show active-toolchain
nightly-2026-04-03-x86_64-unknown-linux-gnu (overridden by 'rust-toolchain.toml')

$ cargo fmt --version
rustfmt 1.9.0-nightly (55e86c9968 2026-04-02)     # matches the pin

$ cargo fmt --all -- --check                                     # root: pass
$ (cd crates/rustc-codegen-cuda && cargo fmt --all -- --check)    # codegen: pass
```

Both workflow files still parse as YAML. No GPU needed.

If you would rather keep the explicit `rustup component add rustfmt` steps as
belt-and-braces, I am happy to restore them and land only the toolchain change.

### DIFF
--- a/.github/workflows/fmt.yml
+++ b/.github/workflows/fmt.yml
@@ -18,11 +18,8 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v4
 
-      # rust-toolchain.toml pins nightly-2026-04-03 but does not list
-      # rustfmt as a component. Add it explicitly so `cargo fmt` resolves
-      # against the same nightly the rest of CI uses.
-      - name: Install rustfmt for the pinned nightly
-        run: rustup component add rustfmt
+      # rustfmt comes from rust-toolchain.toml's component list, like clippy,
+      # so `cargo fmt` resolves against the same nightly as the rest of CI.
 
       - name: Check root workspace formatting
         run: cargo fmt --all -- --check
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -180,8 +180,8 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Install rustfmt for the pinned nightly
-        run: rustup component add rustfmt
+      # cuda-intrinsics-gen shells out to rustfmt when it writes generated
+      # sources; it comes from rust-toolchain.toml's component list.
 
       - name: Check generated intrinsic outputs
         run: cargo run -p cuda-intrinsics-gen -- check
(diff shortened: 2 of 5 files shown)


nihalpasham:
Makes sense, the two CI workarounds deleting cleanly proves the point. 

**I pushed a parity commit:** 
- the nested `crates/rustc-codegen-cuda/rust-toolchain.toml` header demands an exact match with the parent pin, so it gets rustfmt too, plus the same sync for the cargo oxide new scaffold. 

Closes #726.

nihalpasham (approved):

PR #734, opened 2026-08-08
docs(book): wrap barrier::nanosleep example in unsafe

arpanpathak:
## Summary

The API quick reference example calls `cuda_device::barrier::nanosleep(1000);` without an `unsafe` block. The function is generated as `pub unsafe fn nanosleep(ns: u32)` (included in `crates/cuda-device/src/barrier.rs` through `generated/mbarrier_extended.rs`), so the example as written does not compile.

## Change

Wrap the call in an `unsafe` block:

```rust
unsafe { cuda_device::barrier::nanosleep(1000) }; // Sleep ~1μs
```

## Verification

Compiled both forms against the current `cuda-device` crate:

- before: `error[E0133]: call to unsafe function `nanosleep` is unsafe and requires unsafe block`
- after: compiles clean

## Notes

One line, touches only the book. No issue was filed; the drift surfaced while reading the quick reference against the generated API.

### DIFF
--- a/cuda-oxide-book/appendix/api-quick-reference.md
+++ b/cuda-oxide-book/appendix/api-quick-reference.md
@@ -602,7 +602,7 @@ use cuda_device::debug;
 let t = debug::clock64();       // Cycle counter
 debug::trap();                  // Abort kernel
 debug::breakpoint();            // cuda-gdb breakpoint
-cuda_device::barrier::nanosleep(1000); // Sleep ~1μs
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:**
- verified against the shipped signature (`pub unsafe fn nanosleep`); this was the only drift in the snippet, the other four calls are safe functions.

Merging.

**Reviewer note:** Two other snippets on the same page call unsafe functions bare (TMA and cluster DSMEM);

nihalpasham (approved):

PR #736, opened 2026-08-08
test(smoketest): reject an example claimed by two categories

### LINKED ISSUE #735: An example listed in two smoketest category arrays is silently misclassified, and no guard catches it
`classify()` in `scripts/smoketest.sh` returns the **first** category array that
claims an example:

```bash
for cat in "${TCGEN05_EXAMPLES[@]}"; do [[ "$ex" == "$cat" ]] && { echo tcgen05; return; }; done
for cat in "${WGMMA_EXAMPLES[@]}";   do [[ "$ex" == "$cat" ]] && { echo wgmma;   return; }; done
...
```

A second claim is therefore not a conflict it can report — it is silently
ignored, taking the example's compute-capability gate and its verdict rules with
it. Listing `wgmma` in `TCGEN05_EXAMPLES` gates a Hopper `sm_90a` example on
Blackwell `sm_100` instead, and it would simply skip on hardware that should run
it.

**Neither existing guard notices.** Adding `wgmma` to `TCGEN05_EXAMPLES`:

```
$ bash scripts/check-example-smoketest-contract.sh
OK: 11 *_EXAMPLES arrays name only real examples, and all 164 standard examples print a success marker.
exit 0

exit 0
```

The phantom-name check passes because both names *are* real examples, and the
contract script's `categorised` set discards the duplicate by construction:

```python
categorised = {
    entry for name in CATEGORY_LISTS for entry in by_name[name].split()
}
```

so uniqueness is not observable from it. The arrays are disjoint today, so this
is a latent gap rather than an active bug — but it is exactly the failure mode
that script exists to catch, and the same shape as #645 (naming-guard covering
6 of 10 prefixes).

### Suggested fix

Assert uniqueness against the raw lists in
`scripts/check-example-smoketest-contract.sh`, naming the arrays that collide and
which one `classify()` would actually pick.

vyncint:
Closes #735.

`classify()` returns the **first** category array that claims an example, so a
second claim is not a conflict it can report — it is silently ignored, taking the
example's compute-capability gate and verdict rules with it. Listing `wgmma` in
`TCGEN05_EXAMPLES` gates a Hopper `sm_90a` example on Blackwell `sm_100`, so it
skips on hardware that should have run it.

## Neither existing guard saw it

With `wgmma` added to `TCGEN05_EXAMPLES`, on `main`:

```
$ bash scripts/check-example-smoketest-contract.sh
OK: 11 *_EXAMPLES arrays name only real examples, and all 164 standard examples print a success marker.
exit 0

exit 0
```

The phantom-name check passes because both names *are* real examples. And this
script's own `categorised` set discards the duplicate by construction:

```python
categorised = {entry for name in CATEGORY_LISTS for entry in by_name[name].split()}
```

so uniqueness was never observable from it. It has to be asserted against the raw
lists, which is what this adds.

## After

```
error: smoketest example contract violated
  wgmma is claimed by 2 category arrays (TCGEN05_EXAMPLES, WGMMA_EXAMPLES);
  classify() checks TCGEN05_EXAMPLES first and ignores the rest
exit 1
```

## Verified

* clean tree still passes, with the message extended to state the new invariant
* the duplicate above is caught (exit 1)
* a phantom name is still caught — no regression in the existing check

The arrays are disjoint today, so this closes a latent gap rather than fixing a
live misclassification. Same shape as #645. No GPU needed — the script is
source-reading only, and CI already runs it as `status-guard / smoketest example
contract`.

### DIFF
--- a/scripts/check-example-smoketest-contract.sh
+++ b/scripts/check-example-smoketest-contract.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Verify the two things scripts/smoketest.sh assumes about the examples, both
+# Verify the three things scripts/smoketest.sh assumes about the examples, all
 # of which are otherwise only discoverable by running the suite on a GPU:
 #
 #   1. Every name in a *_EXAMPLES array is a real example directory.  These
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Reproduced both directions on a scratch tree:** 
- clean pass on main's arrays, and 
- a planted duplicate names the offender and both claimants. 

Merging; 

closes #735.

PR #744, opened 2026-08-09
ci(book): build the book on pull requests, deploy only from main

### LINKED ISSUE #743: The book is never built on a pull request, so a PR that breaks it cannot be caught before merge
`.github/workflows/book.yml` triggers on:

```yaml
on:
  push:
    branches: [main]
  workflow_dispatch:
```

There is no `pull_request` trigger, so the book is built **only after merge**.
Every other content gate runs on PRs:

| workflow | `pull_request`? |
|---|---|
| `ci.yml` (fmt, clippy, unit-tests, cargo-deny) | yes |
| `docs.yml` | yes |
| `examples-compile.yml` | yes |
| `codeql.yml` | yes |
| `naming-guard.yml` | yes |
| `status-guard.yml` | yes |
| **`book.yml`** | **no** |

A PR that breaks a cross-reference, a directive, or a toctree entry therefore
reaches `main` unchallenged, and the first signal is a failure on the branch
everyone else builds from.

This compounds with the missing `-W` (#732): today the post-merge run would not
fail either, so the breakage lands and then stays silent.

### Why it is not simply "add pull_request"

`book.yml` is a deployment workflow — it holds `pages: write` / `id-token:
write` and a `deploy` job — which is presumably why it is push-only. Adding the
trigger naively would attempt a Pages deploy from every pull request.

The build job is the part worth running on a PR. `deploy` needs gating to
`github.event_name != 'pull_request'`.

The workflow-level `concurrency: group: pages` also wants attention: it exists to
serialise *deploys*, and applying it to every run would queue unrelated PR builds
behind one another. Scoping it per-ref and keeping the `pages` group on the
deploy job preserves the single-deploy guarantee while letting PR builds run in
parallel.

vyncint:
Closes #743.

**Stacked on #733 — please review only the second commit** (`fix the book build
on Sphinx warnings` is that PR). The two are halves of the same thing: #733 makes
a warning fail the build, and this makes that build actually run before merge.

## The gap

`book.yml` had no `pull_request` trigger, so the book was built **only after
merge**. Every other content gate runs on PRs:

| workflow | `pull_request`? |
|---|---|
| `ci.yml`, `docs.yml`, `examples-compile.yml`, `codeql.yml`, `naming-guard.yml`, `status-guard.yml` | yes |
| **`book.yml`** | **no** |

A PR that broke a cross-reference reached `main` unchallenged, and the first
signal was a failure on the branch everyone builds from. Combined with the
missing `-W`, the post-merge run would not have failed either.

## Why this isn't just adding the trigger

`book.yml` is a **deployment** workflow — `pages: write`, `id-token: write`, and a
`deploy` job — which is presumably why it was push-only. A naive trigger would
attempt a Pages deploy from every pull request.

So `deploy` is gated on `github.event_name != 'pull_request'`. A PR builds the
book and stops; it never touches the Pages environment, and those write
permissions go unused on that path.

The workflow-level `concurrency: group: pages` needed the same care. It exists to
serialise *deploys*; left workflow-wide it would queue unrelated PR builds behind
each other. It moves onto the `deploy` job where the constraint belongs, and the
workflow-level group becomes ref-scoped with `cancel-in-progress` only for pull
requests — so superseded PR runs are cancelled, while a publish in flight on
`main` is never interrupted.

## Verified

Parsed the workflow and checked the resulting shape rather than eyeballing it:

```
triggers: ['pull_request', 'push', 'workflow_dispatch']
workflow concurrency: group=${{ github.workflow }}-${{ github.ref }}
                      cancel-in-progress=${{ github.event_name == 'pull_request' }}
job build:  if=None                                    <- runs on PRs
job deploy: if="github.event_name != 'pull_request'"   <- never on PRs
            concurrency={'group': 'pages', 'cancel-in-progress': False}
```

No GPU needed.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -3,6 +3,12 @@ name: Deploy cuda-oxide Book
 on:
   push:
     branches: [main]
+  # Validate the book on PRs too. The build job alone runs there; `deploy` is
+  # gated below, so a pull request never publishes. Without this the book was
+  # only ever built after merge, so a PR that broke it could not be caught by
+  # review or by CI.
+  pull_request:
+    branches: [main]
   workflow_dispatch:
 
 permissions:
@@ -10,9 +16,14 @@ permissions:
   pages: write
   id-token: write
 
+# Serialising every run under one `pages` group is a deploy constraint, and
+# applying it workflow-wide would queue unrelated PR builds behind each other.
+# Scope this to the ref, and keep the single-deploy guarantee on the deploy job
+# itself. Superseded PR runs are cancelled; pushes to main are not, so a
+# publish in flight is never interrupted.
 concurrency:
-  group: pages
-  cancel-in-progress: false
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   build:
@@ -45,6 +56,13 @@ jobs:
 
   deploy:
     needs: build
+    # Publishing is for main only. A pull request builds the book to prove it is
+    # sound and stops there; it never touches the Pages environment, and the
+    # `pages: write` / `id-token: write` permissions go unused on that path.
+    if: github.event_name != 'pull_request'
+    concurrency:
+      group: pages
+      cancel-in-progress: false
     runs-on: ubuntu-latest
     environment:
       name: github-pages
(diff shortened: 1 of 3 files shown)


nihalpasham:
The trigger, deploy gating, and concurrency logic all checked out line by line, and the `-W` build passes with today's Sphinx (9.1.0). 

**I pushed a follow-up moving pages:** 
- write / id-token: write onto the deploy job so PR runs hold a read-only token, and bounding sphinx below the next major, the pin you offered, since -W now gates every PR.

This closes #743 and #732 together; #733 (whose commit this carries) will be closed as already-fixed.

nihalpasham (approved):

Issue #745, opened 2026-08-09
Reading a lane of an over-aligned aggregate drops its abi_align, costing 2x on cache-resident lookups


vyncint:
## The gap

A whole-element copy of an over-aligned struct vectorizes. Reading its fields
does not — same type, same alignment, twice the loads.

```rust
#[repr(C, align(8))]
struct AlignedPoint { x: f32, y: f32 }

// 2x ld.global.b32
let p = pts[i];
*o = p.x + p.y;
```

The `vectorization` example's `f32x2` stays wide only because it does
`output[i] = input[i]` and never touches a lane. Spelling the same read
field-wise — or even as `c.0[0] + c.0[1]` through an array newtype — splits it
and loses the guarantee.

## Root cause

`convert_load` takes the load's alignment from its own result type:

```rust
if let Some(align) = mir_type_abi_align(ctx, result_ty) {
    set_op_alignment(ctx, llvm_load.get_operation(), align as u32);
}
```

`mir_type_abi_align` records `abi_align` for aggregates and answers `None` for
scalars. So a field load exports with LLVM's default `align 4`, and the
aggregate's `align(8)` — which rustc did record — is dropped at the projection.
LoadStoreVectorizer will not widen past an alignment it cannot prove, so the
pair never fuses.

## The fix works

Patching one attribute into the emitted IR, then running the project's own
`opt -O3` + `llc -march=nvptx64 -mcpu=sm_86`:

| first field load | resulting PTX |
|---|---|
| `align 4` (today) | `2x ld.global.b32` |
| `align 8` | `1x ld.global.v2.b32` |

The aggregate is aligned to `abi_align`, so its field at byte `offset` is
aligned to `gcd(abi_align, offset)` — and to `abi_align` itself at offset zero.
Both numbers come from rustc's layout, so nothing would be claimed that the
source did not already guarantee.

## What it is worth, measured

A10G (sm_86, driver 595.71.05), 4194304 threads, 30 timed launches after 5
warmup, every kernel bit-correct against a host reference. `one_wide_load` reads
the identical 8 bytes in a single load and splits in registers — the instruction
sequence the fix would produce:

| kernel | global loads | us/launch | GB/s |
|---|---:|---:|---:|
| `struct_two_fields` (streaming) | 2 | 104.7 | 480.5 |
| `one_wide_load` (streaming ceiling) | 1 | 104.8 | 480.4 |
[...]

PR #746, opened 2026-08-09
perf(mir-lower): keep an over-aligned aggregate's alignment on its field loads

vyncint:
Partially addresses #745: this fixes the named-field spelling (`c.x + c.y`). The array-newtype lane spelling (`c.0[0] + c.0[1]`) still splits, because array-element addresses do not stamp their proved alignment yet; that is a follow-up.

A load takes its alignment from its own result type, and `mir_type_abi_align`
records one only for aggregates — a scalar answers `None`. So `p.x` on an
`#[repr(C, align(8))]` struct exported with LLVM's default `align 4`, and the
guarantee rustc gave the aggregate was dropped at the projection.
LoadStoreVectorizer will not widen past an alignment it cannot prove, so two
adjacent field reads stayed two loads.

The same struct **copied whole** already vectorized, because there the load's
result type *is* the aggregate. The cost depended on how the read was spelled
rather than on the data.

## The arithmetic

The aggregate is aligned to `abi_align`, so its field at byte `offset` is aligned
to `gcd(abi_align, offset)`, and to `abi_align` itself at offset zero. Both
numbers are rustc's own layout, so this claims nothing the source did not already
guarantee. Every uncertain case — no recorded `abi_align`, no recorded offset —
declines and leaves the previous, weaker-but-sound behaviour.

## Where it had to go

`convert_load` cannot recompute this: by the time it runs, `mir.field_addr` is
already a GEP and the MIR aggregate type is no longer reachable from the load's
operands. (I verified that by forcing a distinctive alignment through each step —
the GEP matches and its indices destructure, but
`operands_info.lookup_most_recent_of_type` returns `None` for the GEP's *base*,
which is not an operand of the load.)

So `convert_field_addr`, which still holds the aggregate, stamps what the address
proves onto the GEP, and the load reads it back. The attribute is
lowering-internal and never exported — it deliberately uses a key **distinct**
from the one the exporter prints as `align N`, so a future `emit_gep` cannot
mistake it for the GEP's own access alignment.

## PTX

A `(f32, f32)` element read field-wise:

| element | before | after |
|---|---|---|
| `repr(C)` — align 4 | 2 `ld.global.b32` | **2 `ld.global.b32`** |
| `repr(C, align(8))` | 2 `ld.global.b32` | **1 `ld.global.v2.b32`** |

The align-4 element is the control and **must not change**: nothing proves
8-byte alignment there, and widening it would be wrong.

## Measured

A10G (sm_86, driver 595.71.05), 4194304 threads, 30 timed launches after 5
warmup, every kernel bit-correct against a host reference. Same example built by
both compilers, nothing else changed:

| kernel | stock `main` | this branch | speedup |
|---|---:|---:|---:|
| `packed_pair` (streaming) | 104.8 us | 104.6 us | 1.00x |
| `aligned_pair` (streaming) | 104.8 us | 104.8 us | 1.00x |
| `hot_packed` (**control**) | 1254.0 us | 1253.7 us | **1.00x** |
| `hot_aligned` | 1253.9 us | **628.8 us** | **1.99x** |

Evidence: stock `.gpu-evidence/20260809T010634Z.log`, this branch
`20260809T005928Z.log`.

**Streaming does not move, and should not.** At ~480 GB/s it is bandwidth-bound,
and two loads from consecutive addresses already coalesce into the same
transactions — halving the instruction count buys nothing there, and it would be
wrong to claim otherwise. The win is where load count is the limit rather than
bandwidth: a 4096-entry table with 64 dependent lookups per thread, which is the
lookup-table shape #684 and #694 already optimised. There it tracks the 2:1 load
ratio almost exactly.

`hot_packed` holds at ~1254 us across **both** builds, so this is a controlled
comparison rather than two unrelated timings — and it is the same control that
shows the change does not over-claim alignment.

## Gates

* Full smoketest on the A10G: **192 / 192 pass, 0 fail**, 722s
  (`.gpu-evidence/20260809T012219Z.log`). A codegen change touches every field
  load in the corpus, so this is the full sweep rather than a scoped one.
* Locally: `cargo fmt --check` (root, codegen workspace, and the new example),
  `clippy --all-targets --lib --tests -D warnings`, and the unit-test set
  including `-p mir-lower` and `-p mir-importer` (11 suites).

## Note on a neighbour

#716 also adds an attribute key to `crates/llvm-export/src/lib.rs`
(`GLOBAL_RETAINED_KEY`). It is an independent change in the same region — whichever
lands first, the other rebases trivially. Flagging it so it is not a surprise at
merge time.


nihalpasham:
Really nice work. 

**GPU-verified on an RTX 5090:** 
- 1.96x on the hot lookup kernel (you measured 1.99x on A10G), 
- the PTX fusion with the align-4 control unchanged, and 
- a full 192/192 smoketest sweep on this branch. 

**I pushed a hardening commit:** 
- a power-of-two guard on the stamp, the `&raw` fidelity corner documented, 
- four mir-lower unit tests, and a verify-code-shape.sh so the fusion is CI-checked without a GPU.

One scope note: issue #745 shows two ways of writing the same slow read. This PR fixes the first one, reading named fields (`c.x + c.y`). The second one wraps the values in an array (`c.0[0] + c.0[1]`), and indexing into an array goes through a different address-computation path that doesn't yet record the alignment, so those reads still come out as two narrow loads instead of one wide one. That makes this a partial fix for #745; doing the same trick for array elements in a follow-up would finish it.

nihalpasham (approved):

PR #747, opened 2026-08-09
fix(codegen): support core::sync::atomic::fence in device code

### LINKED ISSUE #723: core::sync::atomic::fence does not translate in device code
## What happens

A plain standard-library fence inside a kernel fails device codegen:

```rust
#[kernel]
pub fn f(flag: &[u32]) {
    core::sync::atomic::fence(core::sync::atomic::Ordering::Release);
}
```

```text
  Unsupported construct: could not extract element type from core atomic intrinsic generics
```

## Why

- The core-atomic dispatcher in `crates/mir-importer/src/translator/terminator/intrinsics/atomic.rs` extracts an element type from the intrinsic's generics before routing by operation name.
- `atomic_fence` has no element type (only the ordering const generic), so every source-level fence hits the bail-out.
- There is also no fence operation in the dialect today; the fence lowering added by #695 exists only as the internal helper the RMW ordering workaround uses.

## Expected

- A fence arm ahead of the element-type extraction: read the ordering, emit a first-class fence op, lower it through the path #695 established:

```text
fence(Release / Acquire / AcqRel)  ->  fence.acq_rel.sys
fence(SeqCst)                      ->  llvm.nvvm.membar (as today for SeqCst)
```

- System scope, matching how every other core atomic in that file is imported.
- Needs the op + importer arm + lowering, a lowering test, and an example touch.

Found while landing #695: its example briefly used a source-level fence and hit this bail-out.

uurl:
Fixes #723

## Summary

Fixes translation of `core::sync::atomic::fence` in device code.

`atomic_fence` differs from the other core atomic intrinsics because it has no element-type generic. The MIR importer previously attempted to extract an atomic element type before dispatching the intrinsic, causing fence translation to fail.

This change handles fences before the common atomic type extraction path and introduces a first-class `NvvmAtomicFenceOp` carrying ordering and scope.

## Changes

* Dispatch `atomic_fence` before element-type extraction in the MIR importer.
* Extract and validate ordering-only const generics for fence intrinsics.
* Add `NvvmAtomicFenceOp` with explicit ordering and scope attributes.
* Import `core::sync::atomic::fence` with system scope.
* Reuse the existing fence lowering infrastructure introduced for atomic ordering handling.
* Lower:

  * `Acquire`, `Release`, and `AcqRel` to `fence.acq_rel.sys;`
  * `SeqCst` to `llvm.nvvm.membar.sys`
* Reject `Relaxed`, which is not a valid Rust fence ordering.
* Extend `scoped_atomic_load_store` to exercise source-level `Release` and `SeqCst` fences.

## Validation

Added lowering coverage verifying that:

* the first-class fence op is fully consumed during lowering;
* `Acquire`, `Release`, and `AcqRel` emit exactly one `fence.acq_rel.sys;`;
* the inline PTX fence carries the `~{memory}` clobber and side-effect semantics;
* `SeqCst` emits exactly one `llvm_nvvm_membar_sys` call and no inline PTX fence;
* `Relaxed` fences are rejected.

Validated with:

```text
cargo fmt --all
git diff --check

cargo test \
  -p dialect-nvvm \
  --test ops_test \
  handwritten_ops_match_reviewed_allowlist \
  -- --nocapture

cargo test \
  -p mir-lower \
  test_first_class_atomic_fence \
  -- --nocapture

cargo test
```

Also validated end-to-end by materializing and executing the updated example on `sm_120`:

```text
cargo oxide run scoped_atomic_load_store \
  --materialize-cubin \
  --arch sm_120
```

Result:

```text
source-level core fences        Release + SeqCst
=== SUCCESS: scoped atomic load/store and fences all correct ===
```


nihalpasham:
**Gist:** `fence` never had a route because the core-atomic dispatcher extracts an element type before dispatching by name, and `atomic_fence` only carries an ordering. This routes fence ahead of the type extraction as a first-class op and lowers it through the exact routes #695 established, which is the same table libcu++ uses for the scope-less C++ fence, at system scope for the same reason.

- 3411 workspace tests green, including both new lowering tests and the allowlist update,
- `scoped_atomic_load_store` passes on GPU; the PTX shows `fence.acq_rel.sys` sitting between the release publish and the neighbour read, and `membar.sys` for the SeqCst fence,
- the legacy sm_90 NVVM path still compiles this example and libNVVM accepts the LTOIR, so the compatibility sentence you dropped was actually still true; deleting it is conservative and harmless.

Merging. 

This closes #723. 

One follow-up to file separately: `compiler_fence` still dies with the misleading element-type message this PR fixed for `fence`.

nihalpasham (dismissed):

nihalpasham (approved):

Issue #722, opened 2026-08-08
cargo-oxide: device artifacts go stale when --materialize-cubin is toggled


nihalpasham:
## What happens

Run an example normally, then again with `--materialize-cubin`, without touching the source in between:

```text
cargo oxide run atomics                                      # llc path, OK
cargo oxide run atomics --materialize-cubin --arch sm_120    # DriverError(301, "file not found")
```

## Why

- The mode flag is not part of the rebuild fingerprint, so the second run reuses the stale host binary and the stale llc-path artifacts.
- The two modes want different artifact sets, and the loader gets the wrong one:

```text
normal build:        atomics.ll  atomics.ptx
materialize build:   atomics.ll  atomics.options  atomics.target
stale mix at load:   .ptx present, .target missing  ->  CUDA error 301
```

- Any source touch before the second run fixes it (forces device re-codegen with the right artifacts).

## Expected

- Toggling `--materialize-cubin` (and likely `--emit-nvvm-ir` or an `--arch` change) invalidates the device build, so artifacts always match the requested mode.

Found while landing #695; the mode only became reachable for the atomics example once that PR made atomic load/store/fence compile under libNVVM.

vyncint:
I cannot reproduce this any more — the mode is part of the rebuild fingerprint now, and the artifact sets swap cleanly in both directions.

**Runtime, A10G (sm_86, driver 595.71.05).** Used `vecadd` rather than `atomics` for the reason in the last section:

```console
$ cargo oxide run vecadd                                    # -> SUCCESS: All 1024 elements correct!
$ cargo oxide run vecadd --materialize-cubin --arch sm_86    # -> SUCCESS: All 1024 elements correct!   (no source touch)
$ cargo oxide run vecadd                                    # -> SUCCESS: All 1024 elements correct!
```

No `DriverError(301)` at any state.

**Artifacts, locally.** The "stale mix" this issue describes cannot form, because the toggle removes the artifacts belonging to the other mode:

| state | artifacts beside the example | `.target` |
|---|---|---|
| normal | `.ll` `.opt.ll` **`.ptx`** | — |
| `--materialize-cubin --arch sm_86`, no source touch | `.ll` **`.options` `.target`** | `sm_86` |
| back to normal | **`.ptx`** | — |
| `--materialize-cubin --arch sm_80` | `.options` `.target` | **`sm_80`** |
| `--arch sm_86`, no source touch | `.options` `.target` | **`sm_86`** |

[...]

PR #748, opened 2026-08-09
fix(atomics): load the embedded CUDA module

uurl:
Refs #722

### Summary

The `atomics` example explicitly loads `atomics.ptx` from the filesystem:

```rust
ctx.load_module_from_file("atomics.ptx")
```

That works for the normal PTX path, but it is incompatible with `--materialize-cubin`, where the device artifact is embedded as a cubin rather than consumed from an external PTX file.

Use the generated `kernels::load(&ctx)` loader instead. This delegates artifact selection to the embedded-module loader and works for both PTX and materialized cubin payloads.

### Relation to #722

While investigating #722 on current `main`, I could reproduce its `cargo oxide run atomics --materialize-cubin --arch sm_120` failure, but not the reported cache-invalidation cause.

The output mode transition does rebuild the device owner:

```text
cargo oxide build atomics --arch sm_120
cargo oxide build atomics --materialize-cubin --arch sm_120
cargo oxide build atomics --arch sm_120
```

`atomics` is recompiled on each PTX → cubin → PTX transition.

The same transition also works end-to-end with `vecadd`, which already uses the embedded module loader.

The remaining `atomics` failure comes from the example explicitly opening `atomics.ptx`. Switching it to `kernels::load(&ctx)` makes the original run reproducer succeed in all three states:

```text
cargo oxide run atomics
cargo oxide run atomics --materialize-cubin --arch sm_120
cargo oxide run atomics
```

All 20 atomic runtime tests pass in each mode.

This therefore fixes the current #722 reproducer without changing the existing codegen invalidation mechanism.

### Validation

```text
cargo fmt --all
cargo test -p cargo-oxide
cargo clippy -p cargo-oxide --all-targets -- -D warnings
```

`cargo-oxide` tests:

```text
154 passed; 0 failed
```

Runtime validation:

```text
cargo oxide run atomics
# SUCCESS: All 20 atomic tests passed

cargo oxide run atomics --materialize-cubin --arch sm_120
# SUCCESS: All 20 atomic tests passed

cargo oxide run atomics
# SUCCESS: All 20 atomic tests passed
```

Control using an example that already loads embedded artifacts:

```text
cargo oxide run vecadd --arch sm_120
cargo oxide run vecadd --materialize-cubin --arch sm_120
cargo oxide run vecadd --arch sm_120
```

All three runs succeed.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/atomics/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/atomics/src/main.rs
@@ -44,7 +44,7 @@ use cuda_device::atomic::{
     AtomicOrdering, BlockAtomicU32, DeviceAtomicF32, DeviceAtomicF64, DeviceAtomicI32,
     DeviceAtomicI64, DeviceAtomicU32, DeviceAtomicU64,
 };
-use cuda_device::{kernel, thread, DisjointSlice};
+use cuda_device::{DisjointSlice, kernel, thread};
 use cuda_host::cuda_module;
 
 // =============================================================================
--- a/crates/rustc-codegen-cuda/examples/atomics/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/atomics/src/main.rs
@@ -44,7 +44,7 @@ use cuda_device::atomic::{
     AtomicOrdering, BlockAtomicU32, DeviceAtomicF32, DeviceAtomicF64, DeviceAtomicI32,
     DeviceAtomicI64, DeviceAtomicU32, DeviceAtomicU64,
 };
-use cuda_device::{DisjointSlice, kernel, thread};
+use cuda_device::{kernel, thread, DisjointSlice};
 use cuda_host::cuda_module;
 
 // =============================================================================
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the fingerprint theory in #722 was already moot, the mode flags have been in the rebuild fingerprint since #340; the residual error 301 was this example loading `atomics.ptx` from disk, which materialize mode never writes and the cleanup step deletes.

**Verified on an RTX 5090:** 
- the exact three-state reproducer passes,
- plain run, then `--materialize-cubin` with no loose `.ptx` on disk, then
- plain run again, 20/20 atomic tests each time; smoketest green on the plain path.

**I pushed one commit** restoring the edition-2024 rustfmt import order, which would have failed the per-example fmt gate.

Merging. 

This closes #722. 

Follow-up: 
- 77 other examples still use
- `load_module_from_file` and stay incompatible with `--materialize-cubin`; we'll sweep those separately.

nihalpasham (approved):

PR #750, opened 2026-08-09
fix(mir-lower): support bounded shared-pointer arrays in enum payloads

### LINKED ISSUE #749: Support bounded arrays of shared-memory pointers in enum payloads
### Problem

Enum payload lowering currently supports shared-memory pointer leaves directly and through ordinary struct/tuple nesting by using target-stable generic-pointer physical storage and reconstructing the semantic address space at enum construction and extraction boundaries.

Arrays containing shared-memory pointers are still rejected.

This leaves cases such as:

```rust
struct SharedPointerArrayWrapper {
    pointers: [&'static SharedArray<u32, 1>; 2],
}

Option<SharedPointerArrayWrapper>
```

unsupported even though the existing recursive storage conversion can reconstruct aggregate values element by element.

### Why this needs an explicit bound

Array reconstruction is performed in SSA. Each shared-pointer leaf requires aggregate extraction/insertion around an address-space cast, so accepting arbitrary arrays would allow type size to translate directly into unbounded generated code and verifier work.

The support should therefore remain fail-closed unless the recursive array shape stays within an explicit expansion bound.

The bound should count shared-pointer leaves recursively rather than only the outer array length. For example:

```text
[ptr; 2]                  -> 2 leaves
[[ptr; 2]; 2]             -> 4 leaves
[struct { p1, p2 }; 4]    -> 8 leaves
[struct { p1, p2 }; 9]    -> 18 leaves
```

### Proposed behavior

Use a maximum of 16 shared-pointer leaves for recursive enum-payload array reconstruction.

Within that bound:

* convert address-space-3 pointer leaves to CUDA generic pointers for physical enum storage;
* recursively rebuild arrays and nested structs at construction/extraction boundaries;
* allow bounded pointer arrays to participate in enum pointer-overlap/niche analysis.

Outside that bound:

* reject oversized shared-pointer arrays with a diagnostic reporting the required expansion and supported bound.

Pointer vectors should remain unsupported because they have separate ABI, alignment, and address-space-cast semantics.

### Acceptance criteria

* `[shared_ptr; N]` enum payloads are supported when recursive shared-pointer leaf expansion is at most 16.
* Nested aggregate shapes use the same recursive leaf-count bound.
* Oversized arrays remain fail-closed.
* Pointer vectors remain fail-closed.
* Physical enum storage contains generic pointers rather than target-dependent address-space-3 pointers.
* Semantic payload types retain their original shared address-space semantics.
* `Option`/niche layouts containing bounded shared-pointer arrays lower correctly.
* Construction and extraction round-trip correctly on CUDA hardware.
* Existing enum payload lowering behavior remains unchanged for non-pointer arrays, bool payloads, and existing direct/nested shared-pointer cases.

uurl:
Closes #749 

### Summary

Add bounded recursive support for arrays containing shared-memory pointers in enum payloads.

Enum payload lowering already uses CUDA generic pointers as the target-stable physical representation for direct and struct/tuple-nested address-space-3 pointer leaves. This extends the same representation and reconstruction model to arrays when the recursive expansion contains at most 16 shared-pointer leaves.

### Changes

* Track the number of shared-pointer leaves during enum payload storage type rewriting.
* Allow recursive array reconstruction when the resulting expansion is at most 16 shared-pointer leaves.
* Reject larger arrays with an explicit diagnostic reporting the required conversion count and supported bound.
* Extend enum pointer-storage overlap analysis to bounded pointer arrays so niche layouts such as `Option<SharedPointerArrayWrapper>` can be represented correctly.
* Keep pointer vectors fail-closed because they require separate ABI and address-space-cast handling.
* Add `mir-lower` coverage for bounded shared-pointer arrays in niche enum payloads.
* Add positive CUDA runtime coverage using `Option` with a two-element shared-pointer array.
* Change `error_enum_shared_pointer_layout` to exercise the oversized-array boundary with 17 shared-pointer leaves.
* Update `STATUS.md` to document bounded array support and the remaining vector/oversized-array limitation.

The bound counts recursive pointer leaves rather than only the outer array length. This keeps generated aggregate reconstruction and verifier work explicitly bounded for nested shapes as well.

### Validation

```text
cargo test -p mir-lower
  179 unit tests passed
  100 integration tests passed
  0 failures

  PASS

  PASS

  PASS
  bounded shared-pointer array enum round-tripped on GPU

cargo oxide build error_enum_shared_pointer_layout
  expected failure:
  rewrite requires 17 pointer conversions, supported bound is 16

  PASS
```

### Remaining boundary

Pointer vectors remain unsupported. Their ABI/alignment behavior and address-space-cast semantics should be handled separately rather than implicitly treating them as fixed arrays.


nihalpasham:
**Gist:** this is the representation clang and `rustc_codegen_nvvm` use for aggregate-held pointers, generic in memory with the address space recovered later, extended to arrays under our two extra constraints: Rust niche encodings, and one storage layout that must hold under both the 64-bit legacy and p3:32 modern NVVM modes. Framework-conformant and fails closed everywhere we probed.

**I pushed two commits** closing the review's consistency gaps: 
- the 16-leaf budget is now enforced at the payload root, so struct-of-arrays shapes can no longer slip past the gate in direct layouts, and 
- an exhausted overlap walk now reports the bound instead of a generic provenance error.

**Verified on an RTX 5090:** 
- `addressof_sharedarray` round-trips the bounded array on GPU, the error example still fails with the exact bound
diagnostic under both layout modes, mir-lower and full workspace suites green.

Merging. This closes #749. 

Related to #306 but doesn't close it.

nihalpasham (approved):

PR #752, opened 2026-08-09
fix(mir-importer): lower from-end ConstantIndex on slices

### LINKED ISSUE #751: mir-importer: support from-end ConstantIndex projections on runtime slices
## Description

The MIR importer does not currently lower `ProjectionElem::ConstantIndex { from_end: true, .. }` for runtime slices.

This projection is emitted by rustc for slice pattern matching such as:

```rust
fn kernel(input: &[u32]) {
    if let [.., last] = *input {
        // use last
    }
}
```

For a runtime slice, a from-end constant index cannot be resolved statically. Its zero-based index must be computed from the slice metadata as:

```text
index = slice_len - offset
```

where `offset` is the `ConstantIndex` offset emitted by rustc.

This differs from fixed-size arrays: rustc normally resolves from-end indexing on arrays before this reaches the importer, so deriving the index from an array pointee length would not address the actual MIR case.

## Important correctness case

Slices whose elements are themselves arrays must use the **outer slice length**, not the array element width.

For example:

```rust
fn kernel(rows: &[[u32; 3]]) {
    if let [.., row] = *rows {
        let value = row[2];
    }
}
```

The from-end projection selects the final `[u32; 3]` row. The runtime index must therefore be based on `rows.len()`.

Using the pointee array length (`3`) would instead compute an index inside the wrong conceptual dimension and can silently generate incorrect addressing.

## Expected behavior

The importer should:

1. Recognize `ConstantIndex { from_end: true, .. }` when it immediately follows a fat-slice dereference.
2. Extract the runtime slice length from the fat value.
3. Compute `slice_len - offset`.
4. Use that result as the element index for the slice data pointer.
5. Preserve existing behavior for ordinary forward `ConstantIndex` projections.
6. Continue to fail closed for from-end projections where runtime slice-length provenance is unavailable.

## Regression coverage

A regression example should cover at least:

* `offset = 1` (`[.., last]`)
* `offset > 1` (`[.., penultimate, _]`)
* nested array elements such as `&[[u32; 3]]`
* mutable access to the selected final element

uurl:
Closes #751 

## Summary

Add MIR importer support for `ProjectionElem::ConstantIndex { from_end: true, .. }` when indexing runtime slices.

For these projections, rustc provides an offset from the end of the slice, so the importer now materializes the runtime index as:

```text
slice_len - offset
```

using the length metadata from the slice fat value.

## Implementation

The place-address walker now carries slice-length provenance across a fat-slice dereference when the following projection is a from-end `ConstantIndex`.

The lowering:

* extracts the slice data pointer,
* extracts the runtime length from field 1 of the slice value,
* materializes the constant offset as `usize`,
* emits `mir.sub` to compute `len - offset`,
* and uses the resulting index with the existing indexed-address lowering.

The runtime length is intentionally tied to the immediately preceding slice dereference and is consumed by the following projection. From-end indexing without that slice provenance continues to fall back instead of inferring a length from the pointee type.

This is important for cases such as:

```rust
&[[u32; 3]]
```

where the data pointer points to `[u32; 3]` elements. The relevant length is the number of rows in the outer runtime slice, not the inner array width `3`.

Forward `ConstantIndex` lowering is unchanged.

## Regression coverage

Adds `constant_index_from_end`, covering:

* last-element access (`offset = 1`)
* penultimate-element access (`offset = 2`)
* from-end access on `&[[u32; 3]]`
* mutable access to the last slice element

The nested-array case specifically guards against using the element array width in place of the runtime slice length.

## Validation

Validated with:

```text
cargo fmt --all
cargo test -p mir-importer

cargo oxide run slice_get_mut
cargo oxide run constant_index_from_end
```

Results:

* 928 `mir-importer` unit tests passed
* doctests passed
* `clippy -D warnings` passed
* all `slice_get_mut` kernels passed
* all `ref_index_projections` kernels passed
* `constant_index_from_end` passed on GPU

With `CUDA_OXIDE_DUMP_MIR=1`, the nested-array regression produces the expected sequence:

```text
extract slice data pointer
extract runtime slice length
constant offset
sub length, offset
ptr_offset data, computed_index
```

with the `ptr_offset` operating on a pointer to the slice element type (`[u32; 3]` in the nested-array case).


nihalpasham:
**Gist:** byte-for-byte the computation rustc_codegen_ssa performs for `from_end` ConstantIndex (extract len, subtract, index), transplanted into the importer's fat-pointer scheme, and the classifier/walker lockstep held up under an adversarial pass: every residual from-end path fails loudly rather than miscompiling.

**Verified on an RTX 5090:** 
- `constant_index_from_end` plus the two neighbouring slice examples pass on GPU; 
- the IR derives the row index from an extractvalue of the fat pair's len then a sub feeding the GEP, never a constant; 
- 928 mir-importer tests and the full workspace suite green.

**I pushed two small cleanups:** 
- the pre-existing tail-borrow site now reuses your `emit_slice_len_extract` helper, and 
- the example gained a README for parity with `slice_get_mut`.

Merging. 

This closes #751. 

Subslice (#753) stays separate and lands with #754.

nihalpasham:
**Gist:** a faithful transplant of rustc's own Subslice lowering, project at `from`, retype, `len - (from + to)` for the unsized tail, with the lookahead machinery our Deref-time fat-pointer scalarization forces. The verifier also confirmed the `[T; 0]` shape lowers as well-formed, already-supported IR, so no extra guard was needed.

**Verified on an RTX 5090:** 
- `subslice_projection` 5/5 on GPU, 
- full smoketest 193/193, workspace suite green on a quiet host.

**I force-pushed a restack:** 
- rebased onto current main (the base had drifted), added the LICENSE file sibling examples carry, and one fmt fix
for the per-example edition-2024 gate. 
- Diff content otherwise unchanged.

Merging. 

This closes #753.

nihalpasham (approved):

PR #754, opened 2026-08-09
fix(mir-importer): lower `Subslice` projections

### LINKED ISSUE #753: MIR importer: support `ProjectionElem::Subslice`
### Problem

The MIR importer does not currently lower `ProjectionElem::Subslice`.

`rustc` emits this projection for subslice patterns such as:

```rust
let [_, middle @ .., _] = values;
```

The projection appears in two relevant forms:

* Arrays use `Subslice { from, to, from_end: false }`, producing a sized `[T; N]` projection.
* Slices use `Subslice { from, to, from_end: true }`, with semantics equivalent to `slice[from..len - to]`.

Optimized MIR may also continue projecting from a slice subslice, for example:

```text
(*slice)[1:-1][0 of 1]
```

so handling `Subslice` only as a terminal projection is insufficient.

### Expected behavior

The MIR importer should support:

* array subslices as values;
* shared references to array subslices;
* mutable references to array subslices;
* slice subslices while preserving and updating fat-pointer metadata;
* subsequent `Index` / non-from-end `ConstantIndex` projections from a slice subslice;
* mutable write-through through a projected slice.

For slice subslices, lowering must:

1. advance the data pointer by `from`;
2. compute the new length as `len - (from + to)`;
3. reconstruct the slice fat value when metadata is required.

### Reproducer

A CUDA example can exercise the following cases:

* array value;
* array shared reference;
* array mutable reference;
* slice metadata;
* slice mutable reference.

The issue is complete when these cases compile through the CUDA Oxide pipeline and execute successfully on the GPU.

uurl:
Closes #753 

### Summary

Adds MIR importer support for `ProjectionElem::Subslice` for both arrays and slices.

`rustc` emits two forms that require different lowering strategies:

* `from_end: false` for sized array subslices;
* `from_end: true` for slice subslices.

### Implementation

For array subslices, this PR:

* computes the projected `[T; to - from]` type;
* lowers address projections starting at element `from`;
* preserves pointer address space and mutability;
* materializes array subslices when the projection is evaluated as an SSA value.

For slice subslices, this PR preserves the fat slice long enough to update both components:

```text
data' = data + from
len'  = len - (from + to)
```

It then reconstructs the slice value with the adjusted data pointer and metadata.

The lowering also supports optimized MIR where a slice `Subslice` is followed by `Index` or a non-from-end `ConstantIndex`, such as:

```text
(*slice)[1:-1][0 of 1]
```

This is required for mutable write-through cases emitted by rustc.

### Regression coverage

Adds the `subslice_projection` CUDA example covering:

* array subslice value;
* shared reference to an array subslice;
* mutable reference to an array subslice;
* slice subslice metadata;
* mutable slice subslice write-through.

### Validation

Validated with:

```text
cargo fmt --all
cargo check
cargo test -p mir-importer
cargo oxide run subslice_projection
cargo oxide pipeline subslice_projection
```

Results:

```text
mir-importer: 928 passed, 0 failed

array value          PASS
array shared ref     PASS
array mutable ref    PASS
slice metadata       PASS
slice mutable ref    PASS
```

The pipeline also completes MIR translation, LLVM lowering, and PTX generation successfully.


nihalpasham:
**Gist:** a faithful transplant of rustc's own Subslice lowering, project at `from`, retype, `len - (from + to)` for the unsized tail, with the lookahead machinery our Deref-time fat-pointer scalarization forces. The verifier also confirmed the `[T; 0]` shape lowers as well-formed, already-supported IR, so no extra guard was needed.

**Verified on an RTX 5090:** 
- `subslice_projection` 5/5 on GPU, 
- full smoketest 193/193, workspace suite green on a quiet host.

**I force-pushed a restack:** 
- rebased onto current main (the base had drifted), 
- added the LICENSE file sibling examples carry, and 
- one fmt fix for the per-example edition-2024 gate. 
- Diff content otherwise unchanged.

Merging. 

This closes #753.

nihalpasham (approved):

Issue #708, opened 2026-08-07
Warn when #[launch_bounds] can only be satisfied by spilling registers


nh13:
**What problem does this solve?**

`#[launch_bounds(max_threads, min_blocks_per_sm)]` constrains the register allocator. When the requested occupancy is not achievable within the register budget, `ptxas` satisfies the bound by spilling — and the attribute then makes the kernel slower, silently. There is no compile-time signal; you find out by benchmarking.

We hit both outcomes from the *same* attribute with the *same* bound, on two kernels in one build:

| kernel | without | with `#[launch_bounds(256, 2)]` | delta | spill after |
|---|---|---|---|---|
| A | 4.346 / 4.359 s | 3.925 / 3.940 s | **-0.42 s** | **0 bytes** |
| B | 5.268 / 5.274 s | 6.803 / 6.799 s | **+1.53 s** | **24 bytes** |

Both kernels landed on exactly 128 registers and both had their PTX `.maxntid`/`.minnctapersm` directives verified. The occupancy argument for applying the attribute was identical for the two and predicted nothing. The **only** thing that separated them was `ptxas -v`'s spill counter.

So the information that decides the outcome is already computed — it is just not surfaced anywhere a user of the attribute will see it.

**Proposed API / behavior**

Warn at build time when a `#[launch_bounds]` request results in nonzero spill stores/loads:

```
warning: `#[launch_bounds(256, 2)]` on `seed_intv_kernel` forced 24 bytes of register spill
   = note: the requested 2 blocks/SM needs <= 128 registers; the kernel wants more
   = help: relax min_blocks_per_sm, or reduce register pressure
```

**Alternatives considered**

Running `ptxas -v` and reading spill counters per kernel per configuration, which is what we now do, and which requires a toolkit install plus knowing to look. Benchmarking every kernel with and without the attribute, which is what we did first and which costs a build and a measurement per kernel.

**Additional context**

Companion to the local-memory-promotion diagnostic I am filing alongside this. Both are "the register allocator lost and only the assembler knows".

nihalpasham:
Implemented by #757 (merged 2026-08-05): ptxas resource reports are parsed on cubin-materialization builds and every `#[launch_bounds]` kernel that spills registers gets a warning at its definition span, with `CUDA_OXIDE_NO_SPILL_WARN=1` as the measured-and-accepted escape hatch. The book section documenting the warning landed with #798.

Issue #707, opened 2026-08-07
No diagnostic when a local falls to local memory because it is dynamically indexed


nh13:
**What problem does this solve?**

When a small fixed-size local array is indexed by a value the compiler cannot see as constant, SROA cannot promote it and it is placed in local memory. On a CPU that is a stack slot and usually cheap. On a GPU it is a performance cliff, and there is currently no signal that it happened short of reading `ptxas -v` output or the PTX itself.

This cost us **31% of one of our hottest kernels**, unnoticed for months. The pattern:

```rust
let ok = idx.extend(&p, b);         // -> [Sai; 4], 128 bytes
if ok[c].size < min_occ { break }   // `c` is a query base -- a RUNTIME index
p = ok[c];
```

What made it hard to find is that the cost is invisible at the definition. `extend` itself reports a **0 byte stack frame** — every index inside it is a literal, so SROA promotes it fine — and the entire penalty lands at the *call sites*, which is the last place you look after reading a clean callee. Restructuring so the runtime index is resolved once into four literal-index calls gave, on `sm_75`:

| kernel | stack frame | registers |
|---|---|---|
| `seed_intv_kernel` | 464 -> **144 B** | 138 -> **126** |
| `seed_intv_forward_kernel` | 304 -> 272 B | 122 -> 110 |
| `smem_seed` | 432 -> 400 B | 104 -> 92 |

and took that kernel's device time down 31.5% (3.624/3.627/3.635/3.643 s -> 2.474/2.491/2.496/2.491 s, 4 interleaved reps per arm, disjoint). It also brought the register count under the threshold for 2 blocks/SM with zero spill, which was a second win on top.

We found it by taking a full `ptxas -v` census across every kernel and asking which ones had nonzero stack frames — not by reading the source, which looks entirely reasonable.

**Proposed API / behavior**

A diagnostic, off by default or behind a flag, along the lines of:

```
warning: local `ok` ([Sai; 4], 128 bytes) could not be promoted to registers
  --> src/walker.rs:214:9
   = note: indexed by a non-constant value; the allocation is placed in local memory
```

Even a summary at the end of codegen — per kernel, which locals stayed in local memory and how large — would be enough to make this findable. Today it surfaces only as an unexplained `N bytes stack frame` line from the assembler, with no mapping back to a Rust binding.

[...]

PR #756, opened 2026-08-09
Expose ptxas resource diagnostics from native finalization

uurl:
Part of #708 

## Summary

Adds structured access to `ptxas` resource diagnostics produced during nvJitLink finalization.

This is the first part of #708. It provides the resource-reporting infrastructure needed for a later change to warn when kernels compiled with `#[launch_bounds]` spill registers.

## Changes

* preserve the nvJitLink informational log alongside linked cubin/PTX output;
* request verbose nvJitLink / `ptxas` reporting when resource diagnostics are explicitly collected;
* parse per-kernel:

  * register usage;
  * spill store bytes;
  * spill load bytes;
* expose the parsed resource data through `LinkReport`;
* keep diagnostic-only nvJitLink options separate from semantic link options so they do not affect provenance or artifact cache keys.

Existing `finish`, `finish_ptx`, and normal finalization APIs retain their previous behavior.

## Testing

Added unit coverage for:

* zero and non-zero spill reports;
* multiple kernels in one log;
* repeated resource sections;
* unrelated informational output;
* spill detection in either direction;
* separation of diagnostic and semantic nvJitLink options.

Validated with:

```text
cargo fmt --all
cargo check
cargo test
```

All checks pass.

### DIFF
--- a/crates/cuda-artifact-finalizer/src/diagnostics.rs
+++ b/crates/cuda-artifact-finalizer/src/diagnostics.rs
@@ -1,0 +1,188 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Parsing of non-semantic CUDA compiler resource diagnostics.
+//!
+//! nvJitLink can forward `ptxas -v` output through its InfoLog. This module
+//! converts that human-readable output into a stable internal representation
+//! without making the raw text part of artifact provenance.
+
+/// Per-kernel resource usage reported by ptxas.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct KernelResourceUsage {
+    /// Kernel entry name as emitted in PTX/cubin.
+    pub kernel: String,
+    /// Register count reported by ptxas, when present.
+    pub registers: Option<u32>,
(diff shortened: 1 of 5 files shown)


nihalpasham:
**Gist:** this is the right layer for a build-time resource report. `XLA` does the same thing, drive ptxas through the toolchain and read its diagnostics back; Triton's structured route reads driver attributes off a loaded module, which needs a GPU and a context a build host can't assume. The structured parse also captures the spill byte counts #708's warning text needs, which a substring check can't.

**I pushed five commits on top:** 
- typed detection of `NVJITLINK_ERROR_UNRECOGNIZED_OPTION` with graceful degradation to an empty report on older toolkits, 
- stack-frame-bytes parsing plus the colon-terminated kernel-name fix, 
- a JIT-cache bypass so report collection can't be skipped by a warm cache, and 
- an `#[ignore]`d live test that runs the real toolkit.

**Verified here with CUDA 13.3:** 
- the live test links a genuinely spilling kernel and 
- the parsed report matches the raw ptxas log line for line; 
- artifact digests provably exclude the diagnostic flags.

Merging. 

Partially addresses #708, the warning itself lands with #757; related to #707. 

Follow-up we'll file: the crate's pre-existing live test has been validating an empty module all along, its NVVM IR lacks
`@llvm.used` so LTO dead-strips the kernel.

nihalpasham (approved):

PR #757, opened 2026-08-09
Warn when launch-bounded kernels spill registers

uurl:
Depends on #756.
Please review/merge #756 first.
Refs #708.

## Summary

Adds a rustc warning when a kernel compiled with `#[launch_bounds]` reports register spills during native cubin materialization.

This builds on the resource-reporting infrastructure introduced in #756 and addresses the warning path described in #708.

## Changes

* preserve per-kernel launch-bound metadata from the MIR importer through device codegen;
* associate launch bounds with exported kernel names;
* collect `ptxas` resource usage during cubin materialization;
* detect kernels with non-zero spill stores or spill loads;
* emit a span-based warning for launch-bounded kernels that spill registers;
* include spill store/load bytes and register usage when available;
* provide remediation guidance based on whether `min_blocks_per_sm` was specified.

Example diagnostic:

```text
warning: kernel `foo` compiled with `#[launch_bounds(256, 2)]` and spills registers
note: ptxas reports 32 bytes spill stores and 32 bytes spill loads
note: ptxas allocated 64 registers per thread
help: relax `min_blocks_per_sm` or reduce register pressure
```

The warning intentionally does not claim that `#[launch_bounds]` caused the spill. It reports only the facts available from compilation: the kernel has launch bounds and `ptxas` reported register spilling.

## Scope

The diagnostic is emitted when native cubin materialization runs, because that is the stage where `ptxas` resource information becomes available.

Normal PTX-only compilation does not run `ptxas`, so it cannot provide exact spill information.

## Testing

Validated with:

```text
cargo fmt --all
cargo check
cargo test
git diff --check
```


nihalpasham:
**Gist:** ptxas's `--warn-on-spills` diagnostic lifted to a rustc span warning, gated on `#[launch_bounds]`, at the only stage that has the mangled name, the span, and the ptxas text all in hand, which is good; but the backend crate didn't compile as submitted, `rustc-codegen-cuda` sits outside the root workspace so the stated validation never built it.

**I force-pushed a restack** onto the updated #756 plus three commits:
- the missing `KernelLaunchBounds` re-export (the compile break), a `CUDA_OXIDE_NO_SPILL_WARN` escape hatch since span warnings can't be `allow`ed away, and a doc note recording that reported and 
- plain links are code-identical (SASS equal; only ptxas's self-recorded option line in `.note.nv.tkinfo` differs, and that section is outside the digest).
- A fourth commit satisfies the edition-2024 fmt gate.

**Verified on an RTX 5090 with CUDA 13.3:** 
- a deliberately spilling launch-bounded kernel produces exactly one warning whose 428/436 spill bytes match a manual `ptxas -v` run; 
- the non-spilling kernel is silent; workspace, clippy, and the finalizer's live tests all green.

Merging after #756. 

Partially addresses #708 within the disclosed `--materialize-cubin` scope; related to #707.

nihalpasham (approved):

PR #758, opened 2026-08-09
feat(codegen): diagnose locals that survive register promotion

### LINKED ISSUE #707: No diagnostic when a local falls to local memory because it is dynamically indexed

uurl:
Closes #707

## Summary

Adds an opt-in diagnostic for named Rust locals whose allocations remain stack-backed after LLVM optimization.

The diagnostic runs on the final optimized LLVM IR passed to `llc`, avoiding reports for allocations that are subsequently removed by mem2reg or SROA.

## Implementation

* Attach source-level provenance to named Rust MIR locals, including the binding name, type, and allocation size.
* Preserve that provenance through MIR-to-LLVM lowering.
* Encode the provenance into the exported LLVM SSA name so surviving allocations can be identified after `opt`.
* Inspect the optimized LLVM IR immediately before `llc`.
* Report surviving tagged allocations in verbose optimized builds.
* Provide specific diagnostics for dynamic indexing and obvious address escapes, with a conservative fallback when the blocker cannot be identified.
* Skip the diagnostic for non-optimized and full variable-debug builds, where stack-backed storage may be intentional.

Example:

```text
warning: local `scratch` (`[u32; 4]`, 16 bytes) in `kernel` could not be promoted to registers
  = note: indexed by a non-constant value; the allocation survives LLVM scalar replacement and remains stack-backed
  = help: use constant indexing or a small match/select when the index range is bounded
```

## Testing

Added coverage for:

* dynamic-indexed locals
* surviving alloca names with optimizer suffixes
* pointer/address escapes
* LLVM lifetime intrinsics
* untagged compiler-generated allocas
* successfully promoted locals
* verbose and non-verbose codegen behavior
* MIR-to-LLVM provenance preservation

Validated with:

```text
cargo fmt --all -- --check

cargo test \
  -p dialect-mir \
  -p mir-lower \
  -p cuda-oxide-codegen \
  -p mir-importer

cargo clippy \
  -p dialect-mir \
  -p mir-lower \
  -p cuda-oxide-codegen \
  -p mir-importer \
  --all-targets -- -D warnings

```


nihalpasham:
**Gist:** exceeds what #707 asked for, and the placement is right: the diagnostic reads post-opt IR through the one channel that survives O2, since full variable debug info is only available when we skip opt entirely. `ptxas -v` tells you local memory exists; this tells you which binding and why.

**I pushed five commits:**
- provenance is now a first-class pliron attribute on both allocas instead of a tab-separated StringAttr, with the string encoding surviving only at the textual-export boundary where it is genuinely forced; 
- the hex payload is sentinel-terminated with a collision test so LLVM's name-uniquing suffixes can't garble a report; warning text is demangled and bounded; and 
- the new rustc-demangle dependency is recorded in the license inventory. 
- One review suggestion was dropped with evidence: modern SROA has no aggregate-size threshold, so the help text says something true instead.

**Verified with real `opt -O2` on this host:** 
- dynamically indexed locals warn with the right binding, size, and reason; 
- vecadd and six aggregate-heavy examples build warning-free; 
- compile-only smoketest 191/191; workspace suite green.

Merging. 

This closes #707. Related to #708 (ptxas-introduced spills are out of scope here) and #399.

nihalpasham (approved):

PR #760, opened 2026-08-09
feat(cuda-core): expose the primary context's device limits

### LINKED ISSUE #759: cuda-core exposes only the stack-size context limit and cannot raise a truncating printf FIFO
`CudaContext` reaches only one of the seven context limits the CUDA driver defines, which leaves device `printf` output truncating with no supported way to stop it.

## What is missing

`CudaContext` wraps one of the seven `CUlimit` values, `CU_LIMIT_STACK_SIZE`, through `stack_size` and `set_stack_size`. The other six have no wrapper, so reaching them means going through `cuda_core::sys` and calling `cuCtxSetLimit` by hand with a raw `CUlimit`.

## Why the printf FIFO is the case that costs users output

`CU_LIMIT_PRINTF_FIFO_SIZE` backs the device `printf` this repository already supports (`crates/rustc-codegen-cuda/examples/printf`). The FIFO is circular. Once a launch fills it the driver overwrites the oldest entries, prints what remains, and reports nothing at all: the write returns success and the output carries no marker where the loss happened. A kernel whose output is cut therefore looks like a kernel that stopped early, and the only fix is to raise the limit before the launch.

Two further properties make this awkward to reach for by hand:

- The driver refuses `CU_LIMIT_PRINTF_FIFO_SIZE` and `CU_LIMIT_MALLOC_HEAP_SIZE` with `CUDA_ERROR_INVALID_VALUE` once any kernel using `printf` or device `malloc` has run.
- `CudaContext` retains the device's **primary** context, which is process-wide, so that ordering counts launches made through any handle in the process rather than through the handle doing the write.

Neither fact is discoverable from the raw entry point, and both belong in the wrapper's documentation.

## Observed on an RTX 5060 Laptop GPU at compute capability 12.0 under CUDA 13.3

```text
CU_LIMIT_STACK_SIZE                       = 1024
CU_LIMIT_PRINTF_FIFO_SIZE                 = 1277952
CU_LIMIT_MALLOC_HEAP_SIZE                 = 8388608
CU_LIMIT_DEV_RUNTIME_SYNC_DEPTH           -> CUDA_ERROR_UNSUPPORTED_LIMIT (215)
CU_LIMIT_DEV_RUNTIME_PENDING_LAUNCH_COUNT = 2048
CU_LIMIT_MAX_L2_FETCH_GRANULARITY         = 64
CU_LIMIT_PERSISTING_L2_CACHE_SIZE         = 6291456
```

`CU_LIMIT_DEV_RUNTIME_SYNC_DEPTH` answering `CUDA_ERROR_UNSUPPORTED_LIMIT` is the documented behaviour above compute capability 9.0, so a wrapper has to treat that status as an ordinary answer rather than a fault.

## Expected

A `ContextLimit` enum naming the limits, with `CudaContext::limit` and `CudaContext::set_limit`, and `stack_size` and `set_stack_size` kept as they are on top of it.

Three values should stay out. `CU_LIMIT_SHMEM_SIZE` and `CU_LIMIT_CIG_ENABLED` are query-only, all three CIG values concern graphics-interop contexts, and none of them predates CUDA 12.5, so naming them needs a header probe and a `cfg` in the manner of `cuda_has_cuEventElapsedTime_v2` in `crates/cuda-bindings/build.rs`.

alejandro-soto-franco:
Closes #759.

`CudaContext` wrapped one of the seven `CUlimit` values, the thread stack size. The rest were reachable only through `cuda_core::sys`, so raising the device `printf` FIFO meant calling `cuCtxSetLimit` by hand with a raw `CUlimit`.

## Summary

Adds `ContextLimit` and the `CudaContext::limit` and `CudaContext::set_limit` pair. `stack_size` and `set_stack_size` keep their signatures and now delegate to that pair.

## Changes

- `crates/cuda-core/src/context.rs`: `ContextLimit` with the seven limits `cuCtxSetLimit` documents, a private `to_raw`, and the `limit` and `set_limit` methods. Each variant documents its own restrictions, including which two the driver treats as hints and which one answers `CUDA_ERROR_UNSUPPORTED_LIMIT` above compute capability 9.0.
- `crates/cuda-core/src/lib.rs`: re-exports `ContextLimit` next to `CudaContext` and `SyncPolicy`.
- `crates/cuda-core/tests/context_limits.rs`: new test binary, for the reason `context_sync_policy.rs` has its own.

## Rationale

The printf FIFO is the limit that costs users output. It is circular, so once a launch fills it the driver overwrites the oldest entries and reports nothing: the launch succeeds and the printed output carries no marker where the loss happened. A truncated kernel log reads as a kernel that stopped early.

The wrapper carries two properties of that limit the caller would otherwise have to know. The driver refuses `CU_LIMIT_PRINTF_FIFO_SIZE` and `CU_LIMIT_MALLOC_HEAP_SIZE` with `CUDA_ERROR_INVALID_VALUE` once a kernel using `printf` or device `malloc` has run, and `CudaContext` retains the device's primary context, so that ordering counts every launch in the process rather than the launches made through the handle doing the write.

`CU_LIMIT_SHMEM_SIZE`, `CU_LIMIT_CIG_ENABLED` and `CU_LIMIT_CIG_SHMEM_FALLBACK_ENABLED` stay out. The first two are query-only, all three concern CIG contexts, and none predates CUDA 12.5, so naming them needs a header probe and a `cfg` in the manner of `cuda_has_cuEventElapsedTime_v2` in `crates/cuda-bindings/build.rs`. Adding them without that probe would break the build against every toolkit this repository still supports.

The test asserts an exact round trip on the stack size, pins the inherent accessors to the same value the enum path reports, and then writes `MallocHeapSize` and reads `PrintfFifoSize` back unchanged. The last assertion is the one that discriminates: a `to_raw` mapping two variants onto one `CUlimit` passes every other check in the file. Mapping `MallocHeapSize` onto `CU_LIMIT_PRINTF_FIFO_SIZE` on purpose fails it with `left: 2326528, right: 1277952`.

## Validation

```text
cargo build -p cuda-core
cargo test -p cuda-core                       26 unit, 31 integration, 3 ignored, 0 failed
cargo test -p cuda-core --test context_limits 1 passed
cargo clippy -p cuda-core --all-targets       clean
cargo fmt --check -p cuda-core                clean
RUSTDOCFLAGS="-D warnings" cargo doc -p cuda-core --no-deps   clean
```

Limits read on that device, showing the documented refusal rather than a fault:

```text
StackSize                    = 1024
PrintfFifoSize               = 1277952
MallocHeapSize               = 8388608
DevRuntimeSyncDepth          -> DriverError(215, "limit is not supported on this architecture")
DevRuntimePendingLaunchCount = 2048
MaxL2FetchGranularity        = 64
PersistingL2CacheSize        = 6291456
```

### DIFF
--- a/crates/cuda-core/src/context.rs
+++ b/crates/cuda-core/src/context.rs
@@ -88,6 +88,108 @@ impl PartialEq for CudaContext {
 }
 impl Eq for CudaContext {}
 
+/// A per-context device limit (`CU_LIMIT_*`), read with
+/// [`CudaContext::limit`] and written with [`CudaContext::set_limit`].
+///
+/// Every limit is state on the device's **primary** context, so it is shared
+/// by every [`CudaContext`] for that device, by any runtime-API user of the
+/// same device in this process, and across library boundaries. A limit set
+/// here is not scoped to the handle that set it.
+///
+/// The driver is free to clamp or round a request. Read the limit back after
+/// setting it to observe what was actually applied.
+///
+/// # Ordering constraints
+///
+/// [`PrintfFifoSize`](Self::PrintfFifoSize) and
+/// [`MallocHeapSize`](Self::MallocHeapSize) must be set **before the first
+/// kernel launch in the process that uses `printf` or device `malloc`**;
+/// afterwards the driver rejects the write with `CUDA_ERROR_INVALID_VALUE`.
+/// Because the limit lives on the shared primary context, "first launch"
+/// counts launches made through any handle, not just this one.
+///
+/// # Omitted variants
+///
+/// `CU_LIMIT_SHMEM_SIZE`, `CU_LIMIT_CIG_ENABLED` and
+/// `CU_LIMIT_CIG_SHMEM_FALLBACK_ENABLED` are absent. The first two are
+/// query-only, all three concern CIG (graphics-interop) contexts, and none
+/// predates CUDA 12.5, so naming them would need a header probe and a `cfg`
+/// in the manner of `cuda_has_cuEventElapsedTime_v2`. The seven variants
+/// below have been present since CUDA 11 and are the ones
+/// `cuCtxSetLimit`'s own documentation specifies.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ContextLimit {
+    /// Stack size in bytes for each GPU thread (`CU_LIMIT_STACK_SIZE`).
+    ///
+    /// The driver raises this on its own whenever a launched kernel needs a
+    /// larger frame and **does not lower it again**, so a read reflects the
+    /// high-water mark of everything launched in this context so far. The
+    /// reservation scales as roughly `bytes * mp_count * threads_per_mp` and
+    /// is invisible to the allocation APIs: it simply reduces the device
+    /// memory a later allocation can obtain. Lowering it after a deep-frame
+    /// kernel has raised it is how that memory is handed back.
+    StackSize,
+    /// Size in bytes of the FIFO backing the device `printf`
+    /// (`CU_LIMIT_PRINTF_FIFO_SIZE`).
+    ///
+    /// The FIFO is circular: once a launch fills it, the **oldest** output is
+    /// overwritten and lost silently, with no error and no marker in the
+    /// stream the host prints. Raising this is the only fix for a kernel
+    /// whose output is truncated. Subject to the ordering constraint above.
+    PrintfFifoSize,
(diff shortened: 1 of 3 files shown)


nihalpasham:
**GPU validation on 2x RTX 5090 (CC 12.0, driver 580.173.02):** 
- the new `context_limits` test passes and the full cuda-core suite is green (59 passed, 0 failed). 
- clean wrapper, matches the `SyncPolicy` precedent. 

Merging.

nihalpasham (approved):

PR #762, opened 2026-08-09
feat(cuda-core): create streams at a device priority

### LINKED ISSUE #761: cuda-core cannot create a stream at a device priority
`cuda-core` creates every stream through `cuStreamCreate`, so there is no supported way to order two streams against each other on the device.

## What is missing

`CudaContext::new_stream` and `CudaStream::fork` both call `cuStreamCreate` with `CU_STREAM_NON_BLOCKING`. None of `cuStreamCreateWithPriority`, `cuCtxGetStreamPriorityRange` or `cuStreamGetPriority` is wrapped, and the word "priority" does not appear in the crate. A caller who wants a latency-sensitive kernel to preempt a long background one has to go through `cuda_core::sys` and reproduce the range handling by hand.

## Why the raw API is easy to get wrong

Two properties of the CUDA priority model defeat the obvious reading:

- **Lower numbers are higher priorities.** The range is reported as a `(leastPriority, greatestPriority)` pair where `greatest <= least`, so a caller reading the pair positionally gets the ordering backwards.
- **An out-of-range priority is clamped silently.** `cuStreamCreateWithPriority` does not refuse a value outside the range. It clamps to the nearest end and returns `CUDA_SUCCESS`, so `i32::MIN` produces the greatest supported priority and nothing reports the substitution.

A third property belongs in the documentation rather than the types: priority orders compute kernels only. Host-to-device and device-to-host transfers are unaffected, so a high-priority stream does not jump a queue of copies.

## Observed on an RTX 5060 Laptop GPU at compute capability 12.0 under CUDA 13.3

```text
range: least=0 greatest=-5

requested  -2147483648 -> driver  -5
requested         -100 -> driver  -5
requested           -1 -> driver  -1
requested            0 -> driver   0
requested            1 -> driver   0
requested          100 -> driver   0
requested   2147483647 -> driver   0
```

Every out-of-range request lands on an end of the range with no error.

## Expected

`CudaContext::stream_priority_range` returning a type that names the two ends rather than a bare tuple, `CudaContext::new_stream_with_priority`, and `CudaStream::priority` so a caller can observe what the driver applied.

A device without priority support reports `0` for both ends, which the type should be able to answer directly.

alejandro-soto-franco:
Closes #761.

Every stream in `cuda-core` came from `cuStreamCreate`, so ordering a latency-sensitive kernel against a long background one meant reaching through `cuda_core::sys` and reproducing the range handling by hand.

## Summary

Adds `StreamPriorityRange`, `CudaContext::stream_priority_range`, `CudaContext::new_stream_with_priority` and `CudaStream::priority`.

## Changes

- `crates/cuda-core/src/context.rs`: `StreamPriorityRange` with `least`, `greatest`, `contains`, `clamp` and `is_supported`; `stream_priority_range` wrapping `cuCtxGetStreamPriorityRange`; `new_stream_with_priority` wrapping `cuStreamCreateWithPriority`; a private `create_stream` now shared with `new_stream`.
- `crates/cuda-core/src/stream.rs`: `CudaStream::priority` wrapping `cuStreamGetPriority`.
- `crates/cuda-core/src/lib.rs`: re-exports `StreamPriorityRange`.
- `crates/cuda-core/tests/stream_priority.rs`: four tests.

## Rationale

Two properties of the CUDA priority model defeat the obvious reading of the raw API, and both are things a type can hold.

Lower numbers are higher priorities, so a `(least, greatest)` pair read positionally gives the ordering backwards. `greatest()` and `least()` name the ends instead, and the invariant `greatest <= least` is stated where a caller will read it.

An out-of-range priority is clamped rather than refused: `cuStreamCreateWithPriority` returns `CUDA_SUCCESS` and substitutes the nearest end, so `i32::MIN` becomes the greatest supported priority with nothing to show for the substitution. `clamp` answers that before the call and `CudaStream::priority` observes it after, which is the pair a caller needs since the driver will not say.

`new_stream` now delegates to `create_stream`, so the live-stream count and the first-call `synchronize` cannot drift between the two constructors. The no-priority path keeps calling `cuStreamCreate` rather than passing a literal 0, since the driver documents `cuStreamCreate` as the default request and a device whose range excludes 0 would treat an explicit 0 differently.

`is_supported` reads the both-ends-zero answer `cuCtxGetStreamPriorityRange` gives on a device without priority support. Stream creation still succeeds there and every priority collapses to one value, so this reports a capability rather than an error condition.

The clamp test is the one that discriminates. Asserting only that an out-of-range creation succeeds would pass against any `clamp` at all, so the test asserts the stream lands exactly where `clamp` predicted. Replacing the body of `clamp` with the identity fails it and leaves the other three passing.

A priority-carrying `fork` is the obvious follow-on and is left out of this change to keep it to one idea.

## Validation

RTX 5060 Laptop GPU, compute capability 12.0, CUDA 13.3, driver 610.43.03. Device range `least=0 greatest=-5`.

```text
requested  -2147483648 -> clamp()  -5 -> driver  -5
requested         -100 -> clamp()  -5 -> driver  -5
requested           -1 -> clamp()  -1 -> driver  -1
requested            0 -> clamp()   0 -> driver   0
requested            1 -> clamp()   0 -> driver   0
requested          100 -> clamp()   0 -> driver   0
requested   2147483647 -> clamp()   0 -> driver   0
```

`clamp` matches the driver at all seven probes, both extremes included.

```text
cargo test -p cuda-core                        26 unit, 34 integration, 2 doc, 3 ignored, 0 failed
cargo test -p cuda-core --test stream_priority 4 passed
cargo clippy -p cuda-core --all-targets        clean
cargo fmt --check -p cuda-core                 clean
RUSTDOCFLAGS="-D warnings" cargo doc -p cuda-core --no-deps    clean
```


alejandro-soto-franco:
This applies to `main` on its own, as does #760, and the two are independent in what they change. They do overlap textually in two places: both add a variant to the `pub use context::{..}` list in `crates/cuda-core/src/lib.rs`, and both insert a type immediately above `SyncPolicy` in `crates/cuda-core/src/context.rs`.

Whichever lands second needs a rebase over those two lines. I will rebase this one once the other merges, in whichever order they go in.

nihalpasham:
**Validated on 2x RTX 5090:** 
- all four `stream_priority` tests pass (range ordering, priority readback, out-of-range clamping, default priority) and 
- the full `cuda-core` suite is green. 
- the typed range with named ends fixes the positional footgun the raw API has. 

Merging.

nihalpasham (approved):

PR #764, opened 2026-08-09
feat(cuda-core): report stream completion without blocking

### LINKED ISSUE #763: CudaStream has no non-blocking completion check
`CudaStream` can wait for its work but cannot ask whether that work has finished, so a caller that must not block has no way to find out.

## What is missing

`CudaEvent` wraps both halves of the completion question: `synchronize` blocks, and `query` answers `true` or `false` without blocking by mapping `CUDA_ERROR_NOT_READY` onto `false`.

`CudaStream` wraps only `synchronize`. `cuStreamQuery` has no wrapper, so the same question asked of a stream costs the calling thread its progress.

## Why this shape matters

The `stream` module documents `launch_host_function` as "the primary bridge between CUDA stream completion and Rust `async` futures". A `Future::poll` implementation completes that bridge only if it can ask whether the work has finished and return `Poll::Pending` when it has not. `synchronize` cannot answer that question: parking the executor's thread inside `poll` is the failure the bridge exists to avoid.

The callback route can be made to work by moving a `Waker` into the closure, but that fixes only the case where a waker exists to move. Reading a stream's status during a shutdown path, a progress display, or a test has no callback to hang from.

## Expected

`CudaStream::query`, wrapping `cuStreamQuery` with the mapping `CudaEvent::query` already uses for `cuEventQuery`.

alejandro-soto-franco:
Closes #763.

`CudaEvent` carries both halves of the completion question. `CudaStream` carried only the blocking half, so asking whether a stream had finished cost the calling thread its progress.

## Summary

Adds `CudaStream::query`, wrapping `cuStreamQuery` with the mapping `CudaEvent::query` already uses for `cuEventQuery`.

```rust
pub fn query(&self) -> Result<bool, DriverError>
```

## Changes

- `crates/cuda-core/src/stream.rs`: `CudaStream::query`, mapping `CUDA_SUCCESS` to `Ok(true)` and `CUDA_ERROR_NOT_READY` to `Ok(false)`, with any other status returned as a `DriverError`.
- `crates/cuda-core/tests/stream_query.rs`: two tests.

## Rationale

The `stream` module documents `launch_host_function` as the bridge between stream completion and Rust `async` futures. A `Future::poll` completes that bridge only if it can ask whether work has finished and return `Poll::Pending` when it has not, and `synchronize` cannot answer without parking the executor's thread, which is the failure the bridge exists to avoid.

Mirroring `CudaEvent::query` rather than inventing a shape keeps one reading of `CUDA_ERROR_NOT_READY` in the crate. That status is a completion answer rather than a fault, and a caller that saw it as an error would treat every unfinished stream as broken.

The test enqueues work with `launch_host_function` rather than a kernel, since what is under test is the driver's completion status for a stream and `cuda-core`'s own tests compile no device code.

Asserting only that a busy stream answers `false` would pass against an implementation that blocked first and answered afterwards, so the test also holds `query` to returning inside a fraction of the interval the callback occupies, and checks the callback had not yet run when the answer came. Replacing the body with `self.synchronize()?; Ok(true)` fails it and leaves the idle-stream test passing.

## Validation

```text
cargo test -p cuda-core                     26 unit, 32 integration, 2 doc, 3 ignored, 0 failed
cargo test -p cuda-core --test stream_query 2 passed
cargo clippy -p cuda-core --all-targets     clean
cargo fmt --check -p cuda-core              clean
RUSTDOCFLAGS="-D warnings" cargo doc -p cuda-core --no-deps   clean
```

### DIFF
--- a/crates/cuda-core/src/stream.rs
+++ b/crates/cuda-core/src/stream.rs
@@ -87,6 +87,36 @@ impl CudaStream {
         unsafe { cuda_bindings::cuStreamSynchronize(self.cu_stream) }.result()
     }
 
+    /// Returns `true` when every operation enqueued on this stream has
+    /// completed, `false` while any is still in flight. Never blocks.
+    ///
+    /// Wraps `cuStreamQuery`, mapping `CUDA_SUCCESS` to `Ok(true)` and
+    /// `CUDA_ERROR_NOT_READY` to `Ok(false)`, as
+    /// [`CudaEvent::query`](crate::event::CudaEvent::query) does for
+    /// `cuEventQuery`. Any other code is a real driver error.
+    ///
+    /// This is the completion half of the async bridge that
+    /// [`launch_host_function`](Self::launch_host_function) opens: a
+    /// `Future::poll` implementation needs to ask whether work has finished
+    /// without parking the executor's thread, which
+    /// [`synchronize`](Self::synchronize) cannot answer.
+    ///
+    /// A sticky error from earlier work on the stream surfaces here rather
+    /// than as `Ok(false)`, the same way it would from
+    /// [`synchronize`](Self::synchronize).
+    ///
+    /// On the default stream this reports the completion of every blocking
+    /// stream in the context, since the default stream orders against them
+    /// all.
+    pub fn query(&self) -> Result<bool, DriverError> {
+        self.ctx.bind_to_thread()?;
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Validated on 2x RTX 5090:** 
- both `stream_query` tests pass and stayed green across five reruns of the timing-sensitive case; 
- full `cuda-core` suite green. 
 
Merging.

nihalpasham (approved):

Issue #765, opened 2026-08-09
A bit intrinsic writing through a dereferenced pointer is typed from the pointer rather than the pointee


alejandro-soto-franco:
A Rust bit intrinsic whose call destination is a dereferenced raw pointer is refused, though the same intrinsic writing to a plain local compiles and runs.

```text
error: [rustc_codegen_cuda] Device codegen failed: PTX generation failed: Lowering failed:
       ["src/generated_case.rs": line: 21, column: 13] Compilation error: invalid input program.
       expected integer type for Rust bit intrinsic
```

## Reproducer

```rust
#[custom_mir(dialect = "runtime", phase = "initial")]
fn fn1(mut _1: i32) -> *mut i32 {
    mir! {
        let _2: ();
        {
            RET = core::ptr::addr_of_mut!(_1);
            Call((*RET) = core::intrinsics::bswap(451059808_i32), ReturnTo(bb1), UnwindUnreachable())
        }
        bb1 = {
            Return()
        }
    }
}
```

The operand is an `i32` literal, so the type the message complains about is not the one the intrinsic was given.

## What varies

Four runs on an RTX 5060, compute capability 12.0, CUDA 13.3, changing one thing at a time:

| call destination | callee | result |
|---|---|---|
| plain local `_3: i32` | `bswap` | compiles, CPU and GPU traces match |
| `(*_2)` where `_2: *mut i32` | ordinary `fn` | compiles, CPU and GPU traces match |
| `(*RET)` where `RET: *mut i32` | ordinary `fn` | compiles, CPU and GPU traces match |
| `(*RET)` where `RET: *mut i32` | `bswap` | rejected |

So neither the deref destination nor the intrinsic is refused on its own. Only their combination is. `bitreverse` reproduces identically.

## Mechanism

`cargo oxide pipeline` shows the importer giving the intrinsic call a **pointer** result type:

```text
v327 = mir.constant () [] [value: builtin.integer <451059808: si32>]: <() -> (builtin.integer si32)>;
v328 = mir.call (v327) [] [callee: builtin.string "__cuda_oxide_rust_intrinsic_bswap"]:
       <(builtin.integer si32) -> (mir.ptr <builtin.integer si32,mutable:true,addrspace:0>)>;
mir.store (v323, v328) [] []:
       <(mir.ptr <mir.ptr <builtin.integer si32,mutable:true,addrspace:0>,mutable:true,addrspace:0>,
[...]

PR #766, opened 2026-08-09
fix(fuzzer): parse a Rust type past the semicolon inside an array

alejandro-soto-franco:
The adapter captured Rust types with `[^;]+`, which reads a type as everything up to the first semicolon. An array type carries a semicolon inside its brackets, so `[u128; 1]` was captured as `[u128`.

## Summary

Adds `split_type_at_semicolon`, uses it at both sites that captured a type, and raises `composite_count` from 0 to 3 now that the adapter survives what that produces.

## Changes

- `crates/fuzzer/tools/mir_generator.py`: `split_type_at_semicolon`, tracking bracket depth over `(` and `[` in the idiom `split_args` already uses. The `type RET = ..;` insertion anchor and the local-declaration scan both use it. `composite_count = 3` in `TINY_CONFIG`, and the module docstring no longer says the generated program is scalar-only.
- `crates/fuzzer/README.md`: refreshes the two quoted `--start 0 --count 2` transcripts, which the config change makes stale.

## Rationale

The insertion anchor matched `type RET = [^;]+;`. On an array return type it stopped mid type and left the remainder stranded after the inserted declarations:

```text
type RET = [u128;
        let __rl_dump0: (u64,); 1];
```

rustc reads that as a `let` statement in expression position and refuses the case. The local-declaration scan truncated the same way, so an array-typed local was recorded as `[i32` and reported under that truncated name when the trace API declined to hash it. Since an `UNSUPPORTED` reason is how the next widening of the trace API gets chosen, a truncated type name misdirects that choice.

Neither fault was reachable as checked in. No `fn0` signature in seeds 300-499 contains a semicolon at `composite_count = 0`, because that setting keeps tuples and arrays out of the generated program entirely. The knobs above it in `TINY_CONFIG` (`tuple_max_len`, `array_max_len`) have been inert.

That is the reason to raise it. The fuzzer's most valuable find so far, the silent miscompile behind #393, lived in aggregate constants with padding, and the configuration excludes that whole space. `adt_count` stays at 0: the adapter extracts one function and carries no user-defined type definitions with it, so a generated `Adt20` fails to resolve. Widening to ADTs is a separate change.

## Validation

Seeds 300-359, changing only the adapter, both runs at the `composite_count = 3` and `adt_count = 0` this PR ships:

```text
before:  PASS=40  UNSUPPORTED=18  COMPILE_FAIL=2
after:   PASS=42  UNSUPPORTED=18  COMPILE_FAIL=0
```

Both failures in the before column were the stranded-type syntax error rather than a backend refusal, and both become passing device runs.

Seeds 1000-1999 on the fixed adapter at the same configuration:

```text
PASS=673  UNSUPPORTED=320  COMPILE_FAIL=7  MISMATCH=0
```

Six of those seven are #765. The seventh is a separate backend fault that llc rejects, reported separately.

Aggregate reach over the same 60 seeds, counting programs whose locals or return type hold an array or a multi-field tuple:

```text
composite_count = 0:   0 of 60 seeds
composite_count = 3:  25 of 60 seeds, of which 7 compile and run on the device
```

The remaining 18 are `UNSUPPORTED`, refused at the trace boundary because a dump site or return position holds the aggregate itself. Hashing an aggregate by its scalar leaves would convert most of them and is the obvious next step.

The truncation fix is visible in the reason text, which now names the whole type:

```text
before:  UNSUPPORTED [adapter] unsupported dumped type for Stage 2 adapter: [i8
after:   UNSUPPORTED [adapter] unsupported dumped type for Stage 2 adapter: [i8; 1]
```

#765 was surfaced by that sweep and the checked-in configuration cannot reach it.


nihalpasham:
**Verified:** 
- the depth-0 semicolon fix and reproduced your 300-359 sweep on sm_120: PASS=42, COMPILE_FAIL=0, no mismatches. 
 
**Pushed three follow-ups on the branch:** 
- the adapter now fails loudly when the type `RET` anchor is missing, `run_seed.py` prints repo-relative paths so the README transcripts are literal output, and 
- the seed-specific prose is re-anchored on the shipped config. 
 
One note: the two-seed transcript needs `--keep-going`, since seed 0's `UNSUPPORTED` otherwise stops the run; the README now quotes that command.

nihalpasham (approved):

Issue #767, opened 2026-08-09
An 8-bit bswap emits an invalid bitcast when its result type is an aggregate


alejandro-soto-franco:
Device codegen emits an invalid `bitcast` for an 8-bit `bswap` when the call's result type is not an 8-bit integer. LLVM rejects the module when it is read back, so the failure arrives as an `llc` diagnostic against a generated `.ll` rather than as an error against the source.

```text
error: rustlantis_smoke.ll:342:17: error: invalid cast opcode for cast from 'i8' to '{ double, i8, [7 x i8] }'
  %v4 = bitcast i8 7 to { double, i8, [7 x i8] }
                ^
```

## Reproducer

```rust
#[custom_mir(dialect = "runtime", phase = "initial")]
fn fn1(mut _1: isize) -> (f64, u8) {
    mir! {
        type RET = (f64, u8);
        let __rl_dump0: (isize, isize);
        let _7: ();
        {
            Call(RET.1 = core::intrinsics::bswap(7_u8), ReturnTo(bb1), UnwindUnreachable())
        }
        bb1 = {
            Call(_7 = dump_var(Move(__rl_dump0)), ReturnTo(bb2), UnwindUnreachable())
        }
        bb2 = {
            Return()
        }
    }
}
```

`{ double, i8, [7 x i8] }` is the layout of the `(f64, u8)` return, and `i8 7` is the intrinsic's operand. The second call is load bearing: with nothing keeping the function alive the whole body is eliminated and the case compiles.

## Why this arm and not the others

`bswap` on a single byte is identity, and LLVM has no useful intrinsic for it, so `convert_rust_bit_intrinsic` handles that width with an early bitcast and returns:

```rust
if matches!(intrinsic, RustBitIntrinsic::Bswap) && value_width == 8 {
    let bitcast = llvm::BitcastOp::new(ctx, value, result_type);
    ...
    return Ok(());
}
```

Every other arm falls through to `cast_integer_value_to_type`, which asks `integer_bit_width` for the target width and refuses a type that has none. The early return skips that, and `bitcast` requires both types to be non-aggregate and of one size, so an aggregate result produces IR that cannot be parsed.

## Where the wrong type comes from

`crates/mir-importer/src/translator/terminator/intrinsics/bitops.rs:107` types the call from the destination's **local**:

```rust
let return_type = types::translate_type(ctx, &body.locals()[destination.local].ty)?;
```

[...]

PR #768, opened 2026-08-09
fix(mir-lower): refuse an 8-bit bswap whose result is not an 8-bit integer

alejandro-soto-franco:
Mitigates #767 by failing closed in mir-lower; the importer root-cause fix in #769 carries the close.

An 8-bit `bswap` took an early bitcast path that checked nothing about the type it was casting to, so an aggregate result produced `bitcast i8 7 to { double, i8, [7 x i8] }` and the failure surfaced as an LLVM parse error against a generated `.ll` file rather than against the source.

## Summary

Checks the result width before the bitcast, so the case reports the same `invalid input program` error every other arm of the intrinsic already reports.

## Changes

- `crates/mir-lower/src/convert/ops/call.rs`: `convert_rust_bit_intrinsic` reads the result width through `integer_bit_width` before constructing the `BitcastOp`, and refuses a width that does not match the operand.
- `crates/mir-lower/tests/lowering_test.rs`: a lowering test building a `mir.call` to the `bswap` placeholder with a `ui8` operand and a `mir.tuple` result.

## Rationale

Rust's `bswap` on a single byte is identity and LLVM has no useful intrinsic for it, so that width returns early:

```rust
if matches!(intrinsic, RustBitIntrinsic::Bswap) && value_width == 8 {
    let bitcast = llvm::BitcastOp::new(ctx, value, result_type);
    ...
    return Ok(());
}
```

Every other arm falls through to `cast_integer_value_to_type`, which asks `integer_bit_width` for the target width and refuses a type that has none. The early return skips that check, and `bitcast` requires both types to be non-aggregate and of one size, so the arm could emit IR that cannot be parsed. Reading the width first restores the property that the other arms already hold: a bad result type is reported against the source rather than handed to LLVM.

This does not address where the aggregate type comes from. The importer types an intrinsic call from its destination local rather than the projected place it writes, so `RET.1 = bswap(x)` on a `(f64, u8)` return arrives here carrying the tuple, as #767 sets out. That fix belongs in `mir-importer` and touches the projected-assignment path `emit_function_call` does not share, so it is left separate.

The test asserts the lowering returns an error. Removing the four added lines makes it fail, and the other 100 tests in the file pass either way.

## Validation

The reduced case from #767, before and after:

```text
before:  llc: rustlantis_smoke.ll:342:17: error: invalid cast opcode for cast from 'i8'
              to '{ double, i8, [7 x i8] }'
after:   error: [rustc_codegen_cuda] Device codegen failed: ... ["src/generated_case.rs": line: 27,
              column: 13] Compilation error: invalid input program.
              expected integer type for Rust bit intrinsic
```

An 8-bit `bswap` writing to a plain `u8` local still compiles and runs, CPU and GPU traces matching, so the valid case that reaches this path is unaffected.

```text
cargo test -p mir-lower                    278 passed, 7 ignored, 0 failed
cargo clippy -p mir-lower --all-targets    clean
cargo fmt --check -p mir-lower             clean
```

### DIFF
--- a/crates/mir-lower/src/convert/ops/call.rs
+++ b/crates/mir-lower/src/convert/ops/call.rs
@@ -811,6 +811,20 @@ fn convert_rust_bit_intrinsic(
 
     if matches!(intrinsic, RustBitIntrinsic::Bswap) && value_width == 8 {
         // LLVM has no useful byte swap for a single byte; Rust's semantics are identity.
+        //
+        // The result type is checked first. `bitcast` requires both types to be
+        // non-aggregate and of one size, so a result that is not an 8-bit
+        // integer would leave this arm emitting IR that LLVM rejects. Every
+        // other arm reaches `cast_integer_value_to_type` below, which performs
+        // the same check; this one returns early and would otherwise skip it.
+        let result_width = integer_bit_width(ctx, result_type, loc.clone())?;
+        if result_width != value_width {
+            return pliron::input_err!(
+                loc,
+                "bswap on an {value_width}-bit integer cannot produce a \
+                 {result_width}-bit result"
+            );
+        }
         let bitcast = llvm::BitcastOp::new(ctx, value, result_type);
         rewriter.insert_operation(ctx, bitcast.get_operation());
         rewriter.replace_operation(ctx, op, bitcast.get_operation());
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Validated on 2x RTX 5090:** 
- workspace suite green (3411 passed), and 
- a `u8::swap_bytes` kernel lowers to the identity `bitcast` and matches the CPU reference on device. 

**Pushed two test fixups:** 
- the refusal test now asserts the guard's actual diagnostic, 
- a positive test covers the valid `u8` path, and 
- both were mutation-checked (removing the guard fails one, inverting it fails the other). 

Landing together with #769, which carries the #767 close.

nihalpasham (approved):

PR #769, opened 2026-08-09
fix(mir-importer): write an intrinsic result through its projected destination

### LINKED ISSUE #765: A bit intrinsic writing through a dereferenced pointer is typed from the pointer rather than the pointee

### LINKED ISSUE #767: An 8-bit bswap emits an invalid bitcast when its result type is an aggregate

alejandro-soto-franco:
Closes #765. Closes #767.

An intrinsic call took its result type from the destination's **local** and stored the result into that local, discarding any projection on the place.

## Summary

`destination_type` asks `Place::ty` for the projected type, and `store_result_to_place` writes through the projection instead of over the local.

## Changes

- `crates/mir-importer/src/translator/terminator/helpers.rs`: `destination_type` and `store_result_to_place`; `emit_function_call` uses the latter; `emit_store_result_and_goto` refuses a projected destination.
- `crates/mir-importer/src/translator/terminator/intrinsics/{bitops,saturating,bigint,exact_div}.rs`: take the result type from `destination_type`.
- `crates/mir-importer/src/translator/statement.rs`: `pointer_is_mutable`, `pointer_address_space`, `reject_raw_field_index_on_enum_pointee`, `slot_array_element_ty` and `emit_array_element_store` become `pub(crate)` so the store path can reuse them rather than restate them.
- `crates/rustc-codegen-cuda/examples/mir_projected_call_destination`: new example covering the shapes on the device against a host oracle.

## Rationale

`RET.1 = bswap(x)` on a `(f64, u8)` return carried the whole tuple as the call's result type and aimed its store at the tuple rather than at the field. `(*RET) = bswap(x)` carried the pointer and stored over the pointer rather than through it.

Ordinary calls never showed this, because rustc lowers a projected call destination into a call to a temporary followed by a store:

```text
_9 = compute_stage1_trace(const 10_i32) -> [return: bb3, unwind continue]
(*_8) = move _9
```

An intrinsic keeps its projected destination through to codegen, so `custom_mir` is the only way to reach it.

Both symptoms were compile failures rather than wrong answers, each caught by a different check. A pointer result has no width, so it was refused where the lowering asks for one (#765). A tuple result reached the 8-bit `bswap` arm, which bitcasts without checking, and produced IR that LLVM will not parse (#767).

`Place::ty` is the existing idiom for this question: `intrinsics/atomic.rs:1348` already uses it to type a cmpxchg destination. The store mirrors what `translator/statement.rs` does for an ordinary assignment to a projected place, taking the field's address through the local's slot with `mir.field_addr`, the pointer through the base place for a deref, and the element through `emit_array_element_store` for an index.

`emit_store_result_and_goto` is the other epilogue with the same fault. It has no `Body`, so it cannot compute the address of a place, and 600 of its 634 callers sit in the generated `intrinsics/generated.rs`, so widening its signature would mean changing the factory template. It now refuses a projected destination instead, which is the alternative to aiming a store at the wrong address.

Four single-element projections are modelled: a deref, a field, a runtime index, and a constant index. Each takes the address the ordinary assignment path takes for the same shape, and a constant index from the end is refused there and here alike. Anything longer than one element is refused rather than guessed at.

Two of those four came from running the work rather than from reading the code. Re-running the seeds that found these bugs left seed 1987 refused with `projection [Index(2)]`, and the example below then hit `ConstantIndex` on its first compile, rustc having folded a literal index into one.

## Validation

The new example fails to compile on `main` with `expected integer type for Rust bit intrinsic`, and here reports:

```text
host  deref case: 0x0000000060a0e21a
host  field case: 0x3ff8000000000007
host  index case: 0x00000060a0c31a0b

PASS: device and host agree on all three projections
```

Each value is checkable by hand. `bswap(0x1AE2A060)` is `0x60A0E21A`; the field case is the bit pattern of `1.5f64` against the byte `7`, which a byte swap leaves alone; and the index case keeps `11` and `33` in the elements the call did not name, which a store over the whole array would not.

The two reduced cases from the issues behave the same way, `main` against this branch:

```text
(*RET) = bswap(x), RET: *mut i32     main: refused          branch: traces match
RET.1 = bswap(7_u8), RET: (f64, u8)  main: invalid bitcast  branch: traces match
```

The seven fuzzer seeds that surfaced both issues, re-run with `composite_count` raised so they generate the same programs:

```text
seed 1170  PASS      seed 1792  PASS      seed 1967  PASS
seed 1202  PASS      seed 1825  PASS      seed 1987  PASS
seed 1420  PASS
```

All seven failed to compile on `main`: six with #765 and one with #767.

```text
cargo test --workspace                              exit 0, no failing test
scripts/smoketest.sh --compile-only                 192 / 192
cargo clippy -p mir-importer --all-targets          clean
cargo clippy -- -D warnings (the new example)       clean
cargo fmt --check                                   clean
```


nihalpasham:
**Validated on 2x RTX 5090:** 
- `mir_projected_call_destination` prints PASS on device, 
- `smoketest --compile-only` is 192/192, workspace suite clean. 

**Pushed one fixup:** 
- the new `helpers::destination_type` duplicated `types::translate_destination_type` while dropping the source location from its error, so the four intrinsic sites now use the existing helper, and `store_result_to_place` documents the known eval-order gap. 

Filed a follow-up for the remaining bare-local typing sites.

nihalpasham (approved):

PR #772, opened 2026-08-09
fix(llvm-export): refuse a kernel parameter pointing into shared or local memory

### LINKED ISSUE #771: A kernel parameter pointing into shared memory emits an entry the driver refuses to load
A kernel that takes a pointer to a shared-memory type emits an entry parameter in the `.shared` state space. ptxas assembles the module, and the driver then refuses to load it, so every kernel in that module becomes unreachable.

```text
DriverError(200, "device kernel image is invalid")
```

`generated_intrinsics_blackwell` contains such a kernel, so the module that example builds cannot be loaded on the architecture it pins.

## Reproducer

```rust
#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    pub fn plain(mut out: DisjointSlice<u32>) {
        if let Some(s) = out.get_mut(thread::index_1d()) { *s = 7; }
    }

    /// Empty body. The only unusual thing is the parameter type.
    #[kernel]
    pub unsafe fn takes_barrier(_barrier: *mut Barrier) {}
}
```

With `takes_barrier` present, `kernels::load` fails with error 200. Remove it and the same module loads and `plain` writes `7`. The body is empty, so nothing the kernel does is involved.

## The emitted entry

```text
.visible .entry takes_barrier(
	.param .u64 .ptr .shared .align 1 takes_barrier_param_0
)
```

`.ptr .shared` on an entry parameter is the problem. A kernel receives its parameters in `.param` space and cannot be handed a shared-memory address by the host, since shared memory is allocated per block at launch. The attribute is meaningful on a `.func` parameter, not on a `.entry` one.

`Barrier` lives in shared memory, so `*mut Barrier` carries `addrspace(3)`, and that address space reaches the entry signature rather than being rejected or widened to generic.

ptxas does not object. Assembling the example's own PTX by hand succeeds:

```text
$ ptxas -arch=sm_120a generated_intrinsics_blackwell.ptx -o /tmp/bw.cubin
ptxas ... warning : Advisory: '.multicast::cluster' modifier ... on .target 'sm_120a'
$ echo $?
0
```

So the fault surfaces only when a module is loaded, which is why nothing catches it today.

## Why the example hides it

`generated_intrinsics_blackwell` sits in `BLACKWELL_COMPILE_EXAMPLES`, documented in `scripts/smoketest.sh` as "compile-only coverage pinned to exact sm_120a. These kernels are never launched." Its `main` is:

```rust
fn main() {
    println!("PASS: generated Blackwell sparse MMA compile coverage");
}
```

The smoketest records a pass, and no GPU in CI would change that, since the example never asks the driver for the module.

Isolating it inside that example gives the same answer as the reproducer above. Commenting out `compile_tma_compatibility` and nothing else makes the module load, after which the conversion kernels run and produce correct values on hardware:

```text
e4m3x2(1.0, 2.0) = 0x4038
e5m2x2(1.0, 2.0) = 0x403c
```

Those match the formats by hand: `1.0` is `0x38` in e4m3 and `0x3C` in e5m2, `2.0` is `0x40` in both, and the first argument occupies the low byte as `cvt_rn_satfinite_e4m3x2_f32` documents. Removing the TMA calls one family at a time does not help, and emptying the body does not help either. The sparse MMA kernels load and run on their own.

## Expected

A kernel parameter that would carry a non-generic address space is refused at compile time, in the manner of the other invalid-input diagnostics, rather than assembled into a module the driver cannot load. Widening the pointer to generic would also work if passing such an address is meant to be allowed, though for shared memory there is nothing sensible the host could pass.

## Environment

RTX 5060 Laptop GPU, compute capability 12.0, CUDA 13.3, driver 610.43.03, target `sm_120a`.

alejandro-soto-franco:
Closes #771.

A kernel parameter pointing into shared memory reached the entry signature as `.ptr .shared`. ptxas assembles that, and the driver then refuses the module, so every kernel in it becomes unreachable. `generated_intrinsics_blackwell` contains such a kernel, which is why the module that example builds cannot be loaded on the architecture it pins.

## Summary

Refuses a kernel parameter in a state space the host cannot supply, repairs the kernel that had one, and checks the conversions on the device rather than only compiling them.

## Changes

- `crates/llvm-export/src/export/function.rs`: a kernel parameter that is a pointer into shared or local memory is refused, naming the kernel, the parameter index and the space.
- `crates/rustc-codegen-cuda/examples/generated_intrinsics_blackwell/src/main.rs`: `compile_tma_compatibility` declares its barrier as a shared `static mut` rather than taking one as a parameter, and the example launches the packed-FP8 and TF32 kernels and checks all 42 results against the two formats.

## Rationale

Shared memory is allocated per block and local memory per thread, both by the device at launch, so the host holds no address in either to put in a parameter. Global and constant are left alone: the host allocates there, a pointer parameter in those spaces is something it can pass, and the libdevice linking tests in `cuda-oxide-codegen` rely on exactly that. A device function may carry any state space on its parameters, so the check is for kernels alone.

The example's kernel could never have been launched. Removing the parameter costs nothing, since a barrier has to live in shared memory anyway, and `examples/barrier` already declares one as a `static mut` inside the kernel.

Removing that kernel is what makes the module load. With it repaired, the conversion kernels run, which is what the rest of this change uses. Their results are bit-exact, so there is no tolerance to pick, and every expected value derives from the format:

| value | e4m3 | e5m2 | why |
|---|---|---|---|
| 1.0 | 0x38 | 0x3C | exponent 0, zero mantissa |
| 2.0 | 0x40 | 0x40 | exponent 1 |
| -1.0 | 0xB8 | 0xBC | sign bit over 1.0 |
| 448.0 | 0x7E | 0x5F | 1.75 x 2^8, the largest finite e4m3 |
| 1e30 | 0x7E | 0x7B | saturates to the largest finite value |

TF32 keeps the f32 exponent and truncates the mantissa to 10 bits, so each input chosen is already representable and all ten rounding forms agree, except the four relu forms on a negative input. The relu cases pin the argument order too: `cvt_rn_satfinite_relu_e4m3x2_f32(-1.0, 448.0)` gives `0x7e00`, which is the low lane zeroed and the high lane kept, matching the documented "first argument in the low byte".

The example stays compile-only in CI, since it pins `sm_120a` and no runner has that device. On one that does, the conversions are now verified rather than assumed.

## Validation

RTX 5060 Laptop GPU, compute capability 12.0, CUDA 13.3, driver 610.43.03, target `sm_120a`.

Before this change, on that device:

```text
DriverError(200, "device kernel image is invalid")
```

After it, the example reports what the device computed:

```text
  e4m3x2       (         1,          2) = 0x4038  want 0x4038  ok
  relu_e4m3x2  (        -1,        448) = 0x7e00  want 0x7e00  ok
  e5m2x2       (      1e30,          0) = 0x007b  want 0x007b  ok
  tf32[3]      (      -3.5)             = 0x00000000  want 0x00000000  ok
  ...
PASS: every conversion matched, and the sparse MMA and TMA kernels compiled
```

42 checks, all passing. Changing one expected constant reports `MISMATCH` and exits 1, so a wrong result is caught.

The refusal was narrowed by the test suite rather than by reading: refusing every non-generic space broke three `libdevice_linking` tests whose kernels take a global pointer, which is legitimate.

```text
cargo test --workspace                              exit 0, no failing test
scripts/smoketest.sh --compile-only                 191 / 191
cargo clippy -- -D warnings (the example)           clean
cargo fmt --check                                   clean
```

The reproducer from #771 now fails to compile with:

```text
kernel `@compile_tma_compatibility` parameter 2 is a pointer into shared memory (address
space 3); shared memory is allocated by the device at launch, so the host has no address
to pass, and the driver refuses a module whose entry declares one
```


nihalpasham:
**Verified end-to-end on sm_120:** 
- the repaired `generated_intrinsics_blackwell` module now loads and all 42 FP8/TF32 conversion checks pass. 

**Pushed one fixup:** 
- adding the missing regression coverage: an `error_kernel_shared_param` error example with the diagnostic checked in smoketest, 
- plus three `export_test.rs` unit tests covering the AS3/AS5 refusal and its scope. 

Full workspace tests and the 192-example compile-only smoketest pass.

nihalpasham (approved):

PR #776, opened 2026-08-10
fix(cuda-device): add ptx_asm output markers through arity 32

### LINKED ISSUE #775: `ptx_asm!` can generate output marker arities that `cuda-device` does not define
## Description

`ptx_asm!` can currently generate output marker calls with arities greater than 24, while `cuda-device` only defines `__ptx_asm_out_0` through `__ptx_asm_out_24`.

The macro allows:

* up to 16 explicit input operands;
* up to 16 output operands;
* `inout` operands to contribute hidden tied-input arguments.

When all 16 outputs are `inout` operands and 16 explicit inputs are present, the generated marker receives 32 arguments:

```text
16 explicit inputs + 16 hidden tied inputs = 32 marker arguments
```

However, `cuda-device::ptx` currently stops at:

```text
__ptx_asm_out_24
```

This leaves valid `ptx_asm!` operand combinations unable to compile because the generated marker function does not exist.

## Expected behavior

The `cuda-device` marker surface should cover every arity that `ptx_asm!` can legally generate.

Output markers should therefore be available through:

```text
__ptx_asm_out_32
```

Void markers can remain capped at 16 because `inout` operands necessarily produce outputs.

## Suggested fix

* Add `__ptx_asm_out_25` through `__ptx_asm_out_32` to `cuda-device`.
* Update the marker-arity documentation accordingly.
* Add compile-time coverage against the real `cuda-device::ptx` marker surface.
* Update the MIR importer boundary test to exercise arity 32.

uurl:
Closes #775 

## Summary

Extend the hidden `ptx_asm!` output marker surface from arity 24 through arity 32.

`ptx_asm!` supports up to 16 explicit inputs and up to 16 `inout` operands. Since each `inout` operand generates a hidden tied-input argument, a valid invocation can require:

```text
16 explicit inputs + 16 hidden tied inputs = 32 marker arguments
```

`cuda-device` previously exposed output markers only through `__ptx_asm_out_24`, leaving valid macro expansions for arities 25–32 unresolved.

## Changes

* Add `__ptx_asm_out_25` through `__ptx_asm_out_32`.
* Update the `cuda-device` marker-arity documentation from 24 to 32 arguments.
* Add a `trybuild` regression test against the real `cuda_device::ptx` marker surface.
* Exercise the maximum supported case with 16 `inout` operands and 16 explicit inputs.
* Update the MIR importer marker parsing boundary test from arity 24 to 32.

Void markers remain capped at 16 because `inout` operands always produce an output.

## Validation

```text
cargo fmt --all -- --check
cargo test -p cuda-macros --test ptx_asm
cargo test -p cuda-device --lib --tests
cargo test -p mir-importer parses_output_marker_with_32_inputs
cargo clippy -p cuda-device -p cuda-macros -p mir-importer --all-targets -- -D warnings
```

All checks pass.

### DIFF
--- a/crates/cuda-device/src/ptx.rs
+++ b/crates/cuda-device/src/ptx.rs
@@ -74,8 +74,8 @@ pub const fn __ptx_asm_c<const N: usize>(value: &'static [u8; N]) -> &'static [u
 }
 
 // Rust has no variadic generics, so expose marker stubs for fixed arities.
-// Output markers support up to 24 arguments: 16 explicit inputs plus up to
-// 8 hidden tied inputs generated for `inout` operands. Void markers require
+// Output markers support up to 32 arguments: 16 explicit inputs plus up to
+// 16 hidden tied inputs generated for `inout` operands. Void markers require
 // only 16 arguments because `inout` always produces an output.
 define_ptx_asm_out!(__ptx_asm_out_0;);
 define_ptx_asm_out!(__ptx_asm_out_1; a0: A0);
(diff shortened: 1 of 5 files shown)


nihalpasham:
The arity math checks out (16 inout tied inputs + 16 in = 32) and the test against the crate is the correct guard against the mock drift that let this slip.

Validated
- end-to-end with a 16-inout + 16-in kernel executing on sm_120; one
- test rename layered on top. 

Merging.

nihalpasham:
Thanks @uurl! The arity math checks out (16 inout tied inputs + 16 in = 32 marker arguments), and the contract test against the real crate is the right guard against the mock drift that let this slip. Validated end-to-end with a 16-inout + 16-in kernel executing on sm_120; one test rename layered on top. Merging.

nihalpasham (approved):

PR #777, opened 2026-08-10
perf(mir-lower): carry proved alignment through an array-element address

### LINKED ISSUE #745: Reading a lane of an over-aligned aggregate drops its abi_align, costing 2x on cache-resident lookups

vyncint:
Refs #745 — the array-element half you scoped in the #746 merge review.

#746 taught a *field* address to record what it proves, so `c.x + c.y` on an
`#[repr(C, align(8))]` element fuses into one `ld.global.v2.b32`. An **element**
address recorded nothing, so the same two f32 reached through an array index
still exported two `align 4` loads.

## One correction to the scope note

The note said `c.0[0] + c.0[1]` still emits two narrow loads. On current `main`
that exact spelling **already fuses** — I checked before building anything. The
local copy is what saves it: `let c = pts[i]` lets SROA load the whole 8-byte
aggregate, and the alignment comes from its type.

The shape that genuinely still failed is reading **through the place**:

```rust
let c = pts[i]; c.0[0] + c.0[1]      // already one wide load
pts[i].0[0] + pts[i].0[1]            // two narrow loads
let l = &pts[i].0; l[0] + l[1]       // two narrow loads   <- the table idiom
```

So the cost depended on whether the source copied the element or read through
it, which is not a distinction a programmer should have to know. The example
covers the reading-through form, since the copying form was never broken.

## The arithmetic

Two facts combine. Whatever the base pointer **already proved** beats the array
type alone when the array is a field of an over-aligned aggregate —
`&table[i].lanes` carries the outer struct's alignment, which `[f32; 2]` does
not know. Element `i` then sits at byte `i * stride`:

| index | provable |
|---|---|
| constant 0 | `base` |
| constant `i` | `gcd(base, i * stride)` |
| runtime | `gcd(base, stride)` — what *every* element satisfies |

All from rustc's layout. Uncertain cases decline, and the same power-of-two
guard you added to the field stamp applies here.

## PTX

| kernel | before | after |
|---|---|---|
| `lanes_through_ref` | 2 `ld.global.b32` | **1 `ld.global.v2.b32`** |
| `hot_lanes` | 4 `ld.global.b32` | **2 `ld.global.v2.b32`** |
| `packed_pair` (align-4 control) | 2 `ld.global.b32` | unchanged |
| `hot_packed` (align-4 control) | 4 `ld.global.b32` | unchanged |

## Measured — A10G (sm_86, 595.71.05)

4194304 threads, 30 timed launches after 5 warmup, all bit-correct. Same example
built by both compilers:

| kernel | stock `main` | this | speedup |
|---|---:|---:|---:|
| `packed_pair` | 105.0 | 104.6 | 1.00x |
| `aligned_pair` (#746's) | 104.9 | 104.8 | 1.00x |
| `lanes_through_ref` | 104.9 | 105.0 | 1.00x |
| `hot_packed` (**control**) | 1253.5 | 1253.5 | **1.00x** |
| `hot_aligned` (#746's) | 628.6 | 633.0 | 1.00x |
| **`hot_lanes`** | **1253.6** | **629.5** | **1.99x** |

Evidence: stock `.gpu-evidence/20260810T050351Z.log`, this branch
`20260810T045728Z.log`.

Streaming does not move and should not — bandwidth-bound at ~480 GB/s, where two
consecutive loads already coalesce into the same transactions. `hot_packed` is
identical to the byte across both builds, so the comparison is controlled *and*
it guards against claiming alignment an align-4 base never had.

## Gates

* Full smoketest on the A10G: **196 / 196, 0 fail**, 1224s
  (`.gpu-evidence/20260810T050600Z.log`).
* Four `mir-lower` unit tests mirroring yours on the field path: element zero
  inherits the base, a nonzero constant index narrows to the gcd, a runtime
  index claims only the stride, and a base with no over-alignment changes
  nothing.
* `verify-code-shape.sh` extended to both new kernels, so the fusion stays
  CI-checked without a GPU.
* Locally: fmt (3 scopes), clippy `-D warnings`, 11 unit-test suites.

## Neighbour

#750 also touches `mir-lower/src/convert/ops/aggregate.rs` (enum payload
storage) — different region, no overlap with the alignment machinery; whichever
lands first, the other rebases trivially.

Closes #745.

### DIFF
--- a/crates/mir-lower/src/convert/ops/aggregate.rs
+++ b/crates/mir-lower/src/convert/ops/aggregate.rs
@@ -39,12 +39,12 @@
 use crate::convert::enum_payload_storage::{coerce_enum_payload_value, enum_payload_storage_type};
 use crate::convert::types::{
     EnumSlotMap, StructLayoutInfo, StructSlotMap, build_enum_slot_map, build_struct_slot_map,
-    build_union_storage_type, convert_type, is_zero_sized_type, llvm_byte_faithful_twin,
-    llvm_type_contains_i1, make_slice_struct, mir_type_abi_align,
+    build_union_storage_type, convert_type, get_type_size, is_zero_sized_type,
+    llvm_byte_faithful_twin, llvm_type_contains_i1, make_slice_struct, mir_type_abi_align,
 };
 use dialect_mir::ops::{
-    MirConstructEnumOp, MirEnumPayloadOp, MirExtractFieldOp, MirFieldAddrOp, MirInsertFieldOp,
-    MirSetDiscriminantOp,
+    MirConstantOp, MirConstructEnumOp, MirEnumPayloadOp, MirExtractFieldOp, MirFieldAddrOp,
+    MirInsertFieldOp, MirSetDiscriminantOp,
 };
 use dialect_mir::types::{
     EnumCarrierKind, EnumLayoutKind, MirArrayType, MirDisjointSliceType, MirEnumType, MirPtrType,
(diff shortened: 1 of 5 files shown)


nihalpasham:
The element-address claim and the gcd reasoning are good and the two kernels vectorize as advertised. 

**I layered a stride-sourcing fix on top:** 
- `get_type_size` guesses 8 for MIR aggregate and f16 element types, which over-claimed alignment on nested arrays; 
- strides now come from rustc layout or the claim declines, with five regression tests. 
- Full sweep 196/196 on sm_120.
  
Merging, 

closes #745.

nihalpasham:
Thanks @vyncint, the element-address claim and the gcd reasoning are solid, and the two kernels vectorize as advertised. We layered a stride-sourcing fix on top: get_type_size guesses 8 for MIR aggregate and f16 element types, which over-claimed alignment on nested arrays; strides now come from rustc layout or the claim declines, with five regression tests. Full smoketest sweep 196/196 on sm_120. Merging, closes #745.

nihalpasham (approved):

PR #779, opened 2026-08-10
test(codegen): cover `read_unaligned` and `write_unaligned`

### LINKED ISSUE #778: Add conformance coverage for `ptr::read_unaligned` and `ptr::write_unaligned`
## Summary

Add end-to-end codegen coverage for:

* `core::ptr::read_unaligned`
* `core::ptr::write_unaligned`

cuda-oxide currently supports the lowering path used by these APIs, but there is no dedicated example that verifies their semantics with genuinely unaligned device pointers.

## Motivation

`read_unaligned` and `write_unaligned` must remain valid when the pointer is not naturally aligned for the pointee type.

A regression test should deliberately construct a misaligned `u32` pointer from a byte buffer rather than relying on an address that may happen to be aligned.

For example:

```rust
let ptr = input.as_ptr().add(1).cast::<u32>();
let value = core::ptr::read_unaligned(ptr);
```

and:

```rust
let ptr = out.as_mut_ptr().add(1).cast::<u32>();
core::ptr::write_unaligned(ptr, value);
```

The `+1` byte offset guarantees that the typed pointer is not naturally aligned for `u32`.

## Expected coverage

Add a standard CUDA example that independently verifies:

1. `read_unaligned` correctly reads four bytes beginning at `base + 1`.
2. `write_unaligned` correctly writes four bytes beginning at `base + 1`.
3. Bytes immediately before and after the write remain unchanged.
4. The example succeeds both with the normal optimization pipeline and with `CUDA_OXIDE_NO_OPT=1`.
5. The example passes the regular and compile-only smoketest paths.

Host-side expected values should use `to_ne_bytes` / `from_ne_bytes` so the test does not introduce an unnecessary endianness assumption.

## Acceptance criteria

```text
cargo oxide run unaligned_memory
CUDA_OXIDE_NO_OPT=1 cargo oxide run unaligned_memory
scripts/smoketest.sh -o '^unaligned_memory$'
scripts/smoketest.sh --compile-only -o '^unaligned_memory$'
cargo fmt --all -- --check
```

Expected runtime result:

```text
PASS: read_unaligned from base + 1
PASS: write_unaligned to base + 1
PASS: guard bytes preserved
PASS: unaligned_memory
```

uurl:
Closes #778 

## Summary

Add the `unaligned_memory` example to provide end-to-end conformance coverage for:

* `core::ptr::read_unaligned`
* `core::ptr::write_unaligned`

The example deliberately forms `u32` pointers from byte addresses at offset `+1`, ensuring that the accesses are genuinely unaligned rather than accidentally satisfying the natural alignment of `u32`.

## Coverage

The example uses separate kernels for the read and write paths so failures can be isolated independently.

### `read_unaligned`

The device constructs an unaligned pointer with:

```rust
let ptr = input.as_ptr().add(1).cast::<u32>();
let value = core::ptr::read_unaligned(ptr);
```

The host reconstructs the expected value with `u32::from_ne_bytes` and compares it against the device result.

### `write_unaligned`

The device writes a `u32` beginning at byte offset 1:

```rust
let ptr = out.as_mut_ptr().add(1).cast::<u32>();
core::ptr::write_unaligned(ptr, value);
```

The host verifies:

* the four payload bytes using `to_ne_bytes`;
* the guard byte before the write;
* the guard byte after the write.

This ensures that the unaligned write neither shifts nor extends beyond the expected byte range.

## Scope

This PR adds conformance coverage only.

No compiler lowering changes are required: both operations already execute correctly through the existing cuda-oxide pipeline.

Added files:

```text
crates/rustc-codegen-cuda/examples/unaligned_memory/
├── Cargo.lock
├── Cargo.toml
├── README.md
└── src/
    └── main.rs
```

## Validation

```text
cargo oxide run unaligned_memory
CUDA_OXIDE_NO_OPT=1 cargo oxide run unaligned_memory
scripts/smoketest.sh -o '^unaligned_memory$'
scripts/smoketest.sh --compile-only -o '^unaligned_memory$'
cargo fmt --all -- --check
```

Runtime result:

```text
PASS: read_unaligned from base + 1
PASS: write_unaligned to base + 1
PASS: guard bytes preserved
PASS: unaligned_memory
```

Smoketest results:

```text
[ 1/ 1] unaligned_memory ... PASS

Pass: 1 / 1
Fail: 0 / 1
```

Compile-only:

```text
[ 1/ 1] unaligned_memory ... PASS (compiled)

Pass: 1 / 1
Fail: 0 / 1
```


nihalpasham:
Thanks for clean conformance example. Verified here on sm_120 in both pipelines.

One layered commit sorts an import the per-example fmt CI step checks (repo-root `cargo fmt --all` cannot reach example crates because of their own `[workspace]`). 

Merging.

nihalpasham:
Thanks @uurl! Clean conformance example, verified here on sm_120 in both the optimized and CUDA_OXIDE_NO_OPT pipelines. One layered commit sorts an import the per-example fmt CI step checks (repo-root cargo fmt --all cannot reach example crates because they declare their own [workspace]). Merging.

nihalpasham (approved):

PR #770, opened 2026-08-09
feat(intrinsics): add TMA tensor reduction intrinsics

uurl:
Depends on #721.
Closes #660.

## Summary

Add ABI-stable support for the 64 `cp.reduce.async.bulk.tensor` reduction variants across tile and im2col addressing modes.

This PR is stacked on #721 (`cf96d6c2ae2e3869b6b95f863049cd6a4658fc88`). Until #721 is merged, the GitHub diff against `main` also includes its changes; the #660-specific delta is the single commit on top of that revision.

## Changes

* Add 64 TMA tensor-reduction intrinsics covering:

  * operations: `add`, `and`, `dec`, `inc`, `max`, `min`, `or`, and `xor`;
  * tile modes for dimensions 1D through 5D;
  * im2col modes for dimensions 3D through 5D.
* Extend compact TMA admission with reduction-specific metadata and fail-closed validation.
* Preserve schema 61 compatibility for existing TMA admissions while requiring schema 62 for TMA reductions.
* Add ABI entries `i0923` through `i0986`.
* Add generated `cuda-device` APIs and dialect operations.
* Extend MIR importer and lowering support for TMA reductions.
* Lower through:

  * typed NVVM intrinsics for the LLVM NVPTX backend;
  * inline PTX for the libNVVM backend.
* Map Rust/LLVM `im2col` forms to the PTX `im2col_no_offs` spelling.
* Add target gating for PTX 8.0 / SM90+.
* Add 64 generated intrinsic probes.

## Evidence

TMA reductions use dedicated evidence profiles rather than modifying the existing #721 TMA evidence:

* `rust-llvm-22.1.2-1cb4e383-tma-reduce`
* `cuda-13.3-libnvvm-13.3.73-tma-reduce`

Both selected backend routes were independently validated across all 64 variants.

LLVM TypedNvvm validation:

* 64/64 variants generated PTX successfully.
* 64/64 variants assembled successfully with `ptxas`.

libNVVM validation:

* 64/64 variants generated PTX successfully.
* 64/64 variants assembled and linked successfully.

Hardware runtime execution was not performed; runtime validation remains `unexecuted`.

## Generated artifacts

This change adds:

* 64 new reduction probe files;
* 2 new evidence profiles;
* 64 ABI entries, increasing the registry from 922 to 986 entries.

The catalog schema changes from 45 to 46. As a result, deterministic regeneration updates the catalog schema/hash header across existing generated files.

Of the 1019 modified tracked files:

* 16 contain semantic changes;
[...]

PR #721, opened 2026-08-08
feat(intrinsics): add Blackwell data-path and execution-control intrinsics

lucifer1004:
## Summary

Extend the reviewed intrinsic catalog and lowering pipeline with 65 ABI-stable entries covering Blackwell mixed-format register MMA, counted CTA barriers, TMA descriptor and prefetch control, programmatic dependent launch, and warpgroup register-allocation control.

## Changes

- Add 25 `mma.m16n8k32.mxf8f6f4.f32` register-MMA variants for the supported SM120/SM121 targets.
- Add counted CTA barrier `sync` and `arrive` variants, including aligned forms.
- Extend TMA support with:
  - tensor-map and tile prefetch;
  - gather4 prefetch;
  - tensor-map descriptor replacement operations;
  - tensor-map proxy acquire/release fences; and
  - L2 cache-hint prefetch variants.
- Add programmatic dependent-launch `launch_dependents` and `wait` operations.
- Add warpgroup `setmaxnreg` increment and decrement operations.
- Extend the generated dialect and MIR lowering for the new execution-control and TMA forms.
- Extend the reviewed Blackwell `ldmatrix` variants and their target gating.
- Record portable evidence tool identifiers and resolve bare `ptxas` names through `PATH` during PTX-assembly validation.
- Regenerate the catalog, ABI registry, reference documentation, dialect bindings, lowering tables, and probe corpus from the reviewed overlays and evidence.

## Target policy

- Treat SM110 according to the SM100/SM103 capability policy where the instruction family permits it.
- Keep Blackwell cluster/CTA-group TMA capabilities disabled for SM120 and SM121.
- Gate SM12x-specific register MMA and other instruction families independently according to their documented target support.

## Generated artifacts

This change adds 65 new probe files, corresponding to the 65 new ABI entries.

The remaining modified probe files are deterministic regeneration caused by portable evidence metadata and tool-name normalization; they do not represent additional public intrinsic entries.

## Testing

- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p cuda-intrinsics-gen --all-targets -- -D warnings`
- [x] `cargo test -p cuda-intrinsics-gen` — 289 passed
- [x] `cargo run -p cuda-intrinsics-gen -- check`
- [x] `cargo run -p cuda-intrinsics-gen -- check-abi-history --base-ref upstream/main`
  - preserves all 857 existing ABI entries
  - contains 922 entries after this change
[...]

PR #780, opened 2026-08-10
fix(mir-importer): type six intrinsic results from the projected destination

### LINKED ISSUE #773: mir-importer: intrinsic sites that type the call result from the bare destination local fail as a verifier type-mismatch on a projected destination
PR #769 taught `store_result_to_place` to write an intrinsic result through a projected call destination (`(*p)`, `x.0`, `x[i]`, `x[0]`) and switched the bitops/saturating/bigint/exact_div sites to type the call result from `Place::ty` via `types::translate_destination_type`. Six sites still type it from `body.locals()[destination.local].ty`, i.e. the bare local:

- `crates/mir-importer/src/translator/terminator/intrinsics/float_math.rs:626` (float-math placeholder calls) and `:497` (tuple-result path, destination tuple type)
- `crates/mir-importer/src/translator/terminator/mod.rs:2637` (`select_unpredictable`)
- `crates/mir-importer/src/translator/terminator/intrinsics/memory.rs:861` and `:954` (`DynamicSharedArray::get`/`offset` pointee extraction from the destination type)
- `crates/mir-importer/src/translator/terminator/intrinsics/generated.rs:22382` (`elect_sync` tuple destination; the template that emits it is `crates/cuda-intrinsics-gen/src/render.rs:10976`)

With a projected destination (custom MIR only, e.g. `RET.1 = powf(...)`), the call op is created with the whole aggregate's type (or the pointer's type for `(*p)`), and `store_result_to_place` then stores that mistyped value through the correctly typed projected address. The pipeline dies as a MirStoreOp verifier type-mismatch ("Compilation error: invalid input program.") instead of either a correct store or the clean "unsupported construct" diagnostic the four fixed sites now guarantee.

Fix: switch each site to `types::translate_destination_type(ctx, body, destination, &loc)`; for generated.rs update the render.rs template and regenerate. The two memory.rs sites match on the destination's `RawPtr` kind to extract a pointee, so they want the projected `Place::ty` before that match rather than the local's type. Add custom-MIR coverage alongside the `mir_projected_call_destination` example or as importer unit tests.

Severity: low (reachable only from `#[custom_mir]`), but the failure mode is a confusing verifier error rather than a diagnostic.

vyncint:
Closes #773.

#769 taught `store_result_to_place` to write an intrinsic result through a
projected call destination and switched four sites to type the result from
`Place::ty`. The six you enumerated still typed it from
`body.locals()[destination.local].ty`, so with a projected destination the call
op got the whole aggregate's type (or the pointer's, for `(*p)`) and the store
then pushed that mistyped value through a correctly typed address.

All six converted, matching the four already fixed:

| site | intrinsic |
|---|---|
| `float_math.rs` | placeholder calls, and the `sincos` tuple-result path |
| `terminator/mod.rs` | `select_unpredictable` |
| `memory.rs` (x2) | `DynamicSharedArray::get` / `::offset` |
| `generated.rs` | `elect_sync`, via the `render.rs` template |

The two `memory.rs` sites match on the destination's `RawPtr` kind to pull out a
pointee, so as you noted they want the projected **rustc** type rather than the
dialect one — they take `destination.ty(body.locals())` before the match, with
the same diagnostic `translate_destination_type` emits when a place cannot be
resolved. The `generated.rs` hunk is regeneration output rather than a hand
edit; `cargo run -p cuda-intrinsics-gen -- check` passes.

## No test, deliberately

I tried to build a `#[custom_mir]` reproducer for the `select_unpredictable`
site — `RET.1 = select_unpredictable(..)` on a `(u64, u32)` — and **could not
make it fail on `main`**:

* with a literal condition, rustc folds the call before the importer sees it;
* with the condition derived from an earlier custom-MIR result, it compiles
  clean both before and after.

A case that passes either way would sit in the suite guarding nothing, so I left
it out rather than claim coverage it does not provide. If you know a shape that
does reach these six with a projection, I am happy to add it — the four
`bitops`/`saturating`/`bigint`/`exact_div` sites presumably had one.

Gates: `cargo fmt --check`, `clippy --all-targets -D warnings`, and
`-p mir-importer -p cuda-intrinsics-gen -p mir-lower` unit tests (930 / 288 /
182 / 102, all green). No GPU needed — this is importer typing only.

## Merge order

#770 and #721 both regenerate `generated.rs` (they add intrinsics). Neither
touches the typing pattern, so there is no semantic conflict, but whichever
lands last will collide textually in that file. If this goes second, the fix is
just `cargo run -p cuda-intrinsics-gen -- generate` on top — happy to rebase.

## Aside, possibly worth its own issue

While probing, `core::hint::black_box` failed device codegen outright:

```
Translation failed: ...core/src/hint.rs: line: 491 Compilation error: invalid input program.
```

That is unrelated to this change and I have not investigated it; flagging it in
case it is news.

### DIFF
--- a/crates/cuda-intrinsics-gen/src/render.rs
+++ b/crates/cuda-intrinsics-gen/src/render.rs
@@ -11591,8 +11591,8 @@ fn render_importer_elect_dispatch(
     output.push_str(
         "            require_arity(name, args.len(), 1, &loc)?;\n\
          \n\
-                     let tuple_ty = crate::translator::types::translate_type(\n\
-                         ctx, &body.locals()[destination.local].ty,\n\
+                     let tuple_ty = crate::translator::types::translate_destination_type(\n\
+                         ctx, body, destination, &loc,\n\
                      )?;\n\
                      let (leader_ty, elected_ty) = {\n\
                          let ty = tuple_ty.deref(ctx);\n\
--- a/crates/mir-importer/src/translator/terminator/intrinsics/float_math.rs
+++ b/crates/mir-importer/src/translator/terminator/intrinsics/float_math.rs
@@ -493,8 +493,10 @@ pub fn emit_sincos(
     use pliron::location::Located;
     use pliron::op::Op;
 
-    // Destination tuple type and its scalar element type.
-    let tuple_ty = types::translate_type(ctx, &body.locals()[destination.local].ty)?;
+    // Destination tuple type and its scalar element type. Typed from the
+    // projected place, so a `RET.1 = sincos(..)`-style destination is the
+    // tuple actually written rather than the whole local.
+    let tuple_ty = types::translate_destination_type(ctx, body, destination, &loc)?;
     let scalar_ty = {
         let r = tuple_ty.deref(ctx);
         match r.downcast_ref::<dialect_mir::types::MirTupleType>() {
@@ -623,7 +625,7 @@ pub fn emit_rust_float_math_intrinsic(
     block_map: &[Ptr<BasicBlock>],
     loc: Location,
 ) -> TranslationResult<Ptr<Operation>> {
-    let return_type = types::translate_type(ctx, &body.locals()[destination.local].ty)?;
+    let return_type = types::translate_destination_type(ctx, body, destination, &loc)?;
     helpers::emit_function_call(
         ctx,
         body,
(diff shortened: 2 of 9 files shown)


nihalpasham:
The typing fix is right; 

**I added the following:**
- completed the other half at 3 sites (`sincos`, `fn-item`, `closure`) where the result was typed from the projection but still stored to the bare local, and 
- added four optimizer-proof regression cases to the example (they hard-fail the build on main's importer and pass here on sm_120). 

Merging, 

closes #773.

nihalpasham:
Thanks @vyncint! The six-site typing fix is right; we completed the other half at three sites (sincos, function-item, closure) where the result was typed from the projection but still stored to the bare local, and added four optimizer-proof regression cases to the example (they hard-fail the device build on main's importer and pass here on sm_120). Merging, closes #773.

nihalpasham (approved):

PR #785, opened 2026-08-10
fix(mir-importer): support compiler_fence in device code

### LINKED ISSUE #781: compiler_fence in device code fails with a misleading element-type error
`core::sync::atomic::fence` works in kernels since #747, but its sibling `compiler_fence` still fails device codegen, and with a misleading message. A compiler fence is the weaker of the two: it only stops the compiler from reordering memory operations across it and emits no hardware instruction at all, so plain valid Rust like this should compile:

```rust
#[kernel]
pub fn publish(data: &mut [u32], flag: &mut [u32]) {
    data[0] = 42;
    core::sync::atomic::compiler_fence(Ordering::Release);
    flag[0] = 1;
}
```

Today it dies with (reproduced on current main, sm_120 host):

```text
  Translation failed: std::sync::atomic::compiler_fence:
  [".../core/src/sync/atomic.rs": line: 4468, column: 23]
       Unsupported construct: could not extract element type from core atomic intrinsic generics
```

The message points at a missing element type, but that is not the real problem. `compiler_fence` lowers to the `atomic_singlethreadfence` intrinsic, which (like `atomic_fence` before #747) carries only an ordering in its generics and no data type. The core-atomic dispatcher in `crates/mir-importer/src/translator/terminator/intrinsics/atomic.rs` strips the `core::intrinsics::atomic_` prefix, fails to match `singlethreadfence` by name, and falls through to the element-type extraction, which bails.

**Expected behavior**, mirroring the #747 fix (which #723 specified):

- Route `singlethreadfence` ahead of the element-type extraction, reusing the ordering-only generics extractor #747 added.
- Lower it as a compiler-only barrier: no PTX fence or membar instruction, just something the optimizer cannot reorder memory operations across. On the libNVVM-safe route (which rejects the LLVM `fence` instruction outright) the natural encoding is an empty side-effecting inline asm with a `~{memory}` clobber, exactly the mechanism the #695 fence routes already use, minus the instruction text.
- Reject `Relaxed` loudly, as the fence path already does (safe Rust cannot produce it; only a direct nightly intrinsic call can).
- Keep the existing behavior otherwise, and give the residual unhandled atomics a message that names the operation rather than complaining about generics.

A lowering test in the shape of `test_first_class_atomic_fence_lowers_at_system_scope` (asserting the empty-asm encoding and the absence of any fence/membar instruction) plus a source-level touch in an example would complete it.

Found while triaging #747; the review flagged it as out of that PR's scope.

vyncint:
Closes #781.

Reproduced your diagnosis exactly: `atomic_singlethreadfence` carries only an
ordering, so it fell through to the load/store/RMW generics extraction and got
reported as a missing element type. It is routed ahead of that now, reusing the
ordering-only extractor #747 added.

## No hardware barrier, and that is checked

A compiler fence constrains only the optimizer, so it must not become a PTX
`fence` or `membar`. The encoding is your suggestion — an **empty**
side-effecting inline PTX block with a `~{memory}` clobber, the #695 mechanism
minus the instruction text.

Rather than assert that indirectly, I built the example both ways on the fixed
compiler and diffed the emitted barriers:

| without `compiler_fence` | with it |
|---|---|
| 1 `fence.acq_rel.sys` | 1 `fence.acq_rel.sys` |
| 2 `fence.sc.gpu` | 2 `fence.sc.gpu` |
| 1 `membar.sys` | 1 `membar.sys` |

Identical. The fence costs no instruction.

## It needed no new dialect op or scope

I first went looking at `AtomicScope`, but adding a thread variant ripples into
`LlvmSyncScope` and `map_scope` across three crates. `InlinePtxOp` already
carries a template and constraints and is already emitted from the importer for
`ptx_asm!`, and its verifier pairs results against `=` output constraints — so
zero results with `~{memory}` is valid, and the whole change lands in one file.

## Relaxed, and the diagnostic

`Relaxed` is refused rather than lowered: safe Rust cannot reach it
(`compiler_fence(Relaxed)` panics in core), only a direct nightly intrinsic call
can, and a relaxed compiler fence orders nothing.

The residual diagnostic now names the operation instead of blaming generics,
which is what made this failure mode confusing in the first place.

## Verified

* **Before**, stock `main`: `Unsupported construct: could not extract element
  type from core atomic intrinsic generics`
* **After**: compiles; the barrier diff above; **A10G (sm_86) executes it and
  reports SUCCESS** (`.gpu-evidence/20260810T084433Z.log`)
* Scoped smoketest over the atomics examples: **8/8**
  (`.gpu-evidence/20260810T085042Z.log`)
* Two importer tests pinning that both ordering-only fences are recognised by
  name and that `singlethreadfence` is not conflated with `fence` — the exact
  confusion that caused this
* `cargo fmt --check` (root + example), `clippy --all-targets -D warnings`,
  and `-p mir-importer -p mir-lower -p dialect-nvvm` unit tests (5 suites)

The source-level touch went into `scoped_atomic_load_store`, the example #747
already uses for the fence routes, so both siblings are now covered in one place.

One judgement call worth flagging: you suggested a lowering test shaped like
`test_first_class_atomic_fence_lowers_at_system_scope`. Because this encodes at
import time rather than through a dialect fence op, the equivalent assertion is
the PTX barrier diff above, which I think is the stronger check — it asserts on
real emitted PTX rather than on the op tree. Happy to add a mir-lower test too if
you would rather have both.

### DIFF
--- a/crates/mir-importer/src/translator/terminator/intrinsics/atomic.rs
+++ b/crates/mir-importer/src/translator/terminator/intrinsics/atomic.rs
@@ -883,6 +883,34 @@ fn extract_type_info_from_generics(
     })
 }
 
+/// The route a core atomic intrinsic takes through [`dispatch_core_intrinsic`].
+///
+/// The two fences carry no element-type generic, only an ordering const, so
+/// they must be picked out by name before the type extraction shared by the
+/// load/store/RMW/CAS paths. A fence that misses this routing is reported as
+/// a missing element type, which is not the problem (issue #781 was exactly
+/// that for `compiler_fence`).
+#[derive(Debug, PartialEq, Eq)]
+enum CoreIntrinsicRoute {
+    /// `atomic_singlethreadfence` (`core::sync::atomic::compiler_fence`):
+    /// constrains only the optimizer and must emit no hardware barrier.
+    CompilerFence,
+    /// `atomic_fence` (`core::sync::atomic::fence`): a hardware barrier at
+    /// system scope.
+    HardwareFence,
+    /// Everything else: element-typed load/store/RMW/CAS dispatch.
+    Typed,
+}
+
+/// Classify a core atomic intrinsic op name (see [`parse_core_intrinsic_op`]).
+fn route_core_intrinsic(op_name: &str) -> CoreIntrinsicRoute {
+    match op_name {
+        "singlethreadfence" => CoreIntrinsicRoute::CompilerFence,
+        "fence" => CoreIntrinsicRoute::HardwareFence,
+        _ => CoreIntrinsicRoute::Typed,
+    }
+}
+
 /// Dispatch a `std::intrinsics::atomic_*` / `core::intrinsics::atomic_*` call.
 ///
 /// Extracts the generic args (type, ordering) from the `func` operand and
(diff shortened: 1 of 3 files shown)


nihalpasham:
**Verified on sm_120:**
- atomics smoketest 5/5 including the libNVVM verify route (the empty `call void asm sideeffect "", "~{memory}"()` passes verification), and
- the PTX fence counts match your table exactly: 1 `fence.acq_rel.sys`, 2 `fence.sc.gpu`, 1 `membar.sys`, with nothing emitted for the compiler fence.

**I layered one commit on top:** your two unit tests pass on an unpatched importer, so the routing decision is now a unit-testable `CoreIntrinsicRoute` classifier the dispatcher itself consumes, with tests that fail without the fix (plus a mir-lower encoding test).

Merging,

closes #781.

nihalpasham (approved):

Issue #226, opened 2026-06-15
Tracking: full device debug metadata for variables and types


nihalpasham:
## Summary

PR #225 adds the first useful device-debug tier: line tables. That gives cuda-gdb source breakpoints, stepping, and basic backtraces.

This issue tracks the larger follow-up: full debug metadata for variables, arguments, and Rust types.

## Goal

Make debugger workflows like these possible for device code:

```gdb
(cuda-gdb) info locals
(cuda-gdb) print idx
(cuda-gdb) print some_struct.field
```

## Likely work

- carry rustc `Body::var_debug_info` into the MIR/importer pipeline
- preserve enough variable/place information through lowering
- emit `DILocalVariable` for args and locals
- emit a useful `DIType` tree for the first supported Rust type surface
- emit `llvm.dbg.value` / equivalent variable location metadata
- decide how this interacts with optimization, likely including an opt-skip or debug-preserving mode
- add cuda-gdb validation for local/argument inspection

## Non-goal

Line-table source mapping is not tracked here. That is handled by #225.

This issue is about variable/type inspection, not just file/line/column mapping.

nihalpasham:
Opened #233 as the first `CUDA_OXIDE_DEBUG=full` implementation slice.

What #233 covers:

```text
rustc var_debug_info
  -> tagged MIR local alloca
  -> LLVM alloca + DILocalVariable + DIType + dbg.declare
  -> LLVM opt may salvage into #dbg_value
  -> cuda-gdb can inspect supported locals/args where locations survive
```

A few choices from the implementation:

- full debug skips Pliron mem2reg for now, because Pliron does not yet rewrite `dbg.declare` into `dbg.value`
- LLVM `opt -O2` stays enabled, since LLVM can salvage many debug locations itself
- line-table mode strips only PTX `.target ..., debug` after `llc` so source breakpoints stay fast
- full mode keeps PTX `.target ..., debug` because cuda-gdb needs it for local/argument inspection

What remains here:

- richer Rust type trees: ADTs, tuples, arrays, slices, closures
- fragments/projections like fields, derefs, destructuring
- precise lexical scopes
- constants from `var_debug_info`
- Pliron-side debug salvage so optimized passes can produce `dbg.value` directly

nihalpasham:
cuda-devtools debugger-assessment follow-ups (validated against main `6c5458fe991bbde32c5bee74d87822aef1b5a691`):

Full variable/type debug children:

[...]

PR #786, opened 2026-08-10
debug: expose closure environments in full debug

uurl:
## Summary

Add full-debug support for Rust closure environments.

When full device debugging is enabled, `cargo-oxide` now disables MIR optimization so closure environments remain represented as aggregate MIR locals instead of being split into scalar replacement fragments.

The MIR importer describes closure types as composite debug types using their captured upvar types and rustc's actual layout. This allows cuda-gdb to inspect captured values as `capture_0`, `capture_1`, and so on.

The `compiler_features` debug fixture and `debug-smoketest.sh` validate closure environments end to end.

## Changes

* Add `-Zmir-opt-level=0` for full device debug builds.
* Preserve normal MIR optimization for line-table-only debugging.
* Emit composite debug types for `RigidTy::Closure`.
* Reuse closure upvar extraction for debug type construction.
* Add a deterministic two-capture closure fixture to `compiler_features`.
* Extend the cuda-gdb smoke test to verify that both captures are inspectable.
* Add `cargo-oxide` tests covering full-debug MIR optimization policy and environment precedence.

## Why disable MIR optimization in full debug?

With optimized MIR, rustc can scalar-replace the closure environment and expose its captures as independent debug fragments rather than preserving the closure as a whole local.

Although those fragments can be carried through LLVM IR, the NVPTX debug pipeline does not currently provide the same reliable cuda-gdb experience for the reconstructed closure value.

Disabling MIR optimization in full-debug mode preserves the closure environment as an aggregate local. This allows the existing alloca-based variable-debug path to describe the complete closure and lets cuda-gdb inspect its captures directly.

Line-table-only debugging remains optimized.

## Validation

```text
cargo fmt --check
cargo test -p cargo-oxide
cargo test -p mir-importer
bash -n scripts/debug-smoketest.sh
scripts/debug-smoketest.sh
git diff --check
```

Results:

```text
cargo-oxide:   173 passed
mir-importer: 1059 passed
debug-smoketest: PASS
```

On `sm_120`, cuda-gdb resolves the closure type and both captured values:

```text
type = struct ... {
  capture_0: u32,
  capture_1: u64,
}

$1 = ... {capture_0: 17, capture_1: 4294967328}
```

Advances #226.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -2999,7 +2999,7 @@ pub enum DeviceDebug {
     Off,
     /// Preserve source line mappings without disabling optimization.
     LineTables,
-    /// Emit full debug information; libNVVM finalization runs unoptimized.
+    /// Emit full debug information; MIR optimization is disabled and libNVVM finalization runs unoptimized.
     Full,
 }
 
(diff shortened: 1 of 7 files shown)


nihalpasham:
**Verified on sm_120:**
- `scripts/debug-smoketest.sh compiler_features` passes and cuda-gdb resolves the closure type, printing `{capture_0: 17, capture_1: 4294967328}` at their layout-derived offsets.

**I layered three commits on top:**
- `cargo oxide pipeline` now applies the codegen env before the rustflags decision, so `--device-debug` works there too;
- `CUDA_OXIDE_DEBUG=2` now parses through one alias table shared with the backend instead of missing the `-Zmir-opt-level=0` decision;
- the closure arm gained a CPU-runnable unit test asserting the capture offsets, so CI covers it without a GPU.

Merging,

advances #226.

nihalpasham (approved):

Issue #782, opened 2026-08-10
77 examples still load loose .ptx files and break under --materialize-cubin


nihalpasham:
#748 fixed the atomics example so it loads the GPU module that is already embedded in the host binary instead of reading `atomics.ptx` off disk. 77 other examples under `crates/rustc-codegen-cuda/examples/` still use the on-disk pattern, and every one of them fails the same way atomics did whenever the loose `.ptx` file is absent.

The failing pattern (before, in each of the 77 `src/main.rs` files):

```rust
let module = ctx.load_module_from_file("example.ptx")?;
let kernels = kernels::from_module(module);
```

The fix, byte-identical to what vecadd and (since #748) atomics do:

```rust
let kernels = kernels::load(&ctx).expect("Failed to load embedded CUDA module");
```

Why the on-disk pattern breaks: `--materialize-cubin` builds never write a loose `.ptx` next to the example (the device image is embedded and finalized to a cubin), and the pre-build cleanup deletes any stale `.ptx` from an earlier run. The example then dies at runtime with CUDA error 301 (file not found) even though a perfectly good module is sitting inside the binary it just built. The exact three-state reproducer from #722:

```text
cargo oxide run atomics                                  # worked (writes atomics.ptx)
cargo oxide run atomics --materialize-cubin --arch sm_120 # failed: error 301, no atomics.ptx on disk
cargo oxide run atomics                                  # worked again
```

After #748, atomics passes all three states; the other 77 examples still only pass the first and third.

This is a mechanical sweep: replace the two-line file-loading pattern with the one-line embedded loader in each example, then confirm with `scripts/smoketest.sh` (standard category) and spot-check a few under `--materialize-cubin`. Examples with a `#[launch_contract]` get the fallible variant the macro generates; everything else takes the exact line above. Count as of today: 77 files (`git grep -l "\.load_module_from_file(" -- 'crates/rustc-codegen-cuda/examples/*/src/main.rs'`).

Found while triaging #748; expanding that PR was deliberately avoided to keep it reviewable.

vyncint:
This looks complete now — the count is **6, not 77**, and all six are deliberate.

Re-running the issue's own count command on `15fa9d9d`:

```console
$ git grep -l "\.load_module_from_file(" -- 'crates/rustc-codegen-cuda/examples/*/src/main.rs' | wc -l
6
```

[...]

PR #787, opened 2026-08-10
fix(examples): load the embedded module instead of a loose .ptx

vyncint:
Refs #782 — **66 of the 77 converted**, with the other 11 accounted for
explicitly rather than silently skipped.

The 77 examples that call `load_module_from_file` split three ways:

| | count |
|---|---|
| converted here | **66** |
| deferred — need a per-file decision | 10 |
| out of scope — not this pattern | 1 |

## The fix works

Your three-state reproducer, run on `barrier` before and after:

| | stock `main` | this branch |
|---|---|---|
| `cargo oxide run barrier` | 0 | 0 |
| `... --materialize-cubin --arch sm_86` | **101** — `DriverError(301` / `file not found` | **0** |
| `cargo oxide run barrier` | 0 | 0 |

## Three things were not byte-identical

Worth flagging, since the issue reads as a single substitution across all 77:

* **Three examples took `unsafe { kernels::from_module(..) }`**, and `load`
  inherits that unsafety — my first pass dropped the block and they failed to
  compile with E0133. They keep it. Their SAFETY comments also claimed *"the PTX
  beside this binary is the one built from `kernels`"*, which is no longer the
  invariant, so they now state it against the embedded module.
* **`vectorization` loads twice**, once per typed module; both sites moved.
* Dead `ptx_path` / `ptx_file` / `ptx_str` bindings — and the `println!`
  announcing the path — go with the load they fed.

## The 10 deferred, and why

These are not the mechanical change:

| examples | why |
|---|---|
| `array_index`, `gemm_sol`, `tcgen05`, `tcgen05_matmul`, `tma_multicast` | wrap the load in a `match` whose `Err` arm inspects `CUDA_ERROR_INVALID_PTX` to explain the GPU is too old. `load` returns `EmbeddedModuleError`, not `DriverError`, so keeping those diagnostics means matching `EmbeddedModuleError::Driver(e)` — a real per-file edit |
| `device_ffi_test`, `mathdx_ffi_test`, `small_type_ffi_test` | LTOIR examples that link their own cubin |
| `coop_groups_demo` | loads one module and derives several typed handles via `from_module(module.clone())`; converting it would issue a separate load per handle |
| `cross_crate_kernel` | has no `#[cuda_module]` of its own — the module belongs to a library crate — so `kernels::load` is not generated in the binary |

Happy to take all ten as a follow-up; the five `match` ones are a tidy unit and
the LTOIR three are their own question.

## The 1 out of scope

`interop_cubin_identity` is the only one of the 77 that never calls
`from_module`. It loads a cubin **it produced itself**, passed in by path, and
resolves the entry point by name with `load_function`. Loading that artifact is
precisely what the example tests, so `kernels::load` would delete its subject —
and it is unaffected by the bug anyway, since nothing there expects a `.ptx` to
appear.

## Verified

* **all 66 compile individually** — 66/66, no failures (each is its own
  workspace, so this is 66 separate builds)
* **full smoketest sweep 201/201 on an A10G**
  (`.gpu-evidence/20260810T101719Z.log`)
* three-state repro above (`20260810T100743Z.log`; stock run in the same session)
* every changed example passes its own `cargo fmt --check` — CI checks these per
  example workspace, and my first pass failed 48 of them on line wrapping

Diff is 66 files, +73/−300 — it deletes about four lines per example and adds one.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/array_for_loop/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/array_for_loop/src/main.rs
@@ -94,11 +94,7 @@ fn main() {
     println!("=== array_for_loop regression (issue #138) ===\n");
 
     let ctx = CudaContext::new(0).expect("Failed to create CUDA context");
-    let ptx_path = concat!(env!("CARGO_MANIFEST_DIR"), "/array_for_loop.ptx");
-    let module = ctx
-        .load_module_from_file(ptx_path)
-        .expect("Failed to load PTX");
-    let module = kernels::from_module(module).expect("Failed to initialize typed module");
+    let module = kernels::load(&ctx).expect("Failed to load embedded CUDA module");
     let stream = ctx.default_stream();
 
     const BLOCK: u32 = 32;
--- a/crates/rustc-codegen-cuda/examples/barrier/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/barrier/src/main.rs
@@ -136,11 +136,7 @@ fn main() {
 
     let stream = ctx.default_stream();
 
-    let module = ctx
-        .load_module_from_file("barrier.ptx")
-        .expect("Failed to load PTX module");
-    let module = kernels::from_module(module).expect("Failed to initialize typed CUDA module");
-
+    let module = kernels::load(&ctx).expect("Failed to load embedded CUDA module");
     const N: usize = 256;
 
     let cfg = LaunchConfig {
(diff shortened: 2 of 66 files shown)


nihalpasham:
**Verified on sm_120:**
- 6/6 smoketest over the touched examples, and
- the three-state `--materialize-cubin` repro on barrier and vectorization is clean: no error 301 at any state, honest output at each.

**I layered two commits on top:**
- vectorization now asserts on the embedded PTX payload it actually launches (skipping the textual check with an honest message only when a cubin build has no PTX text to inspect);
- the stale loose-.ptx prints and dead `ptx_path` bindings are gone, including two more the sweep found in clc and mcast_barrier_test.

Merging; #782 stays open for the deferred examples.

nihalpasham (approved):

PR #788, opened 2026-08-10
build(just): run CI's guards from `check`, and let it work without a driver

vyncint:
Two gaps in the one-command gate. Both surfaced by *running* `just check` rather
than reading it — I installed `just` specifically to check, and each attempt
failed for a different reason.

## 1. `check` skipped all four CI guards

It ran fmt, clippy, tests and docs, but none of the source-reading guards the
workflows enforce:

| guard | CI job |
|---|---|
| `check-error-example-status.sh` | `status-guard` |
| `check-example-smoketest-contract.sh` | `status-guard` |
| `check-dependency-licenses.sh` | `cargo-deny / license-manifest` |
| `check-example-license-policy.sh` | `cargo-deny` |

They were reachable only by reading the workflows, so `just check` could pass
while status-guard or cargo-deny failed. None needs a GPU. All four now run from
a new `check-guards`.

They're invoked as `bash scripts/...`, matching the workflows — and that turns
out to be load-bearing: **`check-example-license-policy.sh` and
`check-example-smoketest-contract.sh` carry no exec bit**, so a bare path exits
126 with "Permission denied". My first run hit exactly that.

## 2. `check` couldn't finish without a driver

The header said clippy and doc-check build `cuda-bindings` so a toolkit is needed
either way — true, but those need `cuda.h` at **build** time, while the
`test-cuda` binaries need `libcuda.so.1` at **load** time. Different requirement,
and the one CONTRIBUTING calls the common case: *"Most crates test on a machine
with no GPU and no NVIDIA driver."*

```
$ just test-cuda      # on stock main, toolkit present, no driver
error while loading shared libraries: libcuda.so.1
exit 127
```

`test-cuda` now shadows the toolkit's link-time stub under that name itself —
the same thing `unit-tests.yml` does for driverless runners and CONTRIBUTING
documents by hand. No-op where a real driver provides the name; the shadow is a
`mktemp -d` removed on exit. This keeps the full coverage you wanted when you
added `test-cuda` to `check`, without requiring a driver to get it.

## 3. Incidental: `just --list` was showing sentence fragments

`just` uses a recipe's **last** comment line as its description, so multi-line
rationales leaked:

```
check      # used to skip entirely -- it needs `cargo-deny` on PATH.
test       # CI's own overrides.
doc-check  # valid Rust); its docs still build under the rustdoc allows ...
```

Five recipes now end their comment block with a one-line summary. Two of those
were pre-existing; fixing them alongside keeps the listing consistent.

## Verified

With `just` 1.58 on a toolkit-only, driverless machine:

* `just --list` reads cleanly
* `just check-guards` → exit 0, all four guards report OK
* `just test-cuda` → **exit 127 before, exit 0 after**
* `just check` → **exit 0** end to end

No GPU involved anywhere in this.

One thing I did *not* do: `chmod +x` the two guards missing the bit. Invoking via
`bash` matches how CI runs them and avoids a mode-only change, but say the word
if you'd rather have the bits made consistent instead.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -70,7 +70,9 @@ test-cuda:
                 shadow="$(mktemp -d)"
                 trap 'rm -rf "${shadow}"' EXIT
                 ln -sf "${root}/lib64/stubs/libcuda.so" "${shadow}/libcuda.so.1"
-                export LD_LIBRARY_PATH="${shadow}:${LD_LIBRARY_PATH:-}"
+                # `:+` keeps the joining colon out when the variable is unset:
+                # a trailing colon would put the cwd on the loader search path.
+                export LD_LIBRARY_PATH="${shadow}${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
                 break
             fi
         done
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Validated both ways:**
- with the real driver the stub branch is skipped and the tests run against the real libcuda;
- a forced driverless dry run maps the toolkit stub with no trailing-colon cwd leak, and `just check-guards` / `just check` pass end to end.

**I layered one commit on top:** the root `cargo deny check` joins check-guards so its comment is true, the `LD_LIBRARY_PATH` expansion only adds a colon when the variable is set, and the comments now name precisely what stays CI-only. On your open question: keeping the `bash` prefix is right, it matches CI and two of the scripts are intentionally non-executable.

Merging.

nihalpasham (approved):

PR #790, opened 2026-08-10
docs: sync the book's quoted toolchain pin, and guard every copy

vyncint:
## The defect

#727 added `rustfmt` to `rust-toolchain.toml` and to the nested copy at
`crates/rustc-codegen-cuda/rust-toolchain.toml`. The two `[toolchain]` blocks
the book quotes were left at five components:

```toml
# cuda-oxide-book/appendix/building-from-source.md, before
# rust-toolchain.toml (already in the repo root)
[toolchain]
channel = "nightly-2026-04-03"
components = ["rust-src", "rustc-dev", "rust-analyzer", "clippy", "llvm-tools"]
```

Both are presented as the repo's actual file — that one says so outright — so a
reader copies the block into their own project and gets a pin with no `rustfmt`.
The symptom then surfaces somewhere unrelated: `cargo oxide fmt` fails, or
`just fmt-check` does, and neither points back at the toolchain file.
`cuda-oxide-book/compiler/rustc-public.md` carried the same stale list.

## Why nothing caught it

The pin is duplicated three ways and no build compares them:

- rustup resolves the backend's own `[workspace]` against the **nested** copy,
  so a `rustc_private` build against a different nightly fails to *load* — a
  build break, not a warning. That file's header already says *"Must match the
  parent cuda-oxide toolchain exactly!"*, but a comment enforces nothing, which
  is precisely how #727 slipped past it.
- The book's copies are prose as far as any gate is concerned.

The book's own bump procedure named only the root file — the other half of why
this drifted — so it now names the nested copy and the guard.

## The guard

`scripts/check-toolchain-parity.sh`, alongside the other source-only
consistency guards, wired in as a `toolchain-pin-parity` job in
`status-guard.yml` (no rustup, cargo, or toolkit — it reads files).

It compares channel and components between the two real files and every
`[toolchain]` block fenced in the book. Deliberate non-goals, so it stays quiet
on things that are legitimately free to vary:

- A block may omit either key; the check is that what a copy *does* state is
  right, not that it restates everything.
- Anything else in a `[toolchain]` block (`targets`, `profile`) may differ.
- The `rustup component add` command lines are **not** covered. They are not
  exhaustive by design, because rustup installs whatever the pin names.

It carries the same parse self-tests as its siblings, since a guard whose
failure mode is "silently matched nothing" has to prove it still reads its
inputs before a clean result means anything.

## Verification

Passes on the corrected tree:

```
OK: rust-toolchain.toml pins nightly-2026-04-03 with 6 components, and the
nested pin plus all 2 block(s) quoted in the book agree.
```

Reverting my fix reproduces the real #727 drift, naming the cause — this is the
guard failing on the bug it was written for, not a synthetic case:

```
error: the toolchain pin is not consistent across its copies
  cuda-oxide-book/appendix/building-from-source.md quotes components [...
    'clippy', 'llvm-tools'] (missing rustfmt); rust-toolchain.toml lists [...
  cuda-oxide-book/compiler/rustc-public.md quotes components [... (missing rustfmt); ...
```

The nested-pin half is exercised too, since nothing in the tree currently
violates it — injecting a channel change gives `pins channel
'nightly-2026-05-01', rust-toolchain.toml pins 'nightly-2026-04-03'`, and
dropping `llvm-tools` from it is caught the same way.

`sphinx-build -W --keep-going -b html`, exactly as the book gate runs it:
`build succeeded`, exit 0.

## Note

This is a fifth guard script. #788 (open) adds a `check-guards` recipe that runs
the existing four from `just check`; whichever of the two lands second should
add this one to that list. I have left the Justfile alone here to avoid a
conflict between them.

### DIFF
--- a/.github/workflows/status-guard.yml
+++ b/.github/workflows/status-guard.yml
@@ -43,11 +43,13 @@ jobs:
       - name: Verify smoketest categories and standard-example success markers
         run: bash scripts/check-example-smoketest-contract.sh
 
-  # The pin lives in three places and no build catches them drifting apart:
-  # rustup reads the nested copy for the backend's own [workspace], so a
-  # mismatch there is a load failure rather than a warning, and the book's
-  # quoted blocks are presented to readers as the repo's real file. #727 added
-  # rustfmt to both real files and left both book quotes behind.
+  # The pin is copied into places no build compares: the nested backend pin
+  # (rustup reads it for the backend's own [workspace], so a mismatch there
+  # is a load failure rather than a warning), the scaffold `cargo oxide new`
+  # writes into user projects, the devcontainer's preinstalled toolchain, the
+  # book's quoted [toolchain] blocks, and the dated rustup/cargo commands in
+  # the book and READMEs. #727 added rustfmt to both real files and left both
+  # book quotes behind.
   toolchain-pin-parity:
     name: toolchain pin parity
     runs-on: ubuntu-latest
--- a/.github/workflows/status-guard.yml
+++ b/.github/workflows/status-guard.yml
@@ -42,3 +42,20 @@ jobs:
       # Reads sources only: no cargo, no CUDA toolkit, no LLVM.
       - name: Verify smoketest categories and standard-example success markers
         run: bash scripts/check-example-smoketest-contract.sh
+
+  # The pin lives in three places and no build catches them drifting apart:
+  # rustup reads the nested copy for the backend's own [workspace], so a
+  # mismatch there is a load failure rather than a warning, and the book's
+  # quoted blocks are presented to readers as the repo's real file. #727 added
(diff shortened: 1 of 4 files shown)


nihalpasham:
**I layered one commit on top** to make the title literal; the guard now also covers
- the scaffold `cargo oxide new` writes into user projects (with its own parse self-test),
- the devcontainer feature pin (parsed as JSON, components checked as a subset since the container completes them on first use), and
- all 22 dated nightly references in tracked markdown, position-independent.

13 mutation checks all fail loudly naming the right copy; sphinx builds the book warning-free. Note the OK line now counts more copies than the PR body quotes.

Merging.

nihalpasham (approved):

PR #792, opened 2026-08-10
feat(fuzzer): fold an aggregate into the trace by its leaves

### LINKED ISSUE #791: fuzzer: the trace API hashes scalars only, so a third of every sweep is refused before it reaches the device
The differential fuzzer's trace API hashes 16 scalar types and nothing else. A generated program whose dump site or return position holds a tuple or an array is refused at the adapter, before it ever reaches the device:

```text
UNSUPPORTED [adapter] unsupported dumped type for Stage 2 adapter: [i8; 1]
```

## Scale

Measured on the checked-in configuration, which has had `composite_count = 3` since #766:

| window | PASS | UNSUPPORTED |
|---|---|---|
| seeds 300-359 | 42 | 18 |
| seeds 1000-1999 | 673 | 320 |

Roughly a third of every sweep is refused for this reason alone. The generator produces those programs, the backend compiles them, and the harness declines to compare them.

## Why it matters here rather than as a nicety

The most valuable find this fuzzer has produced, the silent miscompile behind #393, lived in aggregate constants with padding. `composite_count = 3` puts aggregates back into the generated programs, and the trace boundary is now the thing keeping them away from the comparison.

## What a fix has to respect

An aggregate should fold as its leaves rather than as its bytes. Padding is uninitialised, so folding raw bytes would be undefined behaviour in the CPU oracle, and any layout the two backends pad differently would report a MISMATCH on every seed that touched it. That would turn the exact-comparison property the trace rests on into a source of false positives, in the same way a ULP tolerance would have in #395.

A pointer stays refused whatever else is widened. The CPU oracle and the device hold different addresses for the same object by construction, so a pointer folded into the trace could never compare equal.

alejandro-soto-franco:
Closes #791.

The trace API hashed 16 scalar types and nothing else, so a generated program whose dump site or return position held a tuple or an array was refused at the adapter with `unsupported dumped type`. That was 18 of the 60 seeds in the window the README quotes, and 320 of the 1000 seeds in the sweep that found #765 and #767.

## Summary

Folds an aggregate into the trace as its leaves, in declaration order, so `TraceValue` covers arrays and tuples of anything it already covered, to any nesting depth.

## Changes

- `crates/fuzzer/src/trace.rs`: `TraceValue` for `[T; N]` over a const generic, for tuples up to arity 5, and for `()`. Each walks its leaves through the existing scalar writers.
- `crates/fuzzer/src/lib.rs`: `#![cfg_attr(not(test), no_std)]`, so a test build links std on the host while every real build, the device build included, stays `no_std`.
- `crates/fuzzer/tools/mir_generator.py`: `supported_trace_type` becomes a recursive question over `[T; N]` and tuples rather than a set membership test. It governs both existing call sites, the dumped locals and the return type.
- `crates/fuzzer/README.md`: refreshes the status transcripts and the supported-type list, which this change makes stale.

## Rationale

The byte sequence an aggregate produces is exactly the one a scalar-by-scalar dump of the same leaves produces, so this widens what the trace accepts without introducing a second trace model. Nothing that hashed before hashes differently.

The fold reads leaves and never the aggregate's bytes. Padding is uninitialised, so folding raw bytes would be undefined behaviour in the CPU oracle, and any layout the two backends pad differently would report a MISMATCH on every seed that touched it. Padding is also where the silent miscompile behind #393 lived, so a byte fold would bury the signal this fuzzer exists to find. `(u8, u32)` is the discriminating case in the test: it holds only for an implementation that reads the two fields.

The array walk is an indexed `while` rather than a `for` over the array's `IntoIterator`. This crate is compiled for the device as well as the host, and #399 records that small local arrays consumed through iterator adapters stay in local memory. An index loop states the same thing without depending on how that resolves.

The tuple arity stops at 5 because `TraceDump` stops at 5, and the adapter refuses a wider one rather than emitting a program that fails to compile. Every assertion lives in one test function, because the trace is a single `static mut` and cargo runs test functions on parallel threads, so two folds at once would race on it.

## Validation

Seeds 300-359, the window the README quotes, changing only the trace API and the adapter:

```text
before:  PASS=42  UNSUPPORTED=18
after:   PASS=60  UNSUPPORTED=0
```

Every refused seed became an executing differential run, and none of them mismatched. Seeds 0-99 are also 100 PASS with nothing refused.

Seeds 1000-1999, the window that produced #765 and #767:

```text
PASS=999  UNSUPPORTED=1  COMPILE_FAIL=0  MISMATCH=0
```

The recorded figure for that window on 2026-08-09 was `PASS=673 UNSUPPORTED=320 COMPILE_FAIL=7`. The seven compile failures are gone because #769 landed, and that run predates this branch, so it is a recorded baseline rather than a paired arm.

No mismatch, so this widening surfaced no new silent miscompile in that window. What it did was put 319 more programs through the device where the harness previously refused to look at them.

The one remaining refusal is seed 1436, which returns a `*const i8`:

```text
UNSUPPORTED [adapter] unsupported return type for return-value tracing: *const i8
```

That one is permanent rather than a gap to widen next. The CPU oracle and the device hold different addresses for the same object by construction, so folding a pointer into the trace would report a MISMATCH on every seed that dumped one.

The unit test discriminates. Reversing the array walk, so that it folds `[1, 2, 3, 4]` as `4, 3, 2, 1`, fails the first assertion and leaves the rest of the file passing.

```text
cargo test -p fuzzer                    1 passed, 0 failed
cargo clippy -p fuzzer --all-targets    clean
cargo fmt --check -p fuzzer             clean
```

### DIFF
--- a/crates/fuzzer/README.md
+++ b/crates/fuzzer/README.md
@@ -72,24 +72,38 @@ rather than that file. Its header records this.
 - `UNSUPPORTED [adapter]`: rustlantis generated a MIR program, but our Python
   adapter refused to turn it into a cuda-oxide smoke case.
 
-For example, seed `0` dumps a `[i8; 1]`, which the trace API does not hash, so
-`--start 0 --count 2 --keep-going` currently reports:
+For example, seed `1436` returns a `*const i8`, which the trace API does not
+hash, so `--start 1436 --count 2 --keep-going` currently reports:
 
 ```text
 results:
-  seed 0: UNSUPPORTED [adapter] unsupported dumped type for Stage 2 adapter: [i8; 1] (crates/fuzzer/artifacts/seed-0-unsupported.log)
-  seed 1: PASS [run] CPU/GPU traces matched
+  seed 1436: UNSUPPORTED [adapter] unsupported return type for return-value tracing: *const i8 (crates/fuzzer/artifacts/seed-1436-unsupported.log)
+  seed 1437: PASS [run] CPU/GPU traces matched
 summary: PASS=1, UNSUPPORTED=1
 ```
 
+A pointer is a permanent refusal rather than a gap to widen later. The CPU
+oracle and the device hold different addresses for the same object by
+construction, so folding one into the trace would report a MISMATCH on every
+seed that dumped it.
+
 The typical `UNSUPPORTED [adapter]` cause is a generated `dump_var(...)` call
 or function signature that uses a type the adapter cannot rewrite. The trace
-API hashes:
+API hashes these scalars:
 
 ```text
 bool, i8, i16, i32, i64, i128, isize, u8, u16, u32, u64, u128, usize, char,
 f32, f64
 ```
+
+It also hashes an array or a tuple of anything in that list, to any nesting
+depth, by folding the leaves. A tuple is hashed up to arity 5, matching the
+`TraceDump` implementations. An aggregate's padding is never read, so a dumped
+`(u8, u32)` hashes as the two fields and nothing else.
+
+What remains refused at a dump site is a shape with no leaf reading, such as a
+reference or a slice. An aggregate in an *argument* position is refused
+elsewhere, by `literal_for_type`, which has no literal to construct for one.
 
 In many `UNSUPPORTED [adapter]` cases, the MIR can probably be patched by
 widening the adapter and trace API. The adapter stops because it does not yet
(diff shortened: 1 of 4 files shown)


nihalpasham:
**Reproduced on RTX 5090 (sm_120):**
- seeds 300-359: PASS=60 / UNSUPPORTED=0, against 42/18 on unpatched main;
- seeds 1000-1099 spot check: PASS=100 / UNSUPPORTED=0; tests, clippy, and fmt all clean.

The leaf-walk impls and the recursive Python gate accept and refuse identical types on every edge case we threw at them (1-tuples, nested units, zero-length arrays, arity-6 tuples, slices, pointers).

Merging,

closes #791.

nihalpasham (approved):

PR #793, opened 2026-08-10
docs: list every workspace member in README, and guard the inventory

vyncint:
> **Stacked on #790** — that PR adds the sibling toolchain-parity guard to the
> same workflow file, so this branch builds on it to avoid a conflict.
> **Review only the second commit** (`dd3a0eab`). If #790 lands first this
> rebases cleanly; if it is rejected I will re-cut this from `main`.

## The gap

`dialect-iket` and `iket-lower` have no row in README's Crate Overview. That is
1,700 lines of compiler across two crates — a pliron dialect for in-kernel event
tracing and its lowering — invisible to anyone reading the tree's only map of
itself. Both are workspace members and both carry a `description` in their own
manifest; they were simply never added to the table.

23 of the 25 members were listed, so "every crate gets a row" is already the
convention here. These two are the exceptions, not a new rule.

## Why nothing caught it

No build fails when a crate goes unlisted, and the overview is the only place
the members are enumerated for a human, so the omission is silent by
construction — there is no symptom to notice.

`scripts/check-crate-inventory.sh` compares the root manifest's `members` list
against the crate rows in that section, joining the other source-only
consistency guards and `status-guard.yml`.

Direction is deliberate: **every member needs a row, extra rows are fine.**
`rustc-codegen-cuda` is not a workspace member — it carries its own
`[workspace]` for the `rustc_private` dylibs — and is listed anyway, which is
right, because a reader cares about the crate and not about which workspace
resolves it. So the guard reports members with no row and never the reverse.

Two other scoping choices:

- Members come from the root `Cargo.toml`, not `cargo metadata`. The question is
  which crates the workspace *declares*, which is exactly what that list says,
  and reading it needs no cargo, no lockfile and no network.
- Only the `## Crate Overview` section counts. A crate named in passing
  elsewhere in the README — a command line, the pipeline diagram, a prose aside
  — is not an inventory row, and accepting one would let a crate stay unlisted
  forever.

A glob in `members` would name crates the guard cannot resolve, so it refuses
outright rather than skipping them silently.

## Verification

Passes on the fixed tree:

```
OK: README.md's Crate Overview lists all 25 workspace members.
```

Reverting only the README rows reports exactly the real gap — the guard failing
on the bug it was written for:

```
error: README.md's Crate Overview has no row for:
  dialect-iket
  iket-lower
```

Adding a member with no row is caught the same way (`brand-new-crate`). Both
self-tests trip if either input stops parsing, since a guard whose failure mode
is "silently matched nothing" has to prove it still reads its inputs.


nihalpasham:
**Verified:**
- the guard prints OK on the clean tree and fails loudly naming the missing row under mutation, and
- the inventory matches `cargo metadata` for all 25 workspace members, with the two extra rows (rustc-codegen-cuda, cuda-oxide-book) correctly allowed as intentionally out-of-workspace.

Landing after #790 since this branch is stacked on it.

Merging.

nihalpasham (dismissed):

nihalpasham (approved):

PR #797, opened 2026-08-11
docs(book): point the intrinsics guide at op files that exist

vyncint:
## The defect

`adding-new-intrinsics.md` — the page a new contributor follows to add an
intrinsic — tells them to define the `dialect-nvvm` op in:

- Stage 2: `crates/dialect-nvvm/src/ops/thread.rs`
- Stage 2 (second walkthrough): `crates/dialect-nvvm/src/ops/warp.rs`

**Neither file exists.**

Following the guide is worse than hitting a dead path, because both ops it uses
as examples are now *generated*. `read_ptx_sreg_tid_x` and `shfl_sync_bfly_i32`
are both entries in `intrinsics/catalog.json`, so `cuda-intrinsics-gen` emits
their ops into `crates/dialect-nvvm/src/ops/generated/` (`sreg.rs` and
`warp_shuffle.rs`), and every file there opens with:

```
// @generated by cuda-intrinsics-gen 0.1.0. DO NOT EDIT.
```

So a contributor either creates a file nothing includes, or hand-writes an op
into a generated file and loses it on the next generator run. Both failures are
quiet.

## The change

Stage 2 now names the seven modules that really are hand-written — `asm.rs`,
`atomic.rs`, `cluster.rs`, `debug.rs`, `grid.rs`, `memory.rs`, `wgmma.rs` — and
states the split plainly: ops described by the catalog are generated, bespoke
ops are not, and the example code shows the shape either kind takes. The second
occurrence cross-references the first rather than repeating it.

The example code itself is unchanged — it is still a correct illustration of an
op definition, which is what that section is teaching.

## Scope

Deliberately limited to correcting what is wrong. The generator workflow itself
— how to add a catalog entry and regenerate — **is not documented anywhere in
the book** (no page mentions `cuda-intrinsics-gen` or the catalog at all).
Writing that guide is a larger piece than this fix and probably wants your view
on the intended contributor path, so I have left it alone rather than inventing
one. Happy to follow up if you want it.

## Verification

- Every `crates/...` path this page names now resolves on disk (checked
  mechanically; it was these two that did not).
- Both example ops confirmed present in `intrinsics/catalog.json`, and
  `cuda-intrinsics-gen/src/generate.rs` confirmed to write to
  `crates/dialect-nvvm/src/ops/generated`.
- `sphinx-build -W --keep-going -b html`, as the book gate runs it: exit 0.


nihalpasham:
**Gist:** the guide pointed at op files that no longer exist. Since the catalog generator landed, `tid.x`-class ops are generated into `ops/generated/` and only bespoke ops are hand-written; following the old text either creates an orphan file that never registers or edits a DO-NOT-EDIT generated file. Both mistakes fail silently.

**Verified here:**
- all seven hand-written module paths and the generated split match main;
- book builds warning-free under the CI gate.

Merging. Please do file the follow-up issue for documenting the catalog workflow.

nihalpasham (approved):

PR #800, opened 2026-08-11
fix(mir-importer): lower DST size and alignment intrinsics

### LINKED ISSUE #799: Support `size_of_val` and `align_of_val` for slices and `str`
## Summary

Add MIR importer support for the Rust layout intrinsics:

* `core::intrinsics::size_of_val`
* `core::intrinsics::align_of_val`

for dynamically sized slices and `str`.

cuda-oxide already represents slice and `str` references as fat pointers carrying runtime metadata, but these layout intrinsics are not currently dispatched by `mir-importer`.

## Motivation

`core::mem::size_of_val` and `core::mem::align_of_val` are the stable interfaces used to query the runtime layout of dynamically sized values.

For slices, the expected semantics are:

```text
size_of_val(&[T])  = len * size_of::<T>()
align_of_val(&[T]) = align_of::<T>()
```

For `str`:

```text
size_of_val(str)  = UTF-8 byte length
align_of_val(str) = 1
```

The required metadata is already available in cuda-oxide's fat-pointer representation. Slice and `str` pointers are represented using the existing `MirSliceType`, and runtime slice length can already be extracted through the same metadata path used by `PtrMetadata`.

The missing piece is lowering the `size_of_val` and `align_of_val` intrinsics when they reach the MIR importer.

## Proposed implementation

Add a dedicated layout intrinsic handler in `mir-importer` that:

1. Recognizes `core::intrinsics::{size_of_val, align_of_val}` and their `std` paths.
2. Extracts slice/`str` metadata from the existing fat-pointer representation.
3. Lowers slice size as runtime length multiplied by the element size.
4. Lowers slice alignment to the element ABI alignment.
5. Lowers `str` size to its byte-length metadata.
6. Lowers `str` alignment to 1.
7. Stores the result through the existing projected-destination-aware intrinsic result path.

No new MIR dialect operations should be required.

## Scope

This issue covers:

* `[T]`
* `str`

Trait objects, extern DSTs, and arbitrary structs with unsized tails are out of scope.

## Regression coverage

Add a `dst_metadata` codegen example covering:

* `size_of_val` on `&[u32]`
* `align_of_val` on `&[u32]`
* `size_of_val` on `str`
* `align_of_val` on `str`

The `str` case should use a multibyte UTF-8 value so the test verifies byte metadata rather than character count.

For example:

```text
"oxide✓"
```

has 8 UTF-8 bytes, so:

```text
size_of_val("oxide✓") == 8
align_of_val("oxide✓") == 1
```

## Acceptance criteria

The example should pass with both the normal and non-optimized MIR pipelines:

```text
cargo oxide run dst_metadata
CUDA_OXIDE_NO_OPT=1 cargo oxide run dst_metadata
scripts/smoketest.sh -o '^dst_metadata$'
scripts/smoketest.sh --compile-only -o '^dst_metadata$'
cargo fmt --all -- --check
```

Expected runtime output:

```text
=== dst_metadata ===
PASS: size_of_val on &[u32]
PASS: align_of_val on &[u32]
PASS: size_of_val on str
PASS: align_of_val on str
PASS: dst_metadata
```

uurl:
Closes #799 

## Summary

Add `mir-importer` lowering for:

* `core::intrinsics::size_of_val`
* `core::intrinsics::align_of_val`

for slices and `str`.

The implementation reuses cuda-oxide's existing fat-pointer representation rather than introducing new DST machinery.

## Implementation

Add a dedicated layout intrinsic handler under:

```text
crates/mir-importer/src/translator/terminator/intrinsics/layout.rs
```

The handler recognizes the `core` and `std` intrinsic paths and lowers them using existing `dialect-mir` operations.

For slices:

```text
size_of_val(&[T])  = metadata_len * size_of::<T>()
align_of_val(&[T]) = align_of::<T>()
```

The runtime length is extracted from field 1 of the existing `MirSliceType` fat-pointer representation.

For `str`:

```text
size_of_val(str)  = metadata_len
align_of_val(str) = 1
```

`str` is already represented by cuda-oxide as the same fat-pointer shape as `[u8]`, so no type representation changes are required.

Intrinsic results are written using the existing projected-destination-aware result path.

## Regression coverage

Add the `dst_metadata` example with independent checks for:

* `size_of_val` on `&[u32]`
* `align_of_val` on `&[u32]`
* `size_of_val` on `str`
* `align_of_val` on `str`

The `str` test uses a multibyte UTF-8 string so the size check specifically verifies byte metadata.

```text
"oxide✓" -> 8 bytes
```

Expected runtime output:

```text
=== dst_metadata ===
PASS: size_of_val on &[u32]
PASS: align_of_val on &[u32]
PASS: size_of_val on str
PASS: align_of_val on str
PASS: dst_metadata
```

## Scope

This PR supports slices and `str`.

It intentionally does not add support for:

* trait-object DSTs
* extern DSTs
* arbitrary structs with unsized tails

Those require additional dynamic layout semantics and should be handled separately.

## Validation

```text
cargo oxide run dst_metadata
CUDA_OXIDE_NO_OPT=1 cargo oxide run dst_metadata
scripts/smoketest.sh -o '^dst_metadata$'
scripts/smoketest.sh --compile-only -o '^dst_metadata$'
cargo fmt --all -- --check
```

Results:

```text
cargo oxide run dst_metadata
PASS

CUDA_OXIDE_NO_OPT=1 cargo oxide run dst_metadata
PASS

scripts/smoketest.sh -o '^dst_metadata$'
Pass: 1 / 1
Fail: 0 / 1

scripts/smoketest.sh --compile-only -o '^dst_metadata$'
Pass: 1 / 1
Fail: 0 / 1
```

`mir-importer` unit tests and the existing debug smoketest also pass with this change.


nihalpasham:
**Gist:** the missing lowering, built the way it should be: the runtime length comes from the same fat-pointer field the `PtrMetadata` path already reads, results go through the projection-aware store, `str` is byte-length with align 1, and trait objects / unsized tails are cleanly rejected instead of miscompiled. Same formulas as rustc's own DST glue, at our layer.

- `dst_metadata` passes all four checks, optimized and `CUDA_OXIDE_NO_OPT=1` on a forced rebuild (PTX regeneration confirmed);
- `cargo test -p mir-importer` green.

Merging. This closes #799.

nihalpasham (approved):

PR #802, opened 2026-08-11
test(codegen): cover `Cell` and `UnsafeCell` interior mutability

### LINKED ISSUE #801: Add conformance coverage for `Cell` and `UnsafeCell` interior mutability
## Summary

Add device-side conformance coverage for the core interior-mutability paths exercised by:

* `Cell::get`
* `Cell::set`
* `Cell::replace`
* `UnsafeCell::get`

The relevant compiler machinery is already present in cuda-oxide. The missing piece is explicit end-to-end coverage showing that shared-reference projections, `UnsafeCell` raw-pointer access, and the resulting loads/stores behave correctly on the device.

## Motivation

`Cell<T>` is implemented on top of `UnsafeCell<T>` and allows mutation through a shared reference.

For the operations covered here, rustc ultimately lowers the behavior to ordinary MIR operations involving:

```text
shared reference
    -> field projection
    -> UnsafeCell<T>
    -> raw pointer
    -> load / store
```

In particular:

```rust
Cell::get
```

reads through `UnsafeCell::get()`,

```rust
Cell::set
```

delegates to `Cell::replace`, and

```rust
Cell::replace
```

uses `mem::replace` on the pointer obtained from the underlying `UnsafeCell`.

On the pinned toolchain, the relevant `mem::replace` operations are reduced before backend codegen to ordinary MIR reads and writes, so no dedicated `Cell` or `UnsafeCell` intrinsic lowering should be necessary.

cuda-oxide already uses `UnsafeCell` internally in device-side APIs such as constant-memory and atomic wrappers, but there is currently no focused conformance example covering the core `Cell` APIs themselves.

## Proposed coverage

Add an `interior_mutability` example with two independent device kernels.

### `Cell`

Use a struct containing guard fields around a `Cell<u32>`:

```rust
struct CellState {
    left_guard: u32,
    value: Cell<u32>,
    right_guard: u32,
}
```

Exercise through a shared reference:

```rust
Cell::get
Cell::set
Cell::replace
Cell::get
```

and verify both the returned values and that the adjacent guard fields remain unchanged.

This covers mutation through a projected field reached from a shared reference.

### `UnsafeCell`

Use a second struct containing an `UnsafeCell<u32>` surrounded by guard fields.

Obtain its raw pointer with:

```rust
UnsafeCell::get
```

then perform an explicit raw-pointer read, write, and readback on the device.

The guard fields should again remain unchanged.

## Scope

This issue covers only the low-level interior-mutability mechanisms needed for:

* `Cell::get`
* `Cell::set`
* `Cell::replace`
* `UnsafeCell::get`

It intentionally does not cover:

* `RefCell`
* borrow counters
* panic paths
* heap allocation
* `Rc` / `Arc`
* synchronization or atomic semantics

Those introduce additional MIR and runtime behavior outside this conformance lane.

## Acceptance criteria

The example should run successfully under both normal and low-MIR-optimization pipelines:

```text
cargo oxide run interior_mutability
CUDA_OXIDE_NO_OPT=1 cargo oxide run interior_mutability

scripts/smoketest.sh -o '^interior_mutability$'
scripts/smoketest.sh --compile-only -o '^interior_mutability$'
```

The standalone example should also pass:

```text
cargo fmt --all -- --check
```

Expected runtime output:

```text
=== interior_mutability ===
PASS: Cell::get
PASS: Cell::set
PASS: Cell::replace
PASS: Cell guard fields preserved
PASS: UnsafeCell::get raw-pointer read/write
PASS: UnsafeCell guard fields preserved
PASS: interior_mutability
```

uurl:
Closes #801 

## Summary

Add an `interior_mutability` codegen example covering device-side execution of:

* `Cell::get`
* `Cell::set`
* `Cell::replace`
* `UnsafeCell::get`

No compiler changes are required.

The example confirms that cuda-oxide already handles the underlying ordinary MIR paths for interior mutability, including shared-reference field projections, raw-pointer access through `UnsafeCell`, and loads/stores through those pointers.

## Coverage

The example contains two independent kernels.

### `Cell`

The kernel constructs a value with guard fields around a `Cell<u32>` and accesses the cell through a shared reference.

It verifies:

```text
initial get
set
get after set
replace return value
get after replace
```

and checks that the surrounding guard fields remain unchanged.

This exercises the path:

```text
&struct
  -> projected Cell field
  -> UnsafeCell
  -> raw pointer
  -> load / store
```

### `UnsafeCell`

A separate kernel constructs a struct containing an `UnsafeCell<u32>`, obtains the underlying pointer with `UnsafeCell::get`, and performs an explicit raw-pointer read/write/readback.

Adjacent guard fields are checked to ensure the projected mutation reaches only the intended field.

## Why no importer change is needed

`Cell` is implemented in `core` on top of `UnsafeCell`.

The relevant operations eventually become ordinary MIR memory operations rather than requiring dedicated backend intrinsics.

`Cell::replace` goes through `mem::replace`; on the pinned Rust toolchain its internal read/write intrinsics are lowered before backend codegen into ordinary MIR reads and assignments.

`UnsafeCell::get` itself is effectively a transparent pointer conversion.

The successful device execution therefore confirms the existing MIR importer and lowering paths rather than introducing a new special case.

## Scope

This PR intentionally covers only:

```text
Cell::get
Cell::set
Cell::replace
UnsafeCell::get
```

It does not add coverage for `RefCell`, allocation-backed interior mutability, synchronization, or atomic behavior.

## Validation

Both optimized and low-MIR-opt executions pass:

```text
cargo oxide run interior_mutability
CUDA_OXIDE_NO_OPT=1 cargo oxide run interior_mutability
```

Runtime result:

```text
=== interior_mutability ===
PASS: Cell::get
PASS: Cell::set
PASS: Cell::replace
PASS: Cell guard fields preserved
PASS: UnsafeCell::get raw-pointer read/write
PASS: UnsafeCell guard fields preserved
PASS: interior_mutability
```

Scoped smoketests:

```text
scripts/smoketest.sh -o '^interior_mutability$'

Pass:    1 / 1
Fail:    0 / 1
```

```text
scripts/smoketest.sh --compile-only -o '^interior_mutability$'

Pass:    1 / 1
Fail:    0 / 1
```

The standalone example also passes:

```text
cargo fmt --all -- --check
```

and repository-level formatting/diff checks are clean.


nihalpasham:
**Gist:** first device-side coverage for interior mutability, and well built: guard fields on both sides catch wrong-offset writes, the runtime seed keeps everything out of const-folding, and zero compiler changes needed is itself the finding: the existing projection path is correct, and now it's gated.

- all seven PASS lines on the optimized run;
- same with `CUDA_OXIDE_NO_OPT=1` on a forced rebuild;
- joins the standard smoketest category automatically.

Merging. This closes #801.

nihalpasham (approved):

PR #803, opened 2026-08-11
ci: guard the book's CLI command reference against subcommand drift

vyncint:
The follow-up you offered on #794 — *"Happy to take the drift guard as a
follow-up if you are up for it."*

## Why

#794 added the Command reference table because the book documented 11 of the
CLI's 15 subcommands; `test`, `fmt`, `update` and `emit-ltoir` appeared nowhere.
Nothing stops it falling behind again: adding a variant to the clap enum is a
one-line change, and no build compares the two.

`scripts/check-cli-doc-coverage.sh` compares the `Commands` enum in
`crates/cargo-oxide/src/main.rs` against the rows under `### Command reference`,
**in both directions**. A missing row is the common case; a row for a command
that no longer exists is worse, because a reader trusts it and gets
`unrecognized subcommand`.

## Design choices

**Subcommands come from the enum, not from `cargo oxide --help`.** The enum is
the definition, reading it needs no cargo and no build, and it keeps this guard
in the same source-only class as its siblings — a guard that needed a backend
build to answer a documentation question would not get run. I cross-checked the
parse against the built binary: both report the same **15** commands, zero
disagreement.

**Not required to appear:** anything carrying `hide = true`
(`__materializer-provenance` today) and clap's generated `help`, which is not in
the enum at all, so reading the enum excludes it without a special case.

**Descriptions are out of scope.** clap's short help is a variant's doc-comment
first line, while the table's cells deliberately add MyST backticks around names
like `cuda-gdb` and `setup`; comparing them literally would fail on correct
prose, and normalising invites the brittleness the guard exists to avoid.

## A defect found while wiring it up

`check-guards` was running **five of the seven** guards. The `crate-inventory`
and `toolchain-pin-parity` jobs went into `status-guard.yml` in #793 and #790
**without being added to the recipe**, so `just check` could pass while
status-guard failed — precisely the drift #729 and #788 built that recipe to
prevent. That was my omission in those two PRs. All three are in now, and the
stale counts in its comment ("the status-guard pair", "two of the four") are
corrected. Every `scripts/check-*.sh` is now referenced by both the Justfile and
a workflow.

## Verification

Clean tree:

```
OK: cuda-oxide-book/getting-started/installation.md documents all 15
`cargo oxide` subcommands (1 hidden one(s) correctly excluded).
```

Seven mutations, each failing and naming the cause:

| mutation | caught as |
|---|---|
| new visible subcommand added to the enum | `profile` |
| row deleted from the table | `emit-ltoir` |
| table lists a command that does not exist | `benchmark (no such subcommand)` |
| table documents the hidden command | `__materializer-provenance (hidden from --help)` |
| subcommand renamed (`Doctor` → `Diagnose`) | `diagnose` |
| book section heading renamed | parse self-test |
| `enum Commands` renamed | parse self-test |

The fourth case initially **passed** — my row pattern was `[a-z][a-z0-9-]*`, so a
`__`-prefixed name never matched and the hidden-command branch was unreachable
dead code. Widened to `[a-z_][a-z0-9_-]*`, with a comment on why the underscore
is deliberate.

`just check-guards` green end to end (exit 0, all eight legs), and
`status-guard.yml` parses as YAML with five jobs.

No Rust changed, so `cargo fmt` / `clippy` are untouched by this.

### DIFF
--- a/.github/workflows/status-guard.yml
+++ b/.github/workflows/status-guard.yml
@@ -76,3 +76,19 @@ jobs:
       # Reads sources only: no rustup, no cargo, no CUDA toolkit.
       - name: Verify every copy of the toolchain pin agrees
         run: bash scripts/check-toolchain-parity.sh
+
+  # The book's Command reference is the only complete list of the CLI surface in
+  # the docs, and adding a subcommand to the clap enum is a one-line change that
+  # leaves it silently short -- `test`, `fmt`, `update` and `emit-ltoir` were all
+  # missing when that table was written (#794).
+  cli-doc-coverage:
+    name: CLI command reference
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      # Reads the clap enum and the book: no cargo, no build, no network.
+      - name: Verify the book documents every cargo oxide subcommand
+        run: bash scripts/check-cli-doc-coverage.sh
--- a/Justfile
+++ b/Justfile
@@ -125,19 +125,26 @@ smoketest *args:
 check-errors:
     scripts/check-error-example-status.sh
 
-# The status-guard pair (error* examples in STATUS.md, and the smoketest
-# example contract) plus all three cargo-deny jobs: `cargo deny check` enforces
+# Every status-guard job (error* examples in STATUS.md, the smoketest example
+# contract, the README crate inventory, toolchain-pin parity, and the book's CLI
+# command reference) plus all three cargo-deny jobs: `cargo deny check` enforces
 # deny.toml over the root workspace's resolved graph, the license inventory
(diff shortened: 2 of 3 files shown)


nihalpasham:
**Gist:** the command table gets enforcement in both directions (missing row, phantom row), with loud parse self-tests and count floors so a restructured source fails instead of silently matching nothing. Also fixes real drift: `just check-guards` had quietly stopped running two of CI's guards.

**Verified here:**
- re-derived the regex parse independently against main: 15 visible subcommands, 15 table rows, both directions match, hidden command excluded;
- full eight-leg `just check-guards` green.

Merging. One follow-up worth a line: the variant regex skips tuple variants (none exist today); extending it closes the one silent hole.

nihalpasham (approved):

PR #804, opened 2026-08-11
fix(examples): load the embedded module in the two multi-handle examples

vyncint:
Follow-up to #787 / #789, which left these two behind — the deferred examples
#782 stays open for.

## What they were doing

Both still read a loose `<name>.ptx` from the manifest directory, so each only
ran when a previous `cargo oxide build` had left that file beside it.

## Why each was held back, and what it actually needed

**`coop_groups_demo`** needs two things from one load: a typed handle for the
`#[cuda_module]` grid-sync kernels, and the **raw** module for the nineteen
`cuda_launch!` calls that reach kernels declared *outside* it. The generated
`load` returns only the typed handle — but `LoadedModule::as_cuda_module()`
hands the raw module straight back out:

```rust
let grid_sync_module = unsafe { grid_sync_kernels::load(&ctx) }
    .expect("Failed to load embedded grid-sync module");
let module = grid_sync_module.as_cuda_module().clone();
```

One `cuModuleLoad` serves both paths — no second load, and both halves provably
run the same image.

**`cross_crate_kernel`** needed nothing special in the end. `kernels` is
generic, and a generic module's loader merges the PTX bundles of every crate in
the build rather than selecting one by name — it discards the crate-name hint
outright (`let _ = name;`). That merge is what makes the specializations
resolvable, since they are monomorphized in the binary from kernels declared in
kernel-lib. So a plain `kernels::load(&ctx)` is correct.

*(I first wrote this as `load_named(&ctx, env!("CARGO_PKG_NAME"))` on the theory
that `load` would look for a `kernel-lib` bundle. Reading the macro showed the
generic path ignores the name entirely, so that was a misleading way to write the
same thing. Corrected and re-verified.)*

**`verify_generated_ptx` also read the loose file unconditionally**, so it would
have hard-failed under `--materialize-cubin` even after the module load was
converted. It now asserts on the PTX payload of the embedded bundle — the same
bytes the load hands the driver — and says so and continues when a cubin build
carries no PTX text. That message deliberately avoids a leading `skipping:`,
which the smoketest reads as the whole example opting out.

## Verification — A10G (sm_86), driver 595.71.05

Every run executes the built binary with **all loose `.ptx` and `.cubin` deleted
first**, so only the embedded payload can serve it (`cargo oxide run` would
regenerate the file and prove nothing):

| check | result |
|---|---|
| `cross_crate_kernel`, 0 loose artifacts | all 5 cross-crate specialization tests PASSED |
| `coop_groups_demo`, 0 loose artifacts | builds, reaches its own sm_90 gate, declines cleanly |
| `smoketest.sh` over both + `vecadd` | **3/3 PASS** (coop_groups_demo correctly `PASS (skipped)`) |

## Scope, stated plainly

**This device cannot execute `coop_groups_demo`** — its kernels need Hopper, and
the example returns at its capability check *before* the load this PR changes. So
on sm_86 that conversion is compile-and-link evidence only.

To cover the mechanism on hardware I ran a scratch probe with the same shape —
contracted `#[cuda_module]`, embedded `load`, `as_cuda_module()`, and a raw
`cuda_launch!` of a kernel declared outside the module — on sm_86 in all three
artifact states **including `--materialize-cubin --arch sm_86`**. Typed and raw
launches were correct in every state. The probe is not part of this PR; it exists
only so I am not claiming an unobserved result. You have sm_120, so
`coop_groups_demo` itself is one run away on your side.

**`cross_crate_kernel` still cannot build with `--materialize-cubin`,** for a
reason this PR neither touches nor can fix:

```
error: [rustc_codegen_cuda] Cannot materialize this device artifact: build-time
cubin materialization does not yet support generic #[cuda_module] loading
because it merges PTX bundles across crates at run time
```

So the loose-`.ptx` half of #782 is settled for these two; the cubin half needs
that backend limitation lifted first. Worth noting on #782 rather than assuming
these examples are fully done.

## What remains on #782

Six examples still load a module from a path, and all six look deliberate:
`interop_cubin_identity` and the three `*_ffi_test` load a cubin they build
themselves as the point of the test; `tma_copy` and `wgmma` already load the
embedded module and keep a loose read only as an existence-checked fallback that
names the real cause. If you agree, #782 can close once you are satisfied on the
cubin question above.


nihalpasham:
**Gist:** the last two convertible loose-.ptx examples, each done the right way: coop_groups_demo derives the raw module from the typed load via `as_cuda_module()` so all nineteen raw launches share one `cuModuleLoad`, and cross_crate_kernel uses the merged-bundle loader with the PTX name check reading the embedded payload instead of a file.

- smoketest 2/2, including coop_groups_demo end to end (the run your A10G couldn't reach);
- three-state cross_crate_kernel: plain run passes, `--materialize-cubin --arch sm_120` fails with the documented generic-module backend error, plain run passes again.

Merging. #782 stays open for the six deliberate loaders, the generic-module cubin gap, and the hashmap bench bins.

nihalpasham (approved):

PR #805, opened 2026-08-11
ci: cover the codegen backend's own workspace in both license guards

vyncint:
> **Stacked on #803**, which adds the CLI-reference guard to the same
> `check-guards` recipe. **Review only the second commit** (`5a80b3e1`). If #803
> lands first this rebases cleanly; if it is rejected I will re-cut from `main`.

The second pass asked for in the #662 review — *"a second pass over
`crates/rustc-codegen-cuda` (its own workspace, invisible to both cargo-deny and
this guard today)."* Still true, and it had already cost a row.

That crate carries its own `[workspace]` for the rustc-private dylibs, so it is a
**third** resolution alongside the root and the examples. Neither guard reached
it: the inventory guard reads the root workspace via `cargo metadata`, which
stops at that boundary, and the policy side ran `cargo deny check` on the root
plus a script over the examples.

## Inventory — a live omission

**`rustc_codegen_cuda` had no row in `dependency-licenses.csv` at all.** The
largest first-party crate in the tree, absent from the inventory, while every
other member carries a `1st Party (workspace)` row. Nothing existed that could
have caught it.

The guard now unions the declared crates of both first-party workspaces — 41 → 55
— and the row is added.

## Policy — narrower than it looks

`cargo deny` now runs over that workspace too, in the same job as the root run so
the `cargo install` is shared.

I want to be precise about what this actually buys, because the obvious framing
overstates it. All **82** of the backend's third-party crate *names* also appear
in the root graph, so nothing there is unlicensed or ungoverned today. What
differs is **versions**: 17 `(name, version)` pairs resolve in that workspace and
nowhere else — `anyhow 1.0.100`, `libc 0.2.187`, `hashbrown 0.17.0`,
`linkme 0.3.35`, and 13 more. Advisory and license verdicts are per version, so
an advisory against one of those 17 is invisible to the root run. That is the gap
this closes.

## Verification

- Inventory guard passes with the row (55 crates); with the row removed again it
  fails naming exactly `rustc_codegen_cuda`.
- **The policy leg discriminates.** Banning `anyhow =1.0.100` — a backend-only
  version — leaves the root run at `bans ok` and makes the backend run fail with
  `error[banned]: crate 'anyhow = 1.0.100' is explicitly banned`. So it covers
  crates the root run genuinely cannot see, rather than re-checking the same set.
- `cargo deny --manifest-path crates/rustc-codegen-cuda/Cargo.toml --locked check`
  is clean on the current tree, so the policy half is prevention, not a fix.
- `just check-guards` green end to end (9 legs); both workflow files parse as
  YAML.
- `deny.toml` restored bit-identical after the mutation tests.

No `--config` on the new run: cargo-deny walks up from the manifest directory to
the repository-root `deny.toml`, the same resolution `check-example-license-policy.sh`
already relies on.

## Note

`--locked` is on the new run, matching the examples script — a stale lock fails
rather than being silently re-resolved. The root `cargo deny check` above it does
not pass `--locked`; I left that alone rather than widen this PR, but it is a
one-word change if you want the two consistent.


nihalpasham:
**Gist:** the follow-up the #662 review asked for. The backend's own workspace was invisible to both guards: no inventory row for the largest first-party crate, and 17 version-distinct resolutions never checked against advisories. The second cargo-deny run reuses the example-policy invocation (no `--config`, walk-up to the root deny.toml), and the inventory script unions both workspaces instead of special-casing.

**Verified here:**
- full `just check-guards` green on the branch;
- inventory covers all 55 declared crates (was 41);
- `cargo deny --manifest-path crates/rustc-codegen-cuda/Cargo.toml --locked check` passes all four checks.

Merging after #803 (the branch stacks it). Your `--locked` note on the root run is a one-word follow-up; we'll take it.

nihalpasham (approved):

PR #812, opened 2026-08-12
chore: fix Apache-2.0 license notice and attribution gaps

rparolin:
## Why

An open-source license review of this repo turned up four compliance gaps. This fixes all four.

Licensing metadata only. No logic, API, or behavior changes.

## What's in the diff

**1. `LICENSE`.** The copyright line read `Copyright 2024-2026 NVIDIA CORPORATION`, missing `(c)`, `& AFFILIATES.` and `All rights reserved.`. Now matches the required string.

The review cited this against `LICENSE-APACHE`. #331 folded that file into `LICENSE` on 2026-07-03, after the reviewed snapshot was taken, so the fix lands in `LICENSE` instead.

**2. `CONTRIBUTING.md`.** The license-header example showed a bare SPDX identifier with no copyright notice. Replaced with the `/* */` block that 599 of 709 Rust files already use, plus the exact required wording and the comment syntax for non-Rust files.

**3. Headers.** 26 files said `NVIDIA CORPORATION.` without `& AFFILIATES`. Another 47 had no SPDX header at all: 8 trybuild fixtures and 39 non-Rust files (CUDA wrappers, the `scripts/check-*.sh` guards, GEMM benches, book assets). Four more had the notice wrapped across two lines and are now joined so the required sentence appears verbatim.

**4. `THIRD_PARTY_NOTICES`.** New file. `crates/fuzzer/rustlantis/` is third-party source vendored in tree (46 files, Apache-2.0 OR MIT, ETH Zurich) with no attribution at the root. Scoped to that alone — Cargo dependencies aren't redistributed here and stay recorded in `dependency-licenses.csv`. Rustlantis' own files are untouched so their upstream notices survive.


nihalpasham:
**Gist:** covers the compliance review's three remaining items: the 81 per-crate LICENSE copies are gone (every crate declares `license = "Apache-2.0"`; nothing referenced them), the root LICENSE opens with the required copyright line and the APPENDIX is restored, and the remaining source headers are in.

**Verified here:**
- repo-wide sweep at the head: zero non-conforming NVIDIA copyright strings, zero source files missing SPDX headers; the two BSD-3-Clause CUTLASS-derived `.py` files keep their well-formed notices as instructed;
- the deleted LICENSE files were byte-identical copies of the root blob; no `Cargo.toml` uses `license-file`; trybuild `.stderr` expectations regenerated; shebangs intact;
- pushed 47ac27af restoring the trailing newlines the header insertion dropped in 24 files (one byte each, no `.stderr` impact).

Landing once CI is green.

nihalpasham (approved):

Issue #811, opened 2026-08-11
warp_reduce runs a 10-instruction shuffle tree where Ampere's redux.sync does it in one: 4.74x measured, but no way to gate it on the target


vyncint:
`cooperative_groups::warp_reduce<T, Op, N>` is a generic butterfly: `log2(N)` rounds of `shfl_xor_via` + `Op::combine`. For a full warp that is **5 shuffles + 5 combines = 10 instructions**. `redux.sync.add.u32` does the same job in **one**, cuda-oxide already exposes it as `warp::redux_sync_add` (#223, #258) — the generic API just never routes integers to it.

## Measured

A10G (sm_86, driver 595.71.05), 1M threads x 64 full-warp `u32` sum reductions each, 30 timed launches after 5 warmup. Same kernel either way; only the reduction call differs:

| form | us/launch | correct |
|---|---|---|
| `warp_reduce::<u32, Sum, _>(&tile, v)` | 159.4 | yes |
| `warp::redux_sync_add(FULL_MASK, v)` | **33.6** | yes |
| | | **4.74x** |

Both bit-identical against a host replay, which is expected: wrapping `u32` addition is associative mod 2^32, and min/max/and/or/xor are associative and exact, so the reduction order cannot change the result.

PTX confirms the mechanism — 10 `shfl.sync` versus 8 `redux.sync` across the respective unroll factors, i.e. 10 instructions per reduction against 1.

**Scope of that number:** the probe kernel does 64 reductions and almost nothing else, so it is reduction-bound and this is the primitive's cost in isolation, not an application speedup. It sizes the prize; a memory-bound kernel with one reduction at the end would see little.

## The participation contract already matches

This looked like it might be unsound, and it is not. #672 states the existing behaviour plainly: "The warp reductions shuffle with the full 32-lane member mask, so every lane must be launched and converged." #672 then added *separate* `_partial` entry points for short warps rather than relaxing `warp_reduce`.

So `warp_reduce` on a `WarpTile<32>` already requires a fully converged warp and the full mask — exactly `redux.sync`'s requirement. Substituting it does not weaken anything. Sub-warp tiles (`N < 32`) would keep the shuffle path, since each tile carries its own mask and the wrapper's out-of-tile lane substitution is shuffle-specific.

## What actually blocks it

[...]

PR #223, opened 2026-06-14
feat: add redux.sync.add warp reduction (sm_80+)

goog00:
Single-instruction warp sum via `@llvm.nvvm.redux.sync.add`: device API, NVVM `ReduxSyncAddOp`, MIR import dispatch, lowering with `[mask,value]` → `[value,mask]` flip, and convergent marking on export.

## Summary

Adds `warp::redux_sync_add(mask, value)` — a single-instruction warp-wide sum reduction (PTX `redux.sync.add`, sm_80+). Previously a 32-lane warp sum needed 5×`shfl` + 5×`add`; this lowers to one hardware instruction. Covers both `u32` and `i32` (two's-complement add is bit-identical).

## Changes

- **cuda-device**: new `redux_sync_add` device API (mirrors existing `*_sync` collectives).
- **dialect-nvvm**: new `ReduxSyncAddOp` (2 operands `[mask, value]`, 1 result) + register.
- **mir-importer**: `emit_warp_redux` helper + dispatch branch.
- **mir-lower**: `convert_redux` → `@llvm.nvvm.redux.sync.add`, flipping operands to `[value, mask]` (LLVM wants `(src, membermask)`); `MirToLlvmConversion` impl.
- **llvm-export**: mark `llvm.nvvm.redux.*` convergent (warp collective — must not be hoisted into divergent branches).
- **tests + example**: op construct/verify, convergent assertion, and `redux_sum` e2e example (sm_80-gated).

## Testing

Verified end-to-end on **RTX 4080 SUPER (sm_89)** — PTX emits `redux.sync.add.s32` and each warp computes the correct sum:

GPU Compute Capability: sm_89
--- Test 1: redux_sync_add over lane ids ---
Warp sums: [496, 496, 496, 496, 496, 496, 496, 496]
✓ All 8 warp sums correct (each = 496)
--- Test 2: redux_sync_add over input data ---
Warp sums: [496, 496, 496, 496, 496, 496, 496, 496]
✓ All 8 warp sums correct (each = 496)
SUCCESS: redux.sync.add produced correct warp sums

- [x] `cargo oxide run redux_sum` passes (sm_89)
- [x] `cargo test -p dialect-nvvm -p llvm-export` passes (op verify + convergent)
- [x] New example added (`redux_sum`)

## Checklist

- [x] All commits signed off (`git commit -s`)
- [x] No files under `cuda-bindings/` modified
- [x] SPDX headers on new source files

[...]

PR #813, opened 2026-08-12
docs(warp): point integer reductions at redux.sync, with the measurement

vyncint:
Option 3 from #811 — document the gap, change no API.

`warp_reduce` is generic over `T`, so a full warp always costs five shuffles plus
five combines. `redux.sync` does an integer reduction in **one** instruction and
cuda-oxide has exposed it since #223/#258 — but nothing told a reader that, so
the book taught the butterfly and stopped there.

## The number

A10G (`sm_86`), 1M threads × 64 full-warp `u32` sum reductions, 30 timed launches
after 5 warmup, same kernel with only the reduction call swapped:

| form | µs/launch |
|---|---|
| butterfly (`warp_reduce` / hand-written) | 159.4 |
| `warp::redux_sync_add` | **33.6** |
| | **4.74x** |

Bit-identical: wrapping `u32` addition is associative modulo 2³² and
min/max/and/or/xor are exact, so reduction order cannot change the answer.

## Where it went

Both places a reader might look:

- the book's **Warp reduction** section gains a subsection naming all eight
  functions, the measurement, and the constraints;
- **`warp_reduce`'s own rustdoc** carries a short performance note, since that is
  what someone about to call it actually reads.

## Scope, stated in both

The probe kernel does almost nothing but reduce, so **4.74x sizes the primitive,
not an application**. A kernel that reduces once after a memory-bound pass will
see very little. Both texts say so — I would rather a reader calibrate than take
a micro-benchmark as a promise.

Three constraints are spelled out rather than implied:

- **Integers only** — there is no `f32` `redux` before `sm_100`, so floats keep
  the butterfly.
- **A full converged 32-lane warp** — which is the contract the shuffle
  reductions already require (#672: "every lane must be launched and
  converged"), so this asks nothing new. Sub-warp tiles and short tails stay on
  shuffles, with a pointer to `reduce_sum_f32_partial` for the tail case.
- **It will not assemble below `sm_80`** — so the call needs gating the way the
  `redux_sum` example gates it, on `compute_capability`.

## One small correction alongside

The sentence above the new subsection said the butterfly is "five instructions".
It is five shuffles *plus* five combines, and the new subsection counts
instructions a few lines later, so I changed it to "five shuffle steps" to keep
the two consistent.

## Not closing #811

No API or codegen change here. #811 stays open for making `warp_reduce` select
this form automatically, which needs device code to know its target architecture
— the broader design question. This documents what is available today.

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p cuda-device` clean, so the
  intra-doc link to `redux_sync_add` resolves;
- `sphinx-build -W --keep-going -b html` succeeds;
- all eight function names, `reduce_sum_f32_partial`, and the `redux_sum`
  capability gate checked against the source rather than remembered;
- `cargo fmt --check` clean.

### DIFF
--- a/crates/cuda-device/src/cooperative_groups.rs
+++ b/crates/cuda-device/src/cooperative_groups.rs
@@ -998,6 +998,25 @@ impl WarpShuffle for f32 {
 /// the butterfly converges all lanes onto the same answer, which is the
 /// usual ergonomic shape and matches CUB's `WarpReduce`.
 ///
+/// # Performance: integers on `sm_80`+ have a one-instruction form
+///
+/// This is generic over `T`, so a full warp always costs five shuffles plus
+/// five combines. For **integer** reductions on Ampere and newer,
+/// [`crate::warp::redux_sync_add`] and its `min`/`max`/`and`/`or`/`xor`
+/// siblings do the whole reduction in a single `redux.sync` instruction.
+/// Measured on an A10G (`sm_86`) over 1M threads reducing 64 times each:
+/// 159.4 µs per launch here against 33.6 µs for `redux.sync`, **4.74x**, with
+/// bit-identical results. That is the primitive in isolation — a kernel that
+/// reduces once after a memory-bound pass will see far less.
+///
+/// This function does not select that form for you: `redux.sync` will not
+/// assemble below `sm_80`, and device code has no way to ask what target it is
+/// being compiled for (see
+/// [#811](https://github.com/NVlabs/cuda-oxide/issues/811)). Call it directly
+/// when you know the target is Ampere+, gating on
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Gist:** implements #811's option 3: document the gap with the measurement rather than change an API that can't see the target arch from device code.

**Verified here:**
- all eight `redux_sync_*` names in the table exist with the documented semantics (add is wrapping, u32/i32 equivalent);
- `warp_reduce`'s `Sum::combine` is `wrapping_add`, so the bit-identical claim is sound;
- the `redux_sum` example's sm_80 gate and skip message match the note.

Landing once CI is green. Partially addresses #811; the root cause (target-aware device code) stays open.

nihalpasham (approved):

PR #815, opened 2026-08-12
docs(api-ref): the shuffle i32 variants documented here do not exist

vyncint:
The quick reference's warp section advertises an `i32` form for **all four**
shuffles. None exist.

```rust
// i32 variants
let partner_i = warp::shuffle_xor_i32(val, mask);   // does not compile
```

```
| `shuffle_xor_{f32,i32}(val, mask)`    | ...   <- four rows, all claiming i32
| `shuffle_down_{f32,i32}(val, delta)`  | ...
| `shuffle_up_{f32,i32}(val, delta)`    | ...
| `shuffle_{f32,i32}(val, src)`         | ...
```

`cuda_device::warp` has no `_i32` function at any width. The real set is the
**unsuffixed name for `u32`**, plus `_f32`, `_f64` and `_u64`, each with a
`_sync` form taking an explicit member mask.

This is the page most likely to be copied from, and the code line does not
compile.

## Worth more than deleting the wrong line

The page showed only the `_f32` forms, so it **never mentioned the `u32` ones at
all** — a reader shuffling integers had nothing correct to copy, which is
plausibly how the fictional `i32` entry got written in the first place. So the
example now uses the unsuffixed form, and the table lists the widths that exist
with a note that an `i32` should be reinterpreted and passed through the `u32`
form, since a shuffle moves bits without interpreting them.

## Verification

- Every function the page now names checked against
  `crates/cuda-device/src/warp.rs` — `shuffle`, `shuffle_xor`, `shuffle_down`,
  `shuffle_up`, the `_f32`/`_f64`/`_u64` suffixes and a `_sync` sample, all
  present.
- No `_i32` shuffle claim remains in any user-facing page.
- `sphinx-build -W --keep-going -b html` succeeds.

*(Found by diffing every `warp::`/`thread::`-style name the book cites against
what `cuda-device` exports. Three more citations in the compiler-internals pages
also fail to resolve; those are a separate PR, since they name mir-importer
helpers rather than user API.)*

### DIFF
--- a/cuda-oxide-book/appendix/api-quick-reference.md
+++ b/cuda-oxide-book/appendix/api-quick-reference.md
@@ -358,8 +358,8 @@ let from_above = warp::shuffle_down_f32(val, delta);
 let from_below = warp::shuffle_up_f32(val, delta);
 let from_lane  = warp::shuffle_f32(val, src_lane);
 
-// i32 variants
-let partner_i = warp::shuffle_xor_i32(val, mask);
+// u32 is the unsuffixed form
+let partner_u = warp::shuffle_xor(val_u32, mask);
 
 // Vote
 let all_true = warp::all(predicate);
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Gist:** right fix, nearly complete: supported-features.md still carried two i32-shuffle claims, and the same fence showed `warp::popc(mask)` for a function that takes a bool predicate. Pushed 08456dbe extending the diff to cover both.

**Verified here:**
- every name the new table cites exists in warp.rs (u32-canonical unsuffixed, `_f32`/`_u64`/`_f64` variants, `_sync` forms; no `_i32` shuffle at any width);
- post-fix grep: no user-facing i32-shuffle or `popc(mask)` text remains; the surviving i32 mentions are compiler-internal pages naming the real NVVM intrinsic.

Landing once CI is green.

nihalpasham (approved):

PR #816, opened 2026-08-12
docs(compiler): the shuffle dispatch these pages show does not exist

vyncint:
Continuation of #797, which corrected **Stage 2** of the intrinsics guide for the
generated/hand-written split. **Stage 3 has the same rot** and #797 did not touch
it.

## The defect

`adding-new-intrinsics.md` and `mir-importer.md` both show
`"cuda_device::warp::shuffle_xor"` dispatching to
`intrinsics::warp::emit_warp_shuffle_i32`.

That function **does not exist**. The file they cite does —
`crates/mir-importer/src/translator/terminator/intrinsics/warp.rs`, 118 lines —
but it defines only `emit_warp_redux`, and the string
`"cuda_device::warp::shuffle_xor"` appears in **no** hand-written match.

The shuffles are **catalog** intrinsics: `shuffle_xor` is in
`intrinsics/catalog.json`, so its dispatch arm is generated into
`intrinsics/generated.rs` (line 1660 and around), beside the generated ops #797
documented. A reader following either page hunts for a dispatch that was never
there in this form.

## The fix, scoped like #797

The guide is told what is generated rather than rewritten:

- `mir-importer.md`'s match keeps its two genuinely hand-written arms and gains a
  paragraph on where the rest come from;
- the guide's Stage 3 gains the same note, and its illustrative emitter is
  renamed `emit_two_operand_intrinsic` — it is teaching the *shape* of a
  two-argument emitter, so it should not wear the name of a symbol that is not in
  the tree.

The op names these pages cite — `ShflSyncBflyI32Op`, `convert_shuffle_i32` — I
checked and both are **real**, so they are untouched.

## Also on `mlir-dialects.md`

> "intrinsic functions like `thread::index_x()` or `warp::shfl_sync_bfly()`"

Neither exists. The real ones are `thread::index_1d()` and `warp::shuffle_xor()`;
`shfl_sync_bfly` is the PTX instruction name, not the Rust entry point. Its
mangled-symbol example further down said `thread_index_x` and now agrees with the
corrected name.

## Verification

- Every name these pages now cite exists in the tree (`emit_warp_redux`,
  `index_1d`, `shuffle_xor`, `ShflSyncBflyI32Op`, `convert_shuffle_i32`).
- No nonexistent name remains on any page.
- `sphinx-build -W --keep-going -b html` succeeds.

*(Found by diffing every API-shaped name the book cites against what the tree
exports — the same sweep that produced #815.)*


nihalpasham:
**Gist:** right to kill `emit_warp_shuffle_i32`, but several corrected passages were still wrong about how dispatch works: only the `_sync` variants are match strings (bare `shuffle_xor` is a source-level wrapper), `emit_warp_redux` is dead code and the redux family is catalog-generated, the barrier arm is `thread::sync_threads`, and `index_1d` never appears as a lone sreg read. Pushed 336a70b0 aligning all three pages with the generated dispatcher.

**Verified here:**
- every corrected passage backed by dispatcher source: generated.rs match strings, the hand-written `typed_swap` arm now used as the hand-dispatch example, and the real 3-operand shuffle converter in Stage 4;
- two provenance statements fixed alongside (NVVM ops are emitted by mir-importer, not mir-lower);
- Sphinx build with CI's `-W` flags green.

Landing once CI is green.

nihalpasham (approved):

Issue #808, opened 2026-08-11
book: the intrinsics guide covers only hand-written ops; the cuda-intrinsics-gen catalog workflow is undocumented


vyncint:
Filing the follow-up asked for in the #797 review.

## What #797 fixed, and what it left

#797 corrected `cuda-oxide-book/compiler/adding-new-intrinsics.md`, which told contributors to define new `dialect-nvvm` ops in `crates/dialect-nvvm/src/ops/thread.rs` and `ops/warp.rs`. Neither file exists: both example ops are catalog entries, generated into `ops/generated/`. Following the old text either created a file nothing registers or edited a `DO NOT EDIT` generated file, and both fail quietly.

That PR's fix was to scope the page honestly — name the seven hand-written modules, and warn that catalog-described ops are generated. It deliberately did **not** document the generator, because that is a larger piece and the intended contributor path is yours to define.

So the page is now correct but incomplete: it explains the path for bespoke ops and tells you the other path exists without saying how to walk it. For scale, that other path is the majority — **986 catalog entries across 35 generated files**.

## What is undocumented today

No page in the book mentions `cuda-intrinsics-gen` or `intrinsics/catalog.json` except the warning #797 added. A contributor adding a catalog intrinsic gets no guidance on any of the following, all of which exist and are load-bearing:

**The generator's surface**, from its own usage text:

```
cuda-intrinsics-gen extract --intrinsics-json FILE --nvptx-json FILE
cuda-intrinsics-gen extract --llvm-src DIR --llvm-tblgen FILE
cuda-intrinsics-gen generate
cuda-intrinsics-gen check
cuda-intrinsics-gen coverage [--family NAME]
cuda-intrinsics-gen check-abi-history --base-ref REF
cuda-intrinsics-gen probe [--all | --intrinsic ID] [--llc FILE] [--skip-terminal]
cuda-intrinsics-gen probe --candidate --intrinsic ID --llc FILE --gpu-target TARGET --ptx-feature FEATURE (--ptxas FILE | --skip-terminal)
```

**Three CI gates that a catalog change has to satisfy** (`unit-tests.yml`), none of them discoverable from the docs:

| gate | what it enforces |
|---|---|
| `cuda-intrinsics-gen -- check` | the committed generated sources match what the generator produces |
| `cuda-intrinsics-gen -- probe --all --skip-terminal` | the pinned `llc` identity and the exact emitted PTX per route |
[...]

PR #818, opened 2026-08-12
docs: add the six missing crate READMEs

vyncint:
Twenty of the twenty-six workspace crates carry a README. These six did not:
`cuda-artifact-finalizer`, `cuda-intrinsics`, `cuda-intrinsics-gen`,
`cuda-oxide-codegen`, `mir-transforms`, `nvvm-transforms`.

#793 made the README crate table cover every member, which fixed the *map*. This
fills in the six entries that had nothing behind them.

## What each says

Each follows the shape of the existing ones — crate name, what it is, then the
specifics worth knowing on arrival — and is drawn from that crate's own manifest
description and module header rather than invented:

| crate | the thing worth knowing |
|---|---|
| `cuda-artifact-finalizer` | why the libNVVM/nvJitLink policy lives in **one driverless crate**: build-time materialization and the runtime fallback cannot drift apart |
| `cuda-intrinsics` | most code wants `cuda-device`; this is the raw compiler contract, generated from the catalog, not hand-written |
| `cuda-intrinsics-gen` | the seven invocations, **which three gate CI**, the DO-NOT-EDIT banner, the `rustfmt` dependency |
| `cuda-oxide-codegen` | the point is the **absent** rustc linkage; `experimental` is the only supported surface |
| `mir-transforms` | where in the pipeline it runs, and why that position matters — Rust-level structure is still visible there |
| `nvvm-transforms` | what libNVVM will not accept, and that ordinary PTX builds skip the stage entirely |

## Verification

Checked against the tree rather than remembered:

- the `// @generated by cuda-intrinsics-gen ... DO NOT EDIT.` banner text;
- the three `cuda-intrinsics-gen` gates really are in `unit-tests.yml`;
- `pub mod experimental` really is the surface;
- `mir.unroll_hint` is the real op name;
- the generated intrinsic bodies really are `unreachable!` — 986 of them, under
  `src/generated/`, which is why calling one outside device compilation panics.
  (My first check grepped `src/lib.rs`, found zero, and I went and looked rather
  than shipping the claim on a failed check.)

Every workspace crate now has a README. No code touched.


nihalpasham:
**Gist:** six solid READMEs completing a repo convention; three had verified under-inclusions. Pushed df000d86: the eighth usage form (`probe --candidate`), a generated-outputs list matching all nine `MANAGED_GENERATED_ROOTS`, and the two mir-transforms passes the crate header skips (canonicalize, scalarize_borrowed_aggregate_reads).

**Verified here:**
- every added claim quoted from source (`print_usage`, `MANAGED_GENERATED_ROOTS`, the pass doc headers);
- existing README content spot-checked against the crates it describes (finalizer deps and consumers, the three CI gates, `@generated` banners).

Landing once CI is green. Partially addresses #808; the book page it asks for stays open.

nihalpasham (approved):

PR #819, opened 2026-08-12
ci: guard the device-API names the book shows in Rust blocks

vyncint:
> **Stacked on #815 and #816** — this guard needs both of their fixes to be
> green, so the branch carries them. **Review only the third commit**
> (`f89c1944`). If those land first this rebases to one commit.

A Rust code block in the book is something a reader **pastes**. When a device
function is renamed, or never existed, nothing fails: the book renders, CI is
green, and the reader finds out. That has now bitten three times —

- #797, op files that no longer existed;
- `warp::shuffle_xor_i32` in the API quick reference, advertised as an "i32
  variant" that has never existed (#815);
- a dispatch helper in the compiler pages replaced by generated code (#816).

All three were found by hand, months apart.

## What it checks

`scripts/check-book-api-names.sh` resolves every `warp::`, `thread::`, `grid::`
and `cluster::` call the book shows against the functions `cuda-device` exports.
On the current tree: **26 calls across 271 Rust blocks against 1,148 exports.**

Scope is what makes it precise rather than merely broad:

- **Only ```rust fenced blocks.** In prose a name may legitimately be a PTX
  instruction (`shfl.sync.bfly`), a module path, or another project's API; inside
  a Rust block it is a call and has to exist. This restriction is what takes the
  check from dozens of false positives to **none**.
- **Only the four device modules.** Host APIs are already covered by rustdoc,
  which builds under `-D warnings`.
- **Existence only** — never arity or types. Those change for good reasons and
  the compiler catches them; a name that is simply absent is the silent case.

## The lookbehind is load-bearing

I only found this by running the guard against my own branch: `warp` and
`thread` are *also* module names inside the compiler, so the intrinsics guide's
`intrinsics::warp::emit_two_operand_intrinsic(...)` — a mir-importer path with no
business resolving against `cuda-device` — was flagged until the pattern learned
to reject a qualified prefix. An explicit `cuda_device::` prefix stays accepted.

## Verification

| case | result |
|---|---|
| current tree | passes, 26/26 resolve |
| book reverted to pre-fix state | reports `warp::shuffle_xor_i32` — the defect it exists for |
| synthetic rename (`shuffle_xor` → `shuffle_rotate`) | caught |
| compiler-internal `intrinsics::warp::…` | correctly ignored |
| `cuda-device/src` missing | exits non-zero rather than passing vacuously |
| `cuda-device/src` truncated | reports the now-missing names |

`just check-guards` green end to end; every `scripts/check-*.sh` is referenced by
both the Justfile and a workflow.

*(Minor: #806 also edits the `check-guards` recipe, on a different line. They
should merge cleanly; happy to rebase if not.)*


nihalpasham:
**Gist:** right guard at the right tier: the book's Rust blocks are statement excerpts with free variables, so name resolution against cuda-device's export surface is what's checkable without compiling 271 fragments. One hole: both regexes were lowercase-only, leaving the `threadIdx_x` family invisible. Pushed 4960d371 widening the identifier classes; coverage goes from 26 to 33 unique calls with the mixed-case family now guarded.

**Verified here:**
- guard red on main's stale `warp::shuffle_xor_i32`, green on the PR's book content;
- two fabricated-API probes both caught post-fix, then reverted;
- job and Justfile wiring matches the existing guard family (checkout@v4, 5-minute bound).

Landing after #815 and #816: this branch stacks both and will need a rebase once they merge.

nihalpasham (dismissed):

nihalpasham (approved):

PR #826, opened 2026-08-12
test(codegen): cover Option and Result enum state transitions

### LINKED ISSUE #825: Add conformance coverage for Option and Result enum state transitions
## Summary

Add focused device-side conformance coverage for enum state transitions exercised by:

- `Option::take`
- `Option::replace`
- `Option::get_or_insert`
- `Result::map`
- `Result::map_err`
- `Result::and_then`
- `Result::or_else`

cuda-oxide already has the underlying enum machinery required by these operations. The missing piece is an end-to-end regression that exercises the corresponding `core` APIs directly and verifies their resulting discriminants and payloads on the GPU.

## Motivation

These APIs stress more than simple enum construction.

`Option::take` and `Option::replace` are implemented in terms of `mem::replace`, while `Option::get_or_insert` conditionally constructs a `Some` value and returns a mutable reference to its payload.

The resulting device-side behavior combines:

```text
enum construction
    -> discriminant read
    -> conditional control flow
    -> whole-enum replacement
    -> payload extraction/addressing
    -> payload mutation
````

The selected `Result` combinators similarly exercise both payload-preserving and variant-changing control flow:

```text
Ok  -> Ok
Ok  -> Err
Err -> Err
Err -> Ok
```

Existing enum regressions cover the underlying compiler primitives, including enum layout, construction, discriminant handling, niche encoding, payload access, and payload mutation. There is currently no focused example that drives these standard `Option` and `Result` APIs through those paths together.

## Proposed coverage

Add an `enum_state_transitions` example containing two independent kernels.

### `Option`

Use a runtime-selected initial `Option<u32>` so both `Some` and `None` paths remain observable.

Cover:

#### `Option::take`

Verify:

```text
Some(value) -> None, returns Some(value)
None        -> None, returns None
```

#### `Option::replace`

Verify:

```text
Some(old) -> Some(new), returns Some(old)
None      -> Some(new), returns None
```

#### `Option::get_or_insert`

Verify both cases:

```text
None
  -> insert Some(value)
  -> return &mut payload
  -> mutate payload

Some(existing)
  -> preserve existing payload
  -> return &mut existing payload
  -> mutate payload
```

This also exercises payload addressing through the mutable reference returned by `get_or_insert`.

### `Result`

Use a runtime-selected `Result<u32, u32>` and cover:

* `Result::map`
* `Result::map_err`
* `Result::and_then`
* `Result::or_else`

The test should include both starting variants and explicitly verify variant-changing cases:

```text
Ok  -> Err via and_then
Err -> Ok  via or_else
```

Every intermediate value should be matched and encoded into an integer returned to the host so that the resulting discriminant and payload are observed explicitly.

## Scope

This issue is intended as conformance coverage for already-supported enum operations.

It does not require new enum representation or lowering machinery unless the regression exposes a previously unknown compiler defect.

It intentionally does not expand into:

* niche-layout-specific coverage already handled by existing regressions
* custom enum ABI coverage
* panic or unwrap failure paths
* allocator-backed types
* `Result` combinators unrelated to enum construction or variant transitions

## Acceptance criteria

The example should pass both optimized and low-MIR-optimization device execution:

```bash
cargo oxide run enum_state_transitions

CUDA_OXIDE_NO_OPT=1 \
cargo oxide run enum_state_transitions
```

It should also pass both scoped smoketest modes:

```bash
scripts/smoketest.sh -o '^enum_state_transitions$'

scripts/smoketest.sh \
  --compile-only \
  -o '^enum_state_transitions$'
```

The standalone example should pass:

```bash
cargo fmt --all -- --check
```

Expected runtime output:

```text
=== enum_state_transitions ===
PASS: Option::take
PASS: Option::replace
PASS: Option::get_or_insert
PASS: Result::map
PASS: Result::map_err
PASS: Result::and_then
PASS: Result::or_else
PASS: enum_state_transitions
```

uurl:
Closes #825 

## Summary

Add an `enum_state_transitions` codegen example covering device-side execution of:

- `Option::take`
- `Option::replace`
- `Option::get_or_insert`
- `Result::map`
- `Result::map_err`
- `Result::and_then`
- `Result::or_else`

No compiler changes are required.

The example confirms that the existing enum lowering machinery correctly handles the construction, discriminant reads, payload extraction, replacement, and payload mutation produced by these standard `core` APIs.

## `Option` coverage

The `option_transitions` kernel starts from a runtime-selected `Some` or `None`, preventing the initial variant from being fixed at compile time.

It verifies `Option::take` in both directions:

```text
Some(value) -> None, returns Some(value)
None        -> None, returns None
```

It verifies `Option::replace` with both initial variants:

```text
Some(old) -> Some(new), returns Some(old)
None      -> Some(new), returns None
```

It also exercises `Option::get_or_insert`.

For an initial `None`, the operation must construct a new `Some` payload and return a mutable reference to it.

For an existing `Some`, the supplied replacement value must not overwrite the existing payload.

In both cases the returned `&mut` payload is mutated and the final `Option` is observed afterward, covering payload addressing and mutation as well as the state transition itself.

## `Result` coverage

The `result_combinators` kernel starts from a runtime-selected:

```rust
Result<u32, u32>
```

and exercises:

```text
Result::map
Result::map_err
Result::and_then
Result::or_else
```

Both `Ok` and `Err` starting states are tested.

This gives explicit coverage of:

```text
Ok  -> Ok
Ok  -> Err
Err -> Err
Err -> Ok
```

including the two cross-variant transitions:

```text
Ok  -> Err through Result::and_then
Err -> Ok  through Result::or_else
```

Every intermediate `Option` or `Result` is matched and packed into a host-observable integer. This makes the resulting discriminant and payload part of the regression rather than merely checking that the methods compile.

## Why no compiler change is needed

The compiler already has the enum infrastructure required by these APIs:

* field-faithful enum layout and construction
* direct and niche discriminant handling
* payload extraction
* `Downcast -> Field` payload addressing
* enum payload mutation
* whole-enum rebuilding where canonical storage requires it

The standard-library APIs covered here lower onto those existing MIR paths.

This PR therefore adds focused conformance coverage rather than another enum representation or special-case intrinsic lowering.

## Scope

This PR only adds:

```text
crates/rustc-codegen-cuda/examples/enum_state_transitions/
├── Cargo.lock
├── Cargo.toml
├── README.md
└── src/main.rs
```

No compiler, dialect, exporter, smoketest-script, or Rustlantis sources are changed.

## Validation

Optimized execution:

```bash
cargo oxide run enum_state_transitions
```

passes with:

```text
=== enum_state_transitions ===
PASS: Option::take
PASS: Option::replace
PASS: Option::get_or_insert
PASS: Result::map
PASS: Result::map_err
PASS: Result::and_then
PASS: Result::or_else
PASS: enum_state_transitions
```

Low-MIR-optimization execution also passes:

```bash
CUDA_OXIDE_NO_OPT=1 \
cargo oxide run enum_state_transitions
```

Scoped runtime smoketest:

```text
[ 1/ 1] enum_state_transitions ... PASS

Pass:    1 / 1
Fail:    0 / 1
```

Scoped compile-only smoketest:

```text
[ 1/ 1] enum_state_transitions ... PASS (compiled)

Pass:    1 / 1
Fail:    0 / 1
```

The standalone example also passes:

```bash
cargo fmt --all -- --check
```

Repository-level formatting and diff checks are clean.


nihalpasham:
**Gist:** device-side coverage for seven Option/Result state transitions (take, replace, get_or_insert, map, map_err, and_then, or_else), with runtime seeds keeping everything out of const-folding and collision-free packing making every slot host-checkable. Zero compiler changes needed: the SetDiscriminant path already lowers these, and now it's gated.

- `enum_state_transitions`: all eight PASS lines, optimized and `CUDA_OXIDE_NO_OPT=1` on regenerated PTX;
- all 22 host assertions re-derived by hand, including both launch variants.

Landing once CI is green. Closes #825.

nihalpasham (approved):

PR #833, opened 2026-08-12
refactor(mir-importer): stop exporting the translator, and drop what that hid

vyncint:
The #816 merge note said `emit_warp_redux` "is dead code and the redux family
is catalog-generated". It is, and it is still in the tree — because nothing was
ever going to tell us. `mir-importer`'s `lib.rs` declares `pub mod translator`,
so every item under that module counts as reachable public API and `dead_code`
is switched off for the entire translator subtree.

Nothing in the workspace uses that public surface. `rustc-codegen-cuda`, the
only in-tree consumer, imports `mir_importer::X` root re-exports and never
`mir_importer::translator::*`; no test or bench names the path either. So the
module is now `pub(crate)`, the two predicates outsiders do need stay
re-exported from the root, and the compiler starts answering the question.

## What the lint finds

Six hand-written orphans, all removed. None had a caller.

| Item | Why it is there |
|---|---|
| `intrinsics::warp::emit_warp_redux` | the whole 118-line module — the one #816 named |
| `intrinsics::indexing::emit_index_2d` + `generated_sreg_op` | superseded legacy expansion |
| `helpers::emit_nvvm_intrinsic`, `emit_nvvm_intrinsic_u64` | pre-catalog twins |
| `translator::translate_function` | the crate doc example's entry point |
| `types::get_f32_type` | — |

Two of them are worth a sentence each.

**`emit_warp_redux`** advertised itself as *"shared by the whole integer
reduction family"*. It is shared by nothing, and that sentence is plausibly why
the book's Stage 3 pointed at this file twice — #815 and then 336a70b0.

**`emit_index_2d`** documents its own obsolescence: *"the dispatcher currently
translates every index helper as a normal function call instead of routing it
here"*. That is accurate — `try_dispatch_intrinsic` returns `Ok(None)` for all
sixteen paths in that arm. The semantic warning it carried, that `index_2d_runtime`
must never be flat-expanded, already appears in the module header, so nothing
is lost with it. The `generated_sreg_op` helper went with it as a cascade.

**`emit_nvvm_intrinsic{,_u64}`** are the pre-catalog twins of
`emit_generated_nvvm_intrinsic{,_u}`: same private core,
`emit_nvvm_integer_intrinsic`, differing only in passing `None` where the live
pair passes the ABI marker.

**`translate_function`** was the crate doc example's entry point. The example
passed four arguments to a five-argument function, so it had not been compiled
against reality in a while.

## One test changed rather than deleted

`index_1d_sreg_ops_carry_their_exact_generated_markers` built its three ops by
calling `generated_sreg_op` — production-dead. It therefore could not have
caught a regression in the marker the compiler actually stamps, since the
compiler does not take that path.

It now asserts the same three markers against
`generated_intrinsic_marker_by_op_name`, the table the backend reads back at
verification time. Same intent, but the assertion now spans the two generated
artifacts — a renumbering that reached the dispatcher and not the target table
(or the reverse) fails it.

## The seventh orphan is generated, so it is your call

`generated::generated_intrinsic_marker` is a 2142-line path-keyed table whose
only callers are the generated tests in the same file — and those tests assert
it against `intrinsic_marker(catalog, record)`, which is the same expression
that rendered it, for the same record, in the same loop. It compares the
generator to itself.

I gated it on `cfg(test)` in `render.rs` (one line in the output) rather than
decide that for you. Dropping it outright is one more line in the renderer and
about −10k in `generated.rs`; say the word and I will send it separately.

## Docs that named a removed symbol

- `adding-new-intrinsics.md`'s quick-reference checklist still told new
  contributors to *"Use `helpers::emit_nvvm_intrinsic()` for zero-arg
  intrinsics"* — the one spot the 336a70b0 sweep did not reach. It now points
  at the `typed_swap` arm you made the hand-dispatch example, and says
  zero-operand intrinsics are catalog territory.
- `mir-importer.md` and `translator/mod.rs` both drew a call flow starting at
  `translate_function()`; the pipeline calls `body::translate_body` directly.
- The crate doc example now uses `run_pipeline`.

## Verification

macOS (`aarch64-apple-darwin`):

- `cargo clippy -p mir-importer --all-targets -- -D warnings` — clean
- `cargo test -p mir-importer --no-fail-fast` — 1068 passed, 0 failed
- `cargo test -p cuda-intrinsics-gen --no-fail-fast` — 289 passed, 0 failed
- `cargo run -p cuda-intrinsics-gen -- check` — outputs current
- `cargo fmt --all -- --check` — clean

Linux `x86_64`, Tesla T4 (sm_75), driver 595.71.05, CUDA 13.2, on
`nightly-2026-04-03`:

- `cargo clippy --workspace --all-targets -- -D warnings` — clean, which is the
  run that matters here: it is what would have caught these six in the first
  place, and it is green with the lint no longer suppressed
- `cargo test --workspace --no-fail-fast` — every target ok, 0 failed
- `cargo run -p cuda-intrinsics-gen -- check` — outputs current
- `cargo fmt --all -- --check` — clean
- `sphinx-build -W --keep-going -b html` — build succeeded
- `scripts/smoketest.sh` over the eight examples that exercise what changed —
  `row_width_slice`, `sharedmem`, `swizzle_smem`, `thread_runs`, `tiled_gemm`,
  `vecadd`, `warp_reduce`, `warp_sums` — 8/8 PASS, 261s. The indexing and warp
  ones are deliberate: those are the paths whose hand-written emitters this
  removes.

*(Found by asking why a function the #816 note called dead was still
compiling.)*


nihalpasham:
**Gist:** the translator module stops being `pub`, and what that visibility hid gets deleted: six items with no callers anywhere on main. This is the hygiene rustc and LLVM get from non-pub modules and anonymous namespaces, done Rust's way (reachability-based dead-code detection). On your open question: the `cfg(test)` gate on the path-keyed marker table is fine as-is; the follow-up that deletes the table (~10k generated lines) is welcome when you want it.

**Verified here:**
- every deletion provably dead on origin/main: no caller outside definitions, doc mentions, and the one test the PR rewrites;
- the root re-exports the backend imports remain intact on the PR head; renderer and checked-in generated.rs stay byte-consistent (the intrinsics-gen `check` CI leg gates this);
- ran the six examples over the touched paths on an RTX 5090 (sm_120): warp_reduce, warp_sums, vecadd, thread_runs, tiled_gemm, row_width_slice all pass.

Landing once CI is green. Doesn't close #811; the redux gap it references stays open.

nihalpasham (approved):

PR #834, opened 2026-08-12
build(just): lint and document the codegen workspace, as CI does

vyncint:
`just check` claims to run CI's gates minus a named few. Two of its recipes were
narrower than the jobs they mirror, both in the same way: they miss
`crates/rustc-codegen-cuda`, which carries its own `[workspace]` for the
`rustc_private` dylibs, so `--workspace` from the root stops at that boundary.

The `test` recipe already handles that crate (`cd crates/rustc-codegen-cuda &&
cargo test --lib`). These two never did.

## clippy

CI lints **three** scopes; the recipe ran one:

| scope | CI | `just clippy` (before) |
|---|---|---|
| root workspace | `cargo clippy --workspace --all-targets` | `cargo clippy --all-targets --lib --tests` |
| codegen workspace | `cd crates/rustc-codegen-cuda && cargo clippy --all-targets` | — |
| per example | one run per example | — |

Two problems in that first cell: no `--workspace`, and `--all-targets --lib
--tests` **narrows** target selection, so it skips bin and example targets CI
lints — `cuda-host/examples/finalize_nvvm_ir.rs` among them. Now
`--workspace --all-targets`, matching CI, plus the codegen workspace.

## doc-check

The docs gate builds this workspace's rustdoc separately under `-D warnings`
too. **That job is mine, from #725, and I did not add it here** — same omission
as #814's missing timeout.

## The per-example pass stays CI-only

It is one clippy run per example across 200-odd separate workspaces — a CI job,
not something to wait on locally. `check`'s comment now names it as CI-only
rather than implying clippy is covered in full, which is the part that was
actually misleading.

## Verification

Both widened recipes **pass here**, which was the real risk — widening a gate
can surface new lints:

- `just clippy` — clean across both workspaces, 0 issues, exit 0;
- `just doc-check` — builds the codegen rustdoc warning-free, exit 0.

So this closes a coverage gap without landing new work.

## Merge order

**#835 is stacked on this one.** Both edit the same `Still CI-only:` sentence —
this PR adds clippy's per-example pass to it, #835 removes naming-guard — so they
conflicted until #835 was rebased onto this branch. Merging this one first keeps
that resolution valid; if #835 lands first its squash carries this change too and
this PR becomes an empty duplicate.

#836 and #838 also touch the `Justfile`/manifests but at unrelated lines, and I
verified all five merge clean in either order after the rebase.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -15,9 +15,24 @@ fmt:
 fmt-check:
     cargo oxide fmt --check
 
-# Run clippy with warnings as errors
+# Lint every scope CI's clippy job lints, minus the per-example pass.
+#
+# CI runs clippy three times: over the root workspace, over
+# crates/rustc-codegen-cuda (its own [workspace] for the rustc_private dylibs,
+# which `--workspace` from the root cannot reach), and once per example. The
+# first two are here. `--workspace --all-targets` matches CI exactly; the
+# previous `--all-targets --lib --tests` narrowed target selection and so
+# skipped bin and example targets that CI lints, `cuda-host/examples/` among
+# them.
+#
+# The per-example pass stays CI-only: it is one clippy run per example across
+# 200-odd separate workspaces, which is a CI job rather than something to wait
+# on locally.
+# Run clippy with warnings as errors (root + codegen workspaces)
 clippy:
-    cargo clippy --all-targets --lib --tests -- -D warnings
+    cargo clippy --workspace --all-targets -- -D warnings
+    # Its own [workspace], so `--workspace` above stops at that boundary.
+    cd crates/rustc-codegen-cuda && cargo clippy --all-targets -- -D warnings
 
 # Run clippy and auto-fix warnings
 clippy-fix:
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the two codegen-workspace passes (clippy and warning-free rustdoc, cd-into-workspace) genuinely mirror the dedicated CI steps in clippy.yml and docs.yml and close a real local/CI gap for the one crate outside the root workspace. The comment's rationale was wrong, though: cargo target flags are additive, so the old `--all-targets --lib --tests` never skipped examples, and the virtual root workspace means the bare invocation already covered every member. Pushed f7cf4c3b rewording the block so the Justfile doesn't teach wrong cargo semantics; the real change is CI-spelling parity plus the two new passes.

**Verified here:**
- both mirrored steps quoted line-for-line from clippy.yml and docs.yml;
- `just -n clippy` dry-runs cleanly with the reworded comment;
- `clippy-fix` widening left as a follow-up if you want fix mode to reach the codegen workspace too.

Landing once CI is green (before #835, which stacks this branch).

nihalpasham (approved):

PR #835, opened 2026-08-12
ci: move the reserved-prefix guard into a script

vyncint:
> **Stacked on #834.** Both PRs edit the same `Still CI-only:` sentence in the
> `Justfile` — #834 adds clippy's per-example pass to it, this one removes
> naming-guard from it — so they conflicted. Rebasing resolved it with both edits
> kept. **Review only the second commit** (`cc59cfc8`).
>
> **Please merge #834 first.** If this one lands first, its squash carries #834's
> change too and #834 becomes an empty duplicate.

naming-guard was the **last CI gate with real logic and no script**: a ~50-line
shell block inline in the workflow. So it could not be run before pushing, could
not be mutation-tested, and `just check` listed it as CI-only *for exactly that
reason*.

`scripts/check-reserved-prefixes.sh` is that block, moved verbatim in behaviour —
same pattern, same canary self-test, same grep exit-status handling, same two
filters.

## The reasoning comes with it

That block's comments are hard-won and worth keeping next to the code:

- why bare `cuda_oxide_` is **deliberately not matched** — pliron attribute keys,
  rustc `--cfg` names, dlopen probe symbols and temp directories live in that
  word-space and are not reserved link symbols;
- why `codegen_v1` needs its own alternation arm — the pattern is anchored to the
  opening quote, so a hardcoded modern prefix's embedded `cuda_oxide_kernel_`
  segment sits mid-string where the `kernel` arm can never reach it;
- why a search that cannot run has to **fail** rather than report empty — the
  ripgrep incident, where `rg … || true` turned "command not found" into a clean
  result on every commit without reading a file.

## Verified equivalent, not merely similar

| check | result |
|---|---|
| same verdict as the inline pipeline on this tree | both report no violations |
| injected literal in a fresh file | caught, with file and line |
| pure-comment mention of the prefix | still allowed |
| `reserved-oxide-symbols` and its 45 legitimate literals | still exempt |
| all nine guards after the rebase | green |

## One deliberate change

The script reports `error:` on stderr like its eight siblings, rather than
emitting `::error::` annotations. That keeps its output identical locally and in
CI, at the cost of the PR-annotation nicety — the step still fails and the message
is still in the log. Say the word if you would rather keep the annotations; it is
a one-line difference.

The workflow now calls the script, `check-guards` runs it, and `check`'s CI-only
list drops naming-guard.


nihalpasham:
**Gist:** the last inline CI guard becomes a checked-in script, joining the eight `check-*.sh` siblings that `just check-guards` runs. The extraction preserves the regex, canary self-test, grep exit-status discrimination, and both filters byte-for-byte, and adds fail-closed hardening (missing-grep check, `LC_ALL=C`, cd-to-repo-root).

**Verified here:**
- script output equivalent to the inline block on the current tree; canary still catches a planted `cuda_oxide_kernel_` literal with file:line;
- disclosed delta only: plain stderr `error:` lines instead of `::error::` annotations — fine as-is, the annotation add-back is a one-liner if we ever want inline PR annotations.

Lands after #834 (this branch stacks it); one stale Justfile docstring line will be tidied at landing.

nihalpasham (approved):

PR #837, opened 2026-08-12
docs(contributing): point contributors at `just check`

vyncint:
CONTRIBUTING **never mentions `just`** — zero occurrences. So the one-command
gate that #729 and #788 exist to provide is invisible to exactly the person it
was built for.

Worse, the Formatting and Style section lists the individual commands and says of
clippy that *"there is no single command covering them"*. That is true of that
one scope, but a reader comes away thinking hand-assembly is the only route, and
quietly skips the guards, the unit tests CI actually runs, and the docs gate.

## The change

A short **Running the checks** section ahead of it: `just check`, what it needs
(a CUDA toolkit and `cargo-deny`; **no GPU** — the CUDA-linked test packages
shadow the toolkit's `libcuda` stub themselves), and `just --list` for the
individual recipes.

The hand commands below it stay. They are worth knowing when you want one scope
rather than the whole gate.

## One deliberate omission

It does **not** restate which CI jobs stay outside `just check`. That list is
moving right now — #834, #835 and #836 each shift a piece in or out of it — and a
copy here would rot exactly the way the toolchain pin's copies did (#790). The
`check` recipe's own comment is the single place that names them, so this text
points there instead of duplicating it.

## Verification

Every claim checked against the Justfile rather than remembered: `check` exists,
29 recipes carry the `--list` descriptions, the `cargo-deny` prerequisite and the
libcuda-stub shadowing are both real, `check`'s comment does name the CI-only
jobs, and `just --list` runs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,6 +122,26 @@ The repository includes a `flake.nix` that provides a fully reproducible develop
 environment (CUDA 13, LLVM 22, Clang, pinned Rust nightly). If you have Nix with
 flakes enabled, `nix develop` is the quickest way to get everything in place.
 
+### Running the checks
+
+Most of CI is one command. The repository ships a `Justfile` that mirrors the
+workflows:
+
+```bash
+just check
+```
+
+It needs a CUDA toolkit and `cargo-deny` on `PATH`; it does not need a GPU --
+the CUDA-linked test packages shadow the toolkit's `libcuda` stub when no driver
+is present. Individual recipes exist for each piece, and `just --list` shows
+them with a one-line description each.
+
+A few CI jobs deliberately stay outside `just check` -- ones that need the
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the right pointer for contributors: `just check` runs the local mirror of CI (fmt-check, clippy, tests, guards, doc-check), and the section correctly defers the CI-only jobs to the Justfile comment. One omission fixed (b6b90288): the guards need `python3` on PATH, which check-guards' own prerequisites comment lists but the new prose skipped.

**Verified here:**
- every claim in the section checked against main's Justfile (`check` recipe composition, the libcuda-stub note for driverless machines, the CI-only job list);
- the reworded sentence keeps the file's wrap style.

Landing once CI is green.

nihalpasham (approved):

PR #838, opened 2026-08-12
build: link each crate's README from its manifest

vyncint:
Every workspace crate has had a README since #818, and no manifest pointed at
one. This adds an explicit `readme = "README.md"` key to all twenty-six.

What the key changes: Cargo has auto-inferred a root `README.md` since 1.46,
so crates.io already picked these files up without it (and docs.rs front pages
render the crate docs from `lib.rs`, not the README, either way). What the
explicit key adds is robustness and self-documentation: if a named README is
renamed or deleted, `cargo package` fails loudly instead of silently
publishing a bare page, and tools reading crate metadata get a stated link
rather than an inference. Large workspaces (tokio, serde, clap) write the key
for the same reason.

## Per-crate, not inherited — deliberately

`[workspace.package]` already carries `edition`, `authors`, `license` and
`repository`, and each crate inherits them with `.workspace = true`. `readme`
cannot follow that pattern: Cargo resolves a workspace-level `readme`
**relative to the workspace root**, so inheriting it would point all
twenty-six crates at the top-level README instead of their own. There is no
relative-to-member form, so the key belongs in each manifest — placed next to
`description`, the other per-crate metadata key.

Applied to all twenty-six, including the two `publish = false` crates. The key
is inert there, and a uniform rule is easier to keep than an exception list
that invites the question later.

## Verification

- `cargo metadata --locked` parses; every member reports a readme (as before
  via inference; now stated explicitly);
- every `readme` path points at a file that exists (checked, 0 broken);
- `scripts/check-dependency-licenses.sh`, which reads `cargo metadata`, is
  still green — the one guard this could plausibly have disturbed.

No code, no behaviour change.

*(Maintainer edit: reworded the motivation and first verification bullet — the
original text described READMEs as unreachable without the key, but Cargo
auto-infers a root README.md since 1.46; the key's value is the fail-loud
`cargo package` check and explicitness, not new rendering.)*

### DIFF
--- a/crates/cargo-oxide/Cargo.toml
+++ b/crates/cargo-oxide/Cargo.toml
@@ -6,6 +6,7 @@ authors.workspace = true
 license.workspace = true
 repository.workspace = true
 description = "Cargo subcommand for cuda-oxide: build and run Rust GPU programs"
+readme = "README.md"
 
 [[bin]]
 name = "cargo-oxide"
--- a/crates/cuda-artifact-finalizer/Cargo.toml
+++ b/crates/cuda-artifact-finalizer/Cargo.toml
@@ -3,6 +3,7 @@ name = "cuda-artifact-finalizer"
 version = "0.2.1"
 edition.workspace = true
 description = "Driver-independent NVVM IR and LTOIR finalization for cuda-oxide"
+readme = "README.md"
 authors.workspace = true
 license.workspace = true
 repository.workspace = true
--- a/crates/cuda-async/Cargo.toml
+++ b/crates/cuda-async/Cargo.toml
@@ -3,6 +3,7 @@ name = "cuda-async"
 version = "0.2.1"
 edition.workspace = true
 description = "Async execution layer for CUDA device operations"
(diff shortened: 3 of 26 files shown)


nihalpasham:
**Gist:** right change, wrong story. The explicit `readme` key is worth having: `cargo package` fails loudly if a named README goes missing instead of silently shipping a bare page, and it matches the tokio/serde/clap convention. But Cargo has auto-inferred a root `README.md` since 1.46, so the pages were never blank; I've reworded the PR description so the history records what actually changed. The per-crate placement over workspace inheritance is correctly reasoned (a workspace-level `readme` resolves against the workspace root).

**Verified here:**
- all 26 target manifests exist on main with no prior `readme` key and no `workspace.package.readme`;
- inference behavior confirmed on the pinned cargo (1.98-nightly): metadata reports `README.md` with no key present;
- key placement uniform, next to `description`, including the two `publish = false` crates.

Landing once CI is green.

nihalpasham (approved):

PR #841, opened 2026-08-12
ci: stop the error-example status guard passing on an empty read

vyncint:
## The defect

`check-error-example-status.sh` compares three lists: the `error*` directories on disk, the names in `STATUS.md`, and the names in smoketest.sh's `ERROR_EXAMPLES`. Every comparison iterates one of them, so a list that comes back **empty** makes the loops over it do nothing and the script exits 0 with

```
```

having classified nothing. Changing only the find pattern reproduces it — `error*` to anything matching no directory, with `STATUS.md` and `smoketest.sh` read perfectly well:

```console
$ echo $?
0
```

Each list is one expression away from that: the find pattern, the `sed` reading STATUS.md's first column, and the `sed` reading `ERROR_EXAMPLES=(...)` — which needs the array to open and close on **one line**, and it is already 12 entries long. `set -e` does not see any of it, because the failure is a command substitution producing no output rather than a non-zero status.

This is the only one of the nine `check-*.sh` guards without a parse self-test.

## The change

Require all three lists to be non-empty first, each naming the extraction to fix.

## Correction since I opened this

The first version of this PR also added a reverse check for stale names in `ERROR_EXAMPLES`, and claimed that direction was missing. **That was wrong** — `check-example-smoketest-contract.sh` already rejects any name in any `*_EXAMPLES` array that is not a real example directory, across all thirteen arrays, and its message is better than mine:

```console
$ # plant a stale name in ERROR_EXAMPLES, then run the sibling guard
error: smoketest example contract violated
  ERROR_EXAMPLES names 1 example(s) that do not exist: zzz_deleted
```

I have dropped that half and left a comment recording why this script does not take a second owner for the same condition. What remains is only the self-test, which is genuinely absent today.

## Verified by mutation — each exits 1 with the intended message

| mutation | result |
|---|---|
| find pattern matches nothing | `parse self-test failed: found no error* example directories` |
| STATUS.md path/sed reads nothing | `parse self-test failed: found no names in STATUS.md` |
| `ERROR_EXAMPLES` sed matches nothing | `parse self-test failed: found no names in smoketest.sh ERROR_EXAMPLES` |

A clean tree still passes, and the two pre-existing directions (dropping a `STATUS.md` row, dropping an `ERROR_EXAMPLES` entry) still fire. No GPU needed.

### DIFF
--- a/scripts/check-error-example-status.sh
+++ b/scripts/check-error-example-status.sh
@@ -43,6 +43,38 @@ collect < <(
 )
 in_smoketest=("${COLLECTED[@]+"${COLLECTED[@]}"}")
 
+# Parse self-test, the way this guard's eight siblings open.
+#
+# Every comparison below iterates one of the three lists, so a list that comes
+# back empty makes the loops over it do nothing and the guard exit 0 with its
+# success message -- reporting that everything is classified while having
+# classified nothing. Each extraction is one `find` pattern or one `sed`
+# expression away from that: rename the example prefix, reflow STATUS.md's
+# table, or wrap `ERROR_EXAMPLES=(` across lines, and the corresponding list
+# silently becomes empty. `set -e` does not help, because the failure is a
+# command substitution producing no output rather than a non-zero status.
+#
+# So require all three to be non-empty and fail loudly, naming the extraction to
+# fix, rather than trusting a clean result from an empty read.
+self_test_failed=0
+check_nonempty() {
+    local count="$1" what="$2" how="$3"
+    if [[ "${count}" -eq 0 ]]; then
+        echo "error: parse self-test failed: found no ${what}" >&2
+        echo "       ${how}" >&2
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the silent pass is real: an empty on-disk list makes loop 1 vacuous and loop 2 re-checks directories that exist regardless of what the broken `find` returned, so the guard printed OK having classified nothing. The self-test matches how this guard's eight siblings open, and the deliberate non-duplication of the smoketest reverse check keeps one owner per condition. Pushed 246947c4 narrowing the comment: an empty STATUS.md or ERROR_EXAMPLES read alone did fail before, just per-example and blaming the wrong side.

**Verified here:**
- clean pass on the current tree;
- breaking the `find` name pattern makes the run exit 1 naming the extraction to fix.

nihalpasham (approved):

PR #842, opened 2026-08-12
docs(book): document the two device-debug flags, not just the env override

vyncint:
## The gap

The debug-info section of `error-handling-and-debugging.md` reaches all three modes through `CUDA_OXIDE_DEBUG` and `cargo oxide debug`, and never mentions `--lineinfo` or `--device-debug`. That is backwards from how the CLI presents them — their own help text calls the environment variable the alternative:

```
--lineinfo      ... (nvcc `-lineinfo`). Also settable via CUDA_OXIDE_DEBUG=line
--device-debug  ... (nvcc `-G`). Supersedes --lineinfo. Also settable via CUDA_OXIDE_DEBUG=full
```

Seven subcommands take them — `build`, `run`, `pipeline`, `test`, `sanitize`, `inspect`, `emit-ltoir` — and a reader following this page has no way to learn they exist. The adjacent `bounds-checks.md` already documents the analogous pair both ways (`--unchecked-indexing`   # same thing`), so this page is the outlier.

## The change

Adds the flags to the mode table and beside the override, with the two rules that decide a combination:

- `--device-debug` supersedes `--lineinfo` (from `DeviceDebug::from_flags`);
- omitting both is **not** the same as asking for `off`, because `DeviceDebug::Off` deliberately returns `None` from `env_value` rather than exporting `off` — so an absent flag leaves a level the surrounding environment already set alone, while a present flag exports and wins.

I deliberately did not touch the existing sentence listing which subcommands the *env var* works with. The flags' list is directly checkable from `--help`; the env var's is not, and I have not verified it.

## Verified

Empirically on `vecadd`, which is the equivalence the new text claims:

| build | `.loc` | `.file` |
|---|---|---|
| default | 0 | 0 |
| `--lineinfo` | 21 | 2 |
| `CUDA_OXIDE_DEBUG=line-tables` | 21 | 2 |

Flag availability read from `cargo oxide <sub> --help` per subcommand; precedence and the export rule from `DeviceDebug::from_flags` / `env_value`. `just book` builds warning-free (`build succeeded`, exit 0). No GPU needed — compile only.

### DIFF
--- a/cuda-oxide-book/gpu-programming/error-handling-and-debugging.md
+++ b/cuda-oxide-book/gpu-programming/error-handling-and-debugging.md
@@ -192,8 +192,8 @@ cuda-oxide has three device debug modes:
 | Mode | How to enable it | What you get | Cost |
 |:-----|:-----------------|:-------------|:-----|
 | Off | default for normal `build` / `run` | Fastest generated PTX, no source mapping | none |
-| Line tables | `cargo oxide debug`, or `CUDA_OXIDE_DEBUG=line-tables` | Source breakpoints, stepping, backtraces | low |
-| Full | `CUDA_OXIDE_DEBUG=full cargo oxide debug <example>` | Line tables plus basic argument/local inspection | higher |
+| Line tables | `--lineinfo`, `cargo oxide debug`, or `CUDA_OXIDE_DEBUG=line-tables` | Source breakpoints, stepping, backtraces | low |
+| Full | `--device-debug`, or `CUDA_OXIDE_DEBUG=full` | Line tables plus basic argument/local inspection | higher |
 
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the page documented only the env override while `--lineinfo` and `--device-debug` exist on every subcommand that builds device code; the new text matches the code's actual precedence.

**Verified here:**
- both flags declared on all seven subcommands (Run, Sanitize, Build, Test, EmitLtoir, Pipeline, Inspect) and absent from Debug, as the text says;
- `from_flags` supersede rule and the Off-means-no-export behavior match main.rs, with the two unit tests backing them;
- pre-image matches current main, no drift.

Landing once CI is green.

nihalpasham (approved):

PR #843, opened 2026-08-12
fix(cargo-oxide): name the flag that forced build passthrough mode

vyncint:
## The defect

```
$ cargo oxide build vecadd --device-cfg foo
Error: `cargo oxide build` accepts either an example name or passthrough args after `--`, not both
```

There is no `--` in that command line.

`use_build_passthrough` switches mode for five reasons, and only two of them involve a separator. The other three are ordinary flags that turn passthrough on as a side effect:

- `--cargo-target-dir`
- `--device-codegen-crate`
- `--device-cfg`

All three produced a message naming a cause the user never typed, which sends them looking for a separator instead of at the flag they just added. I hit this while trying to document `--device-cfg`.

## The change

Report the signal that actually fired. A separator is still named first when one is present, since it explains the mode best; the flags follow in the order `use_build_passthrough` tests them.

`test` is unaffected — it is always passthrough and takes no example name, so the conflict cannot arise there.

## Verified against the built binary

| command | message ends with |
|---|---|
| `build vecadd --device-cfg foo` | ``or `--device-cfg`, not both`` |
| `build vecadd --cargo-target-dir /tmp/x` | ``or `--cargo-target-dir`, not both`` |
| `build vecadd --device-codegen-crate bar` | ``or `--device-codegen-crate`, not both`` |
| `build vecadd -- --release` | ``or passthrough args after `--`, not both`` |

All four said the last one before this change.

A unit test pins one signal at a time, the separator-plus-flag combination, and — the part that matters for the future — that **every** signal turning the mode on also yields a name, so a sixth signal added to `use_build_passthrough` cannot silently fall through to the generic wording via the call site's `unwrap_or`.

`cargo test -p cargo-oxide`, `cargo fmt --check` and `clippy -D warnings` all clean. No GPU needed.

### DIFF
--- a/crates/cargo-oxide/src/main.rs
+++ b/crates/cargo-oxide/src/main.rs
@@ -458,6 +458,38 @@ fn use_build_passthrough(
         || has_cargo_args
 }
 
+/// What put `cargo oxide build` into passthrough mode, phrased for a diagnostic.
+///
+/// [`use_build_passthrough`] turns on for five reasons and only two of them are
+/// the `--` separator, so a message that blames `--` is wrong for the other
+/// three: `--cargo-target-dir`, `--device-codegen-crate`, and `--device-cfg` are
+/// ordinary flags that switch modes as a side effect. Naming the wrong cause
+/// sends the reader looking for a separator they never typed.
+///
+/// The order is the reporting order, not a precedence: a separator explains the
+/// mode best when one is present, so it is named first, and the flags follow in
+/// the order [`use_build_passthrough`] tests them. Returns `None` when the
+/// build is not in passthrough mode at all.
+fn build_passthrough_trigger(
+    explicit_separator: bool,
+    cargo_target_dir_is_set: bool,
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the diagnostic blamed `--` for three signals that are ordinary flags, and the helper fixes that by naming each trigger. Pushed 6b846307 taking it one step further: `use_build_passthrough` is now defined as `build_passthrough_trigger(...).is_some()`, so the mode and its name share one source of truth and a new signal has to name itself to compile; the test's mode-vs-name agreement loop became structurally redundant and was dropped.

**Verified here:**
- all four command lines reproduce against the built binary, each naming its own trigger (`--device-cfg`, `--cargo-target-dir`, `--device-codegen-crate`, bare `--`);
- 175 cargo-oxide tests, `cargo fmt --check`, and `clippy -D warnings` all clean on the updated head.

nihalpasham (approved):

PR #844, opened 2026-08-12
ci: cover the device-only kernel build that #702 made possible

vyncint:
Following up on [my comment on #701](https://github.com/NVlabs/cuda-oxide/issues/701#issuecomment-5268104434), where I offered to add this.

## The gap

#702 put the generated host surface behind a default-on `host` feature, so a crate that only compiles kernels can take `cuda-macros = { default-features = false }` and stop pulling cuda-host → cuda-core → cuda-bindings → `cuda.h`. It works today — I verified it before commenting. **Nothing exercises it.** No example takes that configuration, and every CI lane that builds a kernel crate installs the toolkit, so the whole point of the feature was reachable only by a user who tried it.

The regression direction is quiet too. Proc-macro features unify per build graph, so one crate enabling `cuda-macros/host` makes every crate expanding these macros get the host-emitting expansion — including one that asked for `default-features = false` and has no cuda-host to resolve the generated `cuda_host::` paths. cuda-macros' own manifest documents that trap; nothing catches it, and it surfaces downstream as E0433 on a path the user never wrote.

## The change

A fixture crate exercising **both** macro entry points that emit host items — a bare `#[kernel]` (the `CudaKernel` impl) and a `#[cuda_module]` (the loader and launchers) — plus a guard asserting the two properties #701 named:

1. `cuda-host`, `cuda-core` and `cuda-bindings` are absent from the resolved graph. Read from `cargo metadata`, not the manifest, so an *indirect* reacquisition counts.
2. It type-checks with `CUDA_HOME` / `CUDA_TOOLKIT_PATH` / `CUDA_TOOLKIT_TARGET_DIR` unset.

The CI job deliberately installs no toolkit — a toolkit on the runner would mask exactly the dependency the job exists to keep out.

The fixture carries its own `[workspace]` and is not a root member, so the root `cargo metadata` still resolves 25 packages and no existing lane builds it.

## Verified

Clean run reports an 8-package host-free graph. Exits 1 under four mutations:

| mutation | why it must fail |
|---|---|
| `default-features = false` dropped | that is the configuration under test |
| `cuda-host` added back as a dependency | the graph property is the point |
| `#[cuda_module]` removed from the fixture | one of the two host-emitting paths stops being covered |
| fixture made not to compile | the type-check assertion must be live |

All nine fast guards pass on the merged tree, and `just --list` renders with the new recipe wired into `check-guards`. No GPU, no toolkit, no backend build.


nihalpasham:
**Gist:** covers the device-only, no-default-features build that #702 made possible and nothing guarded since. The out-of-workspace fixture is the correct mechanism here: cargo-hack cannot express it because cuda-macros dev-depends on cuda-host, which feature-unification turns back on even under `--no-default-features`.

**Verified here:**
- #702's merge commit is in main's lineage; the fixture's dependency graph type-checks the gate it claims to;
- workflow YAML consistent with the sibling guards, no toolkit or GPU needed.

After this lands, #701 can be closed manually: #702 fixed the behavior, this adds the coverage that was the thread's one live item. Landing once CI is green.

nihalpasham (approved):

PR #845, opened 2026-08-12
docs(book): document --device-cfg, the only way to gate device code

vyncint:
## The gap

`cargo oxide build --help` advertises `--device-cfg`, and the book never mentions it (0 occurrences). That is not a minor omission: it is the **only** mechanism for conditionally compiling device code, because the compiler supplies no target-derived `cfg` and so a kernel cannot ask which architecture it is being built for. Arch requirements live in doc comments and are enforced by the caller — `redux_sum` checks `ctx.compute_capability()` on the host and skips, rather than specializing the kernel.

I found this while fixing its diagnostic in #843.

## The change

A "Conditional compilation" section on the page that already covers the other device-code attributes, with the three things `--help` cannot tell you:

- **It is not device-scoped despite the name.** The flag travels as a rustflag, so every crate cargo compiles for that build sees the `cfg`, host code included.
- **It switches `build` and `test` into passthrough mode**, so it cannot be combined with an example name and has to be run from the crate's own directory. (#843 makes the resulting error say which flag did it.)
- **Nothing ties it to `--arch`.** A `cfg` standing for an architecture is the caller's to keep in step; getting it wrong builds the specialized path for whatever target was selected, and PTX the assembler then rejects.

## Scope

Deliberately not covered: `--device-features`, which reads like a sibling but is applied only in `build_interop_device_crate`, for metadata-declared device crates. That belongs with the interop story, not on this page.

The `### Command reference` section in installation.md is also left alone — it closes with "Every command takes `--help`, which lists the flags each one accepts", so per-flag detail there would fight the file's own approach. This flag earns prose because its behaviour is not visible from `--help`.

## Verified

Both snippets use APIs I checked against the tree — `warp::redux_sync_add(u32::MAX, value)` and a `warp::shuffle_xor` butterfly — and they agree numerically, since wrapping `u32` addition is associative mod 2³² so reduction order cannot change the result.

Worth recording: my first draft used a `FULL_MASK` constant that only exists **locally inside the redux_sum example**, not as a `cuda-device` export, plus a `cooperative_groups::warp_reduce` turbofish I had not checked. A reader pasting either would have hit a name error. Both are gone.

`check-book-api-names.sh` resolves all 34 device calls across the book's 272 Rust blocks; `just book` builds warning-free. No GPU needed.

### DIFF
--- a/cuda-oxide-book/gpu-programming/kernels-and-device-functions.md
+++ b/cuda-oxide-book/gpu-programming/kernels-and-device-functions.md
@@ -198,15 +198,30 @@ let total = {
 };
 ```
 
-Three things are worth knowing before reaching for it.
+Four things are worth knowing before reaching for it.
 
 **It is not limited to device code.** The flag travels as a rustflag, so every
 crate cargo compiles for that build sees the `cfg`, host code included.
 
-**It switches `build` and `test` into passthrough mode.** Passing it means the
-invocation no longer takes an example name, so
+**It switches `build` into passthrough mode.** Passing it means `build` no
+longer takes an example name, so
 `cargo oxide build my_example --device-cfg ampere_up` is rejected; run it from
-the crate's own directory instead.
+the crate's own directory instead. (`test` is passthrough already, flag or no
+flag.) If the gate can live in the crate's manifest, an ordinary Cargo feature
+(`#[cfg(feature = "ampere_up")]`) does the same job without giving up
+example-name invocations; `--device-cfg` earns its keep when you need a `cfg`
+injected without touching any Cargo.toml.
+
+**rustc warns about an undeclared `cfg` name.** The `unexpected_cfgs` lint
+checks every `#[cfg(...)]` against the declared set, and an injected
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** documents a flag the book never named, and the arch-gating story around it is real (the compiler supplies no target-derived cfg). Pushed 30f2b174 with the review fixes: the passthrough note is scoped to `build` (`test` is passthrough unconditionally and never takes an example name), the `unexpected_cfgs` lint now gets a manifest `check-cfg` declaration readers will need, and Cargo features are positioned as the manifest-native alternative, leaving `--device-cfg` as the no-manifest-change injection path.

**Verified here:**
- `Commands::Test` dispatches unconditionally to passthrough (main.rs:710), confirming the scoping fix;
- the redux/shuffle snippet's APIs exist in cuda-device and `scripts/check-book-api-names.sh` passes;
- the lint warning was reproduced under injected rustflags on a matching nightly during review.

nihalpasham (approved):

PR #846, opened 2026-08-12
ci: check the device names the book puts in prose and tables

vyncint:
## The gap, with the receipt

#819 reads only ```rust fenced blocks, and said so on purpose: a call is something a reader pastes, while a name in running text might be a PTX mnemonic or another project's function. That scoping left **half of the very bug it was written for** uncovered.

#815 removed two things from the API quick reference:

```diff
-// i32 variants
-let partner_i = warp::shuffle_xor_i32(val, mask);      <- #819 catches this
...
-| `shuffle_xor_{f32,i32}(val, mask)`    | Exchange with lane `id ^ mask`    |
-| `shuffle_down_{f32,i32}(val, delta)`  | Read from lane `id + delta`       |
-| `shuffle_up_{f32,i32}(val, delta)`    | Read from lane `id - delta`       |   <- it does not
-| `shuffle_{f32,i32}(val, src)`         | Read from specific lane           |
```

Only the first is a call. Reinstating one of those table rows today passes the guard.

## The change

A second pass over everything outside a fenced block, restricted to **backticked, module-qualified** names. Backticks are what make it precise: a backticked `warp::foo` is a claim about this API wherever it appears, so the false positives that motivated the code-block-only scope do not arise.

`_{a,b,c}` families are expanded, so a family with one bad member fails on *that member* rather than passing as an unrecognised literal. The notation is live today in `thread::threadIdx_{x,y,z}()`, and it is exactly what carried #815's rows.

## What it still cannot do

Unqualified table entries stay out of scope. #815's rows as literally written (`shuffle_xor_{f32,i32}`, no `warp::`) would need a section-to-module mapping, and inventing one is guesswork — so this catches them only in the qualified form the book now tends to use. I would rather state that limit than imply the hole is fully closed.

## Verified

Passes clean: 34 calls in 271 Rust blocks, 31 qualified names in prose and tables, against cuda-device's 1161 exported functions. Exits 1 under four mutations:

| mutation | reported |
|---|---|
| #815's table row reinstated, qualified | `warp::shuffle_xor_i32` |
| bad axis in the live shorthand | `thread::threadIdx_q` |
| broken call in a code block (pre-existing pass) | `warp::popc_nope` |
| prose regex rotted | `parse self-test failed: found 0 qualified device names` |

Both passes now self-test for a non-empty result, since the new regex is the likelier of the two to break quietly. One implementation note that cost a debug cycle: the brace group must not require its own leading `_`, because the identifier class is greedy and otherwise swallows the `_` in `threadIdx_{x,y,z}`, leaving the truncated `threadIdx_` reported as missing. No GPU needed.

### DIFF
--- a/scripts/check-book-api-names.sh
+++ b/scripts/check-book-api-names.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
-# Verify every device-API call the book shows in a Rust code block resolves to a
-# function `cuda-device` actually exports.
+# Verify every device-API name the book shows resolves to a function
+# `cuda-device` actually exports.
 #
 # The failure this catches is silent and has now happened three times. #797
 # found the intrinsics guide pointing at op files that no longer existed; the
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** extends the book-API guard from ```rust blocks to prose and tables, which is where the #815 ghost (`warp::shuffle_xor_i32`) actually lived. The lookbehind that rejects `::`-prefixed module paths keeps compiler-internal mentions like `intrinsics::warp::...` out of scope, and the reinstated quick-reference row makes the new mode's value concrete.

**Verified here (reproduced from a scratch extraction of main + this head):**
- clean pass matches the PR body: 34 calls in 271 Rust blocks, 31 qualified prose/table names, resolved against 1161 exports;
- all four mutations fail naming the right symbol, including the reinstated #815 row flagged alone;
- the parse self-test thresholds still refuse to pass on a silent no-op scan.

Landing once CI is green.

nihalpasham (approved):

PR #847, opened 2026-08-12
docs: propagate the Discord invite update past README.md

vyncint:
## The defect

`1b4f7e7d` ("docs: update Discord invite") replaced `discord.gg/Fua7DeKnm` with `discord.gg/ZUEr4AhH5C` — and touched **only README.md**:

```console
$ git show 1b4f7e7d --stat
 README.md | 2 +-
```

Six other references kept the superseded link, including the three places a new contributor is most likely to follow one:

| file | context |
|---|---|
| `CONTRIBUTING.md` | the only contact route the file gives |
| `.github/ISSUE_TEMPLATE/bug_report.md` | "Ask in #help on Discord first" |
| `.github/ISSUE_TEMPLATE/feature_request.md` | "discuss the design before opening a full issue" |
| `cuda-oxide-book/appendix/ecosystem.md` | the book's community pointer |

The badge in README.md keeps the same server id (`1515530041767759993`) across that commit, so this is one server with a refreshed invite rather than a move — and the newer link is the one you chose.

## Scope

This updates four of the six. The fifth, `.github/pull_request_template.md`, is handled by a separate PR that also fixes its testing checklist, so **the two changes do not share a file**. After both, no reference to the old invite remains anywhere in the tree.

I have not tested either invite (that would mean joining), so this rests on which one the commit history says is current rather than on a claim that the old one is dead.

`just book` builds warning-free. No GPU needed.

PR #848, opened 2026-08-12
docs(github): point the PR template's testing checklist at `just check`

vyncint:
## The defect

The PR template asks contributors to tick:

```
- [ ] `cargo test --workspace` passes
```

That is not the gate this repo runs, and it quietly covers less than it looks:

1. **It cannot reach `crates/rustc-codegen-cuda`.** That crate carries its own `[workspace]` for the rustc-private dylibs, so it is not a root member — `cargo metadata` lists 25 packages and the backend is not among them — and neither `--workspace` nor `-p` from the root crosses that boundary. CI covers it in a separate job, and the Justfile's `test` recipe says so in a comment.
2. **It skips eight tests.** `oxide-artifacts` has `default = []`; measured on the current tree, `cargo test -p oxide-artifacts --all-targets` runs **14** tests, and `--features object` runs **22** — the ELF emit/extract set.
3. **On a driverless machine the CUDA-linked packages fail to *load*** with `libcuda.so.1`, which reads as a broken checkout rather than a test failure. `just test-cuda` shadows the toolkit's stub under that name, as `unit-tests.yml` does.

`just check` handles all three, and CONTRIBUTING.md has pointed at it since #837 — so the template was the odd one out.

## The change

The checklist now names `just check`, and keeps `cargo oxide run <example>` with the scoped smoketest as an alternative, since that is what actually proves an example (a `cargo check` never links, and a plain `cargo build` in an example dir dies on the artifact anchor).

Also updates the stale Discord invite on the template's last line to the one `1b4f7e7d` chose for README.md. The other five stale references are a **separate PR**, deliberately scoped so the two do not share a file.

No GPU needed.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,9 +8,10 @@
 
 ## Testing
 
-<!-- How did you verify this? Paste relevant `cargo oxide run` / `cargo test` output, or smoketest results. -->
-- [ ] `cargo oxide run <example>` passes
-- [ ] `cargo test --workspace` passes
+<!-- How did you verify this? Paste relevant `cargo oxide run` / `just check`
+output, or smoketest results. -->
+- [ ] `just check` passes (the local mirror of CI: fmt, clippy, tests, guards, docs)
+- [ ] `cargo oxide run <example>` passes, or `scripts/smoketest.sh -o '^<example>$'`
 - [ ] New example added (if applicable)
 
 ## Checklist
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the template asked contributors to run less than CI does; `just check` is the one command that mirrors the gates, and the smoketest line now shows the anchored single-example form.

**Verified here:**
- `Justfile` `check` recipe mirrors the CI workflows per its own comments (fmt-check, clippy, test, test-cuda, check-guards, doc-check);
- `scripts/smoketest.sh -o '^<example>$'` matches the exact-match idiom at smoketest.sh:188;
- invite swap consistent with #847, no overlap (this file is the fifth of five).

Landing once CI is green.

nihalpasham (approved):

PR #849, opened 2026-08-12
fix(codegen): diagnose a mismatched CUDA_OXIDE_LLVM_LINK, as opt already is

vyncint:
## The asymmetry

`llvm_tools.rs` exists because textual IR is not stable across LLVM majors, and it warns loudly when an explicit `CUDA_OXIDE_OPT` disagrees with the chosen `llc`:

```
warning: LLVM version mismatch between opt and llc:
warning:   CUDA_OXIDE_OPT = ... (LLVM 22)
warning:   llc            = ... (LLVM 21)
warning: mixing majors can produce IR the older tool rejects ...
```

`CUDA_OXIDE_LLVM_LINK` had no such warning. `resolve_sibling_tool` returns an explicit override straight from `probe_runnable`, before any major comparison:

```rust
if let Ok(path) = std::env::var(env_var) {
    return probe_runnable(&path);          // no major check, no diagnostic
}
```

Its own docstring says *"Explicit env var (always used if set)"* one line below *"matched to the same LLVM major"* — the env path is where those two statements disagree, and it is silent.

The hazard is not a lesser one than opt's: `llvm-link` writes the libdevice-linked module that `llc` then reads, so a major split across that pair is exactly the "IR the older tool rejects" case. The book already describes this variable as taking "the same four steps as `opt`".

## The change

Adds the missing diagnostic, worded like opt's and naming both binaries. **Nothing is rejected** — an explicit override stays an override, matching how opt treats one.

Silent by design wherever the comparison cannot be made — variable unset, override not runnable, or either major unparseable — mirroring `choose_opt`'s `(Some, Some)` guard rather than guessing. And absent any override, no same-major `llvm-link` still resolves to `None` silently, because that is *not* an error: the backend's path decision switches to NVVM.

## Verified

The decision is a pure function, so it tests without touching the environment or probing binaries. Six cases pinned one condition at a time, and mutation-tested both ways:

| mutation | result |
|---|---|
| predicate made always silent | `panicked: a major mismatch on an explicit override must be reported` |
| predicate made to warn on a matching pair | `FAILED` |

`cargo test -p cuda-oxide-codegen --lib` 170 pass, `cargo fmt --check`, `clippy -D warnings`, and rustdoc under `-D warnings` all clean; `just book` builds warning-free.

Also updates the module docstring and the book's override table, which described the unset case correctly and said nothing about the set one. No GPU needed.

### DIFF
--- a/crates/cuda-oxide-codegen/src/llvm_tools.rs
+++ b/crates/cuda-oxide-codegen/src/llvm_tools.rs
@@ -33,6 +33,15 @@
 //!    every rejected candidate. The experimental API treats requested
 //!    optimization as strict; the legacy rustc path retains its unoptimized
 //!    fallback.
+//!
+//! `llvm-link` follows the same discovery and the same rule for an explicit
+//! override: `CUDA_OXIDE_LLVM_LINK` is respected, with a diagnostic when its
+//! major differs from the chosen `llc`'s. It writes the libdevice-linked module
+//! that `llc` then reads, so it is exposed to the same cross-major hazard as
+//! `opt`. Where it differs: absent any override, no same-major `llvm-link`
+//! resolves to `None` *silently*, because that is not an error -- the backend's
+//! path decision sees IR-level libdevice linking is unavailable and switches to
+//! the NVVM path.
 
 use crate::options::BackendOptions;
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Gist:** real asymmetry, correctly closed: `resolve_sibling_tool` handed back a `CUDA_OXIDE_LLVM_LINK` override with no major-version check while `choose_opt` warns on the same mismatch. The new pure function mirrors `choose_opt`'s (Some, Some) guard and wires in at a single point.

**Verified here:**
- asymmetry confirmed on main (`llvm_tools.rs:324-326` returns unchecked; `198-213` warns);
- diff context byte-identical to current main, including the book's lowering-pipeline table row;
- unit tests cover both the warn and silent paths.

Landing once CI is green.

nihalpasham (approved):

PR #850, opened 2026-08-12
docs(book): document constant memory, which the book never mentioned

vyncint:
## The gap

The book had **no coverage of constant memory at all**. Across every page:

```console
$ grep -rc '#\[constant\]' cuda-oxide-book/     # 0
$ grep -rc 'ConstantMemory'  cuda-oxide-book/   # 0
$ grep -rci 'constant memory' cuda-oxide-book/  # 0
```

Meanwhile the tree ships the `#[constant]` attribute macro, the `ConstantMemory<T>` wrapper, a generated host setter per constant, and four examples (`constant_memory`, `_simple`, `_coeffs`, `_table`). The memory-hierarchy table listed registers, shared, L1, L2 and global — and skipped the tier in between.

## The change

A section in the memory chapter plus the missing hierarchy row, covering what a reader cannot get from `--help` or a rustdoc page they do not know to open:

- declaration inside `#[cuda_module]`, and the generated setter's naming — `SCALE` → `set_scale`, `TABLE_CONST` → `set_table_const`;
- what `ConstantMemoryValue` admits, **and why**: the all-zero bit pattern must be a valid `T` because `UNINIT` starts zeroed, which is what excludes references and `NonZero*` and makes `#[repr(C)]` a precondition for custom structs;
- **`get()` vs `get_ref()`**, the trap worth writing down. `get()` hands back a value, and a value indexed at runtime spills the whole array to the thread's local depot first — one store per element, in every thread — while `get_ref()` keeps the read in constant space as a single `ld.const`;
- the measurement `constant_memory_table` already carries, including the part that makes constant memory a *choice* rather than an upgrade:

| index | `get()` | `get_ref()` | `const [f32; N]` (global) |
|:--|--:|--:|--:|
| divergent | 16048us | 7616us | **322us** |
| warp-uniform | 16065us | **258us** | 322us |

On a divergent index constant memory is **23.7x behind** a plain array constant, because a warp's divergent reads of a small table coalesce and stay in L1 while the constant cache serves distinct addresses in sequence.

## Also: the attribute table was incomplete

The API quick reference listed 9 of the 11 `#[proc_macro_attribute]` macros. `#[constant]` appeared **nowhere on the page**; `#[cuda_module]` and `#[launch_contract]` only inside another row's prose. The table now matches `cuda-macros`' attribute set exactly, plus `#[unroll]`, which is a statement attribute rather than a proc-macro one:

```
real   : 11  [cluster_launch constant convergent cooperative_launch cuda_module
              device kernel launch_bounds launch_contract pure readonly]
missing: []
```

## Verified

Every claim checked against the tree, not from memory: setter naming and both read paths against the examples, the trait bounds against `constant.rs`, and the numbers against `ConstantMemory::get_ref`'s own rustdoc table. `check-book-api-names.sh` passes and `just book` builds warning-free. No GPU needed.

### DIFF
--- a/cuda-oxide-book/appendix/api-quick-reference.md
+++ b/cuda-oxide-book/appendix/api-quick-reference.md
@@ -41,10 +41,13 @@ fn helper(x: f32) -> f32 { x * x }
 
 | Attribute                                   | Purpose                                                             |
 |:--------------------------------------------|:--------------------------------------------------------------------|
+| `#[cuda_module]`                            | Collect a module's kernels into a typed host module with a `load` and launchers |
 | `#[kernel]`                                 | Mark a function as a GPU kernel entry point (`.entry` in PTX)       |
 | `#[device]`                                 | Mark a helper function or `extern "C"` block for device compilation |
 | `#[unroll]` / `#[unroll(N)]`               | Request full unrolling, or unrolling by a factor `N >= 2`            |
 | `#[launch_bounds(max_threads, min_blocks)]` | Occupancy hints for register allocation                             |
+| `#[constant]`                               | Place a `ConstantMemory<T>` static in constant memory, with a host `set_<name>` |
+| `#[launch_contract(...)]`                   | Declare the launch shape a kernel requires, unlocking a safe (non-`unsafe`) launch |
 | `#[cluster_launch(x, y, z)]`                | Set compile-time cluster dimensions (Hopper+)                       |
 | `#[cooperative_launch]`                     | Launch cooperatively via `#[cuda_module]` (enables `grid::sync()`)  |
 | `#[convergent]`                             | Mark as convergent (barrier semantics)                              |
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Gist:** real gap: the book never mentioned constant memory, and the section's content traces cleanly to the rustdoc and the two examples. Pushed 99cb3da3 with the two accuracy fixes: the get()/get_ref() benchmark table now carries its measurement context (A10G sm_86, 256-entry f32 table, 64 dependent lookups per thread over 8388608 threads), and the setter paragraph now says the macro lowercases the static's name (`to_ascii_lowercase`, not snake case) and documents the `set_<name>_blocking` sibling.

**Verified here:**
- benchmark numbers byte-identical to the `ConstantMemory::get_ref` rustdoc they came from;
- `ConstantMemoryValue` impl list and the zeroed-UNINIT rationale match the code;
- `scripts/check-book-api-names.sh` passes with the section added.

nihalpasham (approved):

PR #855, opened 2026-08-13
feat(mir-importer): support bare struct array constants

### LINKED ISSUE #854: Support bare arrays of struct constants
## Summary

Bare array constants currently support primitive scalars, enums, tuples, and nested arrays, but arrays whose element type is a struct are rejected.

This prevents constants such as:

```rust
#[repr(C)]
struct Entry {
    tag: u8,
    value: u32,
}

const TABLE: [Entry; 2] = [
    Entry { tag: 1, value: 10 },
    Entry { tag: 2, value: 20 },
];
```

from being materialized through the MIR importer when used as bare array values.

## Expected behavior

Bare `[Struct; N]` constants should be materialized using the existing layout-aware aggregate constant decoding path.

The implementation should preserve:

* rustc-recorded struct field offsets,
* padding,
* nested aggregate layout,
* zero-sized and over-aligned struct behavior,
* recursive array/aggregate decoding.

## Scope

This issue covers bare array values only.

The following remain out of scope:

* pointer-to-array constants such as `&[Struct; N]`,
* widening immutable-global array promotion to struct elements,
* union constants.

Bare arrays may fall back to element-wise materialization when the element type is not eligible for immutable-global promotion. Pointer-to-array constants do not have that fallback and should retain their existing boundary.

## Acceptance criteria

* `[Struct; N]` bare constants are accepted.
* Nested struct constants inside arrays are supported.
* Struct padding and field offsets follow rustc layout information.
* Over-aligned zero-sized struct arrays behave correctly.
* Existing pointer-to-array struct rejection remains unchanged.
* Existing array constant behavior does not regress.
* Codegen regression coverage validates the generated LLVM shape for padded and nested struct arrays.

uurl:
Closes #854 

## Summary

Adds support for bare arrays whose elements are structs.

The MIR importer already has layout-aware struct constant decoding for allocation-backed constants and byte-backed aggregate decoding. This change extends the bare-array element boundary and byte-array dispatch so struct elements can reuse those existing paths.

## What changed

- Allow `MirStructType` elements in bare array value constants.
- Extend byte-backed array materialization to dispatch struct elements through the existing aggregate constant decoder.
- Keep immutable-global promotion unchanged.
- Keep pointer-to-array struct constants unsupported.
  - padded `#[repr(C)]` structs,
  - nested structs,
  - over-aligned zero-sized structs.
- Add LLVM code-shape checks for padded and nested struct array storage.

## Scope

This intentionally does not change `promotable_array_element`.

Bare array values can fall back to element-wise materialization, while pointer-to-array constants require the immutable-global path. Widening the promotion predicate here would therefore also widen pointer-to-array semantics, which is outside this change.

Unions are also out of scope.

## Validation

```text
cargo test -p mir-importer
1068 passed; 0 failed

PASS

PASS

PASS
```

The GPU regression covers primitive, enum, tuple, padded/nested/ZST struct, pointer-to-array, and tuple-array static-pointer constants.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -11502,11 +11502,14 @@ mod pointer_array_constant_type_tests {
         );
 
         // Widening the shared predicate to tuples deliberately widens this form
-        // too: the doc contract on `promotable_array_element` is that `TABLE[i]`
-        // and `(&TABLE)[i]` cannot drift apart. Nothing here enumerates fields --
-        // the initializer is rustc's evaluated byte image and the size-agreement
-        // check rejects any layout the dialect reproduces differently -- so a
-        // tuple element travels this path exactly as a scalar does.
+        // too: `promotable_array_element` still gates both the value form's
+        // promotion and this reference form, and only bare-value *admission*
+        // is wider now (a struct table falls back to element-wise
+        // materialization, which this form does not have). Nothing here
+        // enumerates fields -- the initializer is rustc's evaluated byte image
+        // and the size-agreement check rejects any layout the dialect
+        // reproduces differently -- so a tuple element travels this path
+        // the same way a scalar does.
         let tuple_ty: TypeHandle = MirTupleType::get(&mut ctx, vec![u32_ty]).into();
         let tuple_array: TypeHandle = MirArrayType::get(&mut ctx, tuple_ty, 2).into();
         assert!(
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -6246,12 +6246,12 @@ pub(crate) fn emit_promoted_immutable_global(
 /// Whether an array's elements are cheap enough to be worth promoting the whole
 /// array to an immutable global and copying it in.
 ///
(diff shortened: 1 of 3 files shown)


nihalpasham:
**Gist:** the project's own missing-arm protocol, done right: the bare-array gate admits `MirStructType` and the byte-path builder routes struct elements through the layout-aware decoders that already served structs inside tuples and enum payloads, so rustc offsets, padding, ZST stripping, and the relocation audit all apply unchanged. Immutable-global promotion and pointer-to-array stay narrower on purpose, and the rewritten doc contracts say why. Pushed c146b31a fixing one stale test comment that still cited the old cannot-drift-apart phrasing. Closes #854.

**Verified here:**
- ran on an RTX 5090 (sm_120): `array_constants` PASS in both optimization modes, `verify-code-shape.sh` PASS, and the padded layout `alloca [2 x { i8, [3 x i8], i32 }]` present in the generated IR;
- all 1068 mir-importer tests pass;
- pointer-to-array rejection confirmed unchanged (structs still refused there, with the fallback rationale now documented).

The perf follow-up (promoting struct tables to one module-image global, as nvcc emits) is real but correctly out of scope; a tracking issue is being filed. Landing once CI is green.

nihalpasham (approved):

PR #857, opened 2026-08-13
fix(mir-importer): materialize initialized union constants

### LINKED ISSUE #856: Support initialized union constants in the MIR importer
## Problem

Initialized union constants are currently not materialized by the MIR importer.

This affects cases such as:

- direct initialized union constants;
- runtime-indexed arrays of unions (`[U; N]`);
- unions nested inside tuples or structs;
- `MaybeUninit<T>`-like storage patterns that rely on preserving partially initialized bytes.

Runtime union construction already carries the active field explicitly, but constant allocations do not preserve that information. Reconstructing a union by guessing an active field would therefore be incorrect.

## Expected behavior

For pointer-free initialized unions, the importer should materialize the evaluated allocation as its physical byte image and transmute that storage into the corresponding `MirUnionType`.

The materialization should preserve rustc's initialization mask:

- initialized bytes become byte constants;
- uninitialized bytes remain `undef`;
- no active union field is inferred.

This allows the representation to remain faithful to the evaluated constant while preserving inactive/uninitialized union storage.

## Scope

Support:

- direct pointer-free initialized union constants;
- unions nested in tuples and structs;
- runtime-indexed `[U; N]` constants;
- partially initialized storage such as `MaybeUninit<T>`.

Keep unsupported for now:

- pointer-bearing union constants or union allocations containing relocations;
- fat-pointer/provenance-sensitive union constants;
- pointer-to-array union constants such as `&[U; N]`;
- device-global initializer relocations.

## Acceptance criteria

- Pointer-free initialized union constants compile and execute correctly.
- Runtime indexing into `[U; N]` works.
- Union constants nested inside tuples and structs work.
- Uninitialized bytes are preserved as `undef` rather than zero-filled.
- Pointer-bearing/provenance-sensitive union constants fail closed.
- Existing constant-array and relocation behavior remains unchanged.
- The behavior is covered by `array_constants` runtime and code-shape tests.

uurl:
## Summary

Closes #856 

Teach the MIR importer to materialize pointer-free initialized union constants from their evaluated rustc allocation.

Instead of trying to reconstruct an active union field, this change preserves the allocation's physical byte image and transmutes that storage into the corresponding `MirUnionType`.

## Implementation

The importer now:

- recognizes `MirUnionType` in constant translation;
- validates that the Rust layout size matches the MIR union storage size;
- rejects pointer-bearing union types and allocations with overlapping relocations;
- materializes initialized bytes as `u8` constants;
- preserves uninitialized bytes as typed `undef`;
- constructs a `[u8; size]` storage value;
- transmutes that byte image into the exact union type.

Union constants are also admitted as bare array elements, enabling runtime-indexed `[U; N]` constants.

The byte-only constant decoding path intentionally does not support non-ZST unions, because that path has already lost rustc's initialization mask and cannot faithfully distinguish initialized bytes from inactive/uninitialized storage.

## Covered cases

The `array_constants` example now exercises:

- direct initialized unions;
- runtime-indexed arrays of unions;
- unions nested inside tuples;
- unions nested inside structs;
- partially initialized unions;
- `MaybeUninit<T>` arrays.

The supported-features documentation has been updated accordingly.

## Out of scope

- pointer-bearing union constants;
- union allocations containing pointer relocations;
- fat-pointer/provenance-sensitive union constants;
- pointer-to-array union constants such as `&[U; N]`;
- device-global initializer relocations.

Those cases require provenance-aware constant materialization rather than a raw byte-image transmute.

## Validation

- `cargo fmt --all`
- `cargo test -p cargo-oxide` - 174/174 passed
- `cargo clippy -p cargo-oxide --all-targets -- -D warnings`
- `cargo test -p mir-importer --lib` - 1068/1068 passed
- `cargo oxide run array_constants`
- `CUDA_OXIDE_NO_OPT=1 cargo oxide run array_constants`
- `scripts/smoketest.sh --only '^array_constants$' --no-color --keep-logs`
- `bash crates/rustc-codegen-cuda/examples/array_constants/verify-code-shape.sh`
- example-local `cargo fmt -- --check`
- `git diff --check`


nihalpasham:
**Gist:** materializes initialized union constants from rustc's evaluated byte image without guessing an active field: initialized bytes preserved, uninitialized inactive bytes stay `undef`, and the image is transmuted into the layout-exact union type; nested, runtime-indexed, and `MaybeUninit<T>` forms included, pointer relocations inside unions still fail closed. Pushed 1dae97e6 restoring the executable bit on `verify-code-shape.sh`, which the union-lane edit flipped to 644 (smoketest runs it through bash, but the example header documents direct `./` invocation). Closes #856.

**Verified here:**
- ran on an RTX 5090 (sm_120): `array_constants` PASS in both optimization modes, including the partially-initialized-union and MaybeUninit lanes where undef handling matters;
- `./verify-code-shape.sh` works again by direct invocation after the mode fix;

One durability note for the record: correctness of the undef bytes rides on `MirUndefOp` lowering to `llvm.undef`; if the exporter ever migrates undef to poison, this path must be revisited. Landing once CI is green, after #855 (the two touch the same decoder comments and example, and the reconciliation is prepared).

nihalpasham (approved):

PR #859, opened 2026-08-13
fix(mir-lower): address packed fields at rustc byte offsets

### LINKED ISSUE #858: Support packed struct field addressing at rustc byte offsets
## Problem

`mir.field_addr` currently addresses struct and tuple fields through the LLVM
struct slot produced by `build_struct_slot_map`.

That works when LLVM's natural struct layout places the selected slot at the
same byte offset recorded by rustc.

It does not work for packed layouts.

For example:

```rust
#[repr(C, packed)]
struct Packet {
    tag: u8,
    value: u32,
}
````

rustc places `value` at byte offset 1, while the corresponding natural LLVM
struct layout `{ i8, i32 }` places the `i32` field at byte offset 4.

As a result, a field projection such as:

```rust
let value_ptr = core::ptr::addr_of!((*packet).value);
```

can produce an address based on LLVM's slot layout rather than rustc's physical
layout.

This affects packed field accesses even though `read_unaligned` and
`write_unaligned` themselves are already supported.

## Expected behavior

Field addresses must correspond to the exact byte offsets recorded by rustc.

This includes:

* `#[repr(C, packed)]`
* `#[repr(C, packed(N))]`
* fields displaced indirectly by an earlier packed field
* preservation of the alignment provable from the aggregate alignment and
  field offset

Normal layouts whose LLVM slot offset already matches rustc should keep their
existing typed GEP lowering.

## Example

For:

```rust
#[repr(C, packed)]
struct Packet {
    tag: u8,
    value: u32,
}
```

`addr_of!((*packet).value)` must address `base + 1`, not the natural LLVM
`i32` slot offset.

For:

```rust
#[repr(C, packed(2))]
struct Packet {
    tag: u8,
    value: u32,
}
```

the field must be addressed at rustc's byte offset and retain the corresponding
two-byte alignment guarantee.

A trailing field must also use rustc's recorded position, for example:

```rust
#[repr(C, packed)]
struct Packet {
    a: u8,
    b: u32,
    c: u8,
}
```

where `c` is placed at byte offset 5 even though its own natural alignment does
not reveal the earlier layout divergence.

## Acceptance criteria

* `mir.field_addr` uses rustc's exact field byte offset when LLVM's natural
  struct slot offset differs.
* Existing typed struct GEPs remain unchanged when both layouts agree.
* Address alignment remains derived from rustc's aggregate alignment and field
  offset.
* End-to-end `addr_of!` / `addr_of_mut!` with `read_unaligned` /
  `write_unaligned` works for packed fields.
* Regression coverage includes `packed`, `packed(2)`, and accumulated layout
  displacement.

## Out of scope

This issue is limited to runtime field addressing.

It does not attempt to add full packed aggregate support for:

* by-value aggregate representation
* kernel or device-function ABI
* initialized statics/constants
* packed relocation slots

uurl:
Fixes #858 

## Summary

Fix `mir.field_addr` lowering when rustc's recorded field offset differs from
the byte offset implied by the lowered LLVM struct layout.

Packed Rust layouts can place fields at offsets that cannot be represented by a
natural LLVM struct GEP. For example, rustc places the `u32` field below at
byte offset 1:

```rust
#[repr(C, packed)]
struct Packet {
    tag: u8,
    value: u32,
}
```

while the natural LLVM layout `{ i8, i32 }` places the `i32` at byte offset 4.

## Implementation

The field-address path now compares:

* the byte offset recorded by rustc, and
* the byte offset of the selected slot in the lowered LLVM struct.

When both offsets match, lowering keeps the existing typed struct GEP.

When they differ, lowering emits an `i8` byte-addressed GEP from the original
aggregate pointer using rustc's exact field offset.

This keeps the normal-layout path unchanged while making packed projections
byte-faithful.

The existing field-address alignment propagation is retained, so the resulting
pointer still carries the alignment provable from rustc's aggregate alignment
and field offset.

## Regression coverage

Adds `mir-lower` tests for:

* `#[repr(C, packed)]`: `u32` field at byte offset 1 with alignment 1
* `#[repr(C, packed(2))]`: `u32` field at byte offset 2 with alignment 2
* accumulated layout divergence:

  ```rust
  #[repr(C, packed)]
  struct Packet {
      a: u8,
      b: u32,
      c: u8,
  }
  ```

  where `c` must be addressed at rustc byte offset 5 rather than the offset
  produced by LLVM's natural struct layout

The `unaligned_memory` example is also extended to cover packed field
projections through:

* `addr_of!`
* `addr_of_mut!`
* `read_unaligned`
* `write_unaligned`

## Validation

```text
cargo fmt --all

cargo test -p mir-lower
  204 unit tests passed
  105 lowering integration tests passed
  0 failures

cargo oxide run unaligned_memory
CUDA_OXIDE_NO_OPT=1 cargo oxide run unaligned_memory

scripts/smoketest.sh -o '^unaligned_memory$'
scripts/smoketest.sh --compile-only -o '^unaligned_memory$'
```

End-to-end output includes:

```text
PASS: repr(packed) field read at rustc offset 1
PASS: repr(packed) field write at rustc offset 1
PASS: repr(packed(2)) field read at rustc offset 2
```

## Scope

This PR fixes runtime packed field addressing only.

It does not change packed aggregate by-value representation, ABI handling,
initialized globals, or packed relocation slots.


nihalpasham:
**Gist:** fixes a real miscompile: a packed field's typed struct GEP landed at LLVM's natural offset (byte 4) instead of rustc's (byte 1), silently corrupting packed reads and writes. The byte-offset fallback with rustc-provable alignment is the right mechanism and matches how rustc itself emits field addresses. Review found the missing half: by-value packed aggregates still used the natural image, so a locally constructed packed struct mixed with the new addresses was a silent-wrong-answer path, and `[Packed; N]` element addressing strided at 8 instead of 5. Pushed 8e05018e closing all of it:

- `build_struct_slot_map` records each slot's natural offset plus a `layout_diverges` fact from the same walk `natural_struct_layout` uses (one layout oracle); `field_addr` reads the map instead of re-walking;
- construction, whole-value load, and whole-value store of a divergent struct fail closed, naming the `addr_of!` + `read_unaligned`/`write_unaligned` workaround;
- packed array element addressing fails closed on stride divergence;
- load/store alignment claims take the weaker of the type's abi alignment and the address's proved alignment, so a naturally aligned inner struct at packed offset 1 is accessed `align 1`, not `align 4`.

**Verified here:**
- ran on an RTX 5090 (sm_120): all unaligned_memory lanes PASS in both optimization modes, including the three new packed projections;
- 209 mir-lower tests pass (the PR's three packed tests plus five new ones covering the four refusals and the alignment narrowing);
- supported-features appendix documents the by-value boundary.

The endgame (packed LLVM struct types upstream in pliron-llvm, then byte-faithful by-value support and rustc-stride element addressing) is being filed as a tracked follow-up. Closes #858. Landing once CI is green.

nihalpasham (approved):

PR #861, opened 2026-08-13
fix(mir-lower): scalarize repr(transparent) kernel parameters

### LINKED ISSUE #860: Kernel ABI should scalarize #[repr(transparent)] parameters
## Problem

Kernel entry parameters whose Rust type is `#[repr(transparent)]` are currently
lowered as aggregate parameters even when rustc classifies their ABI as scalar.

For example, a wrapper such as:

```rust
#[repr(transparent)]
struct Scalar(u32);
```

is ABI-equivalent to `u32`, but the generated PTX currently uses an aggregate
parameter shape such as:

```text
.param .align 4 .b8 [4]
```

instead of the scalar form:

```text
.param .u32
```

The same issue applies to transparent wrappers around pointers and to
transparent newtypes containing zero-sized marker fields.

A concrete existing case is `cuda_device::Uniform<T>`, which is
`#[repr(transparent)]` and is intended to have the same ABI as its scalar
payload.

## Expected behavior

At a kernel boundary, a `#[repr(transparent)]` struct should use its underlying
scalar ABI when rustc reports the type as `ValueAbi::Scalar`.

Examples:

```text
#[repr(transparent)] struct Scalar(u32)
    -> scalar u32 kernel parameter

#[repr(transparent)] struct Pointer(*const u32)
    -> scalar pointer kernel parameter

#[repr(transparent)] struct Marked(u32, PhantomData<()>)
    -> scalar u32 kernel parameter

nested repr(transparent) wrappers
    -> recursively reach the underlying scalar ABI
```

Ordinary one-field structs must remain aggregates.

## Requirements

* Preserve rustc's ABI classification rather than inferring scalar ABI from
  field count alone.
* Only scalarize types that are both:

  * `#[repr(transparent)]`, and
  * classified by rustc as `ValueAbi::Scalar`.
* Ignore zero-sized marker fields when identifying the represented field.
* Support nested transparent wrappers.
* Reconstruct the Rust wrapper value inside the kernel entry body.
* Leave ordinary aggregate kernel parameters unchanged.
* Leave internal device-function ABI behavior unchanged.
* Fail closed for inconsistent transparent-scalar layouts.

## Regression coverage

Add end-to-end coverage for:

* transparent `u32`
* transparent pointer
* transparent scalar plus ZST marker
* nested transparent wrappers
* `Uniform<u32>`
* ordinary one-field struct as a negative control

The regression should verify both runtime behavior and emitted PTX parameter
shapes, including with `CUDA_OXIDE_NO_OPT=1`.

uurl:
Closes #860 

## Summary

Teach kernel ABI lowering to preserve rustc's scalar ABI for
`#[repr(transparent)]` structs.

Transparent wrappers whose rustc layout is `ValueAbi::Scalar` are now lowered
as their underlying scalar or pointer kernel parameter instead of an aggregate
`.param .b8[...]`.

Ordinary one-field structs remain aggregates.

## Implementation

The change carries explicit ABI information from rustc through the MIR dialect
and uses it at the kernel boundary.

- Add an ABI classification to `MirStructType`.
- Mark a struct as transparent-scalar only when:
  - its Rust representation is `#[repr(transparent)]`, and
  - rustc reports `ValueAbi::Scalar`.
- Scalarize transparent kernel parameters during MIR-to-LLVM function type
  conversion.
- Reconstruct the original Rust wrapper in the kernel entry prologue.
- Recursively support nested transparent wrappers.
- Ignore zero-sized marker fields when locating the represented field.
- Fail closed if a transparent-scalar type does not contain exactly one
  non-ZST represented field.
- Keep ordinary kernel structs unchanged.
- Keep internal device-function struct ABI behavior unchanged.

This deliberately does not infer transparent ABI from "one field"; the
classification comes from rustc's representation and ABI metadata.

## Regression coverage

Adds the standalone `repr_transparent_abi` example covering:

- `#[repr(transparent)]` over `u32`
- transparent pointer wrappers
- transparent wrappers with a ZST marker
- nested transparent wrappers
- `Uniform<u32>`
- an ordinary one-field struct as a negative control

The example checks runtime values and verifies the emitted PTX parameter
shapes.

## Validation

```text
cargo fmt --check
PASS

cargo clippy -p dialect-mir -p mir-importer -p mir-lower \
  --all-targets -- -D warnings
PASS

cargo clippy \
  --manifest-path crates/rustc-codegen-cuda/examples/repr_transparent_abi/Cargo.toml \
  --all-targets -- -D warnings
PASS
```

Optimized execution:

```text
cargo oxide run repr_transparent_abi
repr_transparent_abi: PASS (runtime values and PTX parameter shapes)
```

Explicit PTX verification:

```text
cargo oxide run repr_transparent_abi -- --verify-ptx
repr_transparent_abi: PASS (PTX parameter shapes)
```

Low-MIR-optimization execution:

```text
CUDA_OXIDE_NO_OPT=1 cargo oxide run repr_transparent_abi
repr_transparent_abi: PASS (runtime values and PTX parameter shapes)

CUDA_OXIDE_NO_OPT=1 cargo oxide run repr_transparent_abi -- --verify-ptx
repr_transparent_abi: PASS (PTX parameter shapes)
```

Scoped smoketest:

```text
scripts/smoketest.sh --only '^repr_transparent_abi$' --no-color

[ 1/ 1] repr_transparent_abi ... PASS

Pass: 1 / 1
Fail: 0 / 1
```

Repository hygiene:

```text
git diff --check
PASS
```

## Scope

This PR changes kernel-entry parameter ABI lowering only.

It does not change:

* ordinary aggregate kernel parameters
* internal device-function struct flattening
* transparent return-value ABI
* host scalar marshalling


nihalpasham:
**Gist:** transparent kernel parameters now honor rustc's own ABI classification: the importer records `repr(transparent)` + `ValueAbi::Scalar` as a first-class `StructAbiKind` on `MirStructType`, the kernel boundary emits the underlying scalar or pointer parameter, and the entry prologue rebuilds the wrapper. The classification is never inferred from field count, so ordinary one-field structs stay aggregates; ZST markers are skipped, nested wrappers unwrap recursively, and malformed inputs fail closed at both the signature and the prologue. Not a miscompile (the driver copies identical bytes into `.b8[4]` and `.u32` params), but the PTX prototype contradicted the transparent-ABI promise and `Uniform<T>`'s documented guarantee, which matters to anything that reads the kernel signature. Closes #860.

**Verified here:**
- ran on an RTX 5090 (sm_120): `repr_transparent_abi` PASS in both optimization modes; the PTX shows `.param .u32` / `.u64 .ptr` for the transparent kernels while the ordinary-struct negative control keeps its aggregate `.param .b8[4]`;
- `vecadd` and `copy_aggregate_borrow` (existing by-value struct kernel param) still pass, so the ordinary-aggregate ABI is untouched;
- full dialect-mir, mir-importer (1068), and mir-lower (215 + 105) suites green.

Landing once CI is green.

nihalpasham (approved):

PR #863, opened 2026-08-13
test(codegen): cover strict and exposed pointer provenance APIs

### LINKED ISSUE #862: test(codegen): add strict-provenance pointer API coverage
## Problem

CUDA-Oxide currently has lowering support for pointer/address conversions and
wrapping pointer arithmetic, but the regression suite does not explicitly cover
Rust's strict- and exposed-provenance pointer APIs at the source level.

The missing coverage includes:

- `ptr::with_addr`
- `ptr::map_addr`
- `ptr::expose_provenance`
- `ptr::with_exposed_provenance`

These APIs are relevant to allocator, tagging, and low-level pointer manipulation
code where an address is transformed or exposed and then used to reconstruct a
pointer.

## Expected behavior

A CUDA kernel should be able to:

1. Obtain and manipulate a pointer address with strict-provenance APIs.
2. Reconstruct a pointer with the expected provenance semantics.
3. Expose a pointer address and reconstruct it through the exposed-provenance API.
4. Successfully dereference the resulting pointer when the address remains within
   the original allocation.

## Suggested coverage

Extend the existing `ptr_offset_from` regression example with runtime checks for:

- `with_addr`
- `map_addr`
- `expose_provenance`
- `with_exposed_provenance`

The regression should exercise the complete source-level path through MIR lowering
and CUDA execution rather than only checking individual lowering operations.

## Acceptance criteria

- `with_addr` can retarget a pointer to another element in the same allocation and
  the resulting pointer dereferences correctly.
- `map_addr` can transform the address and the resulting pointer dereferences
  correctly.
- `expose_provenance` produces the expected exposed address.
- `with_exposed_provenance` reconstructs a usable pointer from that address.
- The regression passes in both optimized and no-opt codegen paths.
- The focused `ptr_offset_from` smoketest passes.

uurl:
Closes #862 

## Summary

Extend the `ptr_offset_from` regression example with coverage for Rust's strict
and exposed pointer provenance APIs.

The new checks cover:

- `with_addr`
- `map_addr`
- `expose_provenance`
- `with_exposed_provenance`

The tests verify both address manipulation and successful dereference after
pointer reconstruction.

This is regression coverage only; no compiler code is changed.

## Validation

- `cargo oxide run ptr_offset_from`
- `CUDA_OXIDE_NO_OPT=1 cargo oxide run ptr_offset_from`
- `cargo fmt --check`
- `cargo clippy --manifest-path crates/rustc-codegen-cuda/examples/ptr_offset_from/Cargo.toml --all-targets -- -D warnings`
- `scripts/smoketest.sh --only '^ptr_offset_from$' --no-color`

All checks pass.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/ptr_offset_from/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/ptr_offset_from/src/main.rs
@@ -1,14 +1,17 @@
 // SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Regression coverage for raw pointer distance and wrapping-offset intrinsics.
+//! Regression coverage for raw pointer distance, wrapping-offset, and provenance APIs.
 //!
 //! The stable raw pointer methods lower through rustc intrinsics in MIR:
 //! `offset_from_unsigned` uses `ptr_offset_from_unsigned`, while
 //! `offset_from` and the byte-oriented wrappers use `ptr_offset_from`.
-//! Safe `wrapping_offset`/`wrapping_add` use `arith_offset`. cuda-oxide must
-//! preserve signed distances, scale offsets by the pointee type, retain pointer
-//! mutability through libcore's cast, and leave zero-sized pointees stationary.
+//! Safe `wrapping_offset`/`wrapping_add` use `arith_offset`. Strict-provenance
+//! `addr`, `with_addr`, and `map_addr`, plus exposed-provenance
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** test-only, and coverage: extends `ptr_offset_from` with the strict/exposed provenance quartet (`with_addr`, `map_addr`, `expose_provenance`, `with_exposed_provenance`), and every reconstructed pointer is dereferenced afterward, so the checks prove usability rather than just address math. The lowering surface already existed; this makes it part of the executable regression suite. Closes #862.

**Verified here:**
- ran on an RTX 5090 (sm_120): `ptr_offset_from` SUCCESS in both optimization modes;

Landing once CI is green.

nihalpasham (approved):

PR #865, opened 2026-08-13
test(codegen): cover nested place projection composition

### LINKED ISSUE #864: Add regression coverage for nested place projection composition
## Summary

The MIR importer already supports the individual place projections involved in several nested access patterns, but the existing examples do not explicitly pin their composition when another runtime `Index` follows an already-projected place.

Two useful gaps are:

```text
Index -> Field -> Index
Downcast -> Field -> Index
```

Both currently lower and execute correctly, so this is a regression-coverage gap rather than a compiler bug.

## Existing coverage

The repository already exercises shorter projection chains such as:

```text
Index -> Field
ConstantIndex -> Field
Deref -> Field -> Index
Downcast -> Field
```

What is not explicitly covered is continuing those paths through another runtime index.

### `index_field_assign`

Extend the existing example with a local array of structs containing an array field:

```rust
struct NestedCell {
    values: [u64; 4],
}
```

Exercise a runtime place of the form:

```rust
cells[row].values[column]
```

through:

```text
store
load
shared reference
raw address
```

This pins the 3-level:

```text
Index -> Field -> Index
```

composition across both value and address-based consumers.

### `enum_payload_addr`

Add an enum variant containing an array payload:

```rust
enum Bucket {
    Data([u32; 4]),
    Empty,
}
```

and mutate a runtime-indexed payload element through pattern matching:

```rust
if let Bucket::Data(values) = &mut bucket {
    values[column] = value;
}
```

The optimized imported MIR for the store resolves the enum payload address first and then applies the runtime array element address, exercising the effective:

```text
Downcast -> Field -> Index
```

composition.

## Acceptance criteria

* `Index -> Field -> Index` is covered by an executable GPU regression.
* `Downcast -> Field -> Index` is covered by an executable GPU regression.
* Both examples pass with normal MIR optimization.
* Both examples pass with `CUDA_OXIDE_NO_OPT=1`.
* No compiler changes are required if the current place/address walkers already compose these projections correctly.

uurl:
Closes #864 

## Summary

Add explicit regression coverage for nested MIR place projections that are already supported individually but were not pinned in composition:

```text
Index -> Field -> Index
Downcast -> Field -> Index
```

No compiler code changes are required. The existing place/address lowering already handles both chains correctly; this PR makes that behavior part of the executable regression suite.

### `index_field_assign`

Extend the existing `Index -> Field` regression with a local array of:

```rust
struct NestedCell {
    values: [u64; 4],
}
```

and exercise:

```rust
cells[row].values[column]
```

with runtime `row` and `column` values.

The new kernel covers the same 3-level place as:

```text
assignment destination
direct load
shared reference
raw address via addr_of!
```

This pins:

```text
Index -> Field -> Index
```

for both value and address-based consumers.

The host verifies every thread against an exact oracle.

### `enum_payload_addr`

Add:

```rust
enum Bucket {
    Data([u32; 4]),
    Empty,
}
```

and a kernel that mutates an indexed element of the `Data` payload:

```rust
if let Bucket::Data(values) = &mut bucket {
    values[column] = base.wrapping_add(i as u32);
}
```

The imported optimized MIR resolves the payload address directly from the enum and then applies the runtime array index before storing:

```text
enum storage
    -> payload field address
    -> runtime array element address
    -> store
```

This provides regression coverage for the effective:

```text
Downcast -> Field -> Index
```

store path.

The host verifies all `1 << 20` elements exactly.

## Scope

Modified:

```text
crates/rustc-codegen-cuda/examples/index_field_assign/src/main.rs
crates/rustc-codegen-cuda/examples/enum_payload_addr/src/main.rs
```

No changes to:

```text
mir-importer
mir-lower
dialect-mir
rustc-codegen-cuda compiler infrastructure
Cargo manifests
smoketest infrastructure
```

## Validation

Example hygiene:

```bash
(
  cd crates/rustc-codegen-cuda/examples/index_field_assign
  cargo fmt --all -- --check
)

(
  cd crates/rustc-codegen-cuda/examples/enum_payload_addr
  cargo fmt --all -- --check
)

git diff --check
```

All pass cleanly.

Optimized execution:

```bash
cargo oxide run index_field_assign
cargo oxide run enum_payload_addr
```

Low-MIR-optimization execution:

```bash
CUDA_OXIDE_NO_OPT=1 cargo oxide run index_field_assign
CUDA_OXIDE_NO_OPT=1 cargo oxide run enum_payload_addr
```

Both modes pass.

`index_field_assign` reports:

```text
PASS: index_field_assign (Index->Field + ConstantIndex->Field + Index->Field->Index)
```

`enum_payload_addr` includes:

```text
mutate_indexed_payload: 1048576 indexed enum payload elements mutated, exact match
```

and completes with:

```text
SUCCESS
```

The `Downcast -> Field -> Index` probe was also inspected with:

```bash
CUDA_OXIDE_DUMP_MIR=1 cargo oxide run enum_payload_addr
```

The imported MIR for the indexed payload store composes the enum payload field address with a runtime array element address before the final store.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/enum_payload_addr/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/enum_payload_addr/src/main.rs
@@ -21,6 +21,8 @@
 //! - `shared_bytes_no_slot` mutates the variant WITHOUT a slot of its own
 //!   (`Bits` shares `Real`'s bytes), so the byte-offset addressing path runs
 //!   against the original storage.
+//! - `mutate_indexed_payload` mutates a runtime-indexed element inside an enum
+//!   array payload, covering the composed `Downcast -> Field -> Index` store path.
 //! - `shared_borrow_bool_payload` takes a SHARED borrow of a `bool` payload
 //!   and passes it to a `#[device]` helper. A bool payload's bytes use
 //!   canonical i8 storage, so no raw payload address can be handed out; the
@@ -69,6 +71,13 @@ mod kernels {
         Both(bool, f32),
         #[allow(dead_code)]
         Neither,
+    }
+
+    /// Array payload used to cover a nested `Downcast -> Field -> Index` store.
+    pub enum Bucket {
+        Data([u32; 4]),
+        #[allow(dead_code)]
+        Empty,
     }
 
     /// Scale a borrowed payload. Taking `&mut f32` across a call boundary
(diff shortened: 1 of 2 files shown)


nihalpasham:
**Gist:** test-only coverage for composed place projections that were supported individually but never checked in composition: `Index -> Field -> Index` in `index_field_assign` (runtime row and column) and `Downcast -> Field -> Index` in `enum_payload_addr`. No compiler surface touched; none needed. The DST slice-tail chains (#870/#872) are related but separate and close via #873. Closes #864.

**Verified here:**
- ran on an RTX 5090 (sm_120): `index_field_assign` and `enum_payload_addr` PASS in both optimization modes;

Landing once CI is green.

nihalpasham (approved):

PR #867, opened 2026-08-13
test(codegen): cover address-space 3 aggregate device-call round trips

### LINKED ISSUE #866: test(codegen): add AS3 aggregate device-call round-trip coverage
## Summary

Add explicit regression coverage for shared-memory address-space (`addrspace(3)`) pointers crossing internal device-call boundaries inside aggregate values.

The existing `addressof_sharedarray` example already covers shared pointers stored in enum payloads, including nested structs and bounded arrays. What is currently missing is an explicit end-to-end test that passes aggregates containing shared pointers through a non-inlined device function and returns them back to the caller before using the recovered pointer as shared memory.

## Motivation

A shared pointer embedded in an aggregate exercises several lowering boundaries:

1. construct an aggregate containing an AS3 pointer,
2. pass the aggregate through a device-call ABI boundary,
3. reconstruct/use the value in the callee,
4. return the aggregate,
5. project the pointer from the returned aggregate,
6. use the pointer again as shared memory.

This should be covered explicitly for the main aggregate shapes supported by the current lowering.

## Proposed coverage

Extend:

`crates/rustc-codegen-cuda/examples/addressof_sharedarray/src/main.rs`

with `#[inline(never)] #[device]` round-trip helpers for:

- struct
- tuple
- array
- enum

Each returned aggregate should be validated by actually dereferencing or storing through the recovered shared pointer, rather than checking only its integer address.

Using `#[inline(never)]` ensures the device-call ABI boundary remains observable.

## Acceptance criteria

- [ ] Struct containing an AS3 pointer survives device call + return.
- [ ] Tuple containing an AS3 pointer survives device call + return.
- [ ] Array containing AS3 pointers survives device call + return.
- [ ] Enum containing a nested AS3 pointer survives device call + return.
- [ ] Returned pointers are used through shared-memory accesses.
- [ ] Existing `addressof_sharedarray` coverage continues to pass.
- [ ] Test passes with normal optimization.
- [ ] Test passes with `CUDA_OXIDE_NO_OPT=1`.

## Validation

```bash
cargo fmt --check

cargo clippy \
  -p dialect-mir \
  -p mir-importer \
  -p mir-lower \
  --all-targets -- -D warnings

CUDA_OXIDE_NO_OPT=1 cargo oxide run addressof_sharedarray
```

uurl:
Closes #866 

## Summary

Adds explicit regression coverage for shared-memory (`addrspace(3)`) pointers crossing device-call boundaries inside aggregate values.

The test extends `addressof_sharedarray` with non-inlined device helpers that round-trip aggregates containing shared pointers through both argument and return ABIs.

Covered aggregate shapes:

- struct
- tuple
- array
- enum

## What this tests

Each case exercises the following path:

```text
AS3 pointer
    ↓
aggregate construction
    ↓
device-call argument ABI
    ↓
callee
    ↓
aggregate return ABI
    ↓
aggregate projection
    ↓
shared-memory access
```

The returned pointer is used for an actual shared-memory operation, so the test verifies that the pointer remains usable as an AS3 pointer after the complete round trip.

`#[inline(never)]` is used on the helper functions to preserve the call boundary.

## Result

All four aggregate forms already work with the current lowering, so this PR contains no compiler changes. It only adds regression coverage for the existing behavior.

## Validation

```bash
cargo fmt --check
```

PASS.

```bash
cargo clippy \
  -p dialect-mir \
  -p mir-importer \
  -p mir-lower \
  --all-targets -- -D warnings
```

PASS.

```bash
```

PASS.

```bash
CUDA_OXIDE_NO_OPT=1 cargo oxide run addressof_sharedarray
```

PASS.

Runtime coverage confirms that struct, tuple, array, and enum device-call round trips preserve shared pointers correctly.

The existing raw shared-array receiver regression also continues to pass.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/addressof_sharedarray/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/addressof_sharedarray/src/main.rs
@@ -142,17 +142,13 @@ mod kernels {
 
     #[inline(never)]
     #[device]
-    fn bounce_tuple(
-        value: (*mut SharedArray<f32, 1>, usize),
-    ) -> (*mut SharedArray<f32, 1>, usize) {
+    fn bounce_tuple(value: (*mut SharedArray<f32, 1>, usize)) -> (*mut SharedArray<f32, 1>, usize) {
         value
     }
 
     #[inline(never)]
     #[device]
-    fn bounce_array(
-        value: [*mut SharedArray<f32, 1>; 2],
-    ) -> [*mut SharedArray<f32, 1>; 2] {
+    fn bounce_array(value: [*mut SharedArray<f32, 1>; 2]) -> [*mut SharedArray<f32, 1>; 2] {
         value
     }
 
@@ -266,15 +262,14 @@ mod kernels {
                     0.0
                 };
 
-                let enum_round_trip = bounce_enum(SharedPointerPayload::Pointer(
-                    SharedPointerOuter {
+                let enum_round_trip =
+                    bounce_enum(SharedPointerPayload::Pointer(SharedPointerOuter {
                         inner: SharedPointerInner {
                             pointer: raw,
                             cookie: ENUM_COOKIE,
                         },
                         guard: ENUM_GUARD,
-                    },
-                ));
+                    }));
                 *out.get_unchecked_mut(9) = match enum_round_trip {
                     SharedPointerPayload::Pointer(extracted)
                         if extracted.inner.cookie == ENUM_COOKIE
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** test-only, and it checks the part that matters: shared-memory (`addrspace(3)`) pointers packed inside struct, tuple, array, and enum aggregates round-trip through `#[inline(never)]` device-call argument and return ABIs, and each returned pointer is then used for a real shared-memory access, so usability after the round trip is proven rather than assumed. Closes #866.

**Verified here:**
- ran on an RTX 5090 (sm_120): `addressof_sharedarray` PASS in both optimization modes, including the four new round-trip lanes and the raw-receiver check;

Landing once CI is green.

nihalpasham (approved):

PR #869, opened 2026-08-13
fix(mir-importer): lower DST layout for slice-tailed structs

### LINKED ISSUE #868: mir-importer: support dynamic layout intrinsics for slice-tailed structs
## Summary

`mir-importer` currently supports `core::mem::size_of_val` and
`core::mem::align_of_val` for slices, `str`, and sized pointees, but not for
struct DSTs whose trailing field is a slice.

For example:

```rust
#[repr(C)]
struct Header<T: ?Sized> {
    head: u64,
    tag: u8,
    tail: T,
}

let concrete = Header {
    head: 0,
    tag: 0,
    tail: [11_u16, 22, 33, 44],
};

let value: &Header<[u16]> = &concrete;

let size = core::mem::size_of_val(value);
let align = core::mem::align_of_val(value);
```

The fat pointer metadata contains the runtime length of the trailing slice, but
the layout intrinsics do not currently use that metadata to compute the dynamic
layout of the enclosing struct.

## Expected behavior

For a slice-tailed struct such as `Header<[u16]>`, `size_of_val` should combine:

* the statically known offset of the trailing field,
* the runtime length metadata,
* the size and alignment of the trailing element type,
* the struct's alignment requirements,
* any applicable `repr(packed)` alignment constraint,
* final ABI size rounding.

`align_of_val` should return the effective ABI alignment of the dynamically
sized struct.

The results should match Rust's host-side layout for the same DST value.

## Actual behavior

The layout intrinsic lowering rejects unsized struct pointees instead of
computing their dynamic size and alignment from the trailing slice metadata.

## Notes

CUDA Oxide already represents slice-tailed DST references using slice-shaped fat
pointer metadata, so the required runtime length is available during lowering.

This issue is limited to dynamic layout for slice-tailed structs.

Direct indexing of a projected slice tail, for example:

```rust
value.tail[1]
```

is a separate projection-lowering issue tracked in #870

uurl:
Closes #868 

## Summary

Extend `size_of_val` and `align_of_val` lowering to support structs with slice tails, such as:

```rust
#[repr(C)]
struct Header<T: ?Sized> {
    head: u64,
    tag: u8,
    tail: T,
}

let value: &Header<[u16]>;
```

Previously, the layout intrinsics supported slices, `str`, and sized pointees, but rejected unsized struct pointees such as slice-tailed structs.

## What changed

The MIR importer now computes dynamic layout for slice-tailed structs by combining:

* the statically known offset of the trailing unsized field,
* the runtime slice metadata,
* the dynamic size and alignment of the trailing field,
* the struct's static alignment requirements,
* `#[repr(packed)]` alignment constraints where applicable,
* final ABI size rounding.

`align_of_val` uses the same tail-layout rules while keeping the resulting alignment statically known.

## Example coverage

The `dst_metadata` example now includes:

```rust
#[repr(C)]
struct Header<T: ?Sized> {
    head: u64,
    tag: u8,
    tail: T,
}
```

and verifies a coercion from:

```rust
&Header<[u16; 4]>
```

to:

```rust
&Header<[u16]>
```

The device result is compared against the host for:

* `size_of_val`
* `align_of_val`
* trailing slice length metadata
* prefix-field access through the DST reference

The mixed-width prefix intentionally makes padding and final layout rounding observable.

## Scope

This PR only addresses dynamic layout for slice-tailed structs.

Direct constant indexing of a projected `MirSliceType`, for example:

```rust
value.tail[1]
```

is a separate projection-lowering gap tracked in #870 

## Validation

Validated on an NVIDIA GeForce RTX 5060 Laptop GPU with:

```text
cargo oxide run dst_metadata
CUDA_OXIDE_NO_OPT=1 cargo oxide run dst_metadata
scripts/smoketest.sh -o '^dst_metadata$'
scripts/smoketest.sh --compile-only -o '^dst_metadata$'
```

All four pass.

Observed runtime output includes:

```text
PASS: size_of_val on &Header<[u16]>
PASS: align_of_val on &Header<[u16]>
PASS: unsized-tail metadata and prefix fields
PASS: dst_metadata
```

### DIFF
--- a/crates/mir-importer/src/translator/terminator/intrinsics/layout.rs
+++ b/crates/mir-importer/src/translator/terminator/intrinsics/layout.rs
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//! Rust dynamic-layout intrinsics for slices and `str`.
+//! Rust dynamic-layout intrinsics for slices, `str`, and slice-tailed structs.
 //!
 //! `core::mem::size_of_val` and `core::mem::align_of_val` inline to the
 //! corresponding `core::intrinsics::*` calls when their argument is a DST.
@@ -14,6 +14,8 @@
 //!
 //! - `[T]`: size = `len * size_of::<T>()`, align = `align_of::<T>()`
 //! - `str`: size = byte length, align = 1
+//! - `struct S { prefix: ..., tail: [T] }`: size and alignment are computed
+//!   from rustc's field offset/layout plus the runtime tail length metadata
 //!
 //! Sized types are also accepted as a defensive fallback for low-MIR-opt
 //! builds, where the intrinsic may survive even though optimized MIR normally
@@ -24,7 +26,7 @@ use crate::error::{TranslationErr, TranslationResult};
 use crate::translator::values::ValueMap;
 use crate::translator::{rvalue, types};
 use dialect_mir::attributes::FieldIndexAttr;
-use dialect_mir::ops::{MirConstantOp, MirExtractFieldOp, MirMulOp};
+use dialect_mir::ops::{MirAddOp, MirConstantOp, MirDivOp, MirExtractFieldOp, MirMulOp};
 use dialect_mir::types::MirSliceType;
 use pliron::basic_block::BasicBlock;
 use pliron::builtin::attributes::IntegerAttr;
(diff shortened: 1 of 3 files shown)


nihalpasham:
**Gist:** `size_of_val`/`align_of_val` now handle slice-tailed structs, and the algorithm is a one-to-one transliteration of rustc's own `size_and_align_of_dst`: same statically-known tail offset, packed alignment cap, max-align rule, and final ABI rounding, verified against the pinned toolchain source. `align_of_val` stays statically known, as it should. Closes #868.

**Verified here:**
- ran on an RTX 5090 (sm_120): `dst_metadata` PASS in both optimization modes, including the padded `repr(C) { u64, u8, [u16] }` case with the tail at byte offset 10, asserted against host-computed values;

Landing once CI is green.

nihalpasham (approved):

PR #871, opened 2026-08-13
fix(mir-importer): lower ConstantIndex on slice-tailed DST values

### LINKED ISSUE #870: mir-importer: support ConstantIndex projection on MirSliceType
## Summary

`mir-importer` currently rejects a `ConstantIndex` projection when the projected value is represented as `MirSliceType`.

This can be reproduced when accessing an element through the unsized slice tail of a DST struct.

For example:

```rust
#[repr(C)]
struct Header<T: ?Sized> {
    head: u64,
    tag: u8,
    tail: T,
}

let concrete = Header {
    head: 0,
    tag: 0,
    tail: [11_u16, 22, 33, 44],
};

let value: &Header<[u16]> = &concrete;
let x = value.tail[1];
```

Device codegen fails with:

```text
Unsupported construct: ConstantIndex projection on unsupported type.

  pliron type: MirSliceType { element_ty: ... }
  display    : mir.slice <builtin.integer ui16>
  index      : 1

ConstantIndex handles MirArrayType (extractvalue) and MirPtrType
(pointer offset + load, e.g. after Deref on a slice). The type above
matched neither. A new handler may need to be added.
```

## Expected behavior

A `ConstantIndex` projection on a `MirSliceType` should lower to an element access using the slice data pointer, analogous to indexing after dereferencing a slice pointer.

For the example above, `value.tail[1]` should load the second `u16` element from the slice tail.

## Actual behavior

Translation aborts because the `ConstantIndex` projection path only handles `MirArrayType` and `MirPtrType`.

## Notes

This was discovered while adding coverage for slice-tailed DST structs.

The DST itself is already represented as a fat pointer and its metadata can be accessed successfully. `size_of_val`, `align_of_val`, `tail.len()`, and prefix-field projections work correctly.

The failure is specifically the composition:

```text
Field(tail)
  -> MirSliceType
  -> ConstantIndex
```

so this appears to be a projection-lowering gap rather than a DST layout issue.

uurl:
Closes #870

## Summary

Support `ConstantIndex` projections on slice values produced by projecting the
unsized slice tail of a DST struct.

For example:

```rust
#[repr(C)]
struct SliceTail<T: ?Sized> {
    head: u32,
    tail: T,
}

let concrete = SliceTail {
    head: 0,
    tail: [11.0_f32, 16.0],
};

let value: &SliceTail<[f32]> = &concrete;

let x = value.tail[1];
```

Previously, this projection chain could not be lowered correctly:

```text
Deref
  -> Field(tail)
  -> MirSliceType
  -> ConstantIndex
```

## Root cause

A fat reference to a slice-tailed struct carries the trailing slice length as
metadata.

For this projection path, `Deref` scalarizes the fat reference to the struct
data pointer, so the trailing slice length must be preserved for the following
`Field` projection.

Projecting the unsized tail directly as `ptr<MirSliceType<T>>` is also invalid,
because the struct storage model represents the inline slice tail using the
element type `T`, not a fat slice value.

As a result, the projected slice tail reached `ConstantIndex` as a
`MirSliceType`, while that lowering only handled arrays and thin pointers.

## What changed

The iterative place-value walker now:

* preserves the trailing slice length when dereferencing a fat reference to a
  slice-tailed struct,
* projects the unsized tail as a pointer to its inline element type,
* reconstructs the semantic `MirSliceType` from the tail data pointer and the
  preserved length,
* normalizes that `MirSliceType` to its data pointer before lowering a forward
  `ConstantIndex`,
* reuses the existing pointer offset + load path for the final element access.

The existing slice and sized-array projection paths remain unchanged.

## Regression coverage

`ref_index_projections` now includes:

```rust
let concrete = SliceTail {
    head: 0,
    tail: input[i],
};

let value: &SliceTail<[f32]> = &concrete;

let r0 = value.tail[0];
let r1 = value.tail[1];
```

The kernel verifies that the two constant-index loads access different elements
of the trailing slice and produce the expected `+5.0` difference.

The example now contains 21 regression kernels.

## Validation

Validated with:

```text
CUDA_OXIDE_NO_OPT=1 cargo oxide run ref_index_projections
cargo test -p mir-importer
scripts/smoketest.sh -o '^ref_index_projections$'
scripts/smoketest.sh --compile-only -o '^ref_index_projections$'
cargo oxide pipeline ref_index_projections
```

Both optimized and low-MIR-optimization runs pass all 21 kernels, including:

```text
test_slice_tail_constant_index  PASS
```

`mir-importer` also passes all 1068 unit tests.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -5576,6 +5576,11 @@ pub fn translate_place_iterative(
     // data pointer because Index/ConstantIndex do not need metadata.
     let mut preserved_slice_deref_mutability: Option<bool> = None;
 
+    // A fat reference to a slice-tailed struct carries the runtime length of
+    // that tail in field 1. Deref scalarizes the fat value to the struct data
+    // pointer, so preserve the length for the immediately following tail Field.
+    let mut preserved_slice_tail_len: Option<Value> = None;
+
     // Process each projection element iteratively
     for (proj_idx, projection) in place.projection.iter().enumerate() {
         match projection {
@@ -5588,10 +5593,18 @@ pub fn translate_place_iterative(
                     .get_type(ctx)
                     .deref(ctx)
                     .is::<dialect_mir::types::MirSliceType>();
+                let pointer_info = get_static_pointer_info(&current_rust_ty);
                 let slice_deref_mutability =
-                    get_static_pointer_info(&current_rust_ty).and_then(|(pointee, is_mutable)| {
-                        rust_ty_is_slice(&pointee).then_some(is_mutable)
+                    pointer_info.as_ref().and_then(|(pointee, is_mutable)| {
+                        rust_ty_is_slice(pointee).then_some(*is_mutable)
                     });
+                let current_is_slice_tail_ref = pointer_info
+                    .as_ref()
+                    .is_some_and(|(pointee, _)| types::slice_tail_element_ty(pointee).is_some());
+                let next_is_slice_tail_field = matches!(
+                    place.projection.get(proj_idx + 1),
+                    Some(ProjectionElem::Field(_, field_ty)) if rust_ty_is_slice(field_ty)
+                );
 
                 if next_is_slice_subslice
                     && current_is_fat_slice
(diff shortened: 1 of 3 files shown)


nihalpasham:
The fix itself is correct, but #873 is a strict superset: a line-by-line diff of the two shows #873 carries every functional hunk here (Deref-arm length preservation, tail reconstruction, ConstantIndex normalization, refactored into a shared helper) plus the runtime-Index arm and both regression kernels, resolving #870 and #872 together. 

Labeled `already-fixed`; this PR will be closed in favor of #873 when it lands.

PR #873, opened 2026-08-13
fix(mir-importer): lower Index on slice-tailed DST values

### LINKED ISSUE #872: mir-importer: support Index projection on MirSliceType
## Summary

`mir-importer` currently rejects a runtime `Index` projection when the projected value is represented as `MirSliceType`.

This occurs when indexing through the unsized slice tail of a DST struct with a runtime index.

For example:

```rust
#[repr(C)]
struct SliceTail<T: ?Sized> {
    head: u32,
    tail: T,
}

let concrete = SliceTail {
    head: 0,
    tail: [11.0_f32, 16.0],
};

let value: &SliceTail<[f32]> = &concrete;

let k = runtime_index;
let x = value.tail[k];
```

The relevant MIR projection shape is:

```text
Deref
  -> Field(tail)
  -> MirSliceType
  -> Index(k)
```

## Expected behavior

A runtime `Index` projection on a `MirSliceType` should use the slice data
pointer and reuse the existing pointer-based index lowering.

Conceptually:

```text
MirSliceType<T>
  -> extract data pointer
  -> ptr<T>
  -> runtime pointer offset
  -> load T
```

## Actual behavior

The iterative place-value walker handles runtime `Index` for:

```text
MirArrayType
MirPtrType
```

but not for `MirSliceType`.

As a result, the projected DST slice tail reaches the `Index` lowering as a
`MirSliceType` and is rejected as an unsupported indexable value.

## Notes

This is the runtime-index counterpart of #870.

#870 / #871 add support for:

```text
Field(tail)
  -> MirSliceType
  -> ConstantIndex
```

This issue covers the sibling path:

```text
Field(tail)
  -> MirSliceType
  -> Index
```

The existing array and thin-pointer indexing paths should remain unchanged.

Depends on the slice-tail value reconstruction introduced by #871

uurl:
Closes #872 

## Summary

Support runtime `Index` projections on `MirSliceType` values produced by projecting the unsized slice tail of a DST struct.

This is the runtime-index counterpart of the `ConstantIndex` support added for #870.

The relevant projection chain is:

```text
Deref
  -> Field(tail)
  -> MirSliceType
  -> Index(k)
```

## Root cause

The iterative place-value walker already supports runtime `Index` for:

```text
MirArrayType
MirPtrType
```

but a projected slice tail is represented semantically as a `MirSliceType`.

After the slice-tail metadata is preserved and the tail value reconstructed,
the runtime `Index` path still sees the fat slice value directly and therefore
does not enter the existing pointer-based indexing path.

## What changed

The slice-value-to-data-pointer normalization is shared by both runtime
`Index` and `ConstantIndex`.

For a `MirSliceType<T>` value, the importer now:

```text
MirSliceType<T>
  -> extract field 0
  -> ptr<T>
```

and then reuses the existing pointer lowering.

For runtime `Index`, the resulting path is:

```text
MirSliceType<T>
  -> data pointer
  -> MirPtrOffsetOp(runtime_index)
  -> MirLoadOp
```

The existing array and thin-pointer indexing behavior remains unchanged.

## Regression coverage

`ref_index_projections` now includes:

```rust
let concrete = SliceTail {
    head: 0,
    tail: input[i],
};

let value: &SliceTail<[f32]> = &concrete;

let k = (input[0][0] as usize) & 1;
let r0 = value.tail[k];
let r1 = value.tail[k + 1];
```

The runtime-derived `k` forces the MIR path:

```text
Field(tail)
  -> MirSliceType
  -> Index
```

The harness verifies that the two loads access different tail elements and
produce the expected `+5.0` difference.

`ref_index_projections` now contains 22 regression kernels.

## Validation

Validated with:

```text
CUDA_OXIDE_NO_OPT=1 cargo oxide run ref_index_projections
cargo test -p mir-importer
scripts/smoketest.sh -o '^ref_index_projections$'
scripts/smoketest.sh --compile-only -o '^ref_index_projections$'
cargo oxide pipeline ref_index_projections
git diff --check
```

Results:

```text
test_slice_tail_constant_index    PASS
test_slice_tail_runtime_index     PASS
SUCCESS: all kernels passed
```

Both optimized and `CUDA_OXIDE_NO_OPT=1` runs pass all 22 kernels.

`mir-importer` passes all 1068 unit tests.

Both filtered smoketests pass:

```text
Pass: 1 / 1
Fail: 0 / 1
```

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -5515,6 +5515,45 @@ fn rust_ty_is_slice(ty: &rustc_public::ty::Ty) -> bool {
     )
 }
 
+fn normalize_slice_value_to_data_ptr(
+    ctx: &mut Context,
+    value: Value,
+    block_ptr: Ptr<BasicBlock>,
+    prev_op: Option<Ptr<Operation>>,
+    loc: Location,
+) -> (Value, Option<Ptr<Operation>>) {
+    let slice_element_ty = {
+        let value_ty = value.get_type(ctx);
+        let value_ty_ref = value_ty.deref(ctx);
+        value_ty_ref
+            .downcast_ref::<dialect_mir::types::MirSliceType>()
+            .map(|slice_ty| slice_ty.element_type())
+    };
+    let Some(element_ty) = slice_element_ty else {
+        return (value, prev_op);
+    };
+
+    let data_ptr_ty: TypeHandle =
+        dialect_mir::types::MirPtrType::get_generic(ctx, element_ty, false).into();
+    let extract_ptr = Operation::new(
+        ctx,
+        MirExtractFieldOp::get_concrete_op_info(),
+        vec![data_ptr_ty],
+        vec![value],
+        vec![],
+        0,
+    );
+    extract_ptr.deref_mut(ctx).set_loc(loc);
+    MirExtractFieldOp::new(extract_ptr)
+        .set_attr_index(ctx, dialect_mir::attributes::FieldIndexAttr(0));
+    match prev_op {
+        Some(prev) => extract_ptr.insert_after(ctx, prev),
+        None => extract_ptr.insert_at_front(block_ptr, ctx),
+    }
+
+    (extract_ptr.deref(ctx).get_result(0), Some(extract_ptr))
+}
+
 /// Translate a MIR Place using iterative projection processing.
 /// This handles arbitrary depth projections by processing each element in sequence.
 pub fn translate_place_iterative(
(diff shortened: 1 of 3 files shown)


nihalpasham:
**Gist:** closes both DST slice-tail indexing gaps in one place: the value walker preserves the fat reference's tail length across `Deref`, rebuilds a first-class `(ptr, len)` slice at the tail `Field`, and normalizes slice values at `Index`/`ConstantIndex` to the data pointer, so both indexing forms reuse the existing pointer-offset + load path. This mirrors the tail-borrow recipe the address walker already ships, so the two walkers now agree on the tail representation, and it converges on the metadata-carry design rustc_codegen_ssa uses. Strict superset of #871, which will be closed as already-fixed. Pushed 28ab403c adding the padded regression the review asked for: the PR's kernels use an unpadded offset-4 struct, so a wrong-tail-offset bug would slip past; the new kernel uses issue #870's own layout (`head: u64, tag: u8, tail: [u16]`, tail at byte offset 10) with constant and data-derived runtime indices. Closes #872 and #870.

**Verified here:**
- ran on an RTX 5090 (sm_120): `ref_index_projections` SUCCESS in both optimization modes, all 23 kernels including the new padded lane;
- the new kernel has teeth, proven by fault injection: forcing the importer's tail offset to 9 or 12 leaves the offset-4 kernels green and fails only the padded lane (misaligned address, then wrong values), so it catches what the original kernels cannot;

Two loud-failing gaps remain for follow-up, not blockers: nested slice tails (tail behind an inner struct field) defeat the one-projection lookahead, and the write/borrow paths through the address walker are untouched. Landing once CI is green.

nihalpasham (approved):

PR #878, opened 2026-08-13
fix(mir-lower): reject FloatToInt to integer widths wider than 64 bits

### LINKED ISSUE #877: FloatToInt saturating casts do not support `i128` destinations (and non-{f16,f32,f64} sources)
## Background
`FloatToInt` (and the saturation `llvm.fptosi.sat` / `llvm.fptoui.sat` intrinsics it
emits) only handles 16/32/64-bit float sources and, implicitly, integer destinations up
to 64 bits. Two gaps:

1. **`i128` destination is unguarded.** `crates/mir-lower/src/convert/ops/cast.rs:281`
   builds the intrinsic name with `format!(\"i{}\", int_width)` unconditionally, so an
   `f64 as i128` cast emits `llvm.fptosi.sat.i128_f64`, which the NVPTX backend cannot
   legalize. There is no explicit guard, so this fails obscurely downstream instead of
   with a clear message.
2. **Non-standard float sources** already error cleanly at
   `crates/mir-lower/src/convert/ops/cast.rs:287-291`, but `f128` (and the matching
   `IntToFloat` direction) has no lowering path at all.

## Location
`crates/mir-lower/src/convert/ops/cast.rs:271-300` (the `convert_float_to_int` path).

## Proposed fix
- Add an explicit guard for `int_width > 64` that returns a clear
  `FloatToInt: integer destinations wider than 64 bits are not yet supported on the device`
  error (consistent with the existing float-width message), so the failure is immediate
  and actionable.
- (Stretch) Lower `f128`/`i128` casts via the runtime soft-float / `__floatditf`-style
  helpers already shipped for other wide arithmetic.

- An `f64 as i128` cast in a kernel now yields a precise, located error instead of an
  opaque LLVM failure.
- New unit test in `crates/mir-lower` covering the rejected width.
- `cargo test -p mir-lower` stays green.

(Note: broad `f16`/`f128` *type* support is tracked separately; this issue is scoped to
the cast/intrinsic lowering specifically.)

Ayush7614:
## Summary
`convert_float_to_int` in `crates/mir-lower/src/convert/ops/cast.rs` built the
saturating intrinsic name with `format!("i{}", int_width)` unconditionally, so a
`f64 as i128` cast emitted `llvm.fptosi.sat.i128_f64` — an intrinsic the NVPTX
backend cannot legalize. The failure surfaced obscurely downstream instead of at
the cast site.

This adds an explicit `int_width > 64` guard that returns a located, actionable
error, matching the existing float-source-width guard.

Fixes #877.

## Verification
- Added `float_to_int_rejects_integer_destination_wider_than_64_bits` in
  `crates/mir-lower/src/convert/ops/cast.rs` asserting the rejection.
- `cargo test -p mir-lower` is green (76 passed, 0 failed) on this branch.

## Scope note
Broad `f16`/`f128` *type* support is tracked separately; this PR is scoped to the
cast/intrinsic lowering only.


nihalpasham:
**Gist:** the right interim answer for a conversion the device toolchain cannot service: the guard sits at the single site that creates the `fptosi/fptoui.sat` intrinsic names, upstream of both the llc and nvvm-transforms paths, and it fires before the signedness split, so `f64 as i128` and `f64 as u128` both get a located, actionable error instead of an opaque llc failure naming an internal intrinsic. No supported program changes behavior (Rust has no integer types between 65 and 127 bits, so only the never-worked i128/u128 casts hit the guard). One precedent correction from review: nvcc does not reject this case, it open-codes the conversion inline in PTX (verified with nvcc 13.3); the accurate precedent for reject-early is clang CUDA's device-side `__int128` handling, and the same no-device-builtins constraint drives it here. Pushed dba0ccfa adding the unsigned twin test. Closes #877.

**Verified here:**
- mir-lower suite green with both wide-int rejection tests (316 tests); a negative control (loosening the guard) makes both tests fail, proving they reach the guard rather than any earlier rejection;
- ran on an RTX 5090 (sm_120): `vecadd` and `cast_tests` (34/34, the existing example exercising float-to-int casts) unchanged;
- the mirror gap, `i128 as f64` through IntToFloat, still fails obscurely and is now tracked as #882.

Landing once CI is green.

nihalpasham (approved):

PR #879, opened 2026-08-13
fix(mir-lower): thread target mode into shared-pointer transmutes

### LINKED ISSUE #874: Transmute involving a shared-memory pointer is rejected (target-mode-dependent layout)
## Background
`core::mem::transmute` (and derived ops) on a value that contains a shared-memory
pointer (`addrspace(3)`) currently fails codegen with:

> Transmute involving a shared-memory pointer is target-mode dependent (64-bit PTX/legacy, 32-bit modern NVVM) and is not yet supported

This blocks patterns like transmuting a `SharedArray` index handle, or reinterpreting a
shared-memory pointer, inside kernels.

## Location
`crates/mir-lower/src/convert/ops/cast.rs:525-532` — the guard in the transmute lowering
rejects any type that `llvm_type_contains_pointer_in_address_space(..., SHARED)`.

## Why it is hard today
In PTX/legacy mode a shared pointer is 64 bits; in modern NVVM (`p3:32`) it is 32 bits.
The general transmute path uses a target-agnostic 8-byte approximation, which cannot be
sound here, so the lowering bails out before choosing a size.

## Proposed fix
Thread the active NVVM/PTX target mode into the cast lowering stage so the transmute can
pick the correct shared-pointer byte width (4 or 8) and rewrite the value as a sized
integer of the appropriate bit width, then bitcast back. Mirror the address-space
handling already done for generic/global pointers elsewhere in the same file.

- A kernel that transmutes a `SharedArray` handle / `p3` pointer round-trips
  (apply + undo == identity) on both `p3:32` and 64-bit targets.
- New unit coverage in `crates/mir-lower` mirroring the existing scalar/aggregate
  transmute tests, plus a `cargo oxide run` example if warranted.
- `cargo test -p mir-lower` and the workspace `cargo check` stay green.

Ayush7614:
## Summary
`mir-lower` rejected every `Transmute` involving a shared-memory pointer
(`addrspace(3)`) with a target-mode-dependent-width error. The physical width
is target-dependent: legacy NVPTX/PTX uses 64-bit pointers, while modern NVVM
(libNVVM) uses 32-bit `p3:32`. With no target mode available at this stage,
the previous code bailed out before choosing a size.

This adds `shared_address_space_pointer_width` to `LoweringOptions` and uses it
in `llvm_type_size_align` and `scalar_bit_width` so the correct width (4 or 8
bytes) is reported for `addrspace(3)` pointers. The existing transmute paths
then handle shared-pointer casts correctly, and a width-mismatched cast still
fails with a clear size-mismatch error.

`cuda-oxide-codegen::lower_to_llvm` derives the width from `IntrinsicBackend`
(`LibNvvm` -> 32, `LlvmNvptx` -> 64), so the plumbing reaches the real codegen
path automatically. `rustc-codegen-cuda` does not construct `LoweringOptions`
directly, so it needs no source change.

Fixes #874.

## Verification
- `crate::mir-lower`: `cargo test -p mir-lower` green (146 integration + 76 lib tests, 0 failed).
- New/updated tests in `crates/mir-lower/src/convert/ops/cast.rs`:
  - `transmute_involving_shared_pointer_uses_target_dependent_width` (both 32- and 64-bit modes).
  - `transmute_shared_pointer_to_wrong_width_integer_is_rejected`.
- `cargo test -p cuda-oxide-codegen` green except for one pre-existing,
  environment-specific failure (`ptx::tests::legacy_opt_failure_warns_but_experimental_mode_fails`,
  which invokes `/bin/false` and fails identically on `main` without this change).

## Scope / caveat (partially verified)
The 32-vs-64-bit selection is keyed off `IntrinsicBackend::LibNvvm` == modern
NVVM (`p3:32`). This matches the existing data-layout assumptions in
`convert/types.rs`, but the final PTX/object correctness of shared-pointer
transmutes on real hardware still needs validation in CI with the CUDA toolkit
and a GPU, which is unavailable in this dev environment.

### DIFF
--- a/crates/cuda-oxide-codegen/src/lower.rs
+++ b/crates/cuda-oxide-codegen/src/lower.rs
@@ -44,6 +44,16 @@ pub fn lower_to_llvm(
         mir_lower::LoweringOptions {
             allow_fma_contraction,
             intrinsic_backend,
+            // Modern NVVM (libNVVM) uses 32-bit `p3:32` shared pointers; the
+            // legacy NVPTX/PTX path uses 64-bit pointers in every address space.
+            shared_address_space_pointer_width: if matches!(
+                intrinsic_backend,
+                mir_lower::IntrinsicBackend::LibNvvm
+            ) {
+                32
+            } else {
+                64
+            },
         },
     ) {
         Ok(()) => Ok(()),
--- a/crates/mir-lower/src/convert/ops/arithmetic.rs
+++ b/crates/mir-lower/src/convert/ops/arithmetic.rs
@@ -942,6 +942,7 @@ mod tests {
             crate::LoweringOptions {
                 allow_fma_contraction: false,
                 intrinsic_backend: crate::IntrinsicBackend::LlvmNvptx,
+                shared_address_space_pointer_width: 64,
             },
         )
         .expect("lowering failed");
--- a/crates/mir-lower/src/convert/ops/call.rs
+++ b/crates/mir-lower/src/convert/ops/call.rs
@@ -1858,6 +1858,7 @@ mod tests {
             crate::LoweringOptions {
                 allow_fma_contraction: false,
                 intrinsic_backend: crate::IntrinsicBackend::LlvmNvptx,
+                shared_address_space_pointer_width: 64,
             },
         );
 
@@ -2276,6 +2277,7 @@ mod tests {
             crate::LoweringOptions {
                 allow_fma_contraction: true,
                 intrinsic_backend: crate::IntrinsicBackend::LlvmNvptx,
+                shared_address_space_pointer_width: 64,
             },
         );
         assert_eq!(
@@ -2324,6 +2326,7 @@ mod tests {
             crate::LoweringOptions {
                 allow_fma_contraction: true,
                 intrinsic_backend: crate::IntrinsicBackend::LibNvvm,
+                shared_address_space_pointer_width: 32,
             },
         );
         assert_eq!(
--- a/crates/mir-lower/src/convert/ops/cast.rs
+++ b/crates/mir-lower/src/convert/ops/cast.rs
@@ -52,7 +52,7 @@
 //! through an exactly-sized, explicitly aligned stack slot.
 
 use crate::convert::types::{
-    convert_type, llvm_type_contains_pointer_in_address_space, llvm_type_is_byte_faithful,
+    convert_type, llvm_type_is_byte_faithful,
     llvm_type_size_align,
 };
 use crate::helpers;
@@ -526,6 +526,10 @@ fn emit_transmute(
         .map(PointerType::address_space);
     let src_is_int = val_ty.deref(ctx).is::<IntegerType>();
     let dst_is_int = llvm_ty.deref(ctx).is::<IntegerType>();
+    let dst_int_width = llvm_ty
+        .deref(ctx)
+        .downcast_ref::<IntegerType>()
+        .map(IntegerType::width);
 
     // `ptr::addr()` is implemented in the sysroot as a pointer-to-usize
     // transmute. For a named CUDA address space, Rust's logical address is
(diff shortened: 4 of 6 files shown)


nihalpasham:
So as I understand it, this implements issue #874's proposed fix. The problem is with that proposal, not with the execution.

The approach isn't right for three independent reasons:

**1. The width isn't knowable at the point the PR reads it.**

The pipeline decides between the two NVVM dialects only after lowering has already run:

```text
lowering runs ──────► IR rendered ──────► legacy-vs-modern NVVM picked
(the PR needs                             (4-vs-8 is actually
 the width HERE)                           decided HERE)
```

So the PR guesses from `--emit-nvvm-ir` (`IntrinsicBackend::LibNvvm`), and the guess is wrong for a whole class of builds, because the legacy NVVM data layout is 64-bit in every address space; only modern NVVM has `p3:32`:

```text
build                                    PR's guess   physical truth
default llc path                         8            8       ok
--emit-nvvm-ir, Blackwell (modern)       4            4       ok
--emit-nvvm-ir, pre-Blackwell (legacy)   4            8       WRONG
```

**2. Even with the right width, it's the wrong ruler.**

Rust defines transmute legality by rustc's layout, and rustc sizes every nvptx64 pointer at 8 bytes, shared or not:

```rust
// valid Rust: rustc checks 8 == 8, so this must compile on every target
let p: *mut u32 = transmute(bits as usize);

// this PR in p3:32 mode: re-measures the shared pointer at 4 bytes,
// rejects with "size mismatch: 8 vs 4"
```

That rejected round trip is the very thing #874's acceptance criteria ask for on the p3:32 target. Meanwhile the added unit test transmutes `i32 <-> ptr(3)`, a 4-byte pairing rustc can never emit, since Rust sizes that pointer at 8. So the check rejects real code and accepts impossible code.

**3. It turns a loud error into silently wrong bytes.** This is the disqualifying one.

`main`'s blanket guard was crude but safe. With it removed, an aggregate transmute takes the memory round-trip (store to a stack slot, load back), which captures the pointer's raw `addrspace(3)` bytes: an offset into the shared window, not an address. The first shared allocation sits at offset 0:

```text
scalar path (already correct on main, genericizes first):
  shared_ptr as usize                      -> 0x7f31_4000_0000   generic address

aggregate path (was a loud error, now compiles):
  transmute::<Wrapper, u64>(Wrapper(ptr))  -> 0x0                raw offset, reads as NULL
```

Two casts Rust considers identical return different integers, and one of them manufactures null-looking values from valid pointers. This is the hazard `emit_ptr_to_int`'s generic-address invariant exists to prevent.

Two further notes:

- `llvm_type_size_align` gets repurposed as a physical-width oracle, but its callers (union storage, global validation, layout-divergence checks) and its sibling `get_type_size` still assume rustc's 8-byte pointers, so two layout oracles disagree about the same type;
- the PR body cites a test (`transmute_shared_pointer_to_wrong_width_integer_is_rejected`) that is not present at the head commit.

The sound fix needs no width at all:

- keep size checks on rustc layout (pointers are 8 bytes, on every target);
- bridge shared-pointer scalars through the generic address space:

```text
shared -> int    addrspacecast to generic, then ptrtoint    (the existing emit_ptr_to_int invariant)
int -> shared    inttoptr to generic, then addrspacecast
ptr <-> ptr      one addrspacecast
```

- keep aggregates loudly rejected: their memory image is the target-dependent part.

That version is implemented and verified:

- mir-lower green with four new transmute tests, none branching on backend or dialect mode;
- a new `shared_ptr_transmute` example passes on an RTX 5090 through the llc path, the unoptimized path, and the modern-NVVM `p3:32` path this PR's heuristic targeted, with no width knowledge at lowering time.

It lands as its own PR shortly; closing this one with `already-fixed`. Thanks for taking #874 on: the issue's own proposed-fix text is what sent this down the wrong road, and it gets corrected there too.

PR #887, opened 2026-08-14
refactor(cuda-device): move BF16 conversions out of tcgen05

lucifer1004:
## Problem

- Packing two `f32` values into a `bf16` pair works on any sm_80+ GPU, but the only source-level API lived in `cuda_device::tcgen05`, the module for Blackwell-only (sm_100) tensor-core ops. Using it on, say, an sm_120 GPU looked like it required TCGEN05 hardware.
- The same name meant opposite things in two modules, and the docs needed a warning box to keep them apart:

```text
// Before: same name, opposite directions, wrong module
tcgen05::cvt_f32x2_bf16x2(a, b)   // PACKS  (f32, f32) -> u32    [lives in the sm_100 module]
convert::cvt_f32x2_bf16x2(bits)   // UNPACKS u32 -> (f32, f32)

// After: one home, destination-first name, like PTX cvt.rn.bf16x2.f32
convert::cvt_bf16x2_f32(lo, hi)   // PACKS  (f32, f32) -> u32    [works on sm_80+]
convert::cvt_f32x2_bf16x2(bits)   // UNPACKS u32 -> (f32, f32)   [unchanged]
```

- Bonus bug: the round-to-nearest scalar helper turned some NaNs into infinity:

```text
f32_to_bf16_rne(0x7F80_0001)      // NaN whose payload sits only in the 16 discarded bits
Before: 0x7F80                    // +Inf, the NaN vanished
After:  0x7FC0                    // quiet NaN, matches hardware cvt.rn.bf16.f32
```

## Fix

- move every BF16 conversion and packing helper into `cuda_device::convert`, where all other conversions already live
- rename the packer destination-first: `convert::cvt_bf16x2_f32`, matching the module convention and the PTX mnemonic order
- the rename is declared in the intrinsics catalog (overlay TOML + resolve recipe), so every dispatch table regenerates; the renderer now derives compat-wrapper names from their declared paths instead of assuming them
- NaN fix: quiet the NaN (`bits | 0x0040`) instead of letting the payload vanish into `Inf`
- migrate the five in-repo example users; follow-ups 11c48000 and 12c2b47f caught four stale doc references and one rustfmt diff

## Gotchas

- breaking rename with no compat re-export: out-of-tree code importing the old `tcgen05` paths stops compiling (loud error, and the project is pre-1.0)
- the raw ABI is untouched: entry `i0071`, `cuda_intrinsics::convert::cvt_f32x2_bf16x2`, and the nvvm op keep their names; only the cuda-device surface moved
- ~1071 generated files changed, but every one is a catalog-hash header refresh, no hand edits hiding in the churn

## Testing and validation

- cuda-device (97 tests) and cuda-intrinsics-gen (289 tests) suites green
- `cuda-intrinsics-gen check` clean; ABI history preserved at 986 entries
- new unit tests for NaN payloads, ties, signed zero, and infinities
- `bf16x2_arith` executed on SM120: all eight packed ops pass
- CI green (43/43)

---


nihalpasham:
**Gist:** right move. Packing two f32s into bf16 works on any sm_80+ GPU, so it never belonged in the Blackwell-only tcgen05 module, and the old name meant the opposite thing in two different modules. The rename rides through the intrinsics catalog, so the ABI entry and the dialect op stay untouched. The NaN fix in `f32_to_bf16_rne` is a genuine bug fix on top (payload-only NaNs used to come out as infinity).

What I checked and did:
- catalog regen check and ABI history both clean (986 entries preserved); the 1071-file churn is all hash-header refreshes, nothing hiding in it
- verified the new rounding helper by hand for NaN payloads, infinities, ties, and signed zero; it now matches what hardware `cvt.rn.bf16.f32` does
- pushed 11c48000: four stale references to the old name you missed (gemm_sol README, tcgen05_matmul README, the cublas test comment, and one in the book)
- pushed 12c2b47f: rustfmt diff in gemm_sol that the CI fmt gate caught
- CI fully green

Heads-up for release notes: this is a breaking rename with no compat re-export, which I think is fine pre-1.0.

nihalpasham (approved):

PR #888, opened 2026-08-15
feat(codegen): forward compiler-owned multi-result bundles

lucifer1004:
## Problem

- Multi-result inline PTX (a generated intrinsic like `tcgen05_ld` returning `[u32; 32]`) comes back from the importer as an aggregate bundle that gets spilled to stack memory, even though the compiler created it itself and knows it never escapes. Downstream `opt -O2` cleans that up today, so the spill reaches final PTX only in `CUDA_OXIDE_NO_OPT=1` builds; the pass makes the compiler's own MIR output clean without leaning on LLVM to fix it:

```text
let regs = tcgen05_ld_32x32b_x4(addr);   // 1 instruction, 4 registers
out[0] = regs.get::<0>() + regs.get::<3>();

What the compiler hands to LLVM without this pass (and what reaches
final PTX under CUDA_OXIDE_NO_OPT=1):
         tcgen05.ld {%r1,%r2,%r3,%r4}, [taddr];
         st.local x4 ... ld.local x2 ...     // spill and reload
         add.u32  %r7, %r5, %r6;
With the pass:
         tcgen05.ld {%r1,%r2,%r3,%r4}, [taddr];
         add.u32  %r7, %r1, %r4;             // registers used directly

Measured on tcgen05_matmul (sm_100a), pass on vs off:
  default pipeline (opt -O2):  final PTX byte-identical (SROA cleans the
                               spills downstream either way)
  CUDA_OXIDE_NO_OPT=1:         8 fewer st.local + 8 fewer ld.local,
                               local depot 68B -> 36B, 8 fewer .b32 regs
```

- Separately, `ptx_asm!` capped outputs at 16, which is too small for tcgen05-scale loads.

## Fix

- the importer marks bundles it created as compiler-owned
- a new mir-transforms pass (`forward_compiler_result_bundles`) runs before mem2reg and forwards each element read straight to the producing value, then deletes the bundle once every use is forwarded
- anything it cannot prove is safe (dynamic index, escape, partial read) drops the whole plan and keeps the memory path, so behavior is never guessed
- raised the `ptx_asm!` output arity from 16 to 64
- review follow-up 2d42ced4: a whole-array read through a struct-wrapped bundle (the `CuSimd::to_array()` pattern) used to make the pass erase one of its own construct ops while it still had uses, which pliron panics on; that case now bails out to the memory path

## Gotchas

- ordinary Rust arrays never qualify; only the importer can attach the marker. So #399 (keeping iterator-chain arrays in SSA) stays open by design, this pass is not that
- non-constant indices reject the whole bundle rather than partially forwarding
- 14 ldmatrix sites still use the older bundling helper and do not forward yet; reasonable follow-up

## Testing and validation

- mir-transforms suite green, including a regression test for the erase panic (verified it fires when the guard is disabled)
- 64-output `ptx_asm!` accept test sits next to the existing 65-output reject test; full cuda-macros and trybuild suites green
- generator consistency check clean
- CI green (43/43)


nihalpasham:
**Gist:** good pass, and it earns its keep: compiler-owned multi-result bundles stay in registers instead of bouncing through `.local`. One real crash found in review: a whole-array read through a struct-wrapped bundle (the `CuSimd::to_array()` pattern) made the pass hand back one of its own construct ops as the replacement, then erase it while it still had uses, which pliron panics on.

What I checked and did:
- reproduced the panic with the guard disabled, so the regression test provably hits the bug
- pushed 2d42ced4: fail-closed rejection in `analyze_projection` when the replacement is one of the bundle's own construct ops (the memory path plus mem2reg picks those up instead), plus the regression test and positive 64-output `ptx_asm!` coverage next to the 65-reject test
- pushed 79f39bd5: two clippy nits in the new code
- element-wise loads through the struct wrapper still forward; generator consistency check clean; all suites green

Also confirmed #399 stays open: that issue wants ordinary iterator-chain arrays in SSA, which this pass rules out by design.


**Update, measured evidence (pass on vs off, same commit):** under the default pipeline the final PTX is byte-identical for tcgen05_matmul, gemm_sol, tcgen05, and vecadd; downstream `opt -O2` (SROA/mem2reg) already removes every bundle spill. The pass pays off in `CUDA_OXIDE_NO_OPT=1` builds: on tcgen05_matmul it removes 8 `st.local` + 8 `ld.local` round-trips, shrinks the kernel local depot from 68B to 36B, and saves 8 `.b32` registers, with `tcgen05.ld` results feeding `cvt.rn.bf16x2.f32` directly. So the honest framing: no final-PTX change under default settings today; the win is clean MIR-level output that does not depend on LLVM's optimizer, plus the NO_OPT escape hatch, plus the arity raise.

nihalpasham:
**Gist of the decision:** taking this, and I want to be precise about why, since I measured it. Under the default pipeline this pass changes nothing today: final PTX is byte-identical with it on or off, because `opt -O2` already cleans up the bundle spills downstream.

What I'm taking it for:

- the compiler stops depending on LLVM's optimizer to clean up spills it generated itself; the MIR we hand downstream is right by construction, which keeps options open for a pipeline that leans less on LLVM
- `CUDA_OXIDE_NO_OPT=1` builds (the mode you reach for when debugging the optimizer) get the proven win: 8 fewer `st.local` + 8 fewer `ld.local`, local depot 68B down to 36B, 8 fewer registers on tcgen05_matmul
- the `ptx_asm!` arity raise to 64 is needed regardless of any of the above

So: accepted for self-sufficiency and the escape hatch, with eyes open that default-pipeline PTX is unchanged today. If the pipeline ever routes around LLVM's optimizer, this pass is already sitting where it needs to be.

nihalpasham (approved):

PR #889, opened 2026-08-15
feat(device): add native packed f32x2 arithmetic

lucifer1004:
## Summary

- add a native packed f32x2 device type and arithmetic operations
- lower supported operations to the corresponding packed PTX instructions
- encode the PTX ISA and SM capability floors in target detection

## Testing

- cuda-device unit tests
- cuda-oxide-codegen library tests
- ptxas compatibility probes across the supported PTX ISA and SM boundaries

### DIFF
--- a/crates/cuda-device/src/f32x2.rs
+++ b/crates/cuda-device/src/f32x2.rs
@@ -1,0 +1,126 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Native SM100 two-lane `f32` arithmetic.
+//!
+//! Unlike [`crate::vector::F32x2`], which is an over-aligned memory element,
+//! these operations consume and produce [`crate::Float2`] register groups and
+//! lower to one `fma.rn.f32x2` or `add.rn.f32x2` PTX instruction.
+
+use crate::{Float2, ptx_asm};
+
+/// Pack two scalar f32 values into the PTX f32x2 register representation.
+#[inline(always)]
+pub fn pack_f32x2(lo: f32, hi: f32) -> u64 {
+    let result: u64;
+    unsafe {
+        ptx_asm!(
+            "mov.b64 %0, {%1, %2};",
+            out("=l") result,
+            in("f") lo,
+            in("f") hi,
+            options(register_only),
+        );
+    }
+    result
+}
+
+/// Return the low scalar lane from a packed PTX f32x2 register.
+#[inline(always)]
+pub fn unpack_f32x2_lo(value: u64) -> f32 {
+    let result: f32;
+    unsafe {
+        ptx_asm!(
+            "mov.b64 {%0, _}, %1;",
+            out("=f") result,
+            in("l") value,
+            options(register_only),
+        );
+    }
+    result
+}
+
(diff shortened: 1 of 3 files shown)


nihalpasham:
Thanks for the PR 

I don't think we should take this in its current form, and I want to walk through why, because the reasons are about where the code lives, not how it's written.

Every intrinsic family in `cuda-device` is generated from the catalog. This PR would be the first hand-written one.

Here is what the packed-math siblings look like today:

```text
crates/cuda-device/src/f16x2.rs      27 lines of docs + include!("generated/f16x2.rs")
crates/cuda-device/src/bf16x2.rs     same
crates/cuda-device/src/dotprod.rs    same

crates/cuda-device/src/f32x2.rs      (this PR) ~160 hand-written lines
```

A catalog entry buys four things automatically that a hand-written module has to fake or skip:

```text
catalog TOML entry
   ├── ABI ledger id (append-only, i-numbered, tracked forever)
   ├── arch + PTX ISA floor derived from the instruction, not a comment
   ├── a ptxas probe, so CI proves the real assembler accepts the spelling
   └── regenerated dispatch in the importer and collector
```

This PR skips all four, and two of them bit immediately:

1. **The floor comments are wrong:** The code says PTX 8.2 and the test narrates rounding 8.2 up to 8.6. The PTX ISA manual says `add.f32x2` / `fma.f32x2` were introduced in PTX ISA 8.6 and need sm_100 or newer. The emitted values happen to land right, but the stated reason is wrong, and that's the kind of thing floor derivation exists to prevent.
2. **The feature detector is too narrow.:** `contains_f32x2_features` matches only the two spellings this PR emits. All of these are valid f32x2 spellings that would sail past it:

```text
add.f32x2          (rounding omitted)
add.rz.f32x2       add.rm.f32x2      add.rp.f32x2
add.rp.ftz.f32x2   (the manual's own example!)
mul.f32x2          sub.f32x2
```

`target.rs` documents at line ~341 that broad family matching is intentional, because a missed spelling silently selects an architecture or PTX ISA that is too old. This detector goes against that rule.

**And there's a collision waiting:** If the catalog later grows an `f32x2` family (which is where this family belongs), the generator will want to emit `add_f32x2(u64, u64) -> u64` into `cuda_device::f32x2`, the same path where this PR already exports `add_f32x2`(Float2, Float2) -> Float2. Same name, same module, different signatures.

For what it's worth, this is also how nvcc slices it: `__fadd2` /` __fmul2` /` __ffma2` are typed, arch-gated math-API intrinsics in the device library, which is their version of our catalog, not hand-written asm in a leaf header.

**The ask:** Please open a new PR moving the raw packed intrinsics (`add`/`sub`/`mul`/`fma` `.f32x2` plus the `.ftz` variants) into the cuda-intrinsics-gen catalog as a PTX-native family, the way `add.rn.bf16x2` already is. They get ledger ids, derived floors, and probe coverage for free. Keep the nice part of this PR (the `Float2` wrappers and pack/unpack sugar) hand-written on top of the generated functions. 

Closing this one.

lucifer1004:
Thanks, agreed on the ownership boundary. I will replace this with a catalog-native packed-ALU family: raw `u64` intrinsics for `add`/`sub`/`mul`/`fma` and their `.ftz` forms, with append-only ABI IDs, derived PTX 8.6 / sm_100 floors, generated importer/collector dispatch, and terminal probes. The hand-written device module will contain only the `Float2` pack/unpack and typed wrapper layer. I will also replace the exact-spelling detector with broad `.f32x2` instruction-family matching and reference this PR from the replacement.

PR #890, opened 2026-08-15
feat(device): type tcgen05 FP8 descriptors

lucifer1004:
## Problem

- Building a tcgen05 FP8 (`.kind::f8f6f4`) matmul descriptor meant hand-assembling raw bitfields: pick the right 3-bit format code for E4M3 or E5M2, shift it to the right field, mask it in, and hope you read the SM100 layout table correctly.
- Nothing type-checked the formats, so swapping the a-type and b-type codes, or using an FP16 code by mistake, compiled fine and produced a garbage descriptor at runtime:

```text
// Before: raw bits, nothing checks that 0b000 means E4M3 in this field
let desc = Tcgen05InstructionDescriptor::from_raw(
    base | (0b000 << A_TYPE_SHIFT)    // E4M3? E5M2? go read the table
         | (0b001 << B_TYPE_SHIFT),   // swap these and it still compiles
);

// After: typed formats, wrong combinations no longer expressible
let desc = Tcgen05InstructionDescriptor::new_f8(
    shape,
    Tcgen05F8ElementType::E4M3,       // a-type
    Tcgen05F8ElementType::E5M2,       // b-type
    accumulator_type,
);
```

## Fix

- adds a typed `Tcgen05F8ElementType` enum and a `new_f8` constructor, so FP8 descriptors are built from named formats instead of raw bits
- review follow-up d59e3afb: `new_f8` now delegates to the existing descriptor builder the same way `new_f16`/`new_bf16`/`new_tf32` do. The builder stores the 3-bit format code raw internally and gained F8-typed a/b setters, so `build()` is the single place descriptor bits get packed
- public signatures unchanged; the hardware encoding used by existing kernels is preserved bit for bit

## Gotchas

- covers E4M3 and E5M2 only; E2M3/E3M2/E2M1 already have matching ws intrinsics and are a natural follow-up
- format values are interpreted in the FP8 domain only when the instruction kind is `f8f6f4`; the descriptor still has to be issued through `tcgen05_mma_shared::<2, ..>`
- a test asserts the builder path produces the identical raw encoding to the previous direct packing, so there is no silent drift between the two construction routes

## Testing and validation

- cuda-device suite green (97 tests), including the SM100 descriptor-layout test and a test that a-type and b-type fields stay independent
- no compiler-surface changes and no new kernels, so no GPU gate needed
- CI green (43/43)

### DIFF
--- a/crates/cuda-device/src/tcgen05.rs
+++ b/crates/cuda-device/src/tcgen05.rs
@@ -581,13 +581,12 @@ impl Tcgen05InstructionDescriptor {
         b_type: Tcgen05F8ElementType,
         accumulator_type: Tcgen05AccumulatorType,
     ) -> Self {
-        let mut raw = 0u32;
-        raw |= (accumulator_type as u32) << 4;
-        raw |= (a_type as u32) << 7;
-        raw |= (b_type as u32) << 10;
-        raw |= ((shape.n >> 3) & 0x3f) << 17;
-        raw |= ((shape.m >> 4) & 0x1f) << 24;
-        Self { raw }
+        Self::builder()
+            .shape(shape)
+            .a_type_f8(a_type)
+            .b_type_f8(b_type)
+            .accumulator_type(accumulator_type)
+            .build()
     }
 
     /// Create descriptor from raw 32-bit value.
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** nice typed surface for the FP8 descriptors, and the two new tests nail the SM100 bit layout. One structural thing I went ahead and did: `new_f8` was packing bits by hand while `new_f16`/`new_bf16`/`new_tf32` all delegate to the builder, which left two packing sites for one layout.

What I checked and did:
- pushed d59e3afb: builder stores the 3-bit format code raw, gained F8-typed a/b setters, and `new_f8` now delegates like its siblings, so `build()` is the single packing site; public signatures unchanged
- extended the E4M3 test to assert the builder path produces the identical raw encoding
- 97 cuda-device tests green, CI fully green

E2M3/E3M2/E2M1 are a sensible follow-up since the ws intrinsics already exist; left out here on purpose.

nihalpasham (approved):

PR #891, opened 2026-08-15
fix(mir-lower): preserve source locations across lowering

lucifer1004:
## Problem

- Lowering from dialect-mir to the LLVM dialect dropped source locations: an op that knew it came from `examples/foo.rs:23` became an `llvm.call` or branch with an unknown location, so PTX line tables pointed nowhere and located diagnostics from later passes lost their anchor:

```text
// kernel source
let v = a[i] + b[i];        // vecadd.rs:23
out[i] = v;                 // vecadd.rs:24

Before (.ll, line-info build):
  %v = call float @...      ; no !dbg, line table has nothing to say
  store float %v, ptr %o    ; no !dbg
After (.ll, line-info build):
  %v = call float @..., !dbg !12    ; !12 = DILocation(line: 23)
  store float %v, ptr %o, !dbg !13  ; !13 = DILocation(line: 24)
```

## Fix

- copy the source op's location onto every op the lowering creates
- review follow-up 5f514ca7 centralizes it: `inline_asm_convergent` and `inline_asm_sideeffect` now take the source op and stamp the location themselves (the same pattern this PR gave `call_intrinsic`), which covers all ~45 inline-asm construction sites in one place (mbarrier, cp_async, wgmma, warp, cluster, wmma, ldmatrix, tma, and the generated templates), plus `convert_store`, the atomic store, and memcpy/memmove
- dropped one redundant copy: pliron's `replace_operation` already propagates the old op's location for single-result replacements (verified at the pinned pliron rev; a comment records this)
- the generator template was updated and `generated_intrinsics.rs` regenerated, so all 26 generated asm sites pass the source op too

## Gotchas

- `!dbg` only appears in line-info builds (`CUDA_OXIDE_DEBUG=line`); plain builds emit none, by design, so do not be surprised by a bare `.ll` in a default build
- two direct `llvm.CallOp` construction sites in tma.rs keep their explicit location copies because they bypass `call_intrinsic`

## Testing and validation

- mir-lower suites green (221 + 105 tests), including a new test that the shared helper copies the source location
- `cuda-intrinsics-gen check` clean (template and generated file in sync)
- end to end: `CUDA_OXIDE_DEBUG=line cargo oxide pipeline vecadd` shows `!dbg` on the device call, branches, and stores at the right lines
- CI green (43/43)

### DIFF
--- a/crates/cuda-intrinsics-gen/src/render.rs
+++ b/crates/cuda-intrinsics-gen/src/render.rs
@@ -14957,15 +14957,15 @@ fn render_lowering(catalog: &CatalogFile, hash: &str) -> String {
         );
     }
     let void_ty = llvm_types::VoidType::get(ctx);
-    let inline_asm = inline_asm_convergent(
+    inline_asm_convergent(
         ctx,
         rewriter,
+        op,
         void_ty.into(),
         operands,
         template,
         constraints,
     );
-    crate::convert::preserve_location(ctx, op, inline_asm);
     rewriter.erase_operation(ctx, op);
     Ok(())
 }
@@ -14996,6 +14996,7 @@ fn convert_generated_tcgen05_load(
         let inline_asm = inline_asm_convergent(
             ctx,
             rewriter,
+            op,
             scalar_ty,
             operands,
             template,
@@ -15012,6 +15013,7 @@ fn convert_generated_tcgen05_load(
     let inline_asm = inline_asm_convergent(
         ctx,
         rewriter,
+        op,
         result_ty.into(),
         operands,
         template,
@@ -15095,6 +15097,7 @@ fn convert_generated_tcgen05_load(
             inline_asm_convergent(
                 ctx,
                 rewriter,
+                op,
                 void_ty.into(),
                 operands,
                 &template,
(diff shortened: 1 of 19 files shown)


nihalpasham:
**Gist:** real fix for a real annoyance: lowered calls and branches were losing their source locations, so `.ll` and PTX line info pointed nowhere. The per-site copies worked but left ~45 inline-asm construction sites uncovered, so I moved the copy into the shared helpers instead.

What I checked and did:
- pushed 5f514ca7: `inline_asm_convergent` / `inline_asm_sideeffect` now take the source op and stamp the location themselves (same pattern as your `call_intrinsic` change), which covers mbarrier, cp_async, wgmma, warp, cluster, wmma, ldmatrix, tma and the generated templates in one place; also covered `convert_store`, the atomic store, and memcpy/memmove
- verified at the pinned pliron rev that `replace_operation` already propagates locations for the single-result branch, so that redundant copy is dropped with a comment saying why
- regenerated the intrinsics template output so the generator check passes
- pushed 46d2de9d: one clippy nit in the new test
- end-to-end: `CUDA_OXIDE_DEBUG=line cargo oxide pipeline vecadd` shows `!dbg` on the device call, branches, and stores; mir-lower suites green; CI fully green

nihalpasham (approved):

PR #892, opened 2026-08-15
feat(finalizer): add standalone PTX assembly

lucifer1004:
## Summary

- add an independent nvptxcompiler-sys crate with dynamic symbol loading and RAII program ownership
- expose a typed, lazy PtxAssembler policy layer from the artifact finalizer
- report deterministic option digests and compiled resource usage without changing normal finalizer discovery

## Testing

- nvptxcompiler-sys unit tests
- cuda-artifact-finalizer unit tests and clippy
- live PTX assembly on a supported target

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,6 +353,7 @@ version = "0.2.1"
 dependencies = [
  "libnvvm-sys",
  "nvjitlink-sys",
+ "nvptxcompiler-sys",
  "sha2",
  "thiserror",
 ]
@@ -934,6 +935,14 @@ dependencies = [
 
 [[package]]
 name = "nvjitlink-sys"
+version = "0.2.1"
+dependencies = [
+ "libloading",
+ "thiserror",
+]
+
+[[package]]
+name = "nvptxcompiler-sys"
 version = "0.2.1"
 dependencies = [
  "libloading",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ members = [
     "crates/cuda-async",
     "crates/libnvvm-sys",
     "crates/nvjitlink-sys",
+    "crates/nvptxcompiler-sys",
     # Internal naming contract (workspace-private; publish = false)
     "crates/reserved-oxide-symbols",
     # Differential codegen fuzzer support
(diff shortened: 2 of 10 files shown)


nihalpasham:
First, the goal here is a good one, and it's a direction we're actively interested in: assembling PTX to a cubin without the driver, with real ptxas semantics (the `--fmad` behavior difference on PTX inputs is a genuine motivation, our own comments document it). The code is also well built; it mirrors the nvjitlink-sys patterns faithfully.

The problem is the delivery: NVIDIA ships `nvPTXCompiler` as a static library only. There is no `.so` on any stock install, so the runtime dlopen discovery can never succeed:

```text
$ ls /usr/local/cuda/lib64/ | grep nvptx
libnvptxcompiler_static.a     <- static archive only, nothing to dlopen

PtxAssembler::discover() -> looks for libnvptxcompiler.so -> always fails
```

That holds across CUDA 11.x through 13.3. Out of curiosity, what did your live-assembly test load? A hand-built `.so` from the archive would explain it, but users won't have one.

The two deliveries that do work, and how existing tools pick:

```text
static-link libnvptxcompiler_static.a at build time   (RAPIDS model, cargo feature)
find and run the ptxas binary as a subprocess         (nvcc / Triton model)
```

To be concrete about where I'd like this capability to end up, since the delivery rework is also a chance to widen the scope. What `nvPTXCompiler` should eventually buy us as an option:

- **decoupling compilation from module load:** your PR already has this by construction (assembly happens in the finalizer at build time, cached), so this part carries over as-is
- **scheduling independent compilations concurrently:** not addressed yet; the tool caching is load-once, but nothing exercises parallel compiles across modules
- **the broader ptxas option and diagnostic surface:** today the mapping is five flags plus resource-usage parsing; things like -maxrregcount, opt level, -split-compile, and warning controls are the actual payoff of talking to ptxas directly

None of that needs to land in one PR. We can discuss which lane fits before any rework; most of what you built (typed options, digests, identity revalidation, finalizer wiring) carries over mostly unchanged, so this would be a re-plumbing, rather than a rewrite.

Closing this one. Please reference it when you do open a new one.

lucifer1004:
Thanks, agreed. The live test exercised a non-stock shared-library environment, so it did not validate a deliverable configuration for toolkit users. I will replace the dynamic-loading route with a discovered `ptxas` subprocess and retain the useful parts here: typed options, deterministic recipe/tool digests, identity revalidation, cubin validation, and resource diagnostics. Independent invocations also give the finalizer a natural concurrency boundary. I will keep the initial option surface focused and reference this PR from the replacement.

PR #893, opened 2026-08-15
perf(finalizer): reuse pinned tool provenance

lucifer1004:
## Problem

- Every backend materialization recomputed provenance by rehashing the pinned tool binaries (libNVVM, nvJitLink), which are hundreds of MB. Warm builds paid the full SHA-256 cost over and over for tools that had not changed:

```text
Before: every single build
  hash libnvvm.so (~300 MB)  ->  provenance digest     (seconds, every time)

After: first build
  hash libnvvm.so  ->  cache { version, dev/inode/ctime/size, digest }
warm build
  stat libnvvm.so, identity matches  ->  reuse digest  (no hashing)
libnvvm.so touched or replaced
  full rehash; if the content actually changed        ->  ProvenanceMismatch (loud)
```

## Fix

- a versioned handshake cache at `.oxide-artifacts/materializer-handshake/v1.json` stores verified tool provenance with explicit, derived-Serde fields
- each operation revalidates cheaply: handshake version, file identity (device, inode, ctime, size), and the expected semantic digest must all match to reuse the cached hash
- any mismatch falls back safely to full rehashing, and a genuine content change fails loudly with `ProvenanceMismatch`
- review follow-ups: 4a772075 minimizes the isolated backend lockfile churn (only the new serde/serde_json entries, no unrelated bumps), skips digest reuse when the Unix identity fields are absent so non-Unix hosts always rehash, and documents the cache file in the README; 99632bcd refreshes all 209 example lockfiles, which went stale because the new serde dependency reaches every example through its cuda-host path dependency

## Gotchas

- version skew is intentional and loud: after updating cargo-oxide, run `cargo oxide setup` again or materialization fails with `MissingToolIdentityHandshake`
- deleting the cache file just forces a full rehash; a corrupted or inconsistent cache is ignored and self-heals on the next build
- identity is only trusted with complete Unix metadata; anything less rehashes

## Testing and validation

- finalizer, cargo-oxide, and isolated rustc-codegen-cuda materialization suites green, plus four live finalizer integration tests
- cold vs warm timing measured with byte-identical handshake JSON output
- cache behavior spot-checked: touch/replace libnvvm between builds triggers a rehash, content change fails with `ProvenanceMismatch`
- example license-policy script, root and backend cargo-deny all green
- CI green (43/43)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,6 +353,8 @@ version = "0.2.1"
 dependencies = [
  "libnvvm-sys",
  "nvjitlink-sys",
+ "serde",
+ "serde_json",
  "sha2",
  "thiserror",
 ]
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -18,28 +18,40 @@ use std::process::Command;
 
 const MATERIALIZE_ENV: &str = reserved_oxide_symbols::MATERIALIZE_CUBIN_ENV;
 const EXPECTED_PROVENANCE_ENV: &str = reserved_oxide_symbols::MATERIALIZER_PROVENANCE_ENV;
+const MATERIALIZER_HANDSHAKE_ENV: &str = reserved_oxide_symbols::MATERIALIZER_HANDSHAKE_ENV;
 const CODEGEN_FINGERPRINT_ENV: &str = reserved_oxide_symbols::CODEGEN_FINGERPRINT_ENV;
 const DEVICE_CODEGEN_CRATE_ENV: &str = reserved_oxide_symbols::DEVICE_CODEGEN_CRATE_ENV;
 const BACKEND_IDENTITY_CFG: &str = "cuda_oxide_internal_backend_identity";
 const LEGACY_CODEGEN_FINGERPRINT_CFG: &str = "cuda_oxide_internal_codegen_env";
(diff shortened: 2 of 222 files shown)


nihalpasham:
**Gist:** caching design is good: versioned handshake, identity revalidation, and a fail-safe path back to full hashing on any mismatch. Warm builds skip rehashing multi-hundred-MB tool binaries, and a corrupted cache self-heals. I tightened a few edges and cleaned up fallout.

What I checked and did:
- pushed 4a772075: minimized the isolated backend lockfile churn (only the serde/serde_json additions remain, no unrelated version bumps), added a guard so digest reuse only happens when the full Unix identity (device, inode, ctime) is present (non-Unix always rehashes), and documented the cache file in the README (delete it to force a rehash)
- pushed 99632bcd: refreshed all 209 example lockfiles. Your serde addition to cuda-artifact-finalizer reaches every example through their cuda-host path dependency, so their `--locked` resolutions went stale and the example-policy CI job failed. The refresh is purely additive and holds serde at the root lock's 1.0.228 so no extra syn churn sneaks in
- finalizer tests, isolated backend build, root and backend cargo-deny, and the example license-policy script all green; CI fully green

One heads-up on version skew: after updating cargo-oxide, users need `cargo oxide setup` again or materialization fails with a clear handshake error. Seems like the right trade.

nihalpasham (approved):

PR #894, opened 2026-08-15
feat(ptx): add lossless structural syntax model

lucifer1004:
## Summary

- add a dependency-free, lossless PTX lexer with exact byte spans for trivia, comments, strings, preprocessors, words, punctuation, and unknown input
- build a PTX-specific flat `Statement`/`Scope` CST with stable IDs, recovery diagnostics, and byte-accounting coverage proofs
- derive instruction, predicate, modifier, and operand views from CST nodes rather than a second source scan
- migrate `cuda-intrinsics-gen` away from its duplicate PTX scanner

The parser intentionally does not type-check the PTX ISA, so newer instruction spellings remain representable. A release corpus audit over a 14,048,388-byte linked PTX artifact recovered 3,646,260 tokens, 308,854 statements, 5,498 scopes, and 288,994 instructions with zero incomplete files.

## Stack

1/9 — base layer; no dependency on another PR in this stack. The commit introduced by this PR is `e53344ca`.

## Validation

- `cargo test --workspace --locked --offline`
- `cargo clippy --workspace --all-targets --locked --offline -- -D warnings`
- `cargo fmt --all --check`
- release audit against the linked PTX corpus described above

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -420,6 +420,7 @@ name = "cuda-intrinsics-gen"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "ptx-syntax",
  "serde",
  "serde_json",
  "sha2",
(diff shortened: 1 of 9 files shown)


lucifer1004:
Superseded by the consolidated PTX stack #914 → #915 → #916 → #917. The focused replacement for this PR is #914. The original branch is retained for history.

PR #914, opened 2026-08-15
feat(ptx): add lossless structural syntax foundation

lucifer1004:
## What this PR does

Adds `ptx-parse`, a dependency-free crate that parses PTX text into a lossless concrete syntax tree (CST), plus a span-safe edit layer, and migrates every in-tree PTX text consumer onto it.

## Motivation

All PTX in this project is text produced by someone else (llc, external inputs), and main handled it in five places with five independent string hacks:

```text
read the recorded sm target     cargo-oxide      line.starts_with(".target")
merge embedded PTX modules      cuda-host        prefix-match header lines
find libdevice declarations     codegen          misses multi-line .extern .func
strip the debug flag            codegen          substring surgery on .target
match probe instructions        intrinsics-gen   hand-rolled head matcher
```

One of those had a live bug: the debug-strip failed to remove `debug` from a `.target` line with a trailing comment, so "stripped" builds shipped debug-flagged PTX.

## What lossless means

```text
coverage over every PTX artifact our pipeline produces:
  176 files, 2.47 MB, 47,813 instructions -> 0 incomplete, 0 unknown bytes
  concatenating token spans reproduces each input byte-for-byte
```

An instruction the crate has never seen still parses (opcode + operands + semicolon), so new ISA revisions cannot break it. Parser gaps are observable via `Coverage`, never silent.

## What changed

- the `ptx-parse` crate: lexer, statement/scope CST with recovery diagnostics, callables/labels/registers/predicates with byte spans, `EditScript` with overlap and UTF-8 validation
- the five consumer migrations, each replacing an ad-hoc scan with a structural query and surfacing parse failures as typed errors
- ten examples' PTX self-checks migrated from brittle `find()` scans to structural queries (moved here from #915; they depend only on the parser)
- parse errors now propagate in the intrinsics-gen matcher (`Result`, not an empty match list), so "unparseable" can never read as "no match"
- crate renamed from ptx-syntax and made publishable: `ptx-parser` is taken on crates.io, `ptx-parse` is free, and publishable cuda-host now depends on it
- all example lockfiles re-locked against current main

## Testing and validation

- ptx-parse 35 tests (incl. a pseudo-fuzz never-panic test); intrinsics-gen 289; all four consumer crates green
- audit run: the coverage numbers above
- compile smoketest 210/210; workspace clippy and both fmt gates clean
- GPU (sm_120): migrated-assertion examples run and their structural PTX checks pass against freshly built PTX

## Landing

The stack (#914 -> #915 -> #916 -> #917) merges as one unit via #917; this PR then closes as already-fixed.


nihalpasham:
Reviewed in depth, direction endorsed, and I've rebuilt this branch on current main with the review outcomes applied. What changed and why:

- **crate renamed: `ptx-syntax` is now `ptx-parse`**, and `publish = true`. Reasoning: `ptx-parser` is taken on crates.io, `ptx-parse` is free, and cuda-host (publishable) now depends on this crate, so it has to be publishable too or cuda-host is stuck git-only
- **#915's example-migration half moved down here** (commit 0ebc6147, your authorship preserved): the ten examples' PTX self-checks depend only on the parser, not the dialect, so this is their home. Their lockfiles were re-locked against current main rather than carrying the stale copies
- one catch during that fold: #915's copy of `gemm_sol_final/src/main.rs` would have reverted the `cvt_bf16x2_f32` import migration that landed in #887 after you branched; restored main's imports
- **parse errors now propagate in the intrinsics-gen matcher** (5c4a2729): `instructions_with_matching_head` returns `Result` instead of mapping unparseable PTX to an empty match list, so "couldn't parse" can never read as "no match"
- all 201 remaining example lockfiles re-locked (6f3acfb0): every example resolves cuda-host, which now pulls ptx-parse, and the cargo-deny example gate runs `--locked`

Verified on this branch:

- ptx-parse 35 tests, intrinsics-gen 289, all four consumer crates green; workspace clippy and both fmt gates clean
- compile smoketest 210/210
- the audit example over every PTX artifact our pipeline produces: 176 files, 2.47 MB, 47,813 instructions, **0 incomplete**
- GPU (sm_120): three of the migrated-assertion examples run and their structural PTX checks pass against freshly built PTX

Landing plan for the stack: merges as one unit via #917; this one closes as already-fixed at that point.

nihalpasham:

PR #915, opened 2026-08-15
feat(ptx): add structured surface dialect

lucifer1004:
## What this PR does

Adds `dialect-ptx`: a pliron dialect that models a PTX module as first-class IR, a projection that builds it from the ptx-parse CST, and a deterministic canonical emitter.

## Motivation

Strings cannot be safely transformed. To instrument, canonicalize, or one day emit PTX, the structure has to be IR the framework can verify:

```text
ptx.module
 └── ptx.callable [callable_kind: Entry, callable_name: "countdown", ...]
      ├── ptx.directive    ".reg .b32 %r<4>;"
      ├── ptx.instruction  [head: "mov.u32", operands: ["%r1","10"]]
      ├── ptx.label        "$L_loop"
      └── ptx.instruction  [head: "bra", predicate: {%p1}, ...]
```

## What changed (including the review redesigns)

- the dialect: module/callable/scope/directive/instruction/label ops via `#[pliron_op]`, typed `CallableKindAttr`, projection with byte-span lineage, canonical emitter
- **typed attrs are the enforced truth.** Review found the emitter printed only the raw header string, so renaming a callable through its typed attr silently emitted stale PTX:

```text
callable_name   = "saxpy"          <- typed attr, was decorative
callable_header = ".visible ..."   <- what the emitter actually printed
```

  `Verify` now re-parses the header and cross-checks name/kind/external with targeted messages, and the emitter verifies before printing, so a desynced op cannot emit. `ptx.scope` gets the mirror check (including refusing a scope header that smuggles a whole callable). The contract is documented on the ops: typed attrs = queryable truth, header = print carrier for unmodeled syntax, Verify enforces agreement
- **predicates are a typed attribute**: `PredicateAttr { register, negated }` replaces the flattened `instruction_prefix` string entirely; the emitter derives `@%p1` / `@!%p1` from the attr. Nothing but predicate text ever rode in the string (verified), so the field is removed, not deprecated
- instruction interiors stay strings by design: typing is added incrementally as consumers need it, catalog-style, rather than transcribing the full ISA up front

## Testing and validation

- dialect-ptx unit tests incl. predicated round-trips (`@%p1 add`, `@!%p1 bra`), negated projection, and every desync case asserting its error message
- all 176 real generated PTX files canonicalize through the new verify path, 0 failures
- full workspace green; a `demo_pipeline` example walks build -> print -> emit -> reparse

## Landing

Merges as one unit via #917; this PR then closes as already-fixed.


nihalpasham:
This PR is now dialect-only (the example migrations moved down to #914 where they belong) and carries the two redesigns from review. Both address the same disease: load-bearing facts living in strings.

**1. Typed attrs are now the enforced truth on `ptx.callable` / `ptx.scope`** (e2383aca). The review found this:

```text
callable_name   = "saxpy"        <- typed attr, was decorative
callable_header = ".visible..."  <- string, was the only thing the emitter read
set_attr_callable_name("v2")  ->  emitted PTX still said "saxpy". silently.
```

Now `Verify` re-parses the header and cross-checks all three typed attrs against it, with targeted messages ("header names \"kernel\" but callable_name is \"renamed\""), and the emitter verifies before printing, so a desynced op can never emit. `ptx.scope` gets the mirror check, including refusing a scope header that smuggles a whole callable past the typed attrs. The contract is documented on the ops: typed attrs = queryable truth, header = print carrier for unmodeled syntax, Verify enforces agreement.

**2. Predicates are a typed attribute** (238b5aad): `PredicateAttr { register, negated }` replaces the `instruction_prefix` string entirely; the projection maps the parser's already-parsed predicate straight in, and the emitter derives `@%p1` / `@!%p1` from the attr. Grep confirmed nothing but predicate text ever rode in the prefix string, so the field is gone, not deprecated.

Verified:

- dialect-ptx tests incl. predicated round-trips (`@%p1 add`, `@!%p1 bra`), negated projection, and each desync case asserting its error message; full workspace green
- all 176 real generated PTX files canonicalize through the new verify path, 0 failures
- demo example (945ec5a8) walks build -> print -> emit -> reparse end to end

Closes as already-fixed when #917 merges the stack.

nihalpasham:

PR #916, opened 2026-08-15
feat(ptx): analyze CFG and lexical normalization

lucifer1004:
## What this PR does

Recovers control flow from parsed PTX and adds two source-level normalizations (anonymous-scope flattening, scoped-register renaming) that prepare a callable for raising into basic blocks (#917). Everything is fail-closed: any construct the analysis cannot prove is a typed hard error, never a guess.

## Motivation

PTX text has no block structure, just labels and branches. Raising it into real IR (#917) needs a proven CFG first, and needs lexical scopes flattened (pliron blocks have no nested scopes):

```text
bb0  [ld.param / cvta / mov]        ──fallthrough──▶ bb1
bb1  "$L_loop" [sub / setp / bra]   ──branch──▶ bb1   (loop back-edge)
                                    ──fallthrough──▶ bb2
bb2  [st.global / ret]              exit: Return
```

The fail-closed error taxonomy:

```text
bra $L_typo;       target not defined       -> Err(UnknownBranchTarget)
brx.idx %r1, tbl;  table defined after use  -> Err(BranchTableAfterUse)
block ends, nothing follows                 -> Err(OpenFallthrough)
.version 9.4       above tested ceiling     -> Err(Version(Unsupported))
```

## What changed (including the review hardening)

- CFG recovery (`ControlFlow::analyze`): direct `bra`, `brx.idx` via `.branchtargets`, predicated fallthrough, `ret`/`exit`/`trap`; plan-then-apply rewrites composing into one validated `EditScript` transaction
- **the `v.x` hazard is closed.** `.` is a word byte, so `v.x` lexes as one token and never matched the binding `v`; after renaming an inner vector register, a leftover `v.x` silently rebound to an outer register of the same name and assembled with wrong semantics. Rename plans now scan for `base.suffix` tokens (bank forms included, `%r0.x` is caught for `%r<2>`) and hard-error with the token and byte span
- **name-capture guard**: a rename target can no longer collide with any label in the callable or any callable name in the module
- **the version ceiling now gates the rewrites**, not just the read-only CFG (rewriting newer PTX is the risky direction). One shared `version.rs`, one typed error, all three entry points gated identically

## Testing and validation

- 35 dialect-ptx tests (6 new for guards and gating); full workspace green
- round-trip over all 178 real PTX artifacts: 24 genuinely rewritten (gemm_sol alone: 185 renames, 129 scopes flattened), every rewrite reparses complete with identical instruction counts; ptxas accepts the rewritten text
- 8 artifacts fail closed on nested `.param` call frames, correctly: flattening those would change `.param` allocation semantics

## Landing

Merges as one unit via #917; this PR then closes as already-fixed.


nihalpasham:
Rebuilt on the redesigned #915 (your analysis code needed nothing but the crate-rename adaptation, since it reads the parser's `Instruction::predicate()`, never the removed prefix string). Three hardening changes on top, two of which close correctness holes found in review:

**1. The `v.x` hazard is now a plan-time error** (389aedfd). The failure it closes:

```text
.reg .v4 .f32 v;          <- outer v
{  .reg .v4 .f32 v;       <- inner v, renamed to v_1
   mov.f32 %f1, v.x;      <- "v.x" is ONE token, never matched the binding
}                            -> silently read the OUTER register, assembled fine, wrong
```

`RegisterAlphaPlan` now scans for `base.suffix` word tokens before planning; if the base resolves to a binding being renamed (including bank forms, so `%r0.x` is caught for `%r<2>`), the plan errors with the token and byte span. Whole-vector uses still rename fine.

**2. Rename targets can no longer capture label or callable names** (35c38fe0): every candidate is cross-checked against the callable's labels and all module callable names, deliberately broader than strictly-referenced symbols.

**3. The PTX version ceiling now gates the rewrites, not just the read-only CFG** (52db71e6). Rewriting newer PTX is the risky direction; reading it conservatively is not. The check moved to a shared `version.rs` with one typed error, gating all three entry points identically. API note: `CfgError`'s three version variants collapsed into `CfgError::Version(PtxVersionError)`; nothing outside dialect-ptx referenced the old ones.

Verified:

- 35 dialect-ptx tests (6 new for the guards and gates), full workspace green, all CI-mirror scripts pass
- round-trip over all 178 real PTX artifacts: 24 genuinely rewritten (gemm_sol alone: 185 renames, 129 scopes flattened), every rewrite reparses complete with identical instruction counts; 8 fail closed on nested `.param` call frames, which is correct (flattening those would change `.param` allocation semantics); ptxas accepts the rewrites

Closes as already-fixed when #917 merges the stack.

nihalpasham:

PR #917, opened 2026-08-15
feat(ptx): add ptx-parse and the PTX dialect with verified native CFG (stack #914-#917)

lucifer1004:
## What this PR does

Raises analyzed PTX into native pliron basic blocks with verified structure, and canonically re-emits it. Head of the #914 -> #915 -> #916 -> #917 stack: merging this lands all four.

## Motivation

This makes control flow first-class IR instead of substrings, which is the foundation both near-term consumers (IKET PTX instrumentation, Tile-to-SIMT epic #96) and the long-term direct-emission path build on:

```text
now:    llc -> PTX text ──parse──▶ PTX IR ──verify──▶ tools, instrumentation
                                     ▲
later:  LLVM dialect ── direct emission ┘              (same IR, same emitter,
                                                        already battle-tested)
```

The raised form, from a real loop kernel:

```text
^loop:
  ptx.instruction [head: "setp.ne.u32", operands: ["%p1","%r1","0"]]
  ptx.terminator [^exit, ^loop] [terminator_kind: Branch,
                  terminator_predicate: ptx.predicate {register="%p1", negated=false},
                  terminator_has_fallthrough: true]
```

Successors are real pliron block references; the back-edge is IR, not a string.

## What changed (including the review fixes)

- transactional raising: `NativeCfgPlan::analyze` does all fallible work (flatten, rename, CFG) touching nothing; `materialize` only allocates the proven graph
- every structural invariant in `Verify`: fallthrough must name the next emitted block, branch successors need a same-callable source label, `brx.idx` tables defined once, before use, all users agreeing
- the emitter verifies before printing, so unverified IR cannot emit
- **`.pragma` is scope-sensitive now** (fail-closed): pragma strings have placement-dependent meaning (a block-scoped `nounroll` widened to the callable is a behavior change), so nested scopes containing one refuse to flatten. Only `.loc` stays scope-neutral, with the documented reason
- **terminators carry the typed predicate** (`terminator_prefix` string removed, mirroring #915)
- first consumers and the transform convention are on record: passes operating ON the raised dialect use pliron rewriter/interface infrastructure; the text-domain edit layer is pre-raising only

## Testing and validation

- **CUBIN byte-identity over all 175 PTX artifacts our pipeline produces**: original vs parse -> raise -> verify -> emit, both through ptxas: 167/175 byte-identical and emit-twice idempotent, **zero cubin diffs**; all 7 artifacts containing `brx.idx` byte-identical
- the other 8 fail closed at raise time (nested `.param` call frames, the same boundary #916 enforces); no crashes, no wrong output
- full workspace suite green; compile smoketest 210/210; GPU (sm_120) 7/7 incl. mma_mxf8f6f4 with numeric check

## Known boundary

Raising is not universal: callables with nested call-frame scopes refuse to raise rather than guess. Intended fail-closed posture; widening it is future work with its own tests.

## Landing

Merging this PR lands the stack; #914, #915, #916 then close as already-fixed.


nihalpasham:
Head of the stack and the landing vehicle: merging this merges #914 through #917, and those three close as already-fixed. Rebuilt on the hardened #916 with the final review items plus the whole-stack verification on our hardware.

Changes:

- **`.pragma` is scope-sensitive now** (690416db): pragma strings are tool-defined with placement-dependent meaning (a block-scoped `.pragma "nounroll";` widened to the whole callable is a real behavior change), so a nested scope containing one refuses to flatten, fail-closed. Only `.loc` stays scope-neutral, with the reason documented: it sets lexical debug state and braces never bound its effect
- **terminators carry the typed predicate too**: `terminator_prefix` is gone the same way #915 removed `instruction_prefix`. The raised loop terminator now prints as real IR:

```text
ptx.terminator [^exit, ^loop] [terminator_kind: Branch,
                terminator_predicate: ptx.predicate {register="%p1", negated=false},
                terminator_has_fallthrough: true]
```

- **first consumers on record** (ff816980): IKET PTX instrumentation and the Tile-to-SIMT epic #96, with the long-term direct-emission direction named. Also the convention note: transforms operating ON the raised dialect go through pliron rewriter/interface infrastructure; the text-domain EditScript layer is for pre-raising normalization only

The verification that matters, all on our hardware:

- **CUBIN byte-identity over all 175 PTX artifacts our pipeline produces**: original vs parse -> raise -> verify -> canonical-emit, both through ptxas. 167/175 byte-identical AND emit-twice idempotent, **zero cubin diffs**, 8 fail closed at raise time (nested `.param` call frames, same fail-closed boundary as #916; no crashes, no wrong output). All 7 artifacts containing `brx.idx` are byte-identical
- full workspace suite green (93 test-result lines), workspace clippy + both fmt gates clean, compile smoketest 210/210
- GPU (sm_120): 7/7 including mma_mxf8f6f4 with numeric check and two migrated-assertion examples parsing their own PTX with ptx-parse

Known boundary, stated plainly: raising is not universal yet. Callables with nested call-frame scopes refuse to raise rather than guess. That is the intended fail-closed posture; widening it is future work with its own tests.

nihalpasham (approved):

PR #919, opened 2026-08-15
chore(deps): update Pliron for explicit struct layouts

### LINKED ISSUE #918: Update Pliron dependency for explicit LLVM struct layouts
## Summary

Update cuda-oxide to the Pliron API where LLVM `StructType` explicitly carries a
`StructLayout`.

This is a dependency/API migration required before cuda-oxide can consume packed
LLVM struct support. It should remain behavior-neutral with respect to packed
Rust aggregates.

## Motivation

pliron-org/pliron#232 changes the LLVM struct type API so that struct layout is
represented explicitly as either `StructLayout::Unpacked` or
`StructLayout::Packed`.

cuda-oxide currently constructs and rebuilds LLVM structs in several lowering
and transformation paths. After updating Pliron, these sites must either:

- explicitly select `StructLayout::Unpacked` when constructing a new synthetic
  LLVM struct, or
- preserve the existing layout when rebuilding an existing LLVM struct.

Without this migration, cuda-oxide cannot update to the Pliron revision that
provides packed struct representation.

## Proposed work

- Update `pliron`, `pliron-derive`, and `pliron-llvm` to the new Pliron API.
- Update `StructType` construction sites to pass an explicit layout.
- Preserve `StructType::layout()` in transformations that rebuild existing
  structs.
- Update generated intrinsic code and its source generator consistently.
- Adapt other Pliron 0.17 API changes required by the dependency update.
- Update lockfiles and dependency-license metadata.

## Non-goals

This issue does **not** enable packed aggregate lowering in cuda-oxide.

In particular, this change should not:

- select `StructLayout::Packed` for cuda-oxide-generated types,
- change `repr(C, packed)` by-value behavior,
- change kernel ABI lowering,
- remove existing fail-closed checks for layout-divergent aggregates,
- change packed field-addressing behavior,
- change LLVM packed-struct emission.

Packed aggregate representation and ABI support should be implemented separately
once this migration is complete.

## Acceptance criteria

- All newly constructed synthetic LLVM structs explicitly use
  `StructLayout::Unpacked`.
- Transformations rebuilding existing LLVM structs preserve their original
  `StructLayout`.
- No cuda-oxide code selects `StructLayout::Packed`.
- Generated intrinsic sources remain reproducible from the generator.
- Existing packed aggregate fail-closed behavior remains unchanged.
- Formatting, Clippy, tests, docs, intrinsic probes, ABI-history checks, and
  repository guards pass, except for source-policy checks that cannot accept a
  temporary contributor-fork dependency while the upstream Pliron change is
  still under review.

## Dependency

Blocked on pliron-org/pliron#232.

While that PR is under review, development may temporarily use the corresponding
contributor-fork revision. Before this issue is considered complete, the
dependency must be switched back to `pliron-org/pliron` at the resulting
upstream revision and the full dependency source-policy checks must pass.

uurl:
Closes #918

## Problem

- LLVM has two struct layouts, and pliron previously could not tell them apart:

```text
{ i8, i64 }     natural layout: 7 padding bytes after the i8, align 8
<{ i8, i64 }>   packed layout:  no padding, i64 sits at offset 1, align 1
```

- pliron 0.17 (pliron#232) makes the layout explicit: every `StructType` construction now declares `Unpacked` or `Packed`. cuda-oxide had 31 construction sites and 3 struct-rebuild paths that needed to say which one they mean.
- Without this, packed carriers (needed by #930 for pointer relocations in `repr(packed)` statics) are not even representable in the IR.

## Fix

- bump pliron, pliron-derive, and pliron-llvm to the upstream revision containing pliron#232 (both the root workspace and `crates/rustc-codegen-cuda` move together, they must share one rev or types stop unifying)
- all 31 synthetic construction sites pass `StructLayout::Unpacked` explicitly
- the 3 paths that rebuild an existing struct (enum payload storage, type conversion, the byte-faithful twin) preserve the source layout instead of assuming one
- adapt generated intrinsic construction and its generator, plus `Identifier` uses, to the 0.17 API
- review follow-up c17e49e5: the textual exporter now fails closed on a `Packed` struct with a clear error instead of silently printing it as unpacked `{ ... }` (#931 replaces this guard with real `<{ ... }>` printing)

## Gotchas

- this PR makes packed representable but never uses it: no cuda-oxide code selects `Packed` yet, and behavior is unchanged
- the bump carries 28 upstream pliron commits beyond the struct API (dialect conversion, hashing, containers), which is why validation below is the full sweep
- merge order: this lands before #931, which builds the first real packed carriers on top

## Testing and validation

- full workspace tests, clippy on both workspaces, `cargo deny check sources` green on the pliron-org pin
- generated-intrinsics consistency check, ABI history intact at 986 entries, all 986 intrinsic probes
- full GPU smoketest on sm_120: 209/209 pass
- CI green (43/43)


nihalpasham:
**Gist:** clean migration, and thanks for the quick re-pin to upstream after pliron#232 merged. All 31 construction sites pass an explicit layout, the three rebuild paths preserve the source layout, and nothing in cuda-oxide selects `Packed` yet, which is the right scope for this PR.

What I checked and did (this one got the full treatment since the bump carries 28 upstream pliron commits, not just the struct API):
- full workspace tests, clippy on both workspaces, `cargo deny check sources` (green on the pliron-org pin)
- generated-intrinsics regen check, ABI history intact at 986 entries, all 986 intrinsic probes
- full GPU smoketest on sm_120: 209/209 pass
- pushed c17e49e5: a fail-closed arm in the textual exporter so a `Packed` struct errors out instead of silently printing as unpacked `{ ... }`; #931 replaces this guard with real `<{ ... }>` printing
- CI fully green

Merge-order note: this lands before #931 so that branch can rebase down to just its relocation work. Closes #918 on merge.

nihalpasham (approved):

PR #921, opened 2026-08-15
fix(nvvm): legalize legacy integer RMW and compare-exchange

### LINKED ISSUE #920: legacy NVVM rejects integer atomic RMW and compare-exchange operations
## Problem

The legacy NVVM legalization pass currently rejects integer `atomicrmw` and
`cmpxchg` operations even when they can be represented directly by the LLVM 7
IR accepted by legacy libNVVM.

This prevents otherwise valid integer atomics from reaching the legacy NVVM
backend. In particular, code using integer RMW operations or compare-exchange
can lower successfully through MIR and LLVM dialect conversion, but is rejected
during `legalize_for_legacy_nvvm`.

## Current behavior

`legalize_for_legacy_nvvm` handles floating-point atomic add by rewriting it to
the corresponding NVVM intrinsic, but other atomic RMW and compare-exchange
operations are rejected before LLVM IR export.

As a result, legacy NVVM cannot compile supported forms such as:

- `atomicrmw` on `i32` / `i64`
- `cmpxchg` on `i32` / `i64`
- global, shared, and generic address spaces
- device synchronization scope

even though these forms are representable by the legacy LLVM/NVVM dialect.

## Proposed scope

Allow the subset that can be preserved without weakening semantics:

### Integer `atomicrmw`

Support:

- widths: `i32`, `i64`
- address spaces: generic (`0`), global (`1`), shared (`3`)
- synchronization scope: `device`
- ordering: `monotonic` after the existing fence-splitting lowering
- operations:
  - `xchg`
  - `add`
  - `sub`
  - `and`
  - `or`
  - `xor`
  - `max`
  - `min`
  - `umax`
  - `umin`

Continue rejecting unsupported forms such as `nand`, unsupported widths,
address spaces, scopes, or orderings.

### Integer `cmpxchg`

Support:

- widths: `i32`, `i64`
- address spaces: generic (`0`), global (`1`), shared (`3`)
- synchronization scope: `device`
- normal LLVM success orderings
- `monotonic` failure ordering

Stronger failure orderings should continue to fail closed because legacy
libNVVM does not preserve them faithfully.

## Acceptance criteria

- Supported integer `atomicrmw` and `cmpxchg` operations survive legacy NVVM
  legalization unchanged.
- Unsupported combinations fail before the module is mutated.
- Existing floating-point atomic-add legalization remains unchanged.
- LLVM 7 export uses the expected typed-pointer atomic syntax.
- An end-to-end regression example is verified and compiled by real libNVVM.

uurl:
Closes #920

## Problem

- On the legacy NVVM path (needed for LTO-IR interop, GPUs below sm_100), plain integer atomics did not compile at all:

```text
// kernel code
counter.fetch_add(1, Ordering::Relaxed);
slot.compare_exchange(7, 11, Ordering::Relaxed, Ordering::Relaxed);

Before:  error: legacy NVVM cannot legalize atomic read-modify-write operations
         error: legacy NVVM cannot legalize atomic compare-exchange operations
After:   atomicrmw add i32 addrspace(1)* ... syncscope("device") monotonic
         cmpxchg i32* ... syncscope("device") monotonic monotonic
         // both survive to LLVM 7 IR and libNVVM compiles them natively
```

- The blanket rejection was too broad: LLVM 7 can represent a useful subset natively (i32/i64, address spaces 0/1/3, device scope, all ten RMW kinds).

## Fix

- new validators whitelist the representable subset and let it through as bare `atomicrmw`/`cmpxchg`; everything else still fails with a targeted error
- failed validation is transactional: the module is byte-identical after a rejection, nothing half-rewritten
- review follow-up b344b5a4, from measuring what libNVVM actually emits: ordered success orderings on `cmpxchg` are silently dropped by legacy libNVVM (an `acq_rel` CAS came back as a bare `atom.cas` with no fences), so the validator now requires monotonic success ordering rather than accepting semantics the backend cannot deliver:

```text
cmpxchg ... acq_rel monotonic  ->  libNVVM emits: atom.global.cas.b32  (no ordering!)
so:  legacy NVVM compare-exchange requires monotonic success ordering
     because libNVVM lowers ordered cmpxchg to an unordered atom.cas
```

## Gotchas

- ordered RMW is fine: mir-lower fence-splits it before the legalizer, and the PTX shows `fence.acq_rel.gpu` around `atom.global.add` (verified)
- block/system scopes and ordered CAS are the follow-up lane: #922 handles them via inline PTX (see #923)
- the modern dialect (sm_100+) is untouched; it honors ordered CAS natively

## Testing and validation

- PTX inspected from real libNVVM (CUDA 13.3) for the native lane, the fence-split RMW lane, and the modern dialect
- nvvm-transforms suite green; before==after assertions on every rejection
- `legacy_atomic_rmw_cas` example runs on GPU (sm_120) and through the NVVM-verify/LTOIR smoketest categories
- CI green (43/43)


nihalpasham:
**Gist:** the native-subset legalization is right and the transactional validation is appreciated. One thing: whether legacy libNVVM actually honors ordered success orderings on `cmpxchg`. I tested it, and it does not.

What I found and did:
- built the example's legacy IR (`cmpxchg ... syncscope("device") acq_rel monotonic`) and compiled it through real libNVVM (CUDA 13.3) at compute_90: the PTX is a bare `atom.global.cas.b32/.b64` with no ordering qualifier and no fences anywhere in the kernel. So accepting ordered success silently weakened it to relaxed
- contrast checks: the fence-split RMW path emits `fence.acq_rel.gpu` around `atom.global.add` (honored), and the modern dialect emits `atom.acq_rel.gpu.cas` (honored), so the gap is specific to legacy CAS
- pushed b344b5a4: `validate_integer_cmpxchg` now requires monotonic success ordering, mirroring your failure-ordering check and error style; the example's native-lane CAS moved to Relaxed/Relaxed with a comment pointing at #922 for ordered CAS
- pushed d6ae41de: rustfmt on the example (examples sit outside the workspace, easy to miss)
- 27 nvvm-transforms tests green, GPU run passes on sm_120, CI fully green

The strengthening (routing ordered CAS through inline PTX instead of rejecting) lands in #923, which has the machinery for it. Closes #920 on merge.

nihalpasham (approved):

PR #923, opened 2026-08-16
fix(nvvm): legalize scoped integer atomics

### LINKED ISSUE #922: legacy NVVM cannot preserve scoped integer atomics and strong CAS failure orderings
## Problem

With #921 applied, the legacy NVVM path can preserve the native LLVM 7 integer atomic subset handled by #920 / #921, but some valid CUDA atomic semantics still cannot be represented faithfully through legacy LLVM `atomicrmw` and `cmpxchg`.

In particular:

- block-scoped integer atomics require CTA scope;
- system-scoped integer atomics require system scope;
- legacy libNVVM does not faithfully preserve stronger `cmpxchg` failure orderings.

Passing these forms through as ordinary legacy LLVM atomics would either reject valid programs or silently weaken their synchronization semantics.

## Current behavior

The legacy NVVM legalizer currently accepts only the directly representable integer atomic subset:

- `i32` / `i64`;
- generic, global, and shared address spaces;
- device synchronization scope;
- monotonic integer RMW after the existing fence-splitting lowering;
- compare-exchange with monotonic failure ordering.

Block and system scopes, as well as stronger compare-exchange failure orderings, are rejected.

## Proposed scope

Extend legacy integer atomic legalization by rewriting the forms that cannot be represented faithfully by legacy LLVM/NVVM to scoped inline PTX.

### Integer `atomicrmw`

For supported `i32` / `i64` integer RMW operations:

- keep device-scope operations that are directly representable as native LLVM `atomicrmw`;
- rewrite block-scoped operations to CTA-scoped PTX atomics;
- rewrite system-scoped operations to system-scoped PTX atomics;
- preserve stronger source ordering through the existing fence-splitting path.

Supported RMW operations remain:

- `xchg`
- `add`
- `sub`
- `and`
- `or`
- `xor`
- `max`
- `min`
- `umax`
- `umin`

### Integer `cmpxchg`

Keep the directly representable device-scope / monotonic-failure subset as
native LLVM `cmpxchg`.

Rewrite to inline PTX when:

- the synchronization scope is block;
- the synchronization scope is system;
- the failure ordering requires stronger semantics than monotonic.

The rewrite must preserve or strengthen the requested success/failure ordering; it must never silently weaken the source semantics.

### Target requirements

PTX-rewritten scoped/order-aware forms require `sm_70` or newer because they use PTX memory-semantic qualifiers and fences.

The directly representable native LLVM subset remains available independently of this rewrite path.

## Acceptance criteria

- Device-scope atomics already supported by the native LLVM 7 path remain  unchanged.
- Block-scoped integer RMW operations lower to CTA-scoped PTX atomics.
- System-scoped integer RMW operations lower to system-scoped PTX atomics.
- Block- and system-scoped integer compare-exchange operations lower to scoped  PTX CAS.
- Stronger compare-exchange failure orderings are not silently discarded.
- Invalid widths, address spaces, synchronization scopes, ordering pairs, and unsupported target capabilities continue to fail closed.
- Validation remains transactional: an invalid later atomic operation must not leave earlier operations partially rewritten.
- An end-to-end regression fixture executes successfully and is accepted and compiled by real libNVVM.

uurl:
Closes #922

## Problem

- After #921, the legacy NVVM path keeps the natively representable atomics. But CUDA semantics that LLVM 7 cannot carry were still rejected outright:

```text
// kernel code
block_counter.fetch_add(1, scope = block);          // cta scope
sys_flag.compare_exchange(0, 1, AcqRel, Acquire);   // strong failure ordering

Before:  error (legacy NVVM cannot represent scoped or strongly-ordered forms)
After:   atom.relaxed.cta.add.u32  %r1, [%rd1], %r2;      // inline PTX rewrite
         fence.sc.gpu; atom.acq_rel.gpu.cas.b32 ...;      // ordering made explicit
```

## Fix

- forms LLVM 7 cannot carry are rewritten to scoped inline PTX instead of rejected: block scope becomes `.cta`, system becomes `.sys`, and orderings are spelled into the `atom.{sem}.{scope}` instruction itself (with `fence.sc.{scope}` prepended for SeqCst)
- plan-then-apply: analysis is fully separate from mutation, and any validation failure leaves the module byte-identical
- sm_70 floor enforced (scoped atomics with semantics need it); below that, a clear target error
- review follow-up 13d38875: #921 proved libNVVM drops ordered success orderings on native CAS, so `needs_rewrite` now also fires when the success ordering is not monotonic. An ordered device-scope CAS becomes `atom.acq_rel.gpu.cas` through this PR's own rewrite lane instead of being rejected (the failure-ordering trigger is subsumed: LLVM requires failure <= success)

## Gotchas

- the two lanes split cleanly: fully monotonic device-scope forms stay native LLVM 7 (from #921); anything scoped or ordered goes through the inline-PTX rewrite lane
- merge order: after #921 (this branch builds on it)

## Testing and validation

- 33 nvvm-transforms tests, including rewrite assertions for all four ordered success cases and byte-identical-module checks on every rejection
- GPU (sm_120): example PASS including block/system scoped RMW and CAS, strong-failure CAS, and the new ordered-success device CAS
- sm_90 `--materialize-cubin` build: legacy libNVVM accepted the rewritten inline asm end to end (all kernels compiled by ptxas)
- CI green (43/43)


nihalpasham:
**Gist:** the scoped inline-PTX rewrite lane is well built: plan-then-apply, sm_70 floor, `fence.sc` for SeqCst, and byte-identical modules on every validation failure. I rebased it onto #921's finding and made it do one more thing that finding enabled.

What I found and did:
- #921 proved legacy libNVVM silently drops ordered success orderings on native `cmpxchg`, and its fix rejects those forms. But your rewrite lane can express them, so rejection would have been leaving capability on the table
- pushed the rebase plus 13d38875: `needs_rewrite` now also fires when success ordering is not monotonic, so a device-scope AcqRel CAS becomes `atom.acq_rel.gpu.cas` instead of an error (the failure-ordering clause is subsumed since LLVM requires failure <= success; there is a comment explaining that)
- confirmed `cmpxchg_ptx_semantics` already derives the token from the success ordering, and device scope maps to `.gpu`
- new test covers all four ordered successes with monotonic failure; the example gained a device-scope ordered CAS case
- GPU run passes on sm_120 including the new case, and an sm_90 `--materialize-cubin` build proves legacy libNVVM accepts the rewritten inline asm end to end
- pushed f10d8baa: clippy allow on the 8-arg kernel, matching the pattern other examples use
- 33 nvvm-transforms tests green, CI fully green

Merges after #921. Closes #922 on merge.

nihalpasham (dismissed):

nihalpasham (approved):

PR #925, opened 2026-08-16
perf(mir-importer): promote struct array constants

### LINKED ISSUE #924: Support immutable-global promotion for struct array constants
## Summary

Array constant promotion currently supports primitive scalars, field-less enums,
tuples, and recursively supported nested arrays, but structs remain outside the
immutable-device-global promotion path.

This means read-only struct tables can still be materialized through the local
value path even when their complete byte representation is known at compile
time and contains no pointer relocations.

The same limitation also affects pointer-to-array constants such as:

```rust
#[repr(C)]
struct Entry {
    tag: u8,
    value: u32,
}

const TABLE: [Entry; 2] = [
    Entry { tag: 1, value: 10 },
    Entry { tag: 2, value: 20 },
];

const TABLE_REF: &[Entry; 2] = &TABLE;
```

## Proposed behavior

Allow arrays whose element type is a struct to use the existing immutable-global
promotion path when every field is recursively promotable.

The existing promotion invariants should continue to apply:

* the converted storage size must match rustc's layout
* the initializer must be representable as a deterministic byte image
* pointer relocations / provenance-sensitive initializers must remain rejected
* unsupported nested leaves must remain rejected
* immutable globals should continue to use the existing deduplication mechanism

Struct fields may recursively contain already-supported promotable types,
including nested arrays, tuples, primitive scalars, field-less enums, and other
supported structs.

## Important boundary

Zero-byte over-aligned struct leaves such as `#[repr(align(N))]` ZSTs should
remain on the existing alignment-sensitive value path.

Such a type can raise the ABI alignment of an enclosing aggregate without
contributing bytes to its LLVM storage representation, so treating it as an
ordinary promotable struct could lose an alignment constraint.

## Out of scope

This issue does not propose support for:

* pointer-bearing struct fields
* union-bearing struct fields
* payload enums that are not already promotable
* relocation/provenance-sensitive initializers
* pointer-to-array union constants
* packed struct arrays whose rustc stride differs from LLVM's natural layout

## Acceptance criteria

* bare read-only arrays of compatible structs use immutable-device-global
  promotion
* `const R: &[Struct; N] = &TABLE` can reference the same promoted global
* padded and nested structs are covered by runtime tests
* LLVM code-shape tests verify global promotion and deduplication
* existing over-aligned ZST behavior remains unchanged
* existing unsupported aggregate cases remain rejected

uurl:
Closes #924

## Problem

- Constant tables of structs were rebuilt element by element in every kernel invocation instead of living in device memory as one read-only global:

```text
const TABLE: [Pair; 4] = [Pair { key: 1, val: 10 }, ...];

Before (per kernel, per thread):
  %slot = alloca [4 x { i32, i32 }]
  store ... x8                        // rebuild the table in local memory
After (once per module):
  @__const_table = addrspace(1) constant [32 x i8] c"\01\00\00\00\0A..."
  // kernels just index it; duplicate tables dedup to one global
```

- Scalar and tuple element types already promoted this way; struct elements did not.

## Fix

- `promotable_array_element` gains a struct arm: a struct element is promotable when all of its fields are
- `const R: &[Pair; N] = &TABLE` references the same promoted global; duplicates dedup to one initializer
- pointer-bearing, enum-payload, and over-aligned ZST cases stay excluded on purpose (relocation and alignment paths keep their existing behavior)
- review follow-up 3b173084: `repr(C, packed)` struct elements are also excluded. Their LLVM natural layout cannot reproduce rustc's offsets, so promotion would read wrong bytes:

```text
#[repr(C, packed)] struct P { a: u8, b: u32 }   // rustc: b at offset 1
natural LLVM layout of { i8, i32 }              // b at offset 4
// promoted global + natural layout = reads b from the wrong place
// now: fails the layout-agreement check, falls back, still compiles
```

## Gotchas

- the exclusion check recurses, so a packed struct nested anywhere in the element type also blocks promotion
- a shared benefit: pointer-to-array constants of packed structs now get a hard diagnostic instead of quietly using the wrong layout

## Testing and validation

- mir-importer suite green (1069 tests), including new promote, dedup, reference-sharing, and packed-fails-closed tests
- GPU (sm_120): `array_constants` PASS, `verify-code-shape.sh` confirms the promoted byte-image global in the IR
- CI green (43/43)

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -6572,13 +6572,13 @@ fn struct_layout_matches_llvm_natural(
             continue;
         }
         let offset = field_offsets[decl_idx];
-        if offset % align != 0 || offset < end {
+        if !offset.is_multiple_of(align) || offset < end {
             return false;
         }
         end = offset + size;
         max_align = max_align.max(align);
     }
-    total_size >= end && total_size % max_align == 0
+    total_size >= end && total_size.is_multiple_of(max_align)
 }
 
 /// Natural size and alignment of the LLVM storage a dialect type converts to,
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -6421,13 +6421,13 @@ pub(crate) fn emit_promoted_immutable_global(
 /// bare-array value admission: an unpromotable bare value can fall back to
 /// element-wise materialization, while a pointer-to-array constant cannot.
 ///
-/// Admits primitive scalars, enums carrying no payload, tuples whose every field
-/// is itself admissible, and nested arrays of any of those. `ty` is the array
-/// type, and nesting is walked so an unsupported leaf cannot hide inside it. A
-/// zero-length array passes for any element type: its
-/// initializer is empty and nothing can ever be read through it, which is what
-/// admits a promoted empty-slice constant such as `&[]` (rustc promotes it to
-/// `&[T; 0]`) regardless of `T`.
+/// Admits primitive scalars, enums carrying no payload, tuples and structs whose
+/// every field is itself admissible, and nested arrays of any of those. `ty` is
+/// the array type, and nesting is walked so an unsupported leaf cannot hide
+/// inside it. A zero-length array passes for any element type: its initializer
+/// is empty and nothing can ever be read through it, which is what admits a
+/// promoted empty-slice constant such as `&[]` (rustc promotes it to `&[T; 0]`)
+/// regardless of `T`.
 ///
 /// Tuples are admitted when every field is itself promotable. That only became
 /// worth doing once a tuple field read stopped going through a copy of the whole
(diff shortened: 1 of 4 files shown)


nihalpasham:
**Gist:** matches how tuples and enum payloads already promote. One hole closed in review: `repr(C, packed)` struct elements, where LLVM's natural layout cannot reproduce rustc's offsets, so the promotion invariant (byte image and typed layout agree) silently broke.

What I checked and did:
- confirmed the hazard is real: a packed `{u8, u32}` table has its field at rustc offset 1, but the natural-layout global would read it at offset 4, which is wrong bytes, not a compile error
- pushed 3b173084: an importer-side layout-agreement walk (`struct_layout_matches_llvm_natural`) that requires naturally aligned, non-overlapping recorded offsets; anything divergent fails closed and takes the existing fallback path. A shared benefit: pointer-to-array constants of packed structs now get a hard diagnostic instead of a miscompile
- new test: packed struct tables do not promote and still compile via the fallback; naturally laid out structs with full recorded layout keep promoting
- pushed 6ac14e59: two clippy nits (`is_multiple_of`)
- 1069 mir-importer tests green, GPU `array_constants` PASS, `verify-code-shape.sh` PASS, CI fully green

Closes #924 on merge.

nihalpasham (approved):

PR #927, opened 2026-08-16
fix(mir-importer): preserve provenance in thin-pointer union constants

### LINKED ISSUE #926: mir-importer rejects thin-pointer union constants with relocations
**Description**

The MIR importer currently rejects initialized union constants when their storage contains pointer relocations. Pointer-free union constants are materialized from their evaluated byte image, but applying the same strategy to pointer-bearing unions would lose pointer provenance because the allocation bytes only contain the relocation placeholder/addend, not a reconstructible pointer value.

A useful conservative subset can be supported when every non-ZST union field is a compatible thin pointer: reconstruct one typed pointer carrier from the relocation and insert it into the union storage while preserving provenance.

**Minimal reproducer**

```rust
#[derive(Clone, Copy)]
#[repr(C)]
union PointerBits {
    word: *const u32,
    bytes: *const u8,
}

static VALUES: [u32; 3] = [11, 17, 23];

const DIRECT: PointerBits = PointerBits {
    word: &VALUES[0] as *const u32,
};

const TABLE: [PointerBits; 2] = [
    PointerBits {
        word: &VALUES[0] as *const u32,
    },
    PointerBits {
        bytes: &VALUES[2] as *const u32 as *const u8,
    },
];
```

Using these constants from device code requires preserving both the relocation target and, for the second table entry, the non-zero byte addend.

**Expected behavior**

Thin-pointer-only union constants should preserve pointer provenance in the same way that pointer fields in array, tuple, struct, and supported enum constants do.

The importer should:

- reconstruct an exact thin pointer from a relocation at the union storage base;
- preserve non-zero static byte addends;
- materialize the union using a typed pointer carrier rather than raw integer/byte reconstruction;
- continue to reject cases where a single typed carrier is not sufficient to represent the storage safely.

In particular, pointer/integer overlapping unions, fat or nested pointer storage, and over-aligned/padded pointer unions can remain unsupported initially.

**Actual behavior**

Pointer-free initialized union constants are supported, but union constants containing pointer relocations are rejected rather than reconstructed through a provenance-preserving pointer representation.

**Additional context**

This is a follow-up to the initialized union constant support added previously. The intended scope is deliberately narrow: support compatible thin-pointer-only union constants without weakening the existing fail-closed behavior for ambiguous pointer/integer or fat-pointer storage.

uurl:
Fixes #926

## Problem

- A union constant whose fields are all thin pointers could not hold a real pointer:

```text
static POINTER_VALUES: [u32; 3] = [11, 17, 23];

union PointerBits {
    word:  *const u32,
    bytes: *const u8,
}
const P: PointerBits = PointerBits { bytes: &POINTER_VALUES[2] as *const u32 as *const u8 };

Before:  error (importer rejects union constants carrying pointer relocations)
After:   the union materializes as one typed pointer to POINTER_VALUES + 8 bytes
         and the kernel reads 23 through it at runtime
```

- The pointer inside a constant is a relocation (a symbol plus a byte offset), not a number. Dropping it to raw bytes would forge a pointer; the importer used to just refuse instead.

## Fix

- classify the union's storage: if every non-zero-sized field is a thin pointer in the same address space, the whole union is one typed pointer slot
- reconstruct that pointer from the relocation at the union base, keeping non-zero byte addends (the `+ 8` above survives)
- build the union value through the dialect's own undef + insert-field ops, so downstream lowering sees a normal typed value
- everything else stays rejected with targeted errors: pointer/integer mixes, fat or nested pointer storage, mixed address spaces, over-aligned or padded pointer unions

## Gotchas

- this touches only constant-operand materialization; device-global initializers with union relocations are a separate path and stay rejected (documented)
- the issue's reproducer was added verbatim to the `array_constants` example with runtime value checks

## Testing and validation

- mir-importer suite green (1068 tests)
- GPU (sm_120): `array_constants` PASS including the pointer-union provenance check (reads 11 and 23 through the reconstructed pointers)
- `verify-code-shape.sh` PASS, compile smoketest PASS
- CI green (43/43)


nihalpasham:
**Gist:** the classification is properly fail-closed (thin same-address-space pointers only, everything else keeps its explicit error), the typed-carrier reconstruction goes through the existing relocation decoder with addends intact, and the issue's own reproducer runs on GPU inside the example.

What I verified:
- 1068 mir-importer tests green
- GPU on sm_120: `array_constants` PASS including the pointer-union provenance check (the 8-byte addend lands on `POINTER_VALUES[2]` as specified in #926)
- `verify-code-shape.sh` PASS, compile smoketest PASS, CI fully green

Two follow-ups worth issues later, not blockers: fully-uninitialized pointer unions (the MaybeUninit case) and relocation-free pointer/integer unions via a byte image. Closes #926 on merge.

nihalpasham (dismissed):

nihalpasham (approved):

PR #929, opened 2026-08-16
fix(mir-importer): preserve slice provenance in aggregate constants

### LINKED ISSUE #928: mir-importer: preserve slice provenance in aggregate constants
## Problem

Aggregate constants can contain slice fat-pointer fields whose data word carries pointer provenance to a Rust static while the metadata word stores the slice length.

The MIR importer currently preserves provenance for thin pointer fields in aggregate constants, but rejects slice fields when their allocation range contains provenance.

This prevents constants such as:

```rust
#[repr(C)]
struct View {
    tag: u32,
    values: &'static [u32],
}

static TABLE: [[u32; 4]; 3] = [
    [10, 11, 12, 13],
    [20, 21, 22, 23],
    [30, 31, 32, 33],
];

const VIEW: View = View {
    tag: 7,
    values: &TABLE[1],
};
```

from being imported even though standalone static array-to-slice constants and thin aggregate pointer relocations are already supported.

## Expected behavior

The MIR importer should:

* preserve the data-pointer provenance of slice fields stored in aggregate constants;
* decode the literal `usize` slice-length metadata independently;
* support zero and non-zero byte addends into Rust device statics;
* work for nested aggregate fields at non-zero allocation offsets;
* reject malformed or unsupported relocation shapes instead of reconstructing provenance-bearing pointers from raw bytes.

## Scope

This issue only covers slice fat pointers whose data pointer resolves to a Rust device static and whose pointee is a same-element array-to-slice view.

Out of scope:

* trait-object / vtable metadata;
* custom DST metadata;
* anonymous promoted allocation targets;
* pointer-bearing union constants;
* other unsupported fat-pointer representations.

## Acceptance criteria

* aggregate slice constants with static provenance import successfully;
* non-zero static byte addends are preserved;
* nested struct fields are supported;
* metadata-word provenance and overlapping/ambiguous relocations are rejected;
* existing thin-pointer aggregate behavior remains unchanged;
* runtime regression passes on GPU.

uurl:
Closes #928

## Problem

- A slice stored in a struct constant lost its pointer identity:

```text
static TABLE: [u32; 8] = [...];

struct View { name: u32, data: &'static [u32] }
const V: View = View { name: 7, data: &TABLE[2..5] };

Before:  error: aggregate constant contains a fat pointer (slice) with
         provenance; only thin pointers to device statics are supported
After:   data.ptr  -> relocation to TABLE + 8 bytes   (provenance kept)
         data.len  -> 3                               (decoded from the metadata word)
```

- A slice is two words: a data pointer (a relocation, must stay symbolic) and a length (a literal, must be read as a number). The old path refused any aggregate constant containing one.

## Fix

- slice fields in aggregate constants now route to a dedicated decoder: the data word resolves through the existing addend-aware relocation path, the length word is read independently as an integer
- works at any nesting depth (a slice inside a struct inside a struct) and with non-zero addends into the target static
- malformed layouts are rejected with specific errors: provenance on the metadata word, duplicate relocation anchors, relocations crossing field boundaries; each has a unit test
- doc follow-up d2a3ecca: while verifying, the supported-features wording turned out stale in both directions; it now states the boundary the code actually implements (enum payload relocations work, including tuple/struct/array nesting; a relocation-carrying enum nested inside another constant stays rejected)

## Gotchas

- pointer-bearing union constants are the mirror-image case and belong to #927, not this PR
- device-global initializer relocations were already supported on main; the old doc line claiming otherwise was removed

## Testing and validation

- mir-importer suite green (1072 tests, four new slice-relocation tests)
- GPU (sm_120): `struct_constant_provenance` prints PASS with pointer and slice provenance verified at runtime, including a nested field at a non-zero offset
- full compile smoketest 208/208
- CI green (43/43)

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -10228,7 +10228,94 @@ fn interior_array_to_slice_unsize_info(
         .map(|remaining_len| (slice_elem, remaining_len)))
 }
 
-/// Read the slice length from a fat-pointer constant's metadata word.
+/// Validate the relocation shape of one slice fat pointer stored inside an
+/// allocation.
+///
+/// A slice occupies two pointer-width words. Exactly one relocation must back
+/// the data word at `fat_ptr_offset`; the metadata word is a literal `usize`
+/// and therefore must not overlap another relocation. Sibling relocations that
+/// end exactly at the field start or begin exactly at the field end are fine.
+///
+/// Returns `(metadata_offset, fat_pointer_end)` on success. Generic over the
+/// provenance payload so the boundary rules are unit testable without a rustc
+/// session.
+fn validate_slice_relocation_shape<P>(
+    ptrs: &[(usize, P)],
+    fat_ptr_offset: usize,
+    pointer_width: usize,
+) -> Result<(usize, usize), String> {
+    if pointer_width == 0 {
+        return Err("slice fat pointer has zero-width target pointers".to_string());
+    }
+
+    let metadata_offset = fat_ptr_offset
+        .checked_add(pointer_width)
+        .ok_or_else(|| "slice fat-pointer data-word end overflowed".to_string())?;
+    let fat_pointer_end = metadata_offset
+        .checked_add(pointer_width)
+        .ok_or_else(|| "slice fat-pointer metadata-word end overflowed".to_string())?;
+
+    let anchored = ptrs
+        .iter()
+        .filter(|(offset, _)| *offset == fat_ptr_offset)
+        .count();
+    if anchored != 1 {
+        return Err(format!(
+            "Slice fat pointer at byte {fat_ptr_offset} has {anchored} provenance entries at its \
+             data-word start; expected exactly one"
+        ));
+    }
+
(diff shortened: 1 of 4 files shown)


nihalpasham:
**Gist:** the slice-relocation validators are thorough (metadata-word provenance, duplicate anchors, and crossing relocations all rejected, each with a test) and the addend-aware data-word resolution reuses the existing static decoder rather than inventing a parallel path.

What I verified and did:
- 1072 mir-importer tests green, GPU `struct_constant_provenance` PASS on sm_120, full compile smoketest 208/208
- one surprise from probing the actual boundary on GPU: the supported-features wording was stale in both directions. Relocations nested inside tuple/struct/array fields of enum payloads already work (that predates this PR), and what genuinely stays rejected is a relocation-carrying enum nested inside another constant's field
- pushed d2a3ecca restating that boundary in the book so the docs match the code I can demonstrate
- CI fully green

Worth a small follow-up test: an aggregate-nested enum payload relocation exercised at runtime (the current example only covers direct thin-reference payloads). Closes #928 on merge.

nihalpasham (approved):

PR #931, opened 2026-08-16
fix(mir-lower): preserve relocations in packed static initializers

### LINKED ISSUE #930: Support pointer relocations in packed static initializers
## Problem

Static initializers containing pointer relocations currently require relocation slots to be naturally aligned.

This rejects valid Rust statics such as:

```rust
#[repr(C, packed)]
struct Entry {
    tag: u8,
    ptr: &'static u32,
}

static TARGET: u32 = 0x1234_5678;

static ENTRY: Entry = Entry {
    tag: 0x7b,
    ptr: &TARGET,
};
```

For this layout, rustc places:

* `tag` at byte offset 0
* `ptr` at byte offset 1
* total size at 9 bytes
* ABI alignment at 1

The relocation itself is valid, but the current lowering rejects it because the
pointer slot is not naturally aligned.

A naturally laid out LLVM carrier such as:

```llvm
{ [1 x i8], i64 }
```

cannot represent the Rust allocation byte-for-byte because LLVM aligns the
`i64` field naturally.

## Expected behavior

Relocation-bearing packed statics should preserve the exact rustc allocation
layout while keeping pointer provenance as an LLVM relocation.

For the example above, the physical initializer carrier can be represented as:

```llvm
<{ [1 x i8], i64 }>
```

with alignment 1.

The packed LLVM struct should only be used as the physical storage carrier for
the initialized global. The semantic Rust aggregate should remain on the
existing fail-closed path for unsupported packed by-value operations.

## Scope

Initial support should remain limited to:

* top-level structs with explicit rustc layout metadata
* thin-pointer relocations between supported device statics
* non-overlapping relocation and field ranges
* packed or otherwise under-aligned relocation slots
* existing relocation addend and address-space semantics

This should not add support for:

* packed aggregates passed or returned by value
* fat pointers
* trait objects or vtables
* anonymous promoted allocation targets
* function relocations
* unsized pointees
* overlapping field layouts
* changes to the relocation metadata format

uurl:
Closes #930

## Problem

- A device static with `repr(packed)` layout could not hold a pointer, because the pointer lands at an unaligned offset:

```text
#[repr(C, packed)]
struct Entry { tag: u8, ptr: &'static u64 }

#[device_global]
static ENTRY: Entry = Entry { tag: 123, ptr: &VALUE };   // ptr sits at offset 1

Before:  error: 8-byte pointer slots must be naturally aligned
After:   @ENTRY = addrspace(1) global <{ [1 x i8], i64 }>
                  <{ c"\7B", i64 ptrtoint(@VALUE) }>, align 1
         // packed carrier: pointer slot at offset 1, exactly where rustc put it
```

- The initializer needs a relocation (the address of `VALUE` is only known at link time), and the old lowering refused any relocation slot that was not naturally aligned.

## Fix

- storage for relocated initializers switches to a packed LLVM struct whenever the allocation's alignment or a relocation's offset requires it, instead of bailing out
- a layout validator checks the carrier before emission: top-level explicit-layout structs only, non-overlapping ranges; nested packed stays rejected
- the textual exporter learns to print packed structs (`<{ ... }>`), replacing #919's fail-closed guard
- kernels read the fields with unaligned loads, matching the align-1 storage

## Gotchas

- merge order: after #919 (this builds on the explicit StructLayout API it introduced)
- verified against the question the review left open: legacy LLVM-7 libNVVM accepts the packed carrier. An sm_90 `--materialize-cubin` build compiled the typed-pointer form (pointer at offset 1, and a getelementptr addend variant at offset 3) through libNVVM and ptxas with zero errors
- follow-up worth adding later: a kernel that reads a packed pointer field by value (`let p = ENTRY.ptr;`), which exercises a different translator path than the `addr_of! + read_unaligned` route the example covers

## Testing and validation

- mir-importer, mir-lower, and llvm-export suites green
- GPU (sm_120): `device_global` PASS with the expected values `[123, 305419896, 17, 1090519040, 17493]`
- modern NVVM path (sm_120) and legacy path (sm_90) both accept and compile the packed carrier
- CI green (43/43)


nihalpasham:
**Gist:** this is the carrier #930 asked for: packed storage only when needed, layout validation before emission, and the exporter prints the real `<{ ... }>` form. The verify-time invariants and transactional bail-outs made this easy to review.

What I verified and did:
- GPU on sm_120: `device_global` PASS with the expected values `[123, 305419896, 17, 1090519040, 17493]`
- the open review question is settled: legacy LLVM-7 libNVVM accepts the packed carrier. An sm_90 `--materialize-cubin` build compiled the typed-pointer form (`<{ [1 x i8], i64 }>` with the pointer at offset 1, and the getelementptr-addend variant at offset 3) through libNVVM and ptxas with zero errors, so no legacy gating is needed
- modern path at sm_120 emits the same carrier and runs correctly
- mir-importer, mir-lower, and llvm-export suites all green
- pushed c9e4c418: rustfmt on the example
- CI fully green

Two notes for after #919 lands: rebase to drop the duplicated pliron-migration hunks (your relocation work is untouched by it), and consider adding a by-value packed pointer read (`let p = PACKED_RELOCATION.ptr;`) since that MIR path differs from the `addr_of! + read_unaligned` route the example covers today. Closes #930 on merge.

nihalpasham (approved):

PR #933, opened 2026-08-16
test(codegen): validate spill-free two-slot WGMMA counted pipeline

### LINKED ISSUE #932: Validate spill-free two-slot WGMMA counted pipeline lowering
## Summary

Before combining the existing counted K-loop and partial-wait WGMMA lowering paths, validate that the backend can keep two independent WGMMA accumulator slots register-resident across a counted PTX loop without introducing stack traffic or spills.

The existing implementation already validates these properties separately:

- the counted K-loop carrier keeps one 32-value `f32` accumulator register-resident;
- the partial-wait pipeline carrier keeps two 32-value `f32` accumulator slots register-resident with `wait_group<1>`.

What is not yet validated is their composition: 64 tied accumulator values plus counted-loop control and affine descriptor recurrences in the same convergent inline-PTX lifetime.

## Proposed probe

Add a codegen test containing:

- two independent 32 × `f32` WGMMA accumulator slots;
- four descriptor bases;
- four affine descriptor increments;
- one counted loop trip counter;
- one `wgmma.fence.sync.aligned`;
- two static WGMMA instructions per loop iteration;
- one commit per accumulator slot;
- `wgmma.wait_group.sync.aligned 1` inside the loop;
- one final `wgmma.wait_group.sync.aligned 0`.

The test should compile the generated PTX with:

```text
ptxas -arch=sm_90a
```

## Acceptance criteria

The probe must:

* compile successfully through the existing CUDA Oxide codegen path;
* produce PTX targeting `sm_90a`;
* preserve one counted PTX loop;
* contain two static WGMMA accumulator tuples;
* contain the four descriptor recurrences;
* contain partial waits inside the loop and a final full drain;
* be accepted by `ptxas`;
* report `0 bytes stack frame`;
* report `0 bytes spill stores`;
* report `0 bytes spill loads`;
* not materialize `.local`, `ld.local`, or `st.local`.

## Scope

This issue only validates backend feasibility.

It does not add:

* a new NVVM dialect operation;
* MIR recognition for pipelined counted loops;
* a combined counted-loop/pipeline lowering path;
* dynamic `wait_group<N>`;
* arbitrary accumulator scheduling;
* arbitrary loop CFG support;
* F16 or TF32 variants.

If this probe succeeds without spills, a follow-up change can implement the combined counted-loop + partial-wait lowering path with substantially lower backend risk.

uurl:
Fixes #932 

## Summary

Add a backend feasibility probe for composing the existing register-resident counted WGMMA loop with a two-slot partial-wait pipeline.

The test keeps 64 tied `f32` accumulator values live across a counted PTX loop while also carrying:

- four WGMMA descriptor bases;
- four affine descriptor increments;
- one loop trip counter.

Each loop iteration issues two statically assigned accumulator slots, commits both groups independently, and uses `wait_group<1>` before reusing the slots. A final `wait_group<0>` drains the remaining asynchronous work.

## Why

CUDA Oxide already tests the two required properties independently:

- `wgmma_counted_loop_ptx.rs` validates a counted K-loop with one 32-value register-resident accumulator;
- `wgmma_pipeline_carrier_ptx.rs` validates two 32-value accumulator slots with partial waits.

The remaining backend risk is their composition.

This test isolates that question before introducing a combined MIR recognizer or NVVM carrier operation.

## Validation

Tested with CUDA 13.3 `ptxas` targeting `sm_90a`.

```text
wgmma_counted_loop_kernel:
    0 bytes stack frame
    0 bytes spill stores
    0 bytes spill loads
    Used 58 registers

wgmma_pipeline_carrier_kernel:
    0 bytes stack frame
    0 bytes spill stores
    0 bytes spill loads
    Used 90 registers

wgmma_counted_pipeline_carrier_kernel:
    0 bytes stack frame
    0 bytes spill stores
    0 bytes spill loads
    Used 90 registers
```

The combined probe therefore retains the same reported register count as the existing two-slot straight-line pipeline while adding the counted loop and descriptor recurrences.

The new test also checks that the generated PTX contains:

* one WGMMA fence;
* one counted loop;
* two static `wgmma.mma_async` instructions;
* two `wgmma.commit_group` instructions;
* two `wgmma.wait_group 1` instructions;
* one final `wgmma.wait_group 0`;
* four affine descriptor recurrences;
* no `.local`, `ld.local`, or `st.local`.

## Scope

This PR only adds the backend feasibility test.

It does not change:

* WGMMA dialect operations;
* MIR lowering or recognition;
* public APIs;
* accumulator scheduling;
* supported wait depths;
* supported WGMMA data types or shapes.

A follow-up change can use this result to implement the actual pipelined counted K-loop lowering.

### DIFF
--- a/crates/cuda-oxide-codegen/tests/wgmma_counted_pipeline_carrier_ptx.rs
+++ b/crates/cuda-oxide-codegen/tests/wgmma_counted_pipeline_carrier_ptx.rs
@@ -1,0 +1,426 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Proves that a two-slot BF16 WGMMA counted pipeline can keep sixty-four tied
+//! `f32` accumulator values spill-free while loop control and affine descriptor
+//! recurrences remain inside the same convergent inline-PTX lifetime.
+//!
+//! This is a backend feasibility probe. It intentionally bypasses the production
+//! counted-loop recognizer and combined carrier that a follow-up implementation
+//! may add. The test isolates the harder backend question first:
+//!
+//! ```text
+//! 64 tied f32 accumulator values
+//! 4 descriptor bases
+//! 4 descriptor deltas
+//! 1 trip count
+//! one PTX loop
+//! wait_group<1> throttling
+//! final wait_group<0>
+//! ```
+//!
+//! `ptxas -arch=sm_90a` must accept the generated PTX with zero stack frame and
+//! zero spill traffic.
+
+#![cfg(unix)]
+
+use cuda_oxide_codegen::experimental::{CodegenModule, CompileOptions, Compiler, Target};
+use dialect_mir::{
+    ops::{MirConstantOp, MirFuncOp, MirPtrOffsetOp, MirReturnOp, MirStoreOp},
+    types::MirPtrType,
+};
+use dialect_nvvm::ops::InlinePtxOp;
+use pliron::builtin::attributes::IntegerAttr;
+use pliron::builtin::types::{IntegerType, Signedness};
+use pliron::{
+    basic_block::BasicBlock,
+    builtin::{
+        attributes::TypeAttr,
+        op_interfaces::SymbolOpInterface,
+        types::{FP32Type, FunctionType},
+    },
+    linked_list::ContainsLinkedList,
+    op::Op,
+    operation::Operation,
+    utils::apint::APInt,
+};
+use std::num::NonZeroUsize;
+
+const ACCUMULATOR_LEN: usize = 32;
+const SLOT_COUNT: usize = 2;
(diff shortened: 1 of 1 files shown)


nihalpasham:
**Gist:** the feasibility probe #932 asked for: two BF16 WGMMA accumulator slots (64 tied f32 values), four descriptor recurrences, and a trip counter, all live inside one counted inline-PTX loop, checked against ptxas.

What I verified:
- ran the new test locally: ptxas -arch=sm_90a accepts the loop with 0 bytes stack frame, 0 spill stores/loads, no `.local`, and the same 90-register footprint as the straight-line two-slot probe
- operand numbering checks out: 64 `=f` outputs, 64 tied inputs, 9 `l` controls at \$128-\$136, and the descriptor updates go through asm-local `.reg .u64` temporaries so no input constraint gets clobbered
- wait protocol is sound: each `wait_group 1` retires the other slot's group before its registers are reused
- CI fully green

Closes #932 on merge. The lowering itself is #934, reviewed over in #935.

nihalpasham (approved):

PR #941, opened 2026-08-16
fix(mir-lower): support packed aggregates by value

### LINKED ISSUE #940: Support packed aggregates passed by value across the kernel ABI
## Problem

Packed aggregates passed or returned by value are currently rejected because their physical rustc layout cannot be represented faithfully by an ordinary LLVM struct.

For example:

```rust
#[repr(C, packed)]
struct Packed1 {
    a: u8,
    b: u32,
}

#[repr(C, packed(2))]
struct Packed2 {
    a: u8,
    b: u32,
}
```

have the following Rust layouts:

```text
Packed1: size = 5, align = 1, b @ offset 1
Packed2: size = 6, align = 2, b @ offset 2
```

These layouts cannot be represented faithfully by ordinary LLVM structs.

The expected LLVM representations are:

```llvm
<{ i8, i32 }>
<{ i8, [1 x i8], i32 }>
```

respectively.

There is an additional ABI complication for `repr(C, packed(2))`: an LLVM packed struct has natural alignment 1, while the Rust ABI alignment of the aggregate is 2. That alignment must therefore be preserved separately when the aggregate crosses the kernel parameter ABI.

## Scope

Add support for packed aggregates passed by value, including:

* construction of packed aggregate SSA values;
* whole-value loads and stores;
* packed aggregate array stride;
* kernel parameters passed by value;
* device-function aggregate arguments and returns;
* explicit padding required by `repr(packed(N))`;
* preservation of source ABI alignment at NVVM kernel boundaries.

Packed aggregates containing target-dependent shared-memory pointers should remain fail-closed until their physical representation is target-independent.

This should not change the existing packed field-addressing behavior.

## Expected result

For the examples above:

```text
Packed1 -> <{ i8, i32 }>
Packed2 -> <{ i8, [1 x i8], i32 }>
```

and the generated PTX for `Packed2` should retain its Rust ABI alignment:

```ptx
.param .align 2 .b8 ...[6]
```

The behavior should be covered by both lowering tests and an end-to-end CUDA kernel launch.

uurl:
Closes #940 

## Dependency

This PR is stacked on NVlabs/cuda-oxide#931, which in turn is stacked on NVlabs/cuda-oxide#919.

#919 updates cuda-oxide to the explicit `StructLayout::{Packed, Unpacked}` API, and #931 adds the packed LLVM struct sizing/export infrastructure reused by this change.

Once the prerequisite PRs land, this branch will be rebased onto `main`, leaving only the packed aggregate by-value and ABI alignment changes in this PR.

## Summary

Adds by-value support for Rust packed aggregates while preserving their physical rustc layout through MIR lowering, LLVM export, and the NVVM kernel ABI.

This extends the existing packed-field addressing support to whole aggregate values.

### Packed aggregate representation

Struct lowering now distinguishes between:

- layouts that differ from LLVM's natural struct layout; and
- layouts that can actually be represented faithfully by an LLVM packed struct.

A divergent layout is lowered as `StructLayout::Packed` only when sequential LLVM packed layout reproduces the recorded rustc field offsets and total size exactly.

This avoids incorrectly treating overlapping or union-like layouts as ordinary packed structs.

For example:

```rust
#[repr(C, packed)]
struct Packed1 {
    a: u8,
    b: u32,
}
```

lowers to:

```llvm
<{ i8, i32 }>
```

with size 5.

And:

```rust
#[repr(C, packed(2))]
struct Packed2 {
    a: u8,
    b: u32,
}
```

lowers to:

```llvm
<{ i8, [1 x i8], i32 }>
```

with size 6.

The explicit byte slot preserves the rustc field offset of `b`.

### By-value operations

Packed aggregates are now supported through:

* aggregate construction;
* whole-value loads;
* whole-value stores;
* arrays of packed aggregates;
* kernel arguments passed by value;
* internal device-function aggregate arguments and returns.

Layouts that cannot be represented faithfully still fail closed.

### Kernel ABI alignment

LLVM packed structs have natural alignment 1, which is insufficient for layouts such as:

```rust
#[repr(C, packed(2))]
```

whose Rust ABI alignment is 2.

MIR lowering now records non-natural source ABI alignment when it exceeds the natural alignment of the lowered LLVM type.

The LLVM exporter translates that into NVVM's `"align"` function property annotation.

NVVM encodes this as:

```text
(position << 16) | alignment
```

where argument positions are 1-based and position 0 denotes the return value.

For the first `Packed2` kernel parameter:

```text
(1 << 16) | 2 = 65538
```

This produces the expected PTX:

```ptx
.param .align 2 .b8 packed2_param_0[6]
```

rather than the previous:

```ptx
.param .align 1 .b8 packed2_param_0[6]
```

### Shared-memory pointer aggregates

Packed aggregates containing address-space-3/shared pointers remain unsupported by value.

Their physical pointer width is target-dependent between the rustc NVPTX model and modern NVVM, so these cases continue to fail closed rather than silently producing an incorrect ABI.

This does not affect existing packed field addressing.

## End-to-end coverage

Adds the `packed_aggregate_abi` example covering:

* `repr(C, packed)`;
* `repr(C, packed(2))`;
* host-to-kernel by-value argument passing;
* device helper argument/return paths;
* whole-value load/store;
* physical field offsets;
* PTX aggregate size;
* PTX parameter alignment.

The runtime test executes successfully on CUDA and verifies:

```text
Packed1: size 5, align 1
Packed2: size 6, align 2
```

with generated PTX including:

```ptx
.param .align 2 .b8 packed2_param_0[6]
.param .align 2 .b8 store_packed2_param_0[6]
```

## Validation

```text
mir-lower unit tests:        234 passed
mir-lower integration tests: 105 passed

cargo clippy -p llvm-export --all-targets -- -D warnings

cargo fmt --all -- --check
git diff --check
```

End-to-end:

```text

packed_aggregate_abi: PASS
(runtime values, whole-value load/store, and PTX parameter shapes)
```

PTX ABI checks:

```ptx
.param .align 2 .b8 packed2_param_0[6]
.param .align 2 .b8 store_packed2_param_0[6]
```

## Out of scope

This PR does not attempt to support:

* packed aggregates containing target-dependent shared-memory pointers by value;
* arbitrary overlapping/union-like layouts as LLVM packed structs;
* changes to the existing byte-offset packed field addressing path.


nihalpasham:
**Gist:** by-value packed aggregates get the correct param ABI: `.param .align 2 .b8 ...[6]` instead of natural alignment, plus whole-value load/store handling.

What I verified and did:
- rebased onto main and force-pushed: 24df1fc3 and 5e25eade duplicated the already-merged pliron bump and #931, so the branch now carries only your packed-aggregates commit (now 7c234863, 11 files) plus my fixes
- pushed 5efdf840: renamed the attr key from `cuda_oxide_kernel_param_abi_align_` to `cuda_oxide_param_abi_align_`. The reserved-symbols guard polices the `cuda_oxide_kernel_` link-symbol family; pliron op-attribute keys live outside it, matching the sibling `cuda_oxide_return_abi_align`
- pushed 7495e430: rustfmt on the example
- GPU on sm_120: `packed_aggregate_abi` PASS (runtime values, whole-value load/store, PTX parameter checks), with `.param .align 2 .b8 packed2_param_0[6]` in the PTX
- mir-lower tests green; full smoketest 210/210
- CI fully green

Closes #940 on merge.

nihalpasham (dismissed):

nihalpasham (approved):

PR #943, opened 2026-08-16
fix(mir-lower): scalarize repr(transparent) device returns

### LINKED ISSUE #942: repr(transparent) device function returns use aggregate ABI instead of underlying scalar
## Problem

`#[repr(transparent)]` scalar wrappers are currently preserved as LLVM aggregate return types for device functions instead of using the ABI of their underlying scalar or pointer field.

For example:

```rust
#[repr(transparent)]
struct Scalar(u32);

#[device]
fn make_scalar(value: u32) -> Scalar {
    Scalar(value)
}
```

The device function return ABI should be equivalent to returning `u32`, but MIR lowering currently converts the return type as the wrapper aggregate.

This is inconsistent with the transparent scalar kernel parameter handling added previously. Kernel parameter scalarization was added in #861.

## Expected behavior

For rustc-proven transparent scalar ADTs:

* `#[repr(transparent)] struct Scalar(u32)` should return `i32`
* transparent pointer wrappers should return the underlying pointer type
* transparent wrappers containing ZST marker fields should use the non-ZST scalar field
* nested transparent wrappers should recursively resolve to the underlying scalar type

Inside the function body, the MIR aggregate representation can remain unchanged. Scalarization should occur only at the function ABI boundary.

Ordinary one-field structs must remain aggregate returns.

## Scope

This issue covers device function return ABI lowering, including:

* function return type classification
* extracting the underlying scalar when lowering `return`
* reconstructing the transparent wrapper at device call sites
* nested transparent wrappers
* ZST marker fields
* scalar and pointer transparent wrappers
* negative coverage for ordinary one-field structs

Device extern ABI handling is intentionally out of scope and should be handled separately.

uurl:
Closes #942 

## Summary

Extend `repr(transparent)` scalar ABI lowering to device function returns.

Rustc-proven transparent scalar ADTs now use their underlying scalar or pointer type at the LLVM function ABI boundary, while preserving the aggregate wrapper representation inside MIR lowering.

## Changes

- classify transparent scalar device return types using their underlying scalar LLVM type
- extract the underlying scalar when lowering transparent return values
- reconstruct transparent wrappers after scalar-returning device calls
- preserve exact struct field slots, including nested transparent wrappers and ZST marker fields
- keep ordinary one-field structs aggregate
- extend `repr_transparent_abi` with scalar, pointer, marked, nested, and ordinary return cases

## Implementation

Transparent return ABI information records each wrapper layer together with its actual LLVM field slot.

When lowering a return, the transparent wrapper is recursively unwrapped with `extractvalue` until the ABI scalar is reached.

At a call site, the scalar result is rebuilt into the MIR-visible wrapper representation using `undef` plus `insertvalue`, from the innermost transparent layer back to the outermost one.

This keeps ABI scalarization localized to function boundaries rather than changing the internal representation of transparent structs.

## Validation

- `cargo test -p mir-lower`
  - 225 unit tests passed
  - 105 integration tests passed
- `git diff --check`
- `cargo oxide run repr_transparent_abi`
  - `repr_transparent_abi: PASS (kernel parameters, device returns, and runtime values)`

## Scope

This PR intentionally does not change device extern ABI handling.
That should be addressed separately.

### DIFF
--- a/crates/mir-lower/src/convert/ops/call.rs
+++ b/crates/mir-lower/src/convert/ops/call.rs
@@ -63,7 +63,7 @@
 
 use crate::convert::types::{
     StructLayoutInfo, build_struct_slot_map, convert_function_type, convert_type, is_kernel_func,
-    is_zero_sized_type,
+    is_zero_sized_type, transparent_scalar_abi_info,
 };
 use crate::helpers;
 use dialect_mir::ops::{MirCallOp, MirFuncOp};
@@ -639,6 +639,7 @@ pub fn convert(
     };
 
     let mut zst_replacement_type = None;
+    let mut transparent_result_abi = None;
     let result_type = if let Some(mir_ty) = mir_result_ty_ptr {
         // Only the empty tuple `()` is the unit type. `is::<MirTupleType>()`
         // also matches `(T, U, ...)`, so we have to peek at the field count.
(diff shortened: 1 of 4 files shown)


nihalpasham:
**Gist:** repr(transparent) device returns scalarize to the wrapped scalar instead of a one-field aggregate, keyed off rustc layout (StructAbiKind::TransparentScalar), so a plain one-field struct correctly stays an aggregate.

What I verified and did:
- mir-lower tests green; GPU on sm_120: `repr_transparent_abi` PASS (kernel parameters, device returns, and runtime values)
- pushed 478e921e: rustfmt on the example
- the cargo-deny failure was a transient crates.io index fetch; re-ran clean
- CI fully green

Lands before #945. Closes #942 on merge.

nihalpasham (approved):

PR #945, opened 2026-08-16
fix(device-ffi): preserve repr(transparent) scalar ABI

### LINKED ISSUE #944: repr(transparent) device externs reject or mis-lower scalar ABI types
## Problem

`#[device] extern "C"` declarations currently do not preserve the scalar ABI of `#[repr(transparent)]` wrappers.

For example:

```rust
#[repr(transparent)]
struct Scalar(f32);

#[device]
unsafe extern "C" {
    fn external(value: Scalar) -> Scalar;
}
````

A transparent wrapper must use the ABI of its single non-ZST field, so the external declaration above should be equivalent to:

```text
float (float)
```

Today there are two problems:

1. `rustc-codegen-cuda` rejects ADTs in device-extern signatures instead of resolving rustc-proven transparent wrappers to their underlying scalar or pointer ABI type.
2. At device-extern call sites, nested transparent wrappers can remain partially aggregated instead of being recursively projected to the scalar type expected by the external function.

For example:

```rust
#[repr(transparent)]
struct Inner(f32);

#[repr(transparent)]
struct Outer(Inner);
```

must cross a device-extern ABI boundary as `f32`, not `{ f32 }` or `{{ f32 }}`.

## Expected behavior

For rustc-proven `#[repr(transparent)]` scalar ADTs used in `#[device] extern "C"` declarations:

* scalar wrappers use the underlying scalar ABI type
* pointer wrappers use the underlying pointer ABI type
* ZST marker fields are ignored according to rustc layout
* nested transparent wrappers recursively resolve to the underlying scalar
* call arguments are recursively projected to the declared external ABI type
* returned scalar values are reconstructed into the MIR-visible transparent wrapper representation

Ordinary one-field structs must remain unsupported as device-extern scalar ABI types.

## Example

A declaration such as:

```rust
#[repr(transparent)]
struct InnerF32(f32);

#[repr(transparent)]
struct OuterF32(InnerF32);

#[device]
unsafe extern "C" {
    fn __nv_sinf(x: OuterF32) -> OuterF32;
}
```

must lower to an LLVM declaration equivalent to:

```llvm
declare float @__nv_sinf(float)
```

and the call site must extract both transparent wrapper layers before the call, then reconstruct them from the scalar result.

## Scope

This issue covers device-extern ABI handling for rustc-proven transparent scalar wrappers, including:

* device-extern signature classification
* scalar and pointer wrappers
* nested transparent wrappers
* ZST marker fields
* call-site argument projection
* interaction with transparent scalar return reconstruction
* negative behavior for ordinary one-field structs

Transparent device-function return lowering is handled separately by #942 / #943.

uurl:
Closes #944 

Depends on #943.

This branch is currently stacked on #943, so GitHub's diff also includes the transparent device-return changes from that PR. The #945-specific delta is the final commit and touches only:

- `crates/mir-lower/src/convert/ops/call.rs`
- `crates/rustc-codegen-cuda/src/device_codegen.rs`
- `crates/rustc-codegen-cuda/examples/extern_libdevice/src/main.rs`

After #943 lands, this branch will be rebased onto `main`.

## Summary

Preserve the underlying scalar or pointer ABI of rustc-proven `#[repr(transparent)]` wrappers in `#[device] extern "C"` declarations and calls.

This extends transparent scalar ABI support to the device-extern boundary while keeping the aggregate wrapper representation inside MIR.

## Changes

- resolve rustc-proven transparent ADTs in device-extern signatures to their underlying scalar or pointer ABI type
- use rustc layout information rather than inferring transparency from source field count
- support nested transparent wrappers
- support transparent wrappers containing ZST marker fields
- project transparent call arguments recursively when the declared LLVM parameter expects the underlying scalar ABI
- select the outermost matching transparent wrapper from operand type history so nested wrappers are fully projected
- reuse the transparent return reconstruction introduced by #943
- extend `extern_libdevice` with nested and ZST-marker transparent ABI coverage
- preserve existing behavior for ordinary one-field structs

## Implementation

Device-extern signature classification uses rustc layout information to peel `repr(transparent)` wrappers recursively until the ABI-relevant scalar or pointer type is reached.

At call sites, transparent argument lowering is driven by the declared LLVM parameter type rather than by symbol-name classification.

For a nested wrapper such as:

```text
OuterF32(InnerF32(f32))
````

with an external parameter of `f32`, lowering selects the outermost transparent ABI projection whose scalar type matches the declared parameter and emits:

```text
OuterF32
  -> extractvalue
InnerF32
  -> extractvalue
f32
  -> external call
```

The scalar result is then reconstructed into the MIR-visible wrapper using the transparent return ABI machinery from #943.

This keeps scalarization localized to ABI boundaries and avoids changing the internal representation of transparent structs.

## Validation

* `cargo test -p mir-lower`

  * 226 unit tests passed
  * 105 integration tests passed
* `cargo test --manifest-path crates/rustc-codegen-cuda/Cargo.toml`

  * 22 tests passed
* `cargo clippy --manifest-path crates/rustc-codegen-cuda/Cargo.toml --all-targets -- -D warnings`
* `git diff --check`
* `cargo oxide run extern_libdevice`

  * nested `OuterF32(InnerF32(f32))` device-extern argument/return
  * `MarkedF32(f32, PhantomData<()>)` ZST-marker wrapper
  * successful execution against libdevice on GPU
  * 13 cases × 4 functions within 2 ULP

Final E2E result:

```text
SUCCESS: 13 cases x 4 functions within 2 ULP; repr(transparent) device extern ABI preserved
```

## Scope

This PR is limited to device-extern ABI handling.

Transparent device-function return ABI lowering is handled by #943.

### DIFF
--- a/crates/mir-lower/src/convert/ops/call.rs
+++ b/crates/mir-lower/src/convert/ops/call.rs
@@ -62,8 +62,8 @@
 //! mismatch.
 
 use crate::convert::types::{
-    StructLayoutInfo, build_struct_slot_map, convert_function_type, convert_type, is_kernel_func,
-    is_zero_sized_type, transparent_scalar_abi_info,
+    StructLayoutInfo, TransparentScalarAbiInfo, build_struct_slot_map, convert_function_type,
+    convert_type, is_kernel_func, is_zero_sized_type, transparent_scalar_abi_info,
 };
 use crate::helpers;
 use dialect_mir::ops::{MirCallOp, MirFuncOp};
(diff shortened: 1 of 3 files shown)


nihalpasham:
**Gist:** repr(transparent) scalar ABI preserved across device FFI: signature classification peels the wrapper, call arguments project the scalar, and returns reconstruct the wrapper, with nesting and ZST markers covered.

What I verified and did:
- GPU on sm_120: `extern_libdevice` PASS, 13 cases x 4 libdevice functions within 2 ULP through nested OuterF32 and MarkedF32
- `repr_transparent_abi` PASS; full smoketest 209/209, since the change touches the shared flatten_arguments path
- pushed e7447d87: rustfmt on the example
- CI fully green

Stacked on #943: once that merges, rebase so the diff shrinks to call.rs, device_codegen.rs, and the extern_libdevice example. Closes #944 on merge.

nihalpasham (approved):

PR #946, opened 2026-08-16
feat(intrinsics): add catalog-owned f32x2 arithmetic

lucifer1004:
## Summary

- add raw `u64` packed `add`/`sub`/`mul`/`fma` `.f32x2` intrinsics and all `.ftz` forms to the packed-ALU catalog
- append ABI IDs `i0987` through `i0994` and generate device stubs, importer/collector dispatch, dialect ops, lowering, target metadata, and probes
- generalize packed-ALU generation and MIR lowering from a fixed 32-bit carrier to catalog-selected 32/64-bit carriers
- keep `Float2` pack/unpack and typed add/FMA sugar as a small handwritten layer over generated raw functions
- detect the broad `.f32x2` instruction family so external LLVM IR spellings also receive PTX 8.6 / sm_100 floors

## Evidence

- LLVM-NVPTX evidence covers all eight exact pure inline-PTX forms and terminal ptxas assembly
- public CUDA 12.9 libNVVM/nvJitLink evidence covers NVVM IR through final cubin and SASS
- the `f32x2_arith` example verifies all eight operations end to end

## Validation

- `cargo test -p cuda-intrinsics-gen` — 289 passed
- relevant `cuda-device`, `cuda-intrinsics`, `dialect-nvvm`, `mir-lower`, `mir-importer`, and `cuda-oxide-codegen` suites
- exact pinned nightly release builds for `cargo-oxide` and `rustc_codegen_cuda`
- sm_103 default llc/PTX path: all eight mnemonics emitted exactly and runtime checks passed
- sm_103 materialized libNVVM/nvJitLink path: runtime checks passed; final SASS contains the expected `FADD2`, `FMUL2`, `FFMA2`, and FTZ forms

This replaces #889 following the catalog ownership and broad target-detection feedback.


nihalpasham:
This delivers what we asked for in #889: the raw u64 packed forms live in the catalog with explicit ledger IDs, derived PTX 8.6 / sm_100 floors, probes, and regenerated dispatch; the detector does broad boundary-aware `.f32x2` family matching with negative tests; only the `Float2` sugar stays hand-written.

Two things done on our side:

- rebased the branch onto current main (past #945, no conflicts) and force-pushed
- validated the rebased branch on a 5090 (sm_120): generator check clean, all suites pass, `cargo oxide run f32x2_arith` passes end to end with all eight `add/sub/mul/fma{.ftz}.f32x2` mnemonics in the PTX, and the smoketest verdict handles both the pass and sub-sm_100 skip paths

One sequencing note: #967 was written against the same ledger tail and also claims i0987-i0994. #967 lands first, so this branch needs one final rebase that renumbers the family to i0995-i1002 and regenerates. We'll handle that renumber here after #967 merges.

CI runs are approved.

lucifer1004:
For the ledger part, do you think there is any better way to avoid collisions across PRs? Or is the maintenance overhead acceptable at this moment?

nihalpasham:
For now the overhead is acceptable since collisions only bite when two catalog PRs are in flight at once, which is kind of rare. But its worth revisiting properly (some form of ID reservation) once catalog traffic picks up, but not something we want to get into right now.

nihalpasham (approved):

PR #947, opened 2026-08-16
feat(finalizer): assemble PTX with toolkit ptxas

lucifer1004:
## Summary

- add a standalone `PtxAssembler` that discovers the toolkit `ptxas` executable through an explicit override, CUDA toolkit roots, or `PATH`
- translate the existing typed target, FMA, and debug policy to native ptxas options, validate the returned CUBIN, and expose optional resource reports
- bind cache provenance to the logical PTX, semantic options, and exact ptxas executable digest; revalidate tool identity around every operation and execute native Linux tools through the retained descriptor
- use isolated per-invocation temporary directories, so independent assemblies can run concurrently without a global compiler lock

This replaces #892. Unlike that version, it does not depend on a dynamically loadable nvPTXCompiler library, which stock toolkit packaging does not provide.

## Validation

- `cargo test --release --locked --offline -p cuda-artifact-finalizer`
- non-ignored subprocess tests cover option translation, input normalization, deterministic digests, exit diagnostics, invalid CUBIN rejection, resource parsing, and concurrent invocations
- ignored live tests passed with a discoverable toolkit ptxas for both report generation and concurrent assembly
- `cargo check --locked --offline --workspace --all-targets`
- `cargo clippy --locked --offline --workspace --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --locked --offline -p cuda-artifact-finalizer --no-deps`
- dependency-license and SPDX guards passed

No new runtime dependency or lockfile change is introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ cargo oxide run gemm_sol_final
 |---------------------------|----------------------------------------------------------------|
 | `cargo-oxide`             | Cargo subcommand (`cargo oxide run`, etc.)                     |
 | `cuda-intrinsics-gen`     | Extractor and deterministic source generator for the intrinsics |
-| `cuda-artifact-finalizer` | Driver-independent NVVM IR and LTOIR finalization              |
+| `cuda-artifact-finalizer` | Driver-independent NVVM IR, LTOIR, and PTX finalization         |
 | `oxide-artifacts`         | Architecture-neutral embedded device artifact metadata         |
 | `reserved-oxide-symbols`  | Workspace-private `cuda_oxide_*` symbol-name contract          |
 | `fuzzer`                  | Differential codegen fuzzer support (rustlantis adapter)       |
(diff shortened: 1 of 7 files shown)


nihalpasham:
This delivers what we asked for in #892: subprocess ptxas instead of the dlopen route, with the typed options, recipe/tool digests, identity revalidation, cubin validation, and resource diagnostics carried over, plus per-invocation temp dirs as the concurrency boundary.

Validated locally against a real CUDA 13.3 toolkit ptxas:

- all 32 unit tests and all 6 ignored live tests pass, including the concurrent-invocation one
- also live-checked `DebugPolicy::Full` (`--device-debug`) on our llc-produced PTX: warnings only (no fatal), and the cubin gains the debug sections
- clippy clean

Follow-ups from the #892 discussion, for later PRs: wiring `PtxAssembler` into cargo-oxide / the runtime fallback, and widening the option surface (-maxrregcount, opt level, -split-compile, warning controls).

nihalpasham (approved):

PR #948, opened 2026-08-16
fix(mir-lower): reject IntToFloat from integer widths wider than 64 bits

### LINKED ISSUE #882: mir-lower: i128 as f64 dies deep in llc instead of a located error (mirror of #877)
## The gap

\#878 (closing #877) guarded one direction of the 128-bit float conversion
problem. The mirror direction is still wide open:

```text
f64 as i128   (FloatToInt)  ->  clear, located error at your cast   since #878
i128 as f64   (IntToFloat)  ->  sitofp i128 sails through lowering,
                                then llc dies on a soft-float helper
                                (__floattidf) that does not exist on
                                device, nowhere near your source      still today
```

Neither direction can actually run on the GPU: on CPU targets these
conversions expand to compiler-builtins library calls, and the device
pipeline has no such library to link. The only question is whether the user
gets told at their cast or deep inside llc. Right now it depends on which way
they cast, which is a silly place for the answer to live.

## What to fix

Same five lines as #878, one function over. In `convert_int_to_float`
(`crates/mir-lower/src/convert/ops/cast.rs`), reject integer SOURCE widths
above 64 with a `pliron::input_err!` carrying the cast's location, mirroring
the destination-width guard #878 added to `convert_float_to_int` (and the
float-width guard that was already there). The guard sits before any
signedness split so `i128` and `u128` are both covered.

```rust
// the pattern, verbatim from its FloatToInt twin:
if int_width > 64 {
    return input_err!(loc, "IntToFloat from integer widths wider than 64 bits
                            is not yet supported on the device target");
}
```

Real 128-bit conversion support is a separate, bigger conversation
(open-coded expansion like nvcc does, or a device builtins story), same as it
was for #877.

## Acceptance criteria

- Unit tests assert the located rejection for `i128 as f64` and
  `u128 as f64`, placed next to the #878 pair and using the same helpers.
- `cargo test -p mir-lower` green.
- Supported conversions (i64 and narrower) unchanged.

alejandro-soto-franco:
`convert_int_to_float` in `crates/mir-lower/src/convert/ops/cast.rs` builds `sitofp`/`uitofp` for a source integer of any width, so `i128 as f64` and `u128 as f64` sail through lowering and fail deep inside `llc` on a soft-float helper (`__floattidf`) that does not exist on device, far from the cast that caused it. The mirror direction, `f64 as i128`, is already rejected at the cast site since #878. Neither direction can run on the GPU: on CPU targets both conversions expand to `compiler-builtins` calls, and the device pipeline links no such library. The only open question was whether the user hears about it at their cast or inside `llc`.

## Change

Same shape as the #878 destination-width check, one function over, running before the signedness split so both `i128` and `u128` are covered:

```rust
let int_width = val_ty
    .deref(ctx)
    .downcast_ref::<IntegerType>()
    .map(|t| t.width())
    .ok_or_else(|| pliron::input_error!(op.deref(ctx).loc(), "IntToFloat: operand type is not an integer"))?;
if int_width > 64 {
    return pliron::input_err!(
        op.deref(ctx).loc(),
        "IntToFloat: integer source widths wider than 64 bits are not yet supported on the device"
    );
}
```

`op` now threads into `convert_int_to_float` so the diagnostic carries the cast's location, matching `convert_float_to_int`.

## What this excludes

Real 128-bit conversion support, either an open-coded expansion the way `nvcc` does it or a device builtins story, is a separate and larger piece of work, same as it was for #877. This PR only moves the failure from an unlocated `llc` crash to a located compile error.

## Tests

- Added `int_to_float_rejects_integer_source_wider_than_64_bits` and `int_to_float_rejects_unsigned_integer_source_wider_than_64_bits` in `crates/mir-lower/src/convert/ops/cast.rs`, next to the #878 destination-width pair.
- `cargo test -p mir-lower` - 230 passed, 0 failed.
- `cargo clippy -p mir-lower --lib -- -D warnings` - clean.
- `cargo fmt` on the changed file - clean.
- Supported conversions (source width 64 bits and narrower) are unchanged; existing `int_to_float_signed_lowers_to_si_to_fp` and `int_to_float_unsigned_lowers_to_ui_to_fp` still pass.

Fixes #882.

### DIFF
--- a/crates/mir-lower/src/convert/ops/cast.rs
+++ b/crates/mir-lower/src/convert/ops/cast.rs
@@ -131,7 +131,7 @@ pub fn convert(
         }
 
         MirCastKindAttr::IntToFloat => {
-            convert_int_to_float(ctx, rewriter, val, llvm_ty, mir_opd_ty)?
+            convert_int_to_float(ctx, rewriter, op, val, val_ty, llvm_ty, mir_opd_ty)?
         }
 
         MirCastKindAttr::FloatToInt => {
@@ -214,7 +214,9 @@ fn convert_int_to_int(
 fn convert_int_to_float(
     ctx: &mut Context,
     _rewriter: &mut DialectConversionRewriter,
+    op: Ptr<Operation>,
     val: pliron::value::Value,
+    val_ty: pliron::r#type::TypeHandle,
     llvm_ty: pliron::r#type::TypeHandle,
     mir_opd_ty: pliron::r#type::TypeHandle,
 ) -> Result<Ptr<Operation>> {
@@ -228,6 +230,23 @@ fn convert_int_to_float(
             .signedness()
             == Signedness::Signed
     };
+
+    let int_width = val_ty
+        .deref(ctx)
+        .downcast_ref::<IntegerType>()
+        .map(|t| t.width())
+        .ok_or_else(|| {
+            pliron::input_error!(
+                op.deref(ctx).loc(),
+                "IntToFloat: operand type is not an integer"
+            )
+        })?;
+    if int_width > 64 {
+        return pliron::input_err!(
+            op.deref(ctx).loc(),
+            "IntToFloat: integer source widths wider than 64 bits are not yet supported on the device"
+        );
+    }
 
     if is_signed {
         Ok(llvm::SIToFPOp::new(ctx, val, llvm_ty).get_operation())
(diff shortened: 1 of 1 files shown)


nihalpasham:
Reviewed and validated locally:

- all mir-lower tests pass including the two new i128/u128 guard tests; clippy and fmt clean
- the guard reads the already-converted operand type and fires before any LLVM op is constructed, so the rewriter state stays clean on the error path, same as the #878 FloatToInt twin
- checking source width only also catches i128 -> f32, not just f64

This closes #882 as specced there: located rejection now, real 128-bit expansion stays future work.

nihalpasham (approved):

Issue #881, opened 2026-08-14
mir-importer: DST slice-tail elements can be read but not written or borrowed


nihalpasham:
## The gap

After #873, a DST's slice tail is readable but not writable or borrowable
element-wise:

```rust
#[repr(C)]
struct Packet<T: ?Sized> {
    head: u32,
    tail: T,
}

let p: &mut Packet<[f32]> = &mut concrete;

let x = p.tail[1];     // works        (#873, value walker)
p.tail[1] = 2.0;       // compile error (address walker)
let r = &p.tail[1];    // compile error (address walker)
let whole = &p.tail;   // works        (address walker already has the whole-tail recipe)
```

## Why

Reads and writes take different roads through the importer. `#873` fixed the
value walker (produces a loaded value). Assignments and borrows go through the
address walker (`translate_place_address` in
`crates/mir-importer/src/translator/rvalue.rs`), which was not touched.

The funny part: the address walker already knows the hard trick. Borrowing the
whole tail (`&p.tail`) works today via the recipe around rvalue.rs:4794-4857
(field-addr to the first tail element + length extract + construct a proper
`(ptr, len)` slice). What is missing is only the indexed arm on top of it:

```text
have:   Deref -> Field(tail)              -> whole (ptr, len) slice   works
want:   Deref -> Field(tail) -> Index(k)  -> address of element k     missing
```

Always a loud compile error today, never silent wrongness.

## What to fix

Mirror #873's normalization on the address side: rebuild the `(ptr, len)`
slice at the tail `Field` (the recipe is already there), then let
`Index`/`ConstantIndex` peel the data pointer and reuse the existing
pointer-offset path to hand back an element address. Bounds-check composition
must keep working (the `Len`/`PtrMetadata` reads the checks emit consume the
same tail place).

## Acceptance criteria

- `p.tail[k] = x` stores land where the host can verify them, and
  `&p.tail[k]` dereferences correctly, on GPU in both optimization modes.
- Covered for a padded layout (`u64` + `u8` prefix, `[u16]` tail at byte
  offset 10) as well as the plain one.
- Runtime and constant indices both covered.

PR #950, opened 2026-08-16
fix: carry slice-tail metadata through nested DST fields

### LINKED ISSUE #880: mir-importer: slice-tail indexing loses the length one level down (o.inner.tail[k])
## The gap

\#873 taught the value walker to index a DST's slice tail, so this works now:

```rust
#[repr(C)]
struct Packet<T: ?Sized> {
    head: u32,
    tail: T,
}

let p: &Packet<[f32]> = &concrete;
let x = p.tail[1];   // works since #873
```

But push the slice-tailed struct one level down and it breaks again:

```rust
#[repr(C)]
struct Inner<T: ?Sized> { head: u32, tail: T }

#[repr(C)]
struct Outer<T: ?Sized> { id: u64, inner: Inner<T> }

let o: &Outer<[f32]> = &concrete;
let x = o.inner.tail[1];   // loud compile error today
```

## Why

The #873 fix is a one-step lookahead: at the `Deref`, it peeks at the very
next projection, and if that projection is the slice-typed tail field it
saves the fat reference's length. Nesting defeats the peek:

```text
Deref ─── lookahead sees Field(inner): ordinary struct field,
  │       not a slice  ->  the length is dropped right here
  Field(inner)
    Field(tail) ── needs the length, but it is gone
      Index(1) ─── compile error (currently a generic verifier
                   message, not even the friendly "unsupported" one)
```

It always fails loudly, never with wrong data. But it is valid Rust that
refuses to compile, so per the repo's own rule, it is a compiler bug.

## What to fix

Stop peeking, start carrying. rustc's own backend layer
(`rustc_codegen_ssa::mir::place`) keeps the fat-pointer metadata (`llextra`)
riding alongside the base pointer through every projection step and attaches
it whenever the projected field is unsized, so nesting at any depth works with
zero special cases. The value walker in
`crates/mir-importer/src/translator/rvalue.rs` (`translate_place_iterative`,
the `preserved_slice_tail_len` machinery from #873) should adopt that design:
a `(ptr, optional metadata)` pair that travels the whole walk instead of a
single-step handoff between the `Deref` and `Field` arms.

## Acceptance criteria

- The nested repro above compiles and runs on GPU, including a padded layout
  (`u64` + `u8` prefix so the tail offset is not the naive one).
- The existing 23 `ref_index_projections` kernels stay green in both
  optimization modes.
- Still-unsupported forms keep failing loudly with an actionable message.

uurl:
Closes #880

## Summary

Fix nested DST slice-tail reads such as:

```rust
let value: &Outer<[f32]> = &concrete;
let x = value.inner.tail[k];
```

The MIR value walker previously preserved slice-tail metadata only when the slice field immediately followed `Deref`. A nested `Field(inner)` therefore dropped the length before `Field(tail)` could reconstruct the slice.

This change carries the slice-tail length alongside the projected address through nested slice-tailed fields, following the same general model as rustc's `llextra` handling.

## Changes

* carry slice-tail metadata through nested fields in the MIR value walker
* route nested slice-tailed reads through the value fallback path
* reconstruct generated whole-tail `AddressOf` projections while preserving the existing boundary for indexed address lowering
* recursively recover the concrete array length for nested DST unsize coercions so fat-pointer metadata is initialized correctly
* add constant-index, runtime-index, and padded-layout GPU regressions for nested slice tails

Indexed element writes and borrows remain unsupported and are intentionally left to #881.

## Validation

* `cargo oxide run ref_index_projections`

  * 26/26 kernels pass
* `CUDA_OXIDE_NO_OPT=1 cargo oxide run ref_index_projections`

  * 26/26 kernels pass
* `cargo oxide pipeline ref_index_projections`
* `scripts/smoketest.sh -o '^ref_index_projections$'`
* `scripts/smoketest.sh --compile-only -o '^ref_index_projections$'`
* `cargo test -p mir-importer`
* `cargo test -p mir-lower`
* `git diff --check`

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -2976,11 +2976,15 @@ fn classify_place_read_strategy(
 
                     match &place.projection[proj_idx + 1] {
                         mir::ProjectionElem::Field(_, field_rust_ty) => {
-                            if rust_ty_is_slice(field_rust_ty) {
-                                // Borrow/read of an unsized tail constructs a
-                                // fat value, not a thin final address.
+                            if rust_ty_is_slice(field_rust_ty)
+                                || types::slice_tail_element_ty(field_rust_ty).is_some()
+                            {
+                                // A direct slice tail or a nested slice-tailed DST needs
+                                // fat-pointer metadata that the address read path does not carry
+                                // across fields. Keep it on the value path.
                                 return Ok(PlaceReadStrategy::ValueFallback);
                             }
+
                             current_ptr_ty = dialect_mir::types::MirPtrType::get_generic(
                                 ctx, elem_ty, /* is_mutable */ false,
                             )
@@ -4604,6 +4608,10 @@ fn translate_place_addr_from_slot(
     // a following Index/ConstantIndex.
     let mut consumed_slice_subslice = false;
     let mut current_slice_len: Option<Value> = None;
+    // A fat reference to a slice-tailed struct carries the runtime length of
+    // its final slice field. Preserve that metadata while walking nested
+    // slice-tailed struct fields until the actual `[T]` field is reached.
+    let mut carried_slice_tail_len: Option<Value> = None;
     // Set by a `Downcast` and consumed by the `Field` that follows it, which
     // is the only projection pair that can name an enum payload.
     let mut pending_variant: Option<usize> = None;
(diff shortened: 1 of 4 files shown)


nihalpasham:
**Gist:** slice-tail metadata now survives nested DST field projections: the value walker extracts the tail length eagerly on every fat slice-tail deref, so `Field(inner)` no longer drops the outer fat pointer's length.

What I verified and did:
- GPU on sm_120: `ref_index_projections` all kernels PASS + SUCCESS, in both default and CUDA_OXIDE_NO_OPT=1 modes
- full smoketest 209/209, since the eager extraction touches every by-value array iteration
- mir-importer tests green
- pushed b4b93549: rustfmt on the example kernels
- CI fully green

Closes #880 on merge.

nihalpasham (approved):

PR #951, opened 2026-08-16
test(examples): pin DPX-shaped integer min/max chains to native PTX min/max

almightychang:
Part of #338

## Summary

No compiler code is changed. Hopper DPX operations are SASS `VIMNMX`-family instructions without dedicated PTX instructions; ptxas recognizes and fuses adjacent native integer min/max chains. Therefore, the five s32/u32 items in Phase 4 of #338 are covered here as code-generation regression tests rather than new compiler intrinsics.

The example contains ten kernels: `vimax3`, `vimin3`, `vimnmx`, `viaddmax`, and `viaddmin`, each in s32 and u32 forms. `vimnmx` is not a CUDA API name; it is the issue's name for the mixed min/max SASS pattern.

The runtime test validates every lane numerically and includes hand-computed pins for explicit `MAX + 1` and `MIN - 1` wrapping cases. `verify-code-shape.sh` checks each kernel's PTX body for the exact expected count and signedness of native `min.*` and `max.*` instructions, preventing regressions to compare/select or branch lowering.

## SASS evidence

With CUDA 12.0.140 ptxas targeting `sm_90`, the Rust chains and the corresponding `__vimax3_s32`-family CUDA intrinsics produced the same SASS shape: `VIMNMX` x2 for three-input min/max and `IADD3` + `VIMNMX` for add/min-max. All ten Rust kernels used `VIMNMX`/`VIMNMX.U32` for their result calculation; none degraded to compare/select.

The CI contract intentionally stops at the canonical PTX shape. SASS fusion is a property of the ptxas version in use and may select newer fused opcodes in other toolkit versions.

The ReLU and 16x2 variants have no dedicated NVVM intrinsics and remain tracked by #338.


nihalpasham:
Validated the latest push on a 5090:

- `scripts/smoketest.sh dpx_minmax_chains` passes: runtime check across all 256 lanes plus the code-shape script
- the shape script holds under our pinned llc/opt: exactly the expected native `min/max.{s32,u32}` counts per kernel, none of the opposite signedness, no setp/selp lowering
- re-derived the wrapping edge cases by hand from the EDGES tables; they all match
- the SPDX copyright line you added in the last push was the one gap we'd flagged, so nothing left on our side

The exact-count assertions will fail loudly if a future opt/llc bump changes instruction selection. That's the point of the test, so fine as is.

nihalpasham (approved):

PR #952, opened 2026-08-16
fix: address indexed DST slice-tail elements

### LINKED ISSUE #881: mir-importer: DST slice-tail elements can be read but not written or borrowed

uurl:
Closes #881

## Summary

Allow the MIR address walker to compute element addresses for indexed DST slice tails, enabling assignments and element borrows such as:

```rust
p.tail[k] = value;
let r = &p.tail[k];
```

The address walker already supported borrowing the whole unsized tail by reconstructing its `(data_ptr, len)` slice value. However, that recipe required `Field(tail)` to be the final projection, so a following `Index` or `ConstantIndex` caused address lowering to fail.

This change allows the reconstructed slice tail to continue directly into an element index. The walker then normalizes the slice to its data pointer and reuses the existing indexed-address path to return the real element address.

## Changes

* allow `Deref -> Field(tail) -> Index(k)` in the address walker
* allow `Deref -> Field(tail) -> ConstantIndex(...)`
* preserve the existing whole-tail borrow behavior
* keep unsupported continuations fail-closed with `Ok(None)`
* add GPU regressions for:

  * constant-index writes
  * runtime-index writes
  * constant-index borrows
  * runtime-index borrows
  * writes through a padded DST tail
  * borrows through a padded DST tail

The regression suite now contains 29 kernels.

## Validation

* `cargo oxide run ref_index_projections`

  * 29/29 kernels pass
* `CUDA_OXIDE_NO_OPT=1 cargo oxide run ref_index_projections`

  * 29/29 kernels pass
* `cargo test -p mir-importer`

  * 1073 passed, 0 failed
* `cargo oxide pipeline ref_index_projections`
* `scripts/smoketest.sh -o '^ref_index_projections$'`

  * 1/1 pass
* `scripts/smoketest.sh --compile-only -o '^ref_index_projections$'`

  * 1/1 pass
* `git diff --check`


nihalpasham:
**Gist:** indexed DST slice-tail elements now address correctly: constant and runtime indexes into the tail compute element addresses off the tail pointer with the right padded field offsets.

What I verified and did:
- GPU on sm_120: `ref_index_projections` all 29 kernels PASS + SUCCESS, in both default and CUDA_OXIDE_NO_OPT=1 modes
- mir-importer tests green
- pushed e53653db: rustfmt on the example kernels
- CI fully green

One optional cleanup, not blocking: the indexed path does a construct-slice/extract-field round trip where `tail_ptr` could be used directly; it folds away under opt.

Closes #881 on merge.

nihalpasham (dismissed):

nihalpasham (approved):

PR #954, opened 2026-08-16
docs(book): the legacy-NVVM atomics table still says RMW is rejected

vyncint:
## The drift

#921 and #923 legalized integer `atomicrmw` and `cmpxchg` for the legacy NVVM path. The book's table still reports both as rejected:

| LLVM construct | cuda-oxide legacy path today |
|---|---|
| `cmpxchg` | Rejected pending type, address-space, alignment, and ordering validation |
| `atomicrmw` | Rejected pending the same validation |

`reject_unsupported_op` now returns **no** rejection for either when the operands are scalar integers:

```rust
} else if let Some(rmw) = ... {
    ...
    } else if scalar_integer_width(ctx, value_ty).is_some() {
        None                          // <- accepted
```

So a reader is told a kernel cannot compile on that path when it can.

## The rewrite

Both rows now describe the current legalizer, read off `legalize.rs`:

- `atomicrmw` — native for `i32`/`i64`, address spaces 0/1/3, exactly the ten opcodes `integer_rmw_ptx_opcode` maps, at monotonic ordering;
- `cmpxchg` — native for scalar integers whose success/failure pair `valid_cmpxchg_ordering_pair` accepts.

The NVVM-intrinsics row said "other atomic RMW operations remain rejected", true when only `fadd` was legal and now false for integers; scoped to *floating-point* RMW, which is what that row is about.

## Why the surviving rejections aren't arbitrary

I added the two rules that explain the column, because without them the remaining "rejected" cases look like an unfinished list:

- **Nothing is emitted whose type, address space, ordering and scope cannot be shown to survive.** #921/#923 admitted these by proving that, not by relaxing it.
- **Scope is never weakened to fit.** libNVVM's legacy dialect lowers an ordered `cmpxchg` to a bare `atom.cas` with no ordering qualifier and no fence, which PTX treats as relaxed — so block/system-scoped atomics and stronger failure orderings are rewritten to inline PTX (needing sm_70+) rather than handed over.

Monotonic is the only ordering an `atomicrmw` reaches because MIR lowering already split Rust's stronger orderings into fences.

Also drops "and failure ordering" from the ignored-marker sentence: the failure ordering is validated now, so only `weak` is still ignored.

`just book` builds warning-free; `check-book-api-names.sh` passes. Docs only — no GPU.

### DIFF
--- a/cuda-oxide-book/compiler/lowering-pipeline.md
+++ b/cuda-oxide-book/compiler/lowering-pipeline.md
@@ -649,32 +649,37 @@ dialect accepts only a subset:
 |:----------------|:----------------|:-----------------------------|
 | Atomic load or store | Not supported as an LLVM atomic instruction | Rejected; it needs another lowering |
 | `fence` | LLVM `fence` is unsupported; an NVVM memory-barrier operation is required | Rejected pending an exact ordering/scope mapping |
-| `cmpxchg` | `i32`/`i64`, plus `i128` on `compute_90+`; global/shared pointers or generic pointers known to refer there | Kept native for scalar integers whose success/failure ordering pair libNVVM preserves; a stronger failure ordering is rewritten to inline PTX. Non-scalar-integer operands are rejected |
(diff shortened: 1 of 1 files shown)


nihalpasham:
The premise checks out: 
- the legalizer accepts integer RMW/CAS since #921/#923 and the old rows were false. 

**Layered one commit correcting the cmpxchg description:** 
- native is i32/i64 only (i128 rejected on every capability), 
- the inline-PTX rewrite triggers on a non-relaxed success ordering or block/system scope, and failure-stronger-than-success pairs are rejected outright. 
- Book builds warning-free. 
 
Merging.

nihalpasham (approved):

Issue #953, opened 2026-08-16
cutile_inter_kernel is exempt from both license guards, and #663 no longer tracks the two decisions that keep it there


vyncint:
Closing #663 left two policy decisions live but untracked, and both license guards still defer to it as though it were open.

## What the guards say

`scripts/check-dependency-licenses.sh`:

> Whether those belong in the inventory is the open question in #663; until it is settled the example is listed here rather than left silently uncovered.

`scripts/check-example-license-policy.sh`:

> Both are policy calls for a maintainer, and they are **the open question in #663** -- the same reason the example is already exempt from the inventory guard. Delete the entry once that is settled.

#663 closed on 2026-08-10.

## My part in this

I closed it, with a comment saying both halves were enforced in CI. That was **overstated**. #664 and #681 did close the general gap — every other example workspace is now covered by both the inventory guard and `cargo deny check`. But `cutile_inter_kernel` is on *both* exemption lists, so its ~60 git-pulled crates are governed by neither, and my "Resolved" reads as though they were.

## What is actually still open

The policy guard records the two concrete blockers, and neither has moved:

| blocker | current state |
|---|---|
| `error[source-not-allowed]`: git source not allowed (x7) | `deny.toml`'s `allow-git` lists only `https://github.com/pliron-org/pliron`; cutile-rs is not there |
| `error[unlicensed]`: `cuda-bindings = 0.1.0` is unlicensed | that is cutile-rs's own vendored `cuda-bindings`, not this repo's (ours is `0.2.1` with `license.workspace = true`), so the fix is not ours to make |

Both are maintainer calls, exactly as that comment says. Nothing is wrong in the tree today — the exemptions are explicit, verified against disk, and documented. The problem is only that the issue they point at is closed, so the next person to read either guard is sent to a resolved thread.

## What I would like to do

A PR updating both comments to cite this issue instead of #663, and to state plainly that the exemption withholds *both* the inventory rows and the policy check for this one example, rather than implying the policy still covers it. That is a comment-only change; it does not touch either exemption.

The decisions themselves are yours:

- add `https://github.com/NVlabs/cutile-rs` to `[sources] allow-git`, or keep the example exempt;
- decide whether a git dependency's transitive crates belong in the hand-maintained `dependency-licenses.csv` at all.

Happy to implement whichever you pick, or to leave the exemptions untouched and only fix the stale pointers.

nihalpasham:
Where we stand:

[...]

PR #955, opened 2026-08-16
ci: both license guards still defer to #663, which is closed

vyncint:
Closes the documentation half of #953.

## The drift

Each license guard exempts `cutile_inter_kernel` and justifies it by pointing at #663:

```
check-dependency-licenses.sh:
  Whether those belong in the inventory is the open question in #663

check-example-license-policy.sh:
  Both are policy calls for a maintainer, and they are the open question
  in #663 ... Delete the entry once that is settled.
```

**#663 closed on 2026-08-10.** I closed it, saying both halves were enforced in CI — and that was overstated. #664 and #681 did cover every *other* example workspace, but this one is on **both** exemption lists, so its ~60 git-pulled crates get neither a CSV row nor a `cargo deny check`. Two readers are now sent to a resolved thread to find a decision that was never made.

## What is actually still open

Unchanged, and still yours:

| blocker | state |
|---|---|
| `error[source-not-allowed]` (x7) | `deny.toml`'s `allow-git` lists only pliron, not cutile-rs |
| `error[unlicensed]: cuda-bindings = 0.1.0` | that is cutile-rs's own vendored crate, **not** this repo's — ours is `0.2.1` with `license.workspace = true`, so the fix is not ours to make |

## The change

Comment-only. Repoints both at #953, and states plainly what the exemption withholds: not just the inventory rows but the policy check too, making this the one such hole in the tree. The previous wording let a reader of the inventory guard assume `cargo deny` still covered these crates — I made that assumption myself while writing this, and had to correct the patch.

Neither exemption list is touched.

## Verified

Both guards pass unchanged: 55 declared crates recorded, every example workspace's third-party crates covered, and `deny.toml` holding over every example outside `POLICY_EXEMPT_EXAMPLES`. No GPU.

### DIFF
--- a/scripts/check-dependency-licenses.sh
+++ b/scripts/check-dependency-licenses.sh
@@ -141,10 +141,18 @@ echo "OK: ${CSV} records all $(printf '%s\n' "${required}" | grep -c .) declared
 # Examples whose third-party dependencies are deliberately out of inventory
 # scope.  cutile_inter_kernel links cutile-rs by git, which resolves a further
 # ~60 crates (wasm-bindgen, wit-bindgen, wasmparser, windows-targets) that exist
-# in this tree only to build one interop example.  Whether those belong in the
-# inventory is the open question in #663; until it is settled the example is
-# listed here rather than left silently uncovered.  Delete the entry to require
-# the rows.
+# in this tree only to build one interop example.
+#
+# Be clear about what this withholds: the same example is also on
+# check-example-license-policy.sh's POLICY_EXEMPT_EXAMPLES, so those crates get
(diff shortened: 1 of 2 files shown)


nihalpasham:
Comment-only, and every claim in the new comments checks out against the tree:
- deny.toml's allow-git really does list only pliron,
- ours is 0.2.1 with a license, cutile-rs's vendored copy is the unlicensed 0.1.0,
- and the #664/#681 history is right.

#953 stays open as the live tracker - the two policy decisions are still mine to make.

Merging.

nihalpasham (approved):

PR #956, opened 2026-08-16
docs(github): ask bug reports for `cargo oxide doctor`, not four hand-collected facts

vyncint:
## The gap

`bug_report.md` asks reporters to assemble the environment by hand:

```
- GPU:
- CUDA driver version:
- `llc --version` (or `llc-22 --version`):
```

`cargo oxide doctor` already prints all four — and it appears **nowhere** outside a CI job name: not in the template, not in CONTRIBUTING, not in the book.

## What it adds over the four fields

Real output from this machine:

```
Pinned toolchain active... ✓ nightly-2026-04-03-... (overridden by '.../rust-toolchain.toml')
Codegen backend... ✓ /.../librustc_codegen_cuda.so
Shared cache (external projects)... ⚠ /home/.../librustc_codegen_cuda.so
  Older than the backend built here, so projects outside this
  repository would load a different one.
llc (LLVM)... ✓ LLVM version 22.1.2-rust-... (/.../bin/llc)
```

Several of those separate "the compiler is broken" from "this checkout loads a different backend than you think" — a common enough answer to want in the first message rather than the third.

## Scope

The four fields stay, under *"If `doctor` itself is what fails"* — an environment broken enough to stop `doctor` is exactly when they are still needed. The output goes in a `<details>` block so a long paste does not bury the reproducer.

No GPU, no code change.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -22,6 +22,20 @@ What should happen.
 What actually happens (error message, wrong output, panic, etc.).
 
 **Environment**
+Paste the output of `cargo oxide doctor`. It reports the GPU and driver, the
+toolchain and its components, the resolved `llc`, libNVVM, nvJitLink and
+libdevice, and the backend it would load -- which is most of what a triage
+question would otherwise ask for one field at a time.
+
+<details><summary><code>cargo oxide doctor</code></summary>
+
+```text
+
+```
+
+</details>
+
+If `doctor` itself is what fails, the individual facts still help:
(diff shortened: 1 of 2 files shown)


nihalpasham:
One gap: the template says doctor reports "the GPU and driver", but doctor never printed the driver version, so that field would quietly vanish from reports.

**Layered one commit extending doctor instead of rewording the template:**
- the nvidia-smi query now asks for driver_version and the GPU line prints it,
- parse splits on the last two commas (GPU names can contain commas, the other two fields can't),
- 176 crate tests pass, and the line checks out live on an RTX 5090.

Merging.

nihalpasham (approved):

PR #957, opened 2026-08-16
build(just): check-guards runs the naming guard too, and says it does not

vyncint:
## The drift

`check-guards` has run `scripts/check-reserved-prefixes.sh` since #835, but both descriptions of its scope name only two workflows:

```
# Run the status-guard and cargo-deny CI jobs (needs cargo-deny, python3)
# `check-guards` covers the status-guard and cargo-deny workflows in full
```

That script is **naming-guard.yml**'s `reserved-symbols-guard` job, not a status-guard one.

The recipe is correct; the description undersells it, which is the direction that costs a round trip — a contributor touching a `cuda_oxide_*` literal reads this, concludes the reserved-prefix guard is CI-only, and finds out after pushing.

## After

The mapping now checks out job for job:

| workflow | jobs | mirrored by |
|---|---|---|
| status-guard.yml | 7 | 7 scripts |
| naming-guard.yml | 1 | `check-reserved-prefixes.sh` |
| cargo-deny.yml | 4 | 2 `cargo deny` runs + 2 scripts |

= the eleven scripts and two `cargo deny` invocations `check-guards` runs.

Comment-only; `just --list` and `just -n check-guards` are unchanged apart from the wording. No GPU.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -147,10 +147,10 @@ doc-check:
 # is no longer required: `test-cuda` shadows the toolkit's libcuda stub itself.
 # `check-guards` covers the status-guard, naming-guard and cargo-deny workflows
 # in full, and `check-intrinsics` the generated-intrinsics job; see their
-# comments for prerequisites. Still CI-only: clippy's per-example pass (one run per example
-# workspace), examples-compile (needs the CUDA codegen backend), and CodeQL. The
-# book gate has a local mirror too, but `just book` stays out of `check` as the
-# one gate needing a Python virtualenv.
+# comments for prerequisites. Still CI-only: clippy's per-example pass (one run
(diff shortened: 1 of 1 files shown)


nihalpasham:
Verified the recipe against the workflows - it really does mirror all three, job for job.

**Layered one commit, because there was a third stale description:**
- the detail paragraph above the recipe still said "all three cargo-deny jobs" (there are four) and listed 5 of status-guard's 7 jobs, with no mention of the naming guard,
- repaired it against the real job lists and rewrapped the one over-long line,
- `just -n check-guards` confirms the recipe itself is untouched.

Merging.

nihalpasham (approved):

PR #964, opened 2026-08-17
docs(book): the dialect inventories in mlir-dialects have drifted from the tree

vyncint:
## Summary

All four hand-maintained inventories on `cuda-oxide-book/compiler/mlir-dialects.md` are stale, and nothing checks any of them. Every count below was recomputed from the tree at `a766fc26`.

### 1. dialect-mir types: "seven" → **nine**

`mir.union` and `mir.fp16` both have `#[pliron_type(...)]` declarations in `dialect-mir/src/types.rs` and no row. `mir.union` is not a corner case — #857 and #927 are both about union constants.

### 2. dialect-mir operations: 54 across 11 categories → **62 across 12**

| Category | Book | Tree | Missing |
|:--|--:|--:|:--|
| control flow | 5 | 6 | `mir.unroll_hint` |
| memory | 9 | 11 | `mir.memcpy`, `mir.memmove` |
| comparison | 6 | 7 | `mir.cmp` |
| aggregate | 8 | 10 | `mir.construct_slice`, `mir.construct_disjoint_slice` |
| enum | 3 | 4 | `mir.set_discriminant` |
| **debug** | — | 1 | `mir.dbg_value` — no row at all |

### 3. LLVM dialect: 62 → **69**

Counted at the pinned `pliron` revision `8447aa42`, where `pliron-llvm/src/ops.rs` declares 51 ops with `#[pliron_op(` plus 18 through its binop macros. The table gains `indirectbr`, `insert_element`, `shuffle_vector`, `poison`, and a row for the four with no existing category (`freeze`, `va_arg`, `blockaddress`, `blocktag`).

Two names are wrong rather than merely missing:

- **`inline_asm_multi` does not exist.** Not in any version of `pliron-llvm`, and nowhere in this repository — this page is the only place the string appears anywhere. Same shape as the fictional `shuffle_xor_{f32,i32}` rows #815 removed from the API quick reference.
- `extractelement` is spelled `llvm.extract_element`.

Also records that `llvm-export` contributes one op of its own (`llvm.dbg_value`), which the surrounding prose did not mention, and that this table tracks an upstream crate — so it moves when the pin moves (as it did in #919), not when this repository changes.

### 4. dialect-nvvm: 123 → **543**

Not so much drift as a different tree. **Six of the twelve modules listed no longer exist**: `thread` and `warp` are gone (the warp ops now live in `warp_shuffle`, `redux`, `vote`, `warp_match`, `warp_barrier`, `active_mask`, `elect`), and `mbarrier` split into `mbarrier_basic` / `mbarrier_extended`. The ones that do exist are 4×–9× low — `tcgen05` is listed at 24 and has **210**, `tma` at 15 and has **111**. Eighteen modules are missing outright, including `sreg` at 44 ops.

Rewritten around the split `adding-new-intrinsics.md` already documents — 7 hand-written modules (18 ops) against 34 generated ones (525 ops, resolved from 986 catalog entries, since several intrinsics share one structural op). That is what decides where a contributor would add an op, and a table saying `warp` when the file is `ops/generated/warp_shuffle.rs` sends them to the wrong place.

The per-module "Minimum SM / GPU Family" columns go with it. Those requirements are per intrinsic, not per module, and the catalog already records them: `intrinsics/generated-reference.md` renders the PTX version and minimum SM for each of the 986 entries and is regenerated with the ops. The page now points there instead of keeping a second copy by hand.

## Note on the neighbouring PR

`crates/dialect-mir/README.md` carries its own copy of the operation inventory, stale in the same way and by a *different amount* — it says 55, and already had `set_discriminant` the book was missing. Two hand-kept copies drifting apart at different rates is the argument for reading `src/ops/` rather than either. Fixed separately, no file overlap with this PR.

I checked #946 before writing these numbers: it touches all 35 generated `dialect-nvvm` files but adds **zero** `#[pliron_op(` declarations (only the regenerated catalog-SHA headers change), so the counts here are stable against it.

## Testing

Local, no GPU.

- [x] Every count recomputed from the tree by script; all 12 dialect-mir modules, all 7 hand-written and all 34 generated dialect-nvvm modules, and all 10 grouped areas match exactly
- [x] The types table's nine rows match `#[pliron_type(` in `dialect-mir/src/types.rs` exactly
- [x] `check-book-api-names.sh`, `check-toolchain-parity.sh`, `check-cli-doc-coverage.sh` pass

### DIFF
--- a/cuda-oxide-book/compiler/mlir-dialects.md
+++ b/cuda-oxide-book/compiler/mlir-dialects.md
@@ -291,9 +291,12 @@ they become `call` instructions to `@llvm.nvvm.*` intrinsics.
 
 ### Architecture Coverage
 
-The dialect holds 543 operations across 41 modules, and they come from two
-different places. The split is the first thing to know about it, because it
-decides where -- and whether -- you would add one.
+At catalog SHA-256 `df42ef97` (the stamp in every `ops/generated/` file
+header), the dialect holds 543 operations across 41 modules, and they come
+from two different places. The split is the first thing to know about it,
+because it decides where -- and whether -- you would add one. If the header
+stamp no longer starts with `df42ef97`, the counts on this page predate the
+catalog you are reading.
 
 **Hand-written**, directly under `crates/dialect-nvvm/src/ops/`. These are the
 ops with bespoke verification or lowering that the intrinsic catalog does not
--- a/cuda-oxide-book/compiler/mlir-dialects.md
+++ b/cuda-oxide-book/compiler/mlir-dialects.md
@@ -50,7 +50,7 @@ to LLVM's type system.
 
 ### Types
 
-The dialect defines seven custom types that mirror Rust's compound types:
+The dialect defines nine custom types that mirror Rust's own:
 
 | Type                 | Example                                                   | Description                                                   |
 | :------------------- | :-------------------------------------------------------- | :------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


nihalpasham:
Recomputed every figure on the page at a766fc26 - all exact, including the 543/525/986 split.

**One correction, and a commit layered for it:**
- #946 does add 8 f32x2 ops; `gh pr diff` returns empty on that oversized diff, which is probably what the cross-check hit,
- so the dialect-nvvm counts are now stamped to the catalog SHA the generated headers print, same as the pliron-pin stamp on the LLVM table,
- refreshing the five affected numbers becomes an obligation on #946 when it lands.

Merging.

nihalpasham (approved):

PR #965, opened 2026-08-17
docs(dialect-mir): the crate README's type and operation tables are behind the code

vyncint:
## Summary

Both inventories in `crates/dialect-mir/README.md` are hand-maintained and nothing checks them. Counts recomputed from the tree at `a766fc26`.

### Types: "seven" → **nine**

Two types have `#[pliron_type(...)]` declarations in `src/types.rs` and no row:

- `MirUnionType` — Rust unions, every field a view of the same bytes. Not a corner case: #857 and #927 are both about union constants, so a reader who came to this README to understand that work found no mention of the type carrying it.
- `MirFP16Type` — IEEE 754 binary16, Rust's `f16`.

### Operations: 55 across 11 modules → **62 across 12**

| Module | README | Tree | Missing |
|:--|--:|--:|:--|
| `control_flow` | 5 | 6 | `unroll_hint` |
| `memory` | 9 | 11 | `memcpy`, `memmove` |
| `comparison` | 6 | 7 | `cmp` |
| `aggregate` | 8 | 10 | `construct_slice`, `construct_disjoint_slice` |
| **`debug`** | — | 1 | `dbg_value` — no row at all |

The header now says "one row per file in `src/ops/`", since that is the invariant that makes the table checkable at a glance.

## Note on the neighbouring PR

The book keeps a second copy of this same operation inventory in `cuda-oxide-book/compiler/mlir-dialects.md`. It is stale too, and by a **different amount** — it says 54 where this file says 55, because it never picked up `set_discriminant` which this one already had. Two hand-kept copies drifting apart at different rates is the argument for reading `src/ops/` rather than trusting either. That file is fixed separately; the two changes share no files, and each stands alone.

## Testing

Local, no GPU.

- [x] Every row recomputed from `src/types.rs` and `src/ops/`; both tables now match the tree exactly (9/9 types, 62 ops across 12 modules, per-module counts all equal)
- [x] `check-crate-inventory.sh` and `check-toolchain-parity.sh` pass

### DIFF
--- a/crates/dialect-mir/README.md
+++ b/crates/dialect-mir/README.md
@@ -8,7 +8,7 @@ rustc MIR ──► mir-importer ──► dialect-mir ──► mir-lower ─�
 
 ## Types
 
-The dialect defines seven types that preserve Rust-level semantics:
+The dialect defines nine types that preserve Rust-level semantics:
 
 | Type                  | Description                                        | Example                               |
 |-----------------------|----------------------------------------------------|---------------------------------------|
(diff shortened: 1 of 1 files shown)


nihalpasham:
Re-derived both tables from git objects at a766fc26:
- every cell matches, all seven new op names exist verbatim,
- and the one-row-per-module rule in the header makes the next recount a ten-second job.

Merging.

nihalpasham (approved):

PR #967, opened 2026-08-17
feat(intrinsics): add extended integer min/max (DPX relu and 16x2 forms)

almightychang:
Part of #338

## Summary

Add the `integer_minmax` intrinsic family with eight PTX ISA 8.0 / SM90+ integer min/max extensions: min/max across `relu.s32`, `s16x2`, `u16x2`, and `relu.s16x2`.

The pinned LLVM import has no NVVM intrinsic records for these operations; LLVM reaches them through generic `smin`/`smax` selection. Following the existing `packed_alu` precedent, this family therefore enters the catalog under a reviewed `ptx_native` contract.

Plain scalar `s32` min/max is deliberately excluded because ordinary Rust `min`/`max` already lowers to native `min.s32`/`max.s32`. That path is handled separately in #951.

## Changes

- Add a closed eight-variant `integer_minmax` catalog recipe and ABI entries `i0987` through `i0994`.
- Generate public `cuda_intrinsics::int` and `cuda_intrinsics::i16x2` APIs, with matching `cuda_device` compatibility modules.
- Generate dialect operations and importer dispatch for all eight variants.
- Lower scalar and packed carriers through the same pure two-operand inline-PTX path.
- Gate the family on PTX 8.0 / SM90+.
- Add a MIR lowering test that pins the exact instruction and register constraints for every variant.
- Add `examples/integer_minmax`, which launches all eight intrinsics and checks scalar and individual packed-lane results against host references.

## Evidence

All eight variants were probed successfully with the pinned Rust LLVM 22.1.2 `llc` at `sm_90 +ptx80`, producing the expected PTX instruction heads.

Only the `llvm_nvptx` evidence route is declared. This does not make the family unavailable in a `lib_nvvm` build: `requirement_for_backend` falls back to the top-level requirement, and the lowering does not consume the backend argument, so that build emits the same inline PTX. The route works, but remains unaudited until a CUDA 13.3 libNVVM evidence profile is recorded and added as a second route.

Evidence status is `lowered` and `runtime_validation` is `unexecuted`, matching the repository convention: all 547 existing evidence entries are `unexecuted`, with no `executed` precedent. Recorded terminal `ptxas` validation and runtime testing on physical SM90 hardware were not performed because the local toolkit is CUDA 12.0 and the available GPU is SM120.

## Generated artifacts

The catalog grows from 986 to 994 ABI entries and adds eight probe files. Most of the 1,100-file diff is deterministic catalog SHA regeneration churn across checked-in generated outputs, following the same pattern as #721.

## Testing

- [x] All eight pinned-`llc` probes passed.
- [x] Two consecutive generator regenerations were identical.
- [x] `cargo run -p cuda-intrinsics-gen -- check`
- [x] `cargo test -p cuda-intrinsics-gen` — 289 passed
- [x] `cargo test -p mir-lower` — 334 passed, including the new lowering coverage
- [x] `cargo oxide run integer_minmax` on an RTX 5080 (`sm_120`) — all eight intrinsics passed value verification
- [x] `scripts/check-example-smoketest-contract.sh`


nihalpasham:
Reviewed and validated locally on a 5090 (sm_120):

- catalog check clean, regen is byte-identical, intrinsics-gen / mir-lower / mir-importer suites all pass
- the example passes on GPU and the PTX carries exactly the eight native instructions (`min/max.relu.s32`, `min/max{.relu}.s16x2`, `min/max.u16x2`)
- relu semantics, the u32 packed carriers, and the PTX 8.0 / sm_90 floors all match the ISA

One coordination note: this PR and #946 were both written against the same ledger tail and both claim ABI IDs i0987-i0994. This one lands first and keeps the numbers; #946 renumbers.

Follow-ups we'll track separately, not blockers here:
- a CUDA 13.3 lib_nvvm evidence profile as a second audited route
- the three-input DPX forms (vimax3 / viaddmax / vibmax) stay on #338

CI runs are approved. Merging once they're green.

nihalpasham (approved):

PR #935, opened 2026-08-16
feat(wgmma): pipeline two accumulator groups across counted K-loops

uurl:
Fixes #934 

## Summary

Add production lowering for a conservative two-slot BF16 WGMMA pipeline inside canonical counted K-loops.

The new path combines the existing register-resident counted-loop model with the existing partial-wait pipeline model while keeping the complete asynchronous WGMMA lifetime inside one convergent inline-PTX scope.

## What changed

Add an internal value-form counted-pipeline carrier with:

```text
64 x f32 accumulator inputs
9 x u64 loop/control inputs
64 x f32 accumulator results
```

The nine control operands are:

```text
desc_a0_base
desc_b0_base
desc_a1_base
desc_b1_base

desc_a0_step
desc_b0_step
desc_a1_step
desc_b1_step

trip_count
```

The lowering emits the following static pipeline inside the counted PTX loop:

```text
mma slot0
commit_group
wait_group<1>

mma slot1
commit_group
wait_group<1>

advance four descriptors
decrement trip counter
branch
```

A final `wait_group<0>` drains the remaining asynchronous work before either accumulator becomes visible to LLVM.

## Recognition

The deferred-accumulator fusion pass recognizes this shape before falling back to the existing single-accumulator counted-loop matcher.

The accepted form is intentionally narrow:

* one canonical counted loop;
* one latch;
* one exit;
* compile-time positive trip count;
* two distinct canonical `[[f32; 8]; 4]` accumulators;
* exactly two pointer-form BF16 WGMMA operations in the latch;
* one commit and `wait_group<1>` after each MMA;
* four distinct `u64` descriptor block arguments;
* affine descriptor recurrences;
* final `wait_group<0>` in the exit;
* no loop values escaping the fused region.

## Validation

Validated with CUDA 13.3 `ptxas` targeting `sm_90a`.

The new end-to-end test starts from pointer-form MIR and exercises the complete production path:

```text
pointer-form MIR
    -> counted-pipeline recognition
    -> 64-value internal carrier
    -> convergent inline PTX
    -> LLVM NVPTX
    -> ptxas sm_90a
```

`ptxas` reports:

```text
Function properties for wgmma_pipelined_counted_loop_kernel
    0 bytes stack frame
    0 bytes spill stores
    0 bytes spill loads
Used 90 registers, used 0 barriers
```

For comparison:

```text
single-slot counted K-loop       58 registers, 0 spills
two-slot straight-line pipeline  90 registers, 0 spills
[...]

PR #969, opened 2026-08-17
docs(book): refresh the dialect-nvvm counts the catalog stamp now invalidates

vyncint:
## Summary

The obligation #964's merge comment attached to #946 landing. #946 has landed (`d485edbb`), and it was not the only batch member to move these numbers.

The page carries the self-invalidating stamp your layered commit added:

> If the header stamp no longer starts with `df42ef97`, the counts on this page predate the catalog you are reading.

Every `ops/generated/` header now prints `20440c06...`, so it does.

## What moved, and which PR moved it

Recomputed at `d485edbb`, with each delta attributed per module via `git ls-tree` + `git show` rather than from a PR diff:

| Module | Before | After | From |
|:--|--:|--:|:--|
| `generated/integer_minmax.rs` | 0 | 8 | **#967** — new module, DPX min/max |
| `generated/packed_alu.rs` | 22 | 30 | **#946** — f32x2 arithmetic |
| `ops/wgmma.rs` | 6 | 7 | **#935** — counted-K pipeline op |

Which moves five figures and adds a row:

| Figure | Before | After |
|:--|:--|:--|
| catalog stamp | `df42ef97` | `20440c06` |
| total | 543 / 41 modules | **560 / 42** |
| hand-written | 18 ops | **19** (`wgmma` 6→7) |
| generated | 34 mods / 525 ops / 986 entries | **35 / 541 / 1002** |
| Packed (SIMD-in-register) | 43 | **51** |
| Integer min/max (DPX) | — | **new row, 8** |

`integer_minmax` gets its own row rather than joining Scalar math beside `extended_minmax`: six of its eight ops are `16x2` packed forms (`max_s16x2`, `min_relu_s16x2`, …) and two are scalar `s32`, so it reads as a feature area, not a scalar-math addition.

Checked and **unchanged** in the same pass: dialect-mir is still 62 ops / 12 modules / 9 types, and the `pliron` pin is still `8447aa42`, so the LLVM dialect's 69 stands.

## A correction to my own #964

That PR asserted #946 added no ops, on the strength of `gh pr diff 946 | grep -c pliron_op` returning `0`. As you noted, `gh pr diff` returns empty on a diff that large — so the grep was counting nothing rather than finding nothing. An unfalsifiable check that reads as a clean result, which is the same shape as the `rg || true` bug `check-reserved-prefixes.sh` documents. Every count here is derived from git objects in the tree instead.

## One editorial change on the same page

The chapter opened "works across three pliron dialects … two defined locally". Two more local pliron dialects exist and the book never mentions either: `dialect-iket` (7 ops, 2 types) and `dialect-ptx` (#917). The claim is now scoped to the lowering path — which is what this chapter covers — with a pointer to both crates' READMEs, rather than shoehorning an event-tracing dialect and a terminal PTX dialect into a "levels of abstraction" table where they do not belong.

The same count appears in `pliron.md` and `architecture-overview.md`; those are a companion PR, and this one touches neither file.

## Testing

Local, no GPU.

- [x] Every figure on the page recomputed by script against the tree (stamp, totals, all 7 hand-written rows, all 11 generated group rows summing to 541)
- [x] `check-book-api-names.sh` passes

### DIFF
--- a/cuda-oxide-book/compiler/mlir-dialects.md
+++ b/cuda-oxide-book/compiler/mlir-dialects.md
@@ -1,11 +1,16 @@
 # Pliron Dialects
 
 cuda-oxide does not lower Rust to PTX in a single, heroic transformation. It
-works across three pliron dialects, each modeling a different level of
-abstraction: two defined locally (`dialect-mir`, `dialect-nvvm`) and the LLVM
-dialect provided by the upstream `pliron-llvm` crate. This chapter walks
-through all three -- their types, their operations, and how they fit together
-to form the compilation pipeline.
+works across three pliron dialects on the way down, each modeling a different
+level of abstraction: two defined locally (`dialect-mir`, `dialect-nvvm`) and
+the LLVM dialect provided by the upstream `pliron-llvm` crate. This chapter
+walks through all three -- their types, their operations, and how they fit
+together to form the compilation pipeline.
+
+Two further pliron dialects live in this tree off that path and are not covered
+here: `dialect-iket`, the compiler-facing form of in-kernel event tracing, and
+`dialect-ptx`, a structured terminal PTX dialect that can be built directly or
+projected from parsed PTX source. Each crate's README is the reference.
 
 If you have not read the [Pliron -- Pliron IR (MLIR-like)](pliron.md) chapter yet, now
 is a good time. The concepts there (operations, types, attributes, regions,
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Recomputed every count on the page at current main (d485edbb2) - all exact, including the catalog-stamp figures.
- pliron.md and architecture-overview.md still say two local dialects; a companion PR close behind this one closes the cross-page window.

PR #970, opened 2026-08-17
docs(book): the Crate Map says "every one" and lists 15 of 28

vyncint:
## The Crate Map

`architecture-overview.md`'s Crate Map opens:

> cuda-oxide is split into focused crates. **Here is every one and its role**:

and then lists 15. The workspace has 27 members plus `rustc-codegen-cuda`, so 13 are absent:

```
cuda-intrinsics          cuda-intrinsics-gen    cuda-artifact-finalizer
cuda-oxide-codegen       oxide-artifacts        reserved-oxide-symbols
dialect-iket             iket-lower             dialect-ptx
ptx-parse                libnvvm-sys            nvjitlink-sys
fuzzer
```

This is the same gap `scripts/check-crate-inventory.sh` exists to prevent, one document over. That guard was written because `dialect-iket` and `iket-lower` were missing from README.md's Crate Overview — *"1,700 lines of compiler across two crates, with no row between them."* **Both are missing here too**, and the guard cannot see this table: it only reads README.md.

Roles are copied verbatim from README.md's Crate Overview rather than written fresh, so the two inventories agree word for word and the guarded one stays the single source of truth.

## The dialect count

Two pages state how many pliron dialects cuda-oxide defines, and both predate `dialect-iket` and `dialect-ptx` (#917):

| Location | Says |
|:--|:--|
| `architecture-overview.md:188` | "cuda-oxide defines **two** dialects on top of pliron" |
| `pliron.md:352` | "cuda-oxide works with **three** dialects" |

Neither is a count of everything; both are counts of the lowering path, which is what the surrounding prose goes on to describe. Rather than fold two off-path dialects into a "levels of abstraction" narrative where they do not belong, each claim is now scoped to the lowering path and names the other two so a reader knows they exist.

`mlir-dialects.md` carries the same claim in its opening and is fixed in a companion PR; this one does not touch that file, so the two do not overlap.

## Testing

Local, no GPU.

- [x] The Crate Map matches the member list exactly by script — **28 listed, 28 actual, none missing and none extra**
- [x] `check-crate-inventory.sh` and `check-book-api-names.sh` pass

### DIFF
--- a/cuda-oxide-book/compiler/architecture-overview.md
+++ b/cuda-oxide-book/compiler/architecture-overview.md
@@ -141,6 +141,19 @@ cuda-oxide is split into focused crates. Here is every one and its role:
 | `cuda-core`          | Safe bindings to the CUDA Driver API (`CudaContext`, `DeviceBuffer`, `CudaStream`)     |
 | `cuda-async`         | Async GPU programming: `DeviceOperation`, combinators, stream pool scheduling          |
 | `cuda-bindings`      | Low-level FFI bindings to CUDA driver (`libcuda.so`)                                   |
+| `cuda-intrinsics`    | Generated low-level CUDA intrinsic declarations                                        |
+| `cuda-intrinsics-gen`| Extractor and deterministic source generator for the intrinsics                        |
+| `cuda-artifact-finalizer` | Driver-independent NVVM IR, LTOIR, and PTX finalization                           |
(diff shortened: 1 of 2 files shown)


nihalpasham:
- Re-diffed all 28 table rows against Cargo.toml's members list at d485edbb2 - complete and accurate.
- No file overlap with #969, land order is free.
- Follow-up worth filing: extend scripts/check-crate-inventory.sh to parse this table so it cannot drift a third time.

PR #972, opened 2026-08-17
docs(dialect-mir): the attributes table and source tree are the two halves #965 missed

vyncint:
## Summary

#965 corrected this README's type and operation tables. It left the other two inventories on the same page, and both are stale in the same way — one of them now **contradicting the tables #965 just fixed**.

## Attributes: "four" → seven

`src/attributes.rs` has seven `#[pliron_attr(...)]` declarations; the table lists four. The three missing ones are all recent, and two pair directly with ops the README already lists:

| Attribute | Rust type | Pairs with |
|:--|:--|:--|
| `mir.fp16_attr` | `MirFP16Attr` | `MirFP16Type`, already in the Types table |
| `mir.unroll` | `UnrollAttr` | `mir.unroll_hint` — **added to the ops table by #965** |
| `mir.compiler_result_bundle` | `CompilerResultBundleAttr` | the multi-result return ABI from #888 |

`mir.unroll` is the clearest case: #965 added `mir.unroll_hint` to the operations table, and the attribute that op exists to carry sat three rows below, undocumented.

## Source Layout tree

It contradicts the page above it:

| Tree entry | Says | Page says |
|:--|:--|:--|
| `types.rs` | `# 7 MIR types` | the Types table lists **nine** |
| `attributes.rs` | `# 4 ... attributes` | see above — **seven** |
| `ops/` | no `debug.rs` | the Operations table lists it |
| `src/` | — | no `const_fold.rs`, `rust_intrinsics.rs`, `side_effects.rs` |

So a reader who trusted the tree came away with different numbers than one who trusted the tables, on a single page. Three module comments were also narrower than their files (`control_flow` also carries unroll hints, `memory` also `memcpy`/`memmove`, `aggregate` also slices) — all ops #965 added to the table without the tree following.

Both inventories now state the rule that makes them checkable, as the operations table does: one row per `#[pliron_attr(...)]`, one tree entry per file.

## Testing

Local, no GPU.

- [x] The attributes table matches all seven `(name, Rust type)` pairs exactly by script — `MirCastKindAttr` is a `pub enum`, the rest `pub struct`
- [x] The Source Layout tree matches `src/*.rs` plus `src/ops/*.rs` exactly: **19 listed, 19 on disk**, none missing and none invented
- [x] `check-crate-inventory.sh` passes

### DIFF
--- a/crates/dialect-mir/README.md
+++ b/crates/dialect-mir/README.md
@@ -102,14 +102,17 @@ This catches mismatches immediately after `mir-importer` translates from rustc, 
 
 ## Attributes
 
-The dialect defines four domain-specific attribute types (following the pliron best practice of avoiding overloaded `IntegerAttr`):
+The dialect defines seven domain-specific attribute types (following the pliron best practice of avoiding overloaded `IntegerAttr`), one row per `#[pliron_attr(...)]` in `src/attributes.rs`:
 
-| Attribute           | Rust Type          | Description                                                                                                          |
-|---------------------|--------------------|----------------------------------------------------------------------------------------------------------------------|
-| `mir.cast_kind`     | `MirCastKindAttr`  | Preserves Rust cast intent (e.g. `IntToFloat`, `PtrToPtr`, `Transmute`) so lowering picks the right LLVM instruction |
-| `mir.mutability`    | `MutabilityAttr`   | Boolean: `&` vs `&mut` for `mir.ref`                                                                                 |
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Recounted at d485edbb2: 7 `#[pliron_attr]`, 9 `#[pliron_type]`, 19 files across src/ - all exact.
- Non-blocking: the last three attribute rows are out of declaration order.

PR #976, opened 2026-08-17
feat: add pinned transfer overlap demo and book chapter

drbh:
this pr adds an example and docs for pinned host transfer APIs added in #42.

the `pinned_overlap` example demonstrates the performance difference between pageable and pinned host transfers, and how to overlap them with kernel execution. it shows:

- pageable vs pinned HtoD/DtoH bandwidth from 1 MiB to 256 MiB
- a serialized pageable upload -> kernel -> download loop vs a three-stage rotating pinned-stager pipeline

the kernel is intentionally just an in-place `+1.0`, so the result is about the data path rather than the workload. each pipeline stage owns a persistent device buffer, a pinned host stager, and a non-default stream. completion events protect ring-slot reuse, and every output is checked before `SUCCESS`.

on an RTX 4090 (sm_89) over PCIe Gen4 x16

```bash
nix develop --command cargo oxide run pinned_overlap
```

outputs

```
=== Pinned host transfer and overlap demo ===

The kernel only adds 1.0 to each element; the measured difference is the data path.

Transfer bandwidth (GB/s, median of 10 timed iterations)
  size       pageable HtoD  pinned HtoD  pageable DtoH  pinned DtoH
      1 MiB          13.73         22.51           9.21         25.42
      4 MiB          17.07         23.57          11.31         26.62
     16 MiB          18.42         23.85          12.34         26.55
     64 MiB          12.88         23.93           3.58         26.54
    256 MiB          11.84         23.73           3.52         26.86

Pipeline impact (8 x 32 MiB chunks)
  serialized pageable pipeline:   181.548 ms
  overlapped pinned pipeline:      90.174 ms
  measured speedup:                  2.01x

SUCCESS: pinned transfers and overlap pipeline verified exactly.
```

at 256 MiB, pinned transfer bandwidth was 23.73 GB/s HtoD and 26.86 GB/s DtoH, compared with 11.84 GB/s and 3.52 GB/s for the pageable paths.

the book chapter explains the safe synchronous helper, the safety contract of the async helpers, and why the completion event is needed even though the upload, kernel, and download are already ordered within each stream. it also calls out that the pipeline number is end-to-end: it includes host staging, event waits, and result copies.

please let me know if anything should be updated or clarified in the docs. Thanks!


nihalpasham:
- Ran on an RTX 5090 host: pinned HtoD holds ~28.7 GB/s across sizes while pageable DtoH collapses to ~3 GB/s at 64+ MiB; the overlap pipeline measured 2.12x with exact verification and the SUCCESS marker. Smoketest passes in execute and compile-only modes; the book builds with the new chapter under `-W`.
- One correction, and a commit layered for it: an early `?` return in `run_overlapped` / `benchmark_bandwidth` could free pinned stagers while a stream still had a transfer in flight. A `DrainOnDrop` guard declared after the buffers synchronizes the streams before they drop.
- Same commit: one README sentence noting the pageable columns include per-iteration allocation while pinned columns reuse buffers, and `completion` -> `completions` in the book snippet to match the example.

nihalpasham (approved):

PR #977, opened 2026-08-18
docs(mir-lower): the intrinsic converter table names two modules deleted in #406

vyncint:
## Summary

`README.md`'s **GPU Intrinsic Converters (`convert/intrinsics/`)** table has 13 rows for a directory holding 24 modules. Two of the 13 do not exist:

| Row | Reality |
|:--|:--|
| `stmatrix` | `convert/intrinsics/stmatrix.rs` **deleted in `6a9a6373` (#406)** |
| `tcgen05` | `convert/intrinsics/tcgen05.rs` **deleted in `6a9a6373` (#406)** |

Both went when *"Generate CUDA intrinsics from pinned LLVM metadata"* moved those families onto the catalog path. The table still describes them as hand-written converters with an "Inline PTX" strategy; `stmatrix` and `tcgen05` appear **460 times in `convert/generated_intrinsics.rs`** today. So a reader looking for where either is lowered is sent to a file that has not existed for many releases — the same shape as #797, one crate over.

A third row, `generated_intrinsics`, is real but **not in that directory** — it is `convert/generated_intrinsics.rs`, one level up. That is the row most worth getting right, since it is where nearly every intrinsic now goes.

Fourteen modules that *do* exist had no row at all:

```
asm  cp_async  dotprod  execution_control  extended_minmax  integer_minmax
ldmatrix  memory  packed  prmt  scalar_arithmetic  scalar_conversion
scalar_math  wmma
```

`integer_minmax` arrived last batch with #967.

## The fix

Rewritten as **one row per file**, with each description taken *verbatim from that module's own `//!` doc comment* — so the table cannot drift from the code without the doc comment drifting first.

The "Strategy" and "GPU" columns are dropped: they were per-module guesses at what are per-intrinsic properties, and the catalog already records the PTX version and minimum SM for each, rendered into `intrinsics/generated-reference.md`. The page now points there instead of keeping a second copy by hand — the same treatment #964 gave the book's dialect-nvvm table.

The **Core Modules** table on the same page was short by four: top-level `type_conversion_interface`, `scalarize_block_args` and `wgmma_deferred_accumulator`, plus `convert/enum_payload_storage`.

## Testing

- [x] By script: the converter table's rows are now **exactly** the set of `.rs` files in `convert/intrinsics/` (24 and 24, no ghost and no missing)
- [x] Every top-level `src/*.rs` module and every `convert/*.rs` module now has a row


nihalpasham:
- Verified the two deleted modules are gone and every remaining row matches a file in convert/intrinsics/ with its doc comment.
- One commit layered on: the Core Modules table was missing `convert/type_interface_impls`, the type-conversion counterpart of `convert/interface_impls`.

PR #978, opened 2026-08-18
docs(mir-importer): five of twelve intrinsic modules in the README do not exist

vyncint:
## Summary

`README.md`'s **"`terminator/intrinsics/` — GPU Intrinsics"** table has 12 rows. Only 7 name a module that exists:

| Row | Deleted in |
|:--|:--|
| `intrinsics/sync.rs` | `6a9a6373` (#406) |
| `intrinsics/cluster.rs` | `6a9a6373` (#406) |
| `intrinsics/tcgen05.rs` | `6a9a6373` (#406) |
| `intrinsics/clc.rs` | `6a9a6373` (#406) |
| `intrinsics/warp.rs` | `4c1d5838` (#833) |

#406 moved those families onto the catalog path and #833 dropped the last of them, so all five are handled by `generated.rs` today. Meanwhile **eight** modules that do exist had no row: `asm`, `bigint`, `bitops`, `exact_div`, `float_math`, `iket`, `layout`, `saturating`.

Between the two, the table describes a directory that has not existed for a long time — and it is the one place a contributor looks to find where an intrinsic is translated.

The book's `compiler/mir-importer.md` **already describes the current model correctly** ("`intrinsics/catalog.json` has its arm generated into … `generated.rs`"), so this README was the last copy still on the old one. Rewritten to match: one row per file, each description verbatim from that module's own `//!` doc comment. The per-module "GPU" column is dropped for the reason #964 dropped its equivalent — minimum-SM is a per-intrinsic property the catalog already records.

## Two smaller gaps on the same page

- The `translator/` table was missing `layout`, `location` and `payload_store`.
- The **Error Types** table lists 7 of `PipelineError`'s 12 variants, missing `Optimization`, `TargetSelection`, `UnsupportedLinking`, `LibdeviceUnavailable` and `InvalidMirPassPipeline`. It also never says that `PipelineError` is defined in `cuda-oxide-codegen` and re-exported here, nor that the crate's own `TranslationErr` — `Unsupported`, `TypeError`, `InvalidOp` — is a different, narrower error. Both are now stated.

> A note on that last one: I first read the table against `error.rs`, found none of the seven, and thought the whole table was fictional. It is not — the variants are real, just defined a crate away. Checking before writing is why this PR says "5 missing of 12" rather than something wrong.

## Testing

- [x] By script, no ghost and no missing row in any of the three tables: intrinsics = exactly the 15 files in `terminator/intrinsics/`; translator = exactly the 9 files in `translator/` plus the `terminator` subdirectory it also lists; errors = exactly the 12 `PipelineError` variants

### DIFF
--- a/crates/mir-importer/README.md
+++ b/crates/mir-importer/README.md
@@ -77,24 +77,37 @@ artifact; a missing required sidecar is an error rather than a silent fallback.
 | `rvalue`    | Expression translation (binops, casts, etc.)   |
 | `types`     | Rust type → `dialect-mir` type conversion      |
 | `values`    | MIR local → alloca-slot mapping + load/store   |
+| `layout`    | Shared readers over rustc's aggregate layout   |
+| `location`  | Source-location helpers for MIR translation    |
+| `payload_store` | Enum-payload stores whose storage type differs from its usage |
 
-### `terminator/intrinsics/` — GPU Intrinsics
+### `terminator/intrinsics/` — intrinsic handlers
 
-| Module     | Intrinsics                                         | GPU       |
-|------------|----------------------------------------------------|-----------|
-| `generated`| Admitted generated intrinsics, including barriers  | Varies    |
-| `indexing` | `threadIdx`, `blockIdx`, `blockDim`, `gridDim`,    | All       |
-|            | `index_1d`/`index_2d`, DisjointSlice helpers       |           |
-| `sync`     | mbarrier ops and fences                            | All       |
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Confirmed the five phantom modules are gone from the tree and re-derived the corrected tables from src/ - every row matches.
- The `tma` row's description is copied verbatim from tma.rs, whose own doc comment says "Tensor Memory Access"; one-line source follow-up to make both say Accelerator.

PR #981, opened 2026-08-18
docs(ptx-parse): the one published crate with no README

vyncint:
## Summary

Every workspace member except `ptx-parse` ships a `README.md` **and** declares `readme = "README.md"` — 26 of 27 with both, and `ptx-parse` with neither.

It is the wrong crate to have missed. `ptx-parse` is the **only member that sets `publish = true`**:

```toml
# Published because cuda-host depends on it (name verified free on crates.io).
publish = true
description = "Lossless structural views over PTX source text"
```

So it is the one crate here whose crates.io page renders from a README — and that page would render with the one-line description and nothing else.

Its sibling `dialect-ptx`, from the same #914–#917 stack, has both, and its README sends readers here:

> `ptx-parse::Document` owns exact external source, trivia, unknown syntax, and byte-preserving edits.

Following that pointer currently lands on nothing.

## What the README says, and where each claim comes from

Written from the crate's own sources, not from an impression of them:

| Section | Source |
|:--|:--|
| Framing, and the deliberate no-ISA-type-checking rule | the `lib.rs` crate doc |
| Six `StatementKind` variants, `Document` accessors, `diagnostics`, `coverage`, `ParseError` | `lib.rs`, `syntax.rs` |
| `EditScript` surface, `apply_with_map`/`AppliedEdits`, `MapBias` at edit boundaries | `edit.rs` |
| The `ptx-parse` / `dialect-ptx` split | restated in the same terms `dialect-ptx`'s README already uses, so the two agree rather than drift |
| Consumer table | the actual `ptx_parse::` call sites in `cuda-oxide-codegen`, `cuda-host`, `cuda-intrinsics-gen`, `cargo-oxide`, `dialect-ptx` |

`readme = "README.md"` is added alongside, matching the other 26 manifests, so the packaged crate carries it too.

Happy to cut or reshape any of it — this is the one PR in this batch that adds prose rather than correcting it, and you or @lucifer1004 may have a different idea of what belongs on that page.

## Testing

Local, no GPU.

- [x] Every backticked API name in the new README resolves in `crates/ptx-parse/src/`
- [x] All 27 members now have both the file and the manifest field
- [x] `cargo metadata --locked` still parses; `cargo test -p ptx-parse` green (35 passed)
- [x] `check-crate-inventory`, `check-dependency-licenses`, `check-spdx-headers` pass

### DIFF
--- a/crates/ptx-parse/Cargo.toml
+++ b/crates/ptx-parse/Cargo.toml
@@ -8,5 +8,6 @@ repository.workspace = true
 # Published because cuda-host depends on it (name verified free on crates.io).
 publish = true
 description = "Lossless structural views over PTX source text"
+readme = "README.md"
 
 [dependencies]
--- a/crates/ptx-parse/README.md
+++ b/crates/ptx-parse/README.md
@@ -1,0 +1,91 @@
+# ptx-parse
+
+Lossless structural views over PTX source text.
+
+`Document::parse` borrows the source and owns only structural indices into it,
+so every byte of the input -- including comments, whitespace and syntax this
+crate does not recognise -- is still reachable through the original `&str`.
+Nothing is reformatted, and nothing is discarded.
+
+```text
+PTX source text
+     │  Document::parse
+     ▼
+Document<'source>        tokens, statements, scopes, diagnostics, coverage
+     │                   labels, directives, callables, instructions
+     │  EditScript
+     ▼
+edited PTX + byte lineage (AppliedEdits)
+```
+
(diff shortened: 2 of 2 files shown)


nihalpasham:
- Re-derived the README from the source: the 26-of-27 count, every named API, and all five consumer rows match the code.
- One correction, and a commit layered for it: the relative LICENSE link 404s on crates.io, which rewrites README links against the repository root - replaced with an absolute URL.

PR #984, opened 2026-08-18
fix(mir-importer): admit thin-pointer union static relocations

### LINKED ISSUE #983: Support thin-pointer relocations in top-level union device-global initializers
## Problem

Device-global initializers can already preserve thin-pointer relocations, including non-zero addends, but initialized union storage is still rejected before the existing relocation machinery can handle it.

This leaves a gap for top-level unions whose complete storage is exactly one thin pointer, for example:

```rust
static TARGETS: [u32; 3] = [7, 11, 23];

#[repr(C)]
union PointerUnion {
    word: &'static u32,
    byte: &'static u8,
}

static VALUE: PointerUnion = PointerUnion {
    word: &TARGETS[2],
};
```

The initializer above contains one pointer relocation at byte zero with a non-zero target addend, but the current device-static path rejects the union before lowering the initializer.

## Expected behavior

cuda-oxide should accept a top-level union device-global initializer when its physical storage can be represented unambiguously by the existing provenance-preserving relocation path.

For an initial implementation, the supported shape can remain intentionally narrow:

* the union is the top-level static type;
* the union size is exactly one target pointer word;
* the union has natural pointer alignment;
* the initializer contains exactly one pointer relocation;
* the relocation starts at byte zero and spans the complete pointer word;
* every non-ZST union alternative is a compatible thin pointer;
* all pointer alternatives use the same address space.

The existing device-global relocation machinery should then preserve both the pointer target and its addend.

## Out of scope

This issue does not require support for:

* pointer/integer union alternatives;
* fat pointers;
* nested pointer-bearing aggregates inside a union;
* mixed pointer address spaces;
* nested unions inside other aggregate storage;
* padded, over-aligned, or under-aligned union storage;
* multiple relocations inside one union.

Those shapes need separate representation and layout analysis rather than guessing an active union field.

## Acceptance criteria

A device-global example containing a top-level thin-pointer union should:

1. compile through the MIR importer and lowering pipeline;
2. preserve the pointer relocation and non-zero addend;
3. execute correctly on the GPU in both optimized and no-opt modes;
4. leave unsupported union layouts rejected explicitly rather than silently lowering them as bytes.

uurl:
Closes #983 

## Summary

Allow a narrowly defined class of top-level union device-global initializers to reuse the existing provenance-preserving static relocation path.

Previously, `ensure_static_global_alloc` rejected any initialized static whose stored type contained a union before the initializer relocation metadata could be lowered.

This change admits the case where the complete union storage is provably one representation-compatible thin pointer.

## What changes

`mir-importer` now validates a top-level union initializer before creating the global allocation.

The accepted shape is intentionally strict:

- the static itself lowers to a MIR union type;
- the union is exactly one target pointer word in size;
- the union has natural pointer alignment;
- every non-ZST union field is a representation-compatible thin pointer;
- pointer alternatives use the same address space;
- the evaluated initializer contains exactly one pointer relocation;
- that relocation starts at byte zero and occupies the complete pointer word.

Once those conditions are proven, the implementation reuses the existing
device-global initializer relocation path unchanged.

Nested union storage remains rejected.

## Why this is safe

A device-global initializer does not retain a source-level notion of the active union field.

Therefore this patch does not attempt to infer or select one.

Instead, it only accepts unions for which all non-ZST alternatives describe the same physical pointer representation and the complete initialized storage is covered by one relocation. The relocation remains the source of pointer provenance, including its target addend.

Unsupported layouts continue to fail closed.

## Example

The `device_global` example now covers a union initialized with a pointer to an array subobject:

```rust
static UNION_RELOCATION_TARGETS: [u32; 3] = [7, 11, 23];

#[repr(C)]
union RelocatedPointerUnion {
    word: &'static u32,
    byte: &'static u8,
}

static UNION_RELOCATION: RelocatedPointerUnion = RelocatedPointerUnion {
    word: &UNION_RELOCATION_TARGETS[2],
};
```

The kernel dereferences the union pointer and returns `23`.

This also exercises a non-zero relocation addend, so the test verifies more than preservation of a pointer-shaped initializer.

## Tests

Passed locally:

```text
cargo fmt --all -- --check

cargo test -p mir-importer --lib
1089 passed; 0 failed

cargo test -p mir-lower --lib
247 passed; 0 failed

cargo oxide run device_global
PASS
Union static initializer relocation: result = 23

CUDA_OXIDE_NO_OPT=1 cargo oxide run device_global
PASS
Union static initializer relocation: result = 23

scripts/smoketest.sh --only '^device_global$' --no-color --keep-logs
1 passed; 0 failed

git diff --check
```

## Scope

* pointer/integer union storage;
* fat-pointer alternatives;
* nested pointer-bearing union fields;
* mixed address spaces;
* nested unions inside aggregate static storage;
* padded or non-naturally-aligned union layouts;
* multiple relocations within union storage.

Those cases require additional representation rules and should be handled separately.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -8234,14 +8234,48 @@ fn validate_device_static_union_initializer(
     initializer: &GlobalInitializerData,
     loc: Location,
 ) -> TranslationResult<()> {
+    let relocation_slots: Vec<(u64, u32)> = initializer
+        .relocations
+        .iter()
+        .map(|relocation| (relocation.source_offset, relocation.width_bytes))
+        .collect();
+    validate_device_static_union_storage(
+        ctx,
+        union_ty,
+        rustc_public::target::MachineInfo::target_pointer_width().bytes(),
+        initializer.bytes.len() as u64,
+        initializer.alignment,
+        &relocation_slots,
+    )
+    .map_err(|message| {
+        input_error!(
+            loc,
+            TranslationErr::unsupported(format!(
+                "device static {} contains {message}",
+                static_def.name()
+            ))
+        )
+    })
+}
+
+/// Pure storage gate behind [`validate_device_static_union_initializer`]:
+/// accept only a union that is one naturally aligned pointer word whose
+/// evaluated initializer is exactly one full-width relocation at byte zero.
+/// The caller attaches the static's name and source location to rejections.
+fn validate_device_static_union_storage(
+    ctx: &Context,
+    union_ty: TypeHandle,
+    pointer_width: usize,
+    initializer_len: u64,
+    initializer_align: u64,
+    relocation_slots: &[(u64, u32)],
+) -> Result<(), String> {
     let (union_name, union_size, union_align) = {
         let ty_ref = union_ty.deref(ctx);
         let union_ty = ty_ref
             .downcast_ref::<dialect_mir::types::MirUnionType>()
             .ok_or_else(|| {
-                input_error_noloc!(TranslationErr::unsupported(
-                    "validate_device_static_union_initializer called on non-union type"
-                ))
+                "a non-union type in the device-static union initializer gate".to_string()
             })?;
         (
             union_ty.name().to_string(),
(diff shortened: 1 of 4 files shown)


nihalpasham:
- Verified on the 5090: device_global prints the union relocation result 23 in both the default and CUDA_OXIDE_NO_OPT=1 pipelines, and the importer suite passes.
- One commit layered on: the new size/offset/count gates were only covered by the example's positive path. Split the pure checks into validate_device_static_union_storage (StaticDef can't be built in a unit test, so the wrapper passes plain values) and unit-tested every reject branch.

nihalpasham (approved):

PR #988, opened 2026-08-18
fix(mir-lower): admit direct nested packed static relocations

### LINKED ISSUE #987: Support direct nested packed relocation carriers in static initializers
## Problem

Relocated static initializers currently support a top-level `repr(packed)` struct, but reject the same packed relocation carrier when it is nested directly inside an otherwise ordinary struct.

For example:

```rust
#[repr(C, packed)]
struct Inner {
    tag: u8,
    ptr: &'static u32,
}

#[repr(C)]
struct Outer {
    head: u32,
    inner: Inner,
}

static VALUE: Outer = Outer {
    head: 0x1122_3344,
    inner: Inner {
        tag: 0x7d,
        ptr: &TARGET,
    },
};
```

The relocation storage path can already represent the pointer slot byte-faithfully, and packed aggregates themselves have a faithful LLVM representation when `build_struct_slot_map` proves the layout can be expressed as a packed struct.

However, `validate_relocated_initialized_global_layout` currently validates nested fields through the ordinary initialized-global validator. That validator intentionally rejects packed aggregate layouts, so the nested `Inner` is rejected even though the relocated physical initializer carrier is representable.

## Expected behavior

Allow the narrow case:

```text
ordinary explicit-layout outer struct
    -> one directly nested packed struct
        -> relocation slot
```

provided that:

* the outer struct has an explicit, valid, non-overlapping rustc layout;
* the nested struct has an explicit, valid, non-overlapping rustc layout;
* `build_struct_slot_map` proves the nested struct has a byte-faithful packed LLVM representation;
* the resulting LLVM packed representation matches rustc's stored size;
* no unsupported deeper packed nesting is introduced.

## Scope

This should not broaden the ordinary initialized-global validator.

In particular, this issue does not propose arbitrary recursive packed-layout support. A shape such as:

```text
outer -> packed -> packed -> relocation
```

should remain fail-closed for now.

## Acceptance criteria

* A relocated initialized global containing one directly nested packed struct is accepted.
* Existing top-level packed relocation support continues to work.
* Deeper packed nesting remains rejected.
* Existing overlap and malformed-layout checks remain unchanged.
* `device_global` covers the nested packed relocation end to end on the GPU.

uurl:
Closes #987 

## Summary

Allow relocated static initializers to contain one directly nested packed struct when its rustc byte layout is proven representable by the existing packed LLVM struct lowering.

This extends the existing top-level packed relocation support to the narrow shape:

```text
ordinary outer -> packed child -> relocation
```

while keeping deeper packed nesting fail-closed.

## Background

`validate_relocated_initialized_global_layout` already has a separate path for relocated globals because their initializer is emitted through segmented physical storage rather than through the semantic aggregate value.

That path previously relaxed layout validation only for the top-level struct. Nested fields were still passed to `validate_initialized_global_type`, whose ordinary initialized-global policy intentionally rejects packed aggregate layouts.

As a result, this valid shape was rejected:

```rust
#[repr(C, packed)]
struct NestedPackedRelocation {
    tag: u8,
    ptr: &'static u32,
}

#[repr(C)]
struct OuterNestedPackedRelocation {
    head: u32,
    inner: NestedPackedRelocation,
}
```

## Change

The relocated-global validator now gives a directly nested struct one additional validation path.

When ordinary initialized-global validation fails, the nested struct is accepted only if:

* it has explicit rustc layout metadata;
* its field ranges are valid and non-overlapping;
* `build_struct_slot_map` reports a byte-faithful representation;
* the selected LLVM representation is packed;
* its LLVM stored size matches rustc's recorded size.

Its children are then validated through the existing ordinary validator.

This last point deliberately prevents the exception from becoming recursively permissive, so:

```text
outer -> packed -> packed -> relocation
```

continues to fail.

The ordinary `validate_initialized_global_layout` behavior is unchanged.

## Tests

Added unit coverage for:

* accepting one directly nested packed relocation carrier;
* rejecting a packed top-level struct containing a packed child;
* continuing to reject deeper packed nesting.

Extended `device_global` with an end-to-end nested packed relocation:

```rust
#[repr(C, packed)]
struct NestedPackedRelocation {
    tag: u8,
    ptr: &'static u32,
}

#[repr(C)]
struct OuterNestedPackedRelocation {
    head: u32,
    inner: NestedPackedRelocation,
}
```

The kernel verifies:

```text
[0x11223344, 0x7d, 0xcafebabe]
```

The packed fields are read with `addr_of!` and `read_unaligned()`.

## Validation

```text
cargo fmt --all -- --check
git diff --check
cargo test -p mir-lower --lib relocated_initialized_global_layout
cargo oxide run device_global
CUDA_OXIDE_NO_OPT=1 cargo oxide run device_global
scripts/smoketest.sh --compile-only --only '^device_global$' --no-color --keep-logs
scripts/smoketest.sh --only '^device_global$' --no-color --keep-logs
```

Results:

* relocated initialized-global tests: 6 passed
* mir-lower unit tests: 247 passed
* mir-lower lowering tests: 107 passed
* clippy with `-D warnings`: passed
* optimized `device_global`: passed
* `CUDA_OXIDE_NO_OPT=1` `device_global`: passed
* compile-only smoketest: passed
* executing smoketest: passed


nihalpasham:
- Verified on the 5090: [0x11223344, 0x7d, 0xcafebabe] in both pipelines, mir-lower suite green, and the legacy sm_90 NVVM build carries the nested relocation.

nihalpasham (approved):

PR #990, opened 2026-08-18
fix(mir-importer): preserve nested enum constant provenance

### LINKED ISSUE #989: Support relocation-carrying enum constants nested inside aggregate constants
## Problem

The MIR importer currently preserves pointer provenance for enum constants when the enum is translated directly, including niche-encoded enums and direct-tagged enums with pointer payloads.

However, a relocation-carrying enum constant is still rejected when the enum is nested inside another aggregate constant such as a struct, tuple, or array.

For example, constants conceptually shaped like these are not currently supported:

```rust
struct Holder {
    marker: u32,
    value: TaggedReference,
}

const STRUCT_VALUE: Holder = Holder {
    marker: 17,
    value: TaggedReference::Present(&TARGETS[1]),
};

const TUPLE_VALUE: (u32, Option<&'static u64>) =
    (23, Some(&TARGETS[0]));

const ARRAY_VALUE: [TaggedReference; 2] = [
    TaggedReference::Empty,
    TaggedReference::Present(&TARGETS[1]),
];
```

The importer currently detects provenance inside the nested enum range and rejects the constant instead of delegating to the allocation-aware enum constant decoder.

## Expected behavior

When a nested field is an enum and its storage contains relocation entries, the importer should decode that enum from the enclosing allocation while preserving the field's absolute byte offset.

This should reuse the existing enum allocation decoder rather than introducing a separate nested-enum decoding path.

## Scope

Support relocation-carrying enum constants nested inside:

* struct constants;
* tuple constants;
* array constants.

The existing handling for relocation-free nested enums should remain unchanged.

## Non-goals

This issue does not attempt to add support for:

* pointer/integer overlapping enum storage;
* unsupported address-space pointer layouts;
* unrelated enum layout or aggregate constant changes.

## Acceptance criteria

* A direct-tagged enum with a pointer payload works when nested inside a struct at a non-zero field offset.
* A niche-encoded `Option<&T>` works when nested inside a tuple.
* A relocation-carrying enum works when nested inside an array.
* Existing enum constant provenance tests continue to pass.
* The regression compiles successfully through libNVVM and produces the expected values at runtime on GPU.

uurl:
Closes #989 

## Summary

Preserve pointer provenance for relocation-carrying enum constants nested inside aggregate constants.

Previously, `translate_constant_value_from_alloc` explicitly rejected a nested enum whenever relocation entries were present in the enum's byte range.

This change delegates that case to the existing allocation-aware enum decoder, using the nested field's absolute byte offset.

## Implementation

When an aggregate field is an enum:

- relocation-free enums continue through the existing path;
- if provenance exists within the enum storage range, the importer now calls `translate_enum_constant_from_alloc`;
- the enclosing allocation and `absolute_byte_offset` are preserved, allowing the existing enum decoder to reconstruct pointer fields from rustc relocation entries.

No new enum decoding mechanism is introduced.

## Regression coverage

Extend `enum_constant_provenance` with:

- a direct-tagged enum nested inside a struct at a non-zero offset;
- a niche-encoded `Option<&T>` nested inside a tuple;
- a relocation-carrying direct-tagged enum nested inside an array.

The runtime regression verifies both the enclosing aggregate values and the
pointer targets reconstructed from relocation provenance.

## Documentation

Update the supported-features documentation to record that relocation-carrying enum constants may now be nested inside tuple, struct, and array constants.

## Validation

- `git diff --check`
- `cargo test -p mir-importer`
- `cargo clippy --manifest-path crates/rustc-codegen-cuda/examples/enum_constant_provenance/Cargo.toml --all-targets -- -D warnings`
- `scripts/smoketest.sh --compile-only -o '^enum_constant_provenance$'`
- `cargo oxide run enum_constant_provenance`
- `just book`

The importer test suite completed with 1089 passing tests, and the documentation build completed successfully.

The `enum_constant_provenance` regression also passes both libNVVM compilation and GPU runtime validation.

## Non-goals

This PR does not change support for:

- anonymous promoted allocations;
- pointer/integer overlapping enum storage;
- unsupported address-space pointer layouts;
- unrelated enum or aggregate constant layouts.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -11651,13 +11651,15 @@ fn translate_constant_value_from_alloc(
     if is_enum {
         let size = rust_type_layout_size(*rust_ty, loc.clone())?;
         if alloc_has_provenance_in_range(alloc, absolute_byte_offset, size) {
-            return input_err!(
+            return translate_enum_constant_from_alloc(
+                ctx,
+                alloc,
+                absolute_byte_offset,
+                rust_ty,
+                ty_ptr,
+                block_ptr,
+                prev_op,
                 loc,
-                TranslationErr::unsupported(
-                    "Nested enum constant contains pointer relocation(s); cuda-oxide cannot yet \
-                     preserve nested enum pointer provenance"
-                        .to_string()
-                )
             );
         }
     }
--- a/crates/rustc-codegen-cuda/examples/enum_constant_provenance/README.md
+++ b/crates/rustc-codegen-cuda/examples/enum_constant_provenance/README.md
@@ -7,12 +7,16 @@ The example covers:
 - a niche-encoded `Option<&T>` pointing at an interior device-static element;
 - the niche variant without a pointer relocation;
 - a direct-tagged `#[repr(u8)]` enum without a payload;
-- a direct-tagged enum with a pointer payload.
+- a direct-tagged enum with a pointer payload;
+- a direct-tagged relocation-carrying enum nested at a non-zero offset in a
+  struct constant;
+- a niche-encoded `Option<&T>` nested in a tuple constant;
+- a relocation-carrying enum nested in an array constant.
 
-The importer must preserve the enum's outer allocation while reconstructing
-pointer fields from rustc relocation entries. The bytes stored under a
-relocation represent the addend into the target allocation, not an exposed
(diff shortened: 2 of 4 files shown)


nihalpasham:
- Verified on the 5090: enum_constant_provenance passes in both pipelines, importer suite green.

nihalpasham (approved):

PR #992, opened 2026-08-18
perf(mir-importer): promote pointer-bearing constant tables

### LINKED ISSUE #991: Promote constant tables containing pointer/reference fields
## Summary

Constant arrays containing aggregates with pointer/reference fields currently miss the immutable-global promotion path.

For scalar-only and supported aggregate tables, MIR importer can promote the entire constant table into a single immutable device global. Pointer-bearing aggregates, however, fall back to element-wise reconstruction, which may recreate pointer values per thread instead of preserving the table as one relocation-aware constant object.

The remaining gap is to make constant-table promotion aware of pointer relocations.

## Motivation

A table such as:

```rust
const TABLE: [(&u32, bool); 2] = [
    (&FIRST_VALUE, false),
    (&VALUES[2], true),
];
```

should be representable as one immutable device global whose initializer contains symbolic relocations to the referenced device statics.

In particular:

* pointer/reference leaves should not automatically disqualify an otherwise promotable aggregate;
* relocation provenance must be preserved;
* non-zero relocation addends, such as a pointer to a static subobject, must be preserved;
* promotion should avoid rebuilding the pointer-bearing table in per-thread local storage;
* deduplication must distinguish byte-identical initializers that refer to different relocation targets.

## Expected behavior

For supported constant tables containing thin pointers/references:

1. Materialize the table as one immutable `addrspace(1)` device global.
2. Preserve pointer fields as symbolic initializer relocations.
3. Preserve non-zero relocation addends.
4. Materialize referenced static globals as required by those relocations.
5. Deduplicate equivalent promoted tables using both byte contents and relocation identity.
6. Preserve the existing behavior for unsupported cases: bare array constants may fall back to element-wise materialization, while pointer-to-array constants continue to fail closed where no correct fallback exists.

Fat pointers, unsupported enum layouts, packed aggregates, and other currently unsupported promotion cases retain their existing behavior rather than being partially promoted.

## Regression coverage

The existing `array_constants` example already exercises a tuple table containing:

* a pointer to a static object with addend `0`;
* a pointer to a static subobject with a non-zero addend.

The code-shape regression should verify that the table is emitted as a relocation-aware immutable global rather than reconstructed per thread.

uurl:
Closes #991 

## Summary

Promote supported constant tables containing thin pointer/reference fields into relocation-aware immutable device globals.

Previously, pointer-bearing aggregates were excluded from the constant-table promotion path and fell back to element-wise reconstruction. This change preserves their evaluated byte image together with symbolic pointer relocations, allowing the whole table to live in one read-only device global.

### Relocation-aware promoted initializers

`promoted_array_initializer` now produces the full initializer representation rather than only raw bytes and alignment.

For promoted tables it preserves:

- initializer bytes;
- alignment;
- relocation source offsets;
- relocation targets;
- target address spaces;
- relocation addends;
- static identity/provenance.

For pointer/reference-to-array constants, relocations from the selected backing allocation range are rebased into the promoted table.

### Pointer-bearing aggregates

Thin MIR pointer types are now accepted as promotable leaves inside supported array element layouts.

Existing exclusions remain in place for layouts that cannot safely use this promotion path, including packed/unsupported aggregate layouts and unsupported enum cases.

### Relocation target materialization

Promoted globals now explicitly ensure that static globals referenced by their initializer relocations are materialized.

This reuses the same relocation-target materialization logic used by ordinary static-global initializers.

### Provenance-aware deduplication

The promoted-global deduplication key now includes relocation metadata in addition to the byte image.

This prevents two byte-identical initializers with different relocation targets from incorrectly aliasing the same promoted global.

### Code-shape regression

The `array_constants` regression now checks that the pointer tuple table:

- is sourced from an immutable `addrspace(1)` global;
- contains symbolic `ptrtoint` relocation expressions;
- preserves the non-zero `+8` static-subobject relocation;
- does not leave a per-thread pointer-table depot after optimization;
- does not reconstruct pointers through `inttoptr`.

## Example LLVM shape

The promoted table is emitted in the form:

```llvm
@__device_global_* = addrspace(1) constant ... {
    i64 ptrtoint (... @__device_global_* ... to i64),
    ...,
    i64 ptrtoint (
        ptr addrspacecast (
            ptr addrspace(1) getelementptr (
                i8,
                ptr addrspace(1) @__device_global_*,
                i64 8
            ) to ptr
        ) to i64
    ),
    ...
}
```

This keeps the relocation symbolic through the device-global initializer instead of rebuilding the pointer value in each thread.

## Validation

Validated with:

```text
cargo fmt --all -- --check
git diff --check
cargo check -p mir-importer

cargo test -p mir-importer pointer_array_constant_type_tests
cargo test -p mir-importer promotable_array_element_tests
cargo test -p mir-importer promoted_immutable_globals_are_marked_and_dedup_by_type_and_initializer
cargo test -p mir-importer promoted_dedup_key_distinguishes_relocation_identity
cargo test -p mir-importer

cargo clippy -p mir-importer --all-targets --all-features -- -D warnings

./scripts/check-error-example-status.sh

cargo oxide build array_constants

./scripts/smoketest.sh
```

Results:

```text
array_constants: PASS
```

Full smoketest:

```text
Pass: 214 / 214
Fail: 0 / 214
```

The GPU runtime validation was performed on the `array_constants` regression and the full repository smoketest completed with no failures. 

## Scope

This PR intentionally limits promotion to relocation forms already representable by the existing device-global relocation infrastructure.

Unsupported relocation targets and aggregate layouts retain their existing behavior rather than being partially promoted. Bare array constants may fall back to element-wise materialization; pointer-to-array constants continue to fail closed where no correct fallback exists.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -6592,16 +6592,21 @@ fn translate_ptr_to_array_constant(
             ))
         );
     }
-    let (bytes, alignment) =
-        promoted_array_initializer(constant, expected_size, "array", loc.clone())?;
+    let initializer = promoted_array_initializer(constant, expected_size, "array", loc.clone())?;
+    let promoted_anchor = materialize_promoted_initializer_targets(
+        ctx,
+        &initializer.relocations,
+        block_ptr,
+        prev_op,
+        loc.clone(),
+    )?;
 
     let global_alloc = emit_promoted_immutable_global(
         ctx,
         array_ty,
-        &bytes,
-        alignment,
+        &initializer,
         block_ptr,
-        prev_op,
+        promoted_anchor,
         loc.clone(),
     );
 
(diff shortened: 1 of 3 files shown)


nihalpasham:
- Verified on the 5090: array_constants passes in both pipelines, and verify-code-shape.sh confirms the relocated addrspace(1) constant with the +8 addend, no post-opt alloca depot, no inttoptr.

nihalpasham (approved):

PR #994, opened 2026-08-18
chore(f32x2_arith): refresh stale lockfile

### LINKED ISSUE #993: f32x2_arith smoketest rewrites stale Cargo.lock
## Problem

Running the `f32x2_arith` example through the repository smoketest rewrites its committed `Cargo.lock` and leaves the working tree dirty.

The lockfile is stale relative to the current dependency graph: `cuda-host` depends on the local `ptx-parse` crate, but the example lockfile does not record that dependency.

This can also be reproduced with:

```text
cargo check \
  --manifest-path crates/rustc-codegen-cuda/examples/f32x2_arith/Cargo.toml \
  --locked
```

which fails because Cargo needs to update the lockfile.

## Expected behavior

The committed lockfile should already match the current dependency graph so `--locked` succeeds and the smoketest does not modify tracked files.

uurl:
Closes #993 

## Summary

Refresh the `f32x2_arith` example lockfile to match the current dependency graph.

`cuda-host` now depends on the local `ptx-parse` crate, but the committed `f32x2_arith/Cargo.lock` did not include that dependency. As a result, running the repository smoketest regenerated the lockfile and left the working tree dirty.

The refreshed lockfile adds `ptx-parse` to `cuda-host`'s locked dependencies and records the local `ptx-parse` package.

## Validation

Before refreshing the lockfile:

```text
cargo check --manifest-path crates/rustc-codegen-cuda/examples/f32x2_arith/Cargo.toml --locked
error: cannot update the lock file ... because --locked was passed
```

After refreshing it:

```text
cargo check --manifest-path crates/rustc-codegen-cuda/examples/f32x2_arith/Cargo.toml --locked
PASS

./scripts/smoketest.sh -o '^f32x2_arith$'
Pass: 1 / 1
Fail: 0 / 1

git diff --check
PASS
```

### DIFF
--- a/crates/rustc-codegen-cuda/examples/f32x2_arith/Cargo.lock
+++ b/crates/rustc-codegen-cuda/examples/f32x2_arith/Cargo.lock
@@ -18,15 +18,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
-name = "f32x2_arith"
-version = "0.1.0"
-dependencies = [
- "cuda-core",
- "cuda-device",
- "cuda-host",
-]
-
-[[package]]
 name = "bindgen"
 version = "0.71.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified: cargo check --locked is clean, the smoketest passes, and the tree stays clean afterwards. Same lockfile content every other example already carries.

nihalpasham (approved):

PR #995, opened 2026-08-18
docs(mir-lower): make the in-source intrinsics table a set of links the docs gate checks

vyncint:
## Summary

#977 fixed this table in `README.md`. **The same table lives in `convert/intrinsics/mod.rs`** as a module doc comment, rustdoc renders it, and it had the same two dead rows and the same fourteen omissions.

The interesting part is how it survived. Every live row was an intra-doc link and the two dead ones were plain code spans:

```rust
//! | [`wgmma`]    | Warpgroup MMA        | Inline PTX | SM 90+  |
//! | `tcgen05`    | 5th-gen Tensor Core  | Inline PTX | SM 100+ |
//! | `stmatrix`   | Matrix store         | Inline PTX | SM 90+  |
```

`6a9a6373` (#406) made exactly that edit when it deleted those two modules:

```diff
-//! | [`tcgen05`]  | 5th-gen Tensor Core       | Inline PTX      | SM 100+ |
+//! | `tcgen05`    | 5th-gen Tensor Core       | Inline PTX      | SM 100+ |
-//! | [`stmatrix`] | Matrix store              | Inline PTX      | SM 90+  |
+//! | `stmatrix`   | Matrix store              | Inline PTX      | SM 90+  |
```

The docs gate runs `RUSTDOCFLAGS="-D warnings"`, so leaving the links in would have failed the build. Downgrading them to plain spans was the smallest edit that kept CI green — and it converted a row rustdoc was checking into a row nobody was. **The gate caught the link; silencing the gate kept the claim.**

## So this puts the rows back under the gate

All 24 modules are `pub mod`, so every row is now a link and the table cannot rot the way it just did without the build going red. Verified both directions:

- **Clean tree** — `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p mir-lower` passes, all 24 links resolving.
- **One linked row for the deleted `stmatrix`**, nothing else changed:

```
error: unresolved link to `stmatrix`
  --> crates/mir-lower/src/convert/intrinsics/mod.rs:86:9
error: could not document `mir-lower`
```

which is precisely the failure #406 had to work around.

Descriptions come verbatim from each module's own `//!` line — the same source #977 used for the README — so the two tables now say the same thing rather than drifting apart again. `Lowering` and `Min SM` go for the reason #977 dropped them: per-module guesses at per-intrinsic properties the catalog already records.

One thing deliberately **not** a link: `convert/generated_intrinsics.rs`, which handles everything the catalog describes. It is a private `mod`, so `[`super::generated_intrinsics`]` fails the same gate with `private-intra-doc-links`. It is named as a path, and the doc says why.

## Testing

Local, no GPU.

- [x] Table rows are exactly the 24 files in `convert/intrinsics/`, no ghost and no missing
- [x] Docs gate passes; the dead-link mutation fails it as quoted above
- [x] `cargo test -p mir-lower` (353 tests) and `clippy --all-targets` clean
- [ ] `cargo oxide run <example>` — not applicable, doc comment only


nihalpasham:
- Verified all 24 rows against convert/intrinsics/ and each description against its module's first doc line. Git history confirms the two ghost rows were unlinked in the same commit that deleted the modules (6a9a6373), which is how the rustdoc gate missed the rot.
- With every row a checked link again, deleting a listed module turns the docs build red instead of leaving a stale table.

nihalpasham (approved):

PR #996, opened 2026-08-18
docs(mir-importer): TMA is the Tensor Memory Accelerator, and three translator rows were missing

vyncint:
## Summary

Two follow-ups from the #978 merge note.

### The TMA expansion

You flagged that #978's `tma` row was copied verbatim from `tma.rs`, whose own doc comment says "Tensor Memory **Access**". The unit is the Tensor Memory **Accelerator**, and the tree already says so in 22 places — including `cuda-device/src/tma.rs`, `dialect-nvvm/src/ops/generated/tma.rs`, and the book chapter of that name. mir-importer was the holdout, in four places:

```
src/translator/terminator/intrinsics/tma.rs:6    module doc
src/translator/terminator/intrinsics/tma.rs:34   fn doc prose
src/translator/terminator/intrinsics/mod.rs:18   table row
README.md:105                                    table row
```

All four now read "Accelerator", and `Tensor Memory Access` no longer appears anywhere under `crates/mir-importer/`. Fixing the source alone would have left the README I just landed disagreeing with it, so both move together.

This is the cost of the verbatim-from-doc-comment rule #977 and #978 adopted: it cannot drift from the source, which also means it faithfully reproduces the source's mistakes. Worth the trade — but the fix has to land in the source.

### The three translator rows

`src/translator/mod.rs` carries the same module table as the README and was short the same three modules #978 added there: `layout`, `location`, `payload_store`.

## One claim I checked and am *not* making

Unlike #977's table, making these rows intra-doc links does **not** put them under the docs gate. `translator` is `pub(crate) mod` since `4c1d5838` (#833, "stop exporting the translator"), so `cargo doc --no-deps` never documents it and an unresolved link there is not an error. Verified by adding a `[`ghost_module`]` row and watching the gate pass.

It *does* fail under `--document-private-items` — but that flag also surfaces three pre-existing unresolved links elsewhere in the crate (`MirFieldAddrOp`, `MirArrayElementAddrOp`, `MirArrayType`), so turning it on is its own piece of work, not a free win here. The rows are links because the seven around them are; the enforcement claim belongs only to #977's table.

## Testing

Local, no GPU.

- [x] The translator table is exactly the 9 files in `translator/` plus the `terminator` subdirectory it also lists
- [x] The README's `tma` row and `tma.rs`'s doc comment are byte-identical again
- [x] `cargo test -p mir-importer` (1090 tests) and the docs gate clean

### DIFF
--- a/crates/mir-importer/README.md
+++ b/crates/mir-importer/README.md
@@ -102,7 +102,7 @@ file:
 | `layout`      | Rust dynamic-layout intrinsics for slices, `str`, and slice-tailed structs  |
 | `memory`      | Memory access and conversion intrinsics                                     |
 | `saturating`  | Rust compiler saturating integer intrinsics                                 |
-| `tma`         | Tensor Memory Access (TMA) intrinsics                                       |
+| `tma`         | Tensor Memory Accelerator (TMA) intrinsics                                  |
 | `wgmma`       | Hopper WGMMA (Warpgroup Matrix Multiply-Accumulate) intrinsics              |
 
 Per-intrinsic PTX and minimum-SM requirements live in the catalog and are
--- a/crates/mir-importer/src/translator/mod.rs
+++ b/crates/mir-importer/src/translator/mod.rs
@@ -19,6 +19,9 @@
 //! | [`rvalue`]     | Expression translation (binops, casts, etc.)      |
 //! | [`types`]      | Rust type → `dialect-mir` type conversion         |
 //! | [`values`]     | MIR local → alloca slot mapping                   |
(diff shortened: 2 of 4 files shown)


nihalpasham:
- Verified the four corrected sites against cuda.h and the modules on disk: NVIDIA's own headers expand TMA as Tensor Memory Accelerator, and the three restored translator rows match real files.
- The fifth tracked "Access" sat in mir-lower's converter table; #995's rewrite retired it, so the tree is now clean of the wrong expansion.

nihalpasham (approved):

PR #997, opened 2026-08-18
ci: guard the book's Crate Map with the inventory check, not just README's

vyncint:
## Summary

The follow-up named in the #970 merge note.

`architecture-overview.md`'s Crate Map claims completeness outright — *"cuda-oxide is split into focused crates. Here is every one and its role"* — and listed **15 of 28** when #970 found it. Eleven of the thirteen absentees were ordinary omissions; two were `dialect-iket` and `iket-lower`, the exact pair whose absence from README.md is the reason this guard exists at all (*"1,700 lines of compiler across two crates, with no row between them"*). The guard could not see the second table because it only ever read README.md.

One guard, two inventories now. The member list is parsed once and checked against both, so a new crate cannot be absent from either without CI failing.

## Design kept identical to the existing half

- **Missing-only, both tables.** A member with no row fails; a row with no member does not. That is the direction the guard was designed around (`rustc-codegen-cuda` is deliberately listed in both and is not a workspace member), and it cannot false-positive on a future legitimate addition.
- **Per-table parse self-tests.** Each section must be found by heading and yield ≥20 rows, so a renamed heading or reflowed table fails loudly instead of passing vacuously — the failure mode this script's own header warns about.
- **Remedy text differs per table**, because the fix does: README rows go in one of three grouped tables, while the book's role text should be copied from README's so the two stay word-for-word identical (which is how #970 wrote them).

## Testing

Local, no GPU — reads three text files, no cargo. Each mutation against an otherwise clean tree:

| Mutation | Result |
|:--|:--|
| drop `dialect-iket` from the book's Crate Map | exit 1, names the crate |
| drop `iket-lower` from README's Crate Overview | exit 1, unchanged behaviour |
| drop one from **each** at once | exit 1, reports both tables |
| rename `## Crate Map` → `## Crates` | self-test fires, exit 1 |
| gut the book table's rows | self-test fires, exit 1 |
| clean tree | OK, all 27 members in both |

- [x] No wiring needed: `check-crate-inventory.sh` is already in `status-guard.yml` and in the Justfile's `check-guards`, so extending the script extends both

### DIFF
--- a/scripts/check-crate-inventory.sh
+++ b/scripts/check-crate-inventory.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-# Verify README.md's Crate Overview still lists every workspace member.
+# Verify both crate inventories still list every workspace member: README.md's
+# Crate Overview and the book's Crate Map.
 #
 # The overview is the only map of the tree a newcomer gets, and a crate absent
 # from it is invisible: nothing else in the repo enumerates the members for a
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Ran the extended script against the real tree: clean pass with all 27 members, and the six mutation cases (removed row, extra row, renamed crate, in either table) fail loudly with table-specific remedy text.
- This is the follow-up named in the #970 merge note; no CI wiring needed since status-guard already runs the script.

nihalpasham (approved):

PR #998, opened 2026-08-18
ci: no workspace member can be silently absent from the test matrix

vyncint:
## Summary

The follow-up named in the #971 merge note.

`unit-tests.yml`'s matrix decides which crates get `cargo test`, and a member missing from it fails nothing — the workflow is green, `just check` is green, and the tests never ran. #971 found four crates in exactly that state holding **107 `#[test]` functions** between them, two of them uncovered for months.

The workflow already carried the intent, as a comment:

```
# Crates intentionally not in this matrix:
#   * `cuda-bindings` - generated sys bindings covered transitively.
#   * `fuzzer` - differential-testing infra with no unit tests.
#   * `rustc-codegen-cuda` - its own [workspace] ...
```

**That list reading as complete is precisely what hid the four.** This guard makes it complete: every member must be in the matrix or named in that comment, and a crate in neither fails the run. Excluding one stays a one-line reviewable act with a written reason — the guesswork goes, the choice stays.

## Four assertions, each with a failure this has already had

1. **Member in neither the matrix nor the exclusion list.** The #971 case.
2. **Exclusion list names a non-member.** A renamed or deleted crate left behind there is a permanent hole. `rustc-codegen-cuda` is the one accepted non-member, since its own `[workspace]` keeps it out by design.
3. **Matrix names a non-member**, which would make the job run `cargo test -p` on a name cargo cannot resolve.
4. **The Justfile mirrors the matrix**, which its own comment already promises (*"Mirrors .github/workflows/unit-tests.yml; keep the two in step"*) and nothing enforced.

Parse self-tests on all three inputs, in the house style: the member list, the matrix entry shape, the exclusion comment, and the Justfile `-p` spelling each fail loudly rather than yielding an empty set that would pass vacuously.

One thing the guard caught **in its own first run**, worth keeping: a naive `-p ([a-z0-9-]+)` reads `ldconfig -p 2>/dev/null` in `test-cuda` as a package named `2`. The pattern requires a leading letter, and the comment says why.

## Testing

Local, no GPU. Each mutation against an otherwise clean tree:

| Mutation | Result |
|:--|:--|
| drop `dialect-ptx` from the matrix | exit 1, names it |
| drop it **and** declare it excluded, Justfile untouched | exit 1, mirror broken |
| drop it, declare it, **and** drop it from the Justfile | **OK** — the real exclusion flow |
| exclusion list names `ghost-crate` | exit 1, names it |
| Justfile drops a package the matrix keeps | exit 1, reports both directions |
| reword the exclusion heading | self-test fires |
| rename `- package:` → `- pkg:` | self-test fires |
| clean tree | OK, 27 = 25 matrix + 3 declared |

- [x] Wired into `status-guard.yml` as a new job and into the Justfile's `check-guards`, whose docstring lists the status-guard jobs and now names this one — the drift that recipe exists to prevent
- [x] All ten guards, the docs gate, the book gate and every workflow parse clean on the merged tree

### DIFF
--- a/.github/workflows/status-guard.yml
+++ b/.github/workflows/status-guard.yml
@@ -109,6 +109,24 @@ jobs:
       - name: Verify every device call the book shows resolves
         run: bash scripts/check-book-api-names.sh
 
+  # `unit-tests.yml`'s matrix is the only thing deciding which crates get
+  # `cargo test`, and a member missing from it fails nothing: the workflow is
+  # green, `just check` is green, and the tests never ran. #971 found four
+  # crates in that state holding 107 tests between them, two of them for
+  # months. The workflow's own "intentionally not in this matrix" comment
+  # reading as complete is what hid them.
+  test-matrix-coverage:
+    name: test-matrix coverage
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      # Reads Cargo.toml, unit-tests.yml and the Justfile: no cargo, no network.
+      - name: Verify every workspace member is tested or declared untested
+        run: bash scripts/check-test-matrix-coverage.sh
+
   # #702 let a kernel-only crate drop the host surface (and with it `cuda.h`) via
   # `cuda-macros = { default-features = false }`, which is what #701 asked for.
   # Nothing exercised it: no example takes that configuration, and every other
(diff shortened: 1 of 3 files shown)


nihalpasham:
- Ran the new lint against main: 27 members resolve as 25 matrix entries plus declared exclusions, and the mutation table replays cleanly: stale exclusions, non-member matrix entries, and Justfile drift all fail with the described messages.
- This is the follow-up requested in the #971 merge note.

PR #1001, opened 2026-08-18
fix(mir-lower): support packed address-space 3 aggregates across the internal device ABI

### LINKED ISSUE #1000: Support packed aggregates containing AS3 pointers across the internal device ABI
## Problem

Packed aggregates are supported by value after #941, but packed aggregates containing shared-memory / address-space-3 pointers still fail closed.

That restriction is necessary for physical memory images because the representation of an AS3 pointer is target-mode dependent. However, the same restriction is currently applied even when the packed aggregate only crosses an internal device-function ABI boundary.

For example:

```rust
#[repr(C, packed)]
struct Cursor {
    tag: u8,
    ptr: *mut SharedThing,
}

#[inline(never)]
#[device]
fn bounce(value: Cursor) -> Cursor {
    value
}
```

The semantic device-side value can remain a packed aggregate containing an AS3 pointer, but the internal function return ABI needs a target-stable physical representation.

## Background

#867 added address-space-aware internal ABI handling for ordinary aggregate arguments and returns.

#884 established the scalar precedent for AS3 values whose physical representation is target dependent: cross the ABI boundary through the generic-pointer representation rather than relying on the backend pointer width.

#941 added packed aggregate by-value support while deliberately keeping packed aggregates containing shared pointers unsupported.

The remaining gap is the combination of those cases: a packed aggregate with an AS3 pointer crossing an internal device ABI boundary.

## Proposed approach

Keep the semantic value unchanged inside the function body:

```text
packed semantic aggregate with AS3 pointer
```

At the internal return boundary, rebuild it into a target-stable packed carrier:

```text
packed semantic <{ scalar, p3 }>
        ↓
packed ABI carrier <{ scalar, p0 }>
        ↓ internal device return
packed ABI carrier <{ scalar, p0 }>
        ↓
packed semantic <{ scalar, p3 }>
```

The address-space conversion should be explicit through `addrspacecast`.

## Initial scope

Keep the first implementation deliberately narrow:

* internal device ABI only;
* packed structs only;
* exactly one direct AS3/shared pointer field;
* remaining non-ZST fields must be scalar;
* arguments may continue to use the existing flattened internal ABI path;
* returns use a target-stable generic-pointer aggregate carrier.

## Out of scope

This issue should not enable:

* packed AS3 kernel parameters;
* whole-value loads or stores of packed aggregates containing AS3 pointers;
* arrays containing AS3 pointers;
* nested aggregate AS3 leaves;
* vectors containing AS3 pointers;
* multiple AS3 pointer leaves;
* enums or arbitrary recursive packed layouts.

Whole-value packed AS3 memory images must remain fail closed because their physical byte representation is target-mode dependent.

## Acceptance criteria

* A packed struct containing one direct AS3 pointer can be constructed as a semantic SSA value.
* An internal device function can return that struct.
* The function ABI uses a packed generic-pointer carrier instead of the semantic AS3 representation.
* The caller reconstructs the semantic AS3 value after the call.
* No change is made to kernel ABI handling.
* Whole-value packed AS3 memory load/store remains rejected.
* Existing shared-pointer aggregate and enum regressions continue to pass.
* End-to-end CUDA execution confirms the shared pointer survives the internal call round trip.

uurl:
Closes #1000 

## Summary

Adds support for the first packed aggregate shape containing an AS3/shared-memory pointer across the internal device ABI.

The implementation keeps the packed aggregate's semantic representation unchanged inside device-function SSA, but uses a target-stable packed carrier containing a generic pointer when the value crosses an internal return boundary.

This extends the aggregate ABI work from #867, the scalar AS3 representation strategy from #884, and the packed aggregate support from #941.

## Problem

#941 intentionally kept packed aggregates containing shared-memory pointers unsupported by value because their physical representation depends on the selected NVPTX target mode.

That restriction is still correct for:

- kernel parameters;
- whole-value memory loads;
- whole-value memory stores.

It is unnecessarily restrictive for an internal device-function return, where the compiler controls both sides of the ABI boundary.

A packed semantic value such as:

```llvm
<{ i8, ptr addrspace(3) }>
```

can cross that boundary through a target-stable carrier:

```llvm
<{ i8, ptr }>
```

and be reconstructed back to the semantic AS3 representation at the caller.

## Design

The new internal return path is:

```text
semantic packed aggregate
<{ scalar, p3 }>
        ↓ target-stable coercion
internal ABI carrier
<{ scalar, p0 }>
        ↓ call / return
internal ABI carrier
<{ scalar, p0 }>
        ↓ target-stable coercion
semantic packed aggregate
<{ scalar, p3 }>
```

The AS3 ↔ generic conversion is represented explicitly with `addrspacecast`.

The function body itself continues to use the semantic packed AS3 type.

## Target-stable aggregate storage

This PR factors the recursive physical-storage rewrite previously used by enum payload storage into a shared helper.

The helper supports:

* shared pointer → generic pointer rewriting;
* optional bool `i1` ↔ `i8` canonicalization;
* recursive structs;
* recursive arrays where allowed by the caller;
* symmetric value reconstruction.

Enum payload storage retains its existing bounded-array and bool-canonicalization policy on top of the shared implementation.

## Packed AS3 internal ABI classification

The new packed-AS3 ABI path is intentionally narrow.

It accepts only:

* a byte-faithful packed struct;
* exactly one direct AS3/shared pointer field;
* scalar non-ZST sibling fields.

It rejects:

* nested aggregate AS3 leaves;
* arrays;
* vectors;
* multiple AS3 pointer leaves;
* layouts that cannot be represented faithfully.

The target-stable carrier must preserve the recorded aggregate size.

## Arguments and returns

Internal aggregate arguments already use the existing flattened field ABI path, so no new aggregate argument carrier is required.

For eligible packed-AS3 returns:

* the callee rebuilds the semantic packed value into the generic-pointer carrier before `llvm.return`;
* the call result is typed as the carrier;
* the caller reconstructs the semantic packed AS3 aggregate after the call.

Kernel entry returns and parameters are unchanged.

## Memory behavior

Whole-value packed AS3 loads and stores remain fail closed.

This PR does not make the target-dependent packed memory image legal.

The new exception applies only to semantic SSA construction and the controlled internal device ABI boundary.

## End-to-end coverage

`packed_aggregate_abi` now includes a device-only packed shared-pointer case:

```rust
#[repr(C, packed)]
struct PackedShared {
    tag: u8,
    ptr: *mut SharedArray<u32, 1>,
}
```

The value crosses internal device-function boundaries while remaining in SSA.

The shared pointer is created on device and is never exposed through the host/kernel parameter ABI.

Runtime validation confirms that the shared-memory pointer survives the round trip.

## Validation

```text
cargo fmt --all -- --check
git diff --check

cargo test -p mir-lower packed_struct_construction -- --nocapture
cargo test -p mir-lower packed_shared -- --nocapture

```

Results:

```text
mir-lower unit tests:        248 passed
mir-lower integration tests: 107 passed
```

End-to-end:

```text
cargo oxide build packed_aggregate_abi
```

Results:

```text
packed_aggregate_abi: PASS
(runtime values, packed shared internal ABI, whole-value load/store, and PTX parameter shapes)

PASS addressof_sharedarray:
shared address is non-null, enum storage and struct/tuple/array/enum
device-call round trips preserved shared pointers
```

## Out of scope

This PR intentionally does not support:

* packed AS3 kernel parameters;
* whole-value packed AS3 memory loads or stores;
* nested AS3 aggregates;
* arrays containing AS3 pointers;
* vectors containing AS3 pointers;
* multiple AS3 pointer leaves;
* broader recursive packed-AS3 ABI legalization.

### DIFF
--- a/crates/mir-lower/src/convert/enum_payload_storage.rs
+++ b/crates/mir-lower/src/convert/enum_payload_storage.rs
@@ -20,17 +20,13 @@
 //! the same contract. Pointer vectors remain fail-closed because they require
 //! separate ABI and address-space-cast semantics.
 
-use llvm_export::op_interfaces::{CastOpInterface, CastOpWithNNegInterface};
-use llvm_export::ops as llvm;
-use llvm_export::types as llvm_types;
-use llvm_export::types::PointerTypeExt;
-use pliron::builtin::types::{IntegerType, Signedness};
+use crate::convert::target_stable_storage::{
+    StorageRewriteOptions, coerce_target_stable_value, target_stable_storage_type,
+};
 use pliron::context::Context;
 use pliron::irbuild::dialect_conversion::DialectConversionRewriter;
-use pliron::irbuild::inserter::Inserter;
-use pliron::op::Op;
 use pliron::result::Result;
-use pliron::r#type::{TypeHandle, Typed};
+use pliron::r#type::TypeHandle;
 use pliron::value::Value;
 
 /// Maximum number of array-expanded shared-pointer leaves one payload rewrite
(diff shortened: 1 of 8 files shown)


nihalpasham:
- Verified on the 5090: packed_aggregate_abi and addressof_sharedarray both pass, mir-lower suite green including the target_stable_storage refactor.
- Tried strengthening the example to read value.ptr inside consume_packed_shared: any field projection forces the whole-value store that #1000 lists as out of scope, and it fails closed with the documented diagnostic. So the runtime check stays as authored; the coercion itself is proven by the unit tests (insertvalue slots, address spaces, both directions).
- Filing a follow-up for the next rung: carrier-typed local slots so field projections of packed AS3 aggregates work.

nihalpasham (approved):

PR #1003, opened 2026-08-18
fix(mir-importer): forward generated ldmatrix result bundles

### LINKED ISSUE #1002: Forward generated ldmatrix result bundles through compiler-owned result forwarding
## Problem

Generated multi-result `ldmatrix` intrinsics are still bundled into a Rust array through the older importer helper without marking that aggregate as compiler-owned.

As a result, the existing `forward_compiler_result_bundles` pass cannot recognize those arrays as safe forwarding candidates.

This is mostly hidden by the normal optimized pipeline because downstream LLVM optimization removes the temporary aggregate. Under `CUDA_OXIDE_NO_OPT=1`, however, the generated bundle is materialized through local memory, introducing unnecessary `st.local` / `ld.local` traffic around `ldmatrix`.

This is the remaining `ldmatrix` follow-up identified in #888.

## Proposed fix

Mark arrays created by `bundle_generated_u32_results_as_array` with the existing `mir.compiler_result_bundle` marker.

The helper is specifically used for compiler-generated multi-result carriers, so ordinary Rust arrays remain unmarked and outside the forwarding pass.

No changes should be required to:

- `forward_compiler_result_bundles`
- generated intrinsic call sites
- `ldmatrix` overlays
- generated source files

## Acceptance criteria

- Generated multi-result `ldmatrix` bundles carry the compiler-owned result marker.
- Result ordering is preserved.
- Normal optimized PTX remains unchanged.
- `CUDA_OXIDE_NO_OPT=1` no longer materializes the generated `ldmatrix` result bundle through local memory.
- Existing importer, transform, generator, formatting, and Clippy checks remain green.
- The existing generated `ldmatrix` runtime oracle continues to validate all returned fragments.

uurl:
Closes #1002 

## Summary

Mark arrays produced by `bundle_generated_u32_results_as_array` as compiler-owned result bundles so the existing `forward_compiler_result_bundles` pass can forward generated multi-result `ldmatrix` values directly from their producer results.

This closes the remaining `ldmatrix` forwarding gap called out in #888.

## Changes

- attach the existing `mir.compiler_result_bundle` marker to generated `u32` result arrays
- document that the helper is restricted to compiler-generated multi-result carriers
- add a regression test verifying:

  - the generated array carries the forwarding marker
  - producer result ordering is preserved

No changes are made to the forwarding pass, generated intrinsic dispatcher, overlays, or generated files.

## Validation

All standard validation passes:

- `cargo test -p mir-importer --all-targets`
  - 1090 tests passed
- `cargo test -p mir-transforms --all-targets`

### PTX validation

Using the existing `generated_ldmatrix` example and an unmodified `HEAD` worktree as the baseline:

**Default pipeline**

```text
PASS: default PTX is byte-identical
```

**CUDA_OXIDE_NO_OPT=1**

Before:

```text
st.local=16
ld.local=15
```

After:

```text
st.local=0
ld.local=0
```

The resulting NO_OPT PTX keeps the `ldmatrix` results directly in registers for the x1/x2/x4 and transpose forms.

### Runtime oracle

On sm_120:

```text
```

The runtime oracle validates all 128 x4 fragments, while the regression test and NO_OPT PTX inspection verify result ordering and direct-register forwarding across the generated ldmatrix path.

### DIFF
--- a/crates/mir-importer/src/translator/terminator/helpers.rs
+++ b/crates/mir-importer/src/translator/terminator/helpers.rs
@@ -376,10 +376,11 @@ pub fn set_compiler_result_bundle_marker(ctx: &mut Context, op: Ptr<Operation>) 
 }
 
 /// Bundle a generated operation's independent `u32` results into the Rust
-/// array value expected by its raw ABI.
+/// array value expected by its raw ABI and mark the compiler-owned adapter
+/// for result forwarding.
 ///
-/// Keeping this adapter here lets later multi-result families reuse the same
-/// SSA-to-array boundary without introducing a stack temporary.
+/// This helper is only for compiler-generated multi-result carriers. Ordinary
+/// Rust arrays must never receive the forwarding marker.
 pub fn bundle_generated_u32_results_as_array(
     ctx: &mut Context,
     producer: Ptr<Operation>,
@@ -400,6 +401,7 @@ pub fn bundle_generated_u32_results_as_array(
         0,
     );
     array.deref_mut(ctx).set_loc(loc);
+    set_compiler_result_bundle_marker(ctx, array);
     array.insert_after(ctx, producer);
     (array.deref(ctx).get_result(0), array)
 }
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified on the 5090: all 128 fragments match, and with CUDA_OXIDE_NO_OPT=1 the PTX has zero st.local/ld.local, so the forwarding now works without leaning on LLVM's optimizer.

nihalpasham (approved):

PR #1005, opened 2026-08-18
feat(codegen): forward bounded dynamic result-bundle indexing

### LINKED ISSUE #1004: codegen: forward bounded dynamic indexing of compiler-owned result bundles
## Problem

`forward_compiler_result_bundles` currently forwards reads from compiler-owned multi-result bundles only when every field/array projection uses a constant index.

This leaves bounded dynamic accesses on the memory path even when the compiler can prove that the index has a small finite candidate set.

For example:

```rust
let regs = intrinsic_returning_array();
let value = regs[index % 4];
```

The importer-created result bundle is already known to be compiler-owned, immutable, non-escaping, and backed by the results of a single producer, but the dynamic `mir.array_element_addr` prevents the existing forwarding pass from recovering the SSA aggregate.

## Existing infrastructure

The MIR dialect and lowering already support bounded dynamic extraction through `mir.extract_array_element`.

When its index has the form:

```text
unsigned_value % constant
```

and the candidate count is small enough, lowering scalarizes the access into fixed `extractvalue` candidates plus a select chain.

`MAX_SCALARIZED_CANDIDATES` currently limits this expansion to 16 candidates.

## Proposed scope

Extend `forward_compiler_result_bundles` to accept a terminal dynamic array projection when all of the following hold:

* the aggregate is an importer-marked compiler result bundle;
* the existing single-producer / single-store / non-escaping boundary proof succeeds;
* the load is non-volatile;
* the dynamic projection is an array element access;
* the index is an unsigned `mir.rem(value, constant)`;
* `constant > 0`;
* `constant <= array length`;
* `constant <= MAX_SCALARIZED_CANDIDATES`;
* the dynamic array projection is terminal.

Instead of resolving the access to one statically selected producer result, rewrite the load to:

```text
mir.extract_array_element %ssa_array, %index
```

and let the existing MIR-to-LLVM lowering decide how to scalarize it.

Unsupported or unsafe cases should continue to fail closed and preserve the complete memory path.

## Expected behavior

Accepted:

```text
bundle[index % 4]
bundle[index % 16]
```

when the corresponding array is large enough.

Rejected:

```text
bundle[index]       // unbounded
bundle[index % 17]  // exceeds scalarization limit
bundle[index % 5]   // array length is 4
signed_index % 4    // signed remainder
```

Existing escape, mutation, volatile, and additional-store rejection rules should remain unchanged.

## Acceptance criteria

* bounded unsigned remainder indexing is forwarded to `mir.extract_array_element`;
* constant-index forwarding remains unchanged;
* candidate count 16 is accepted;
* candidate count 17 is rejected;
* a bound greater than the array length is rejected;
* unbounded dynamic indices are rejected;
* signed remainder indices are rejected;
* rejected cases leave the original memory path intact;
* existing `mir-transforms`, `mir-lower`, and `cuda-oxide-codegen` tests continue to pass.

uurl:
Closes #1004

## Summary

Extend compiler-result bundle forwarding to handle bounded dynamic array indexing.

`forward_compiler_result_bundles` previously required every array projection to use a constant index. As a result, accesses such as:

```rust
let regs = intrinsic_returning_array();
let value = regs[index % 4];
```

kept the importer-created aggregate on the memory path even though the bundle boundary had already been proven safe.

This PR recognizes a narrow bounded dynamic form and rewrites it to the existing SSA aggregate extraction operation.

## What changed

`forward_compiler_result_bundles` now accepts a terminal dynamic array projection when:

* the result bundle passes the existing compiler-owned boundary proof;
* the index type is unsigned;
* the index is produced by `mir.rem(value, constant)`;
* the constant is non-zero;
* the constant does not exceed the array length;
* the constant does not exceed `MAX_SCALARIZED_CANDIDATES`.

For accepted accesses, the pass emits:

```text
mir.extract_array_element %array, %index
```

instead of preserving the `alloca -> store -> array_element_addr -> load` path.

The existing `mir.extract_array_element` lowering already recognizes bounded unsigned remainder indices and lowers them to fixed `extractvalue` candidates plus a select chain when profitable.

No second scalarization implementation is introduced here.

Source-level dynamic indices may initially reach this transform through temporary MIR slots, so the direct `mir.rem` shape is not always visible before mem2reg.

To make the optimization reachable from real compiler output, the same fail-closed compiler-result forwarding pass is run again immediately after mem2reg. At that point, promotable index temporaries have been converted to SSA and bounded forms such as `index % C` are directly visible to the existing recognition logic.

The original pre-mem2reg invocation remains in place for constant/static result bundle forwarding.

## Fail-closed behavior

The pass still leaves the complete memory path untouched for:

* arbitrary dynamic indices;
* signed remainder indices;
* candidate sets larger than `MAX_SCALARIZED_CANDIDATES`;
* bounds larger than the source array;
* volatile loads;
* additional stores or escaping/unknown pointer users;
* unsupported projection shapes.

Constant-index forwarding is unchanged.

## Constructor lifetime

Dynamic extraction retains the SSA `mir.construct_array` needed by `mir.extract_array_element`.

Bundle constructors are now erased only when their result has no remaining uses, while constant forwarding continues to eliminate the complete aggregate construction when possible.

## Tests

Added transform coverage for:

* bounded dynamic indexing;
* the maximum 16-candidate boundary;
* candidate count greater than 16;
* candidate count greater than the array length;
* unbounded dynamic indexing;
* signed remainder indexing.

Added an end-to-end `generated_ldmatrix` regression using a source-level bounded dynamic access:

```rust
let register_index = lane % 4;
let value = registers[register_index];
```

This exercises the complete rustc -> MIR -> mem2reg -> compiler-result forwarding -> LLVM/PTX pipeline.

Validation performed:

```text
cargo test -p mir-transforms
25 passed

cargo test -p mir-lower
252 unit tests passed
117 lowering tests passed

176 unit tests passed
integration tests passed

cargo clippy \
  -p mir-transforms \
  -p mir-lower \
  -p cuda-oxide-codegen \
  --all-targets \
  -- \
  -D warnings

cargo fmt --all -- --check
git diff --check
```

End-to-end validation on sm_120:

```text
cargo oxide run generated_ldmatrix

PASS: all 32 bounded dynamic ldmatrix selections matched on sm_120
```

With LLVM optimization disabled:

```text
CUDA_OXIDE_NO_OPT=1 cargo oxide run generated_ldmatrix

PASS: all 32 bounded dynamic ldmatrix selections matched on sm_120
```

The generated PTX for the bounded dynamic result-bundle path contains no local memory round trip:

```text
st.local = 0
ld.local = 0
```

### DIFF
--- a/crates/cuda-oxide-codegen/src/prep.rs
+++ b/crates/cuda-oxide-codegen/src/prep.rs
@@ -101,6 +101,28 @@ pub fn prepare_mir_module(
     })?;
     verify_operation(ctx, module, "module post-mem2reg")?;
 
+    // Source-level dynamic indices commonly pass through temporary MIR slots.
+    // After mem2reg, bounded forms such as `index % C` are direct SSA values,
+    // while result-bundle allocas blocked by dynamic array projections remain.
+    // Re-run the same fail-closed forwarding proof here so those bundles can
+    // be recovered without teaching the pass to trace arbitrary stack slots.
+    mir_transforms::forward_compiler_result_bundles::forward_compiler_result_bundles(
+        module,
+        ctx,
+        &mut analyses,
+        preparation.verbose,
+    )
+    .map_err(|error| PipelineError::Verification {
+        name: "post-mem2reg compiler-result forwarding".to_string(),
+        message: error.disp(ctx).to_string(),
+        operation: None,
+    })?;
+    verify_operation(
+        ctx,
+        module,
+        "module post-mem2reg-compiler-result-forwarding",
+    )?;
+
     // Formation passes that need promoted SSA values but must still see the
     // original loop CFG run here. In particular, a reduction formation pass
     // cannot safely infer a source loop once generic unrolling has cloned it.
--- a/crates/mir-transforms/src/forward_compiler_result_bundles.rs
+++ b/crates/mir-transforms/src/forward_compiler_result_bundles.rs
@@ -20,7 +20,9 @@
 //! * the construct tree contains every result of exactly one producer, in order;
 //! * the outer bundle has one non-volatile store to one local alloca;
 //! * the alloca has no other stores, copies, escapes, or unknown pointer users;
-//! * every read is a non-volatile load through constant field/array projections;
+//! * every read is a non-volatile load through constant projections, or through
+//!   one terminal array projection whose unsigned runtime index is proven to be
+//!   `value % C` with a small constant candidate set;
 //! * the producer/store dominate every rewritten load.
 //!
 //! Any failed proof leaves the IR untouched.  Ordinary Rust aggregates are
@@ -29,12 +31,15 @@
 use dialect_mir::{
     attributes::{COMPILER_RESULT_BUNDLE_ATTR_KEY, CompilerResultBundleAttr},
     ops::{
-        MirAllocaOp, MirArrayElementAddrOp, MirConstantOp, MirConstructArrayOp,
-        MirConstructStructOp, MirConstructTupleOp, MirFieldAddrOp, MirLoadOp, MirStoreOp,
+        MAX_SCALARIZED_CANDIDATES, MirAllocaOp, MirArrayElementAddrOp, MirConstantOp,
+        MirConstructArrayOp, MirConstructStructOp, MirConstructTupleOp, MirExtractArrayElementOp,
+        MirFieldAddrOp, MirLoadOp, MirRemOp, MirStoreOp,
     },
+    types::MirArrayType,
 };
 use pliron::{
     basic_block::BasicBlock,
+    builtin::types::{IntegerType, Signedness},
     context::{Context, Ptr},
     graph::dominance::DomInfo,
     irbuild::{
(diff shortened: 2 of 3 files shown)


nihalpasham (changes requested):
The select-chain rewrite is sound and everything fails closed, but the feature is unreachable from real compiler output:

- `validate_bounded_dynamic_index` requires the index's defining op to be `MirRemOp` directly. The pass runs pre-mem2reg, where a source-level `regs[i % 4]` arrives as a `mir.load` from the index's temp slot, so the gate never fires outside the hand-built test IR. Confirmed against the importer's `translate_place` address path.
- **Two ways to fix:** 
  - trace the index through its slot (a load from an alloca with a single dominating rem store, mirroring the bundle proof this pass already does), or 
  - run the recognition again post-mem2reg where the rem is a direct operand; `canonicalize_bounded_borrowed_pointer_arguments` already lives there.
- Add an end-to-end test through the real pipeline showing `st.local`/`ld.local` disappear for a source-level modulo index, not just hand-built IR.
- Another point while you're there: assert-derived bounds (the Asserted form the sibling pass supports) can carry the same proof when no rem is present.

uurl:
Addressed the reachability issue by rerunning compiler-result forwarding immediately after mem2reg, where the source-level modulo index is available directly in SSA.

Added an end-to-end `generated_ldmatrix` regression using `registers[lane % 4]`.

Verified on sm_120:

```text
normal pipeline:
PASS: all 128 ldmatrix x4 fragments matched
PASS: all 32 bounded dynamic ldmatrix selections matched

CUDA_OXIDE_NO_OPT=1:
PASS: all 128 ldmatrix x4 fragments matched
PASS: all 32 bounded dynamic ldmatrix selections matched

generated PTX:
st.local = 0
ld.local = 0
```

The existing unit tests and mir/codegen suites remain green.

I kept assert-derived bounds outside this PR for now so #1004 remains scoped to the `% constant` form.

nihalpasham:
- Verified end to end on the current head (which added a post-mem2reg re-run of the forwarding pass plus a real end-to-end kernel): mir-transforms 25 tests green, `cargo oxide run generated_ldmatrix` passes on sm_120 in both the default and `CUDA_OXIDE_NO_OPT=1` pipelines with all 32 bounded dynamic selections matched, per-example fmt/clippy clean, full compile-only smoketest 216/216.
- The second call site is the same fail-closed proof, not a widening: non-load users, type-mismatched loads, and signed/oversized/over-length/unbounded indices all still decline scalarization, and the recognizer shares `MAX_SCALARIZED_CANDIDATES` and the urem-of-constant contract with the mir-lower select-chain lowering.
- The deferred assert-derived bounds are a reasonable follow-up issue rather than a blocker.

nihalpasham (approved):

PR #939, opened 2026-08-16
Add debug info for static MIR place projections

uurl:
Closes #938 

Depends on #937. This PR is currently stacked on top of the enum-debug work and should be rebased onto `main` after #937 lands.

## Summary

Adds full-debug support for source variables represented by statically addressable MIR place projections.

The importer now preserves projected debug bindings for:

- struct fields
- tuple fields
- fixed-array constant indices
- chains composed from those projection kinds

For each supported projection, the importer computes the byte offset from rustc layout metadata. The offset is propagated through MIR lowering and emitted as a projected `llvm.dbg.declare` using a `DIExpression` with `DW_OP_plus_uconst`.

This allows multiple source variables to refer to different offsets within the same underlying MIR local / stack allocation.

This is the projected-source-variable portion of the remaining debug-info work tracked in #226.

## Implementation

- add projected-variable debug metadata alongside existing whole-local metadata
- derive static projection offsets from rustc layout information
- support `Field` projections
- support fixed `ConstantIndex` projections
- preserve projected debug metadata through MIR-to-LLVM lowering
- emit projected `dbg.declare` intrinsics relative to the base allocation
- use `DIExpression(DW_OP_plus_uconst, offset)` for non-zero offsets
- keep the existing whole-local debug path unchanged
- add focused LLVM exporter coverage
- add `compiler_features` fixtures for:
  - struct field projection
  - tuple field projection
  - fixed-array element projection
- extend the `cuda-gdb` smoketest to verify the projected values directly
- update the supported-features documentation

## Validation

Formatting and workspace tests:

```text
cargo fmt --all
cargo fmt --all -- --check
cargo test --workspace
```

Focused exporter test:

```text
cargo test -p llvm-export \
  full_debug_metadata_emits_static_projected_dbg_declares \
  -- --nocapture
```

Result:

```text
test full_debug_metadata_emits_static_projected_dbg_declares ... ok
```

End-to-end hardware validation:

```text
scripts/debug-smoketest.sh compiler_features
```

Tested with `cuda-gdb` on `sm_120`.

Projected source variables were inspectable with the expected values:

```text
$1 = 18
$2 = 4294967329
$3 = 37
```

The full debugger smoketest completed successfully:

```text
debug-smoketest: PASS (source debugging + closure environments + Rust enums + static projections verified on sm_120)
```

`cargo test --workspace` also passes.

## Scope / non-goals

This PR intentionally supports only projections that can be represented as a static byte offset from the base local.

The following remain out of scope:

[...]

PR #937, opened 2026-08-16
Add full debug info for Rust enums

uurl:
Closes #936 

## Summary

Adds full device debug metadata for Rust enums so `cuda-gdb` can identify the active variant and inspect its payload.

This implements the enum-debug portion of #226. Projected source variables remain out of scope for this PR.

### MIR debug type import

Adds enum support to `DebugLocalTypeKind` and imports rustc enum layout information for debug metadata.

The importer now records:

- enum size
- physical discriminant location and type
- variant names
- variant discriminant values
- variant payload members and their byte offsets

Both direct-tag and niche-encoded layouts are supported.

For niche layouts, the untagged variant is emitted without a discriminant value, while tagged niche variants use the physical niche value.

### LLVM debug metadata

Emits Rust enums using the DWARF variant-part representation:

```text
enum DICompositeType
├── DW_TAG_variant_part
│   ├── artificial discriminant member
│   ├── variant member
│   └── variant member
└── variant payload structures
```

The metadata preserves the parent/scope relationships used by rustc's native LLVM debuginfo implementation, allowing debuggers to resolve both the active variant and its payload.

### End-to-end coverage

Adds enum debug coverage to `compiler_features`, including:

* `Option<u32>`
* `Result<u32, u64>`
* a direct-tag custom enum
* niche-encoded `Option<&u32>`

The CUDA GDB smoketest now verifies the actual printed variant and payload and explicitly rejects `<No data fields>` regressions.

## Validation

Focused LLVM exporter test:

```text
full_debug_metadata_emits_rust_enum_variant_parts ... ok
```

Full workspace:

```text
cargo test --workspace
```

passes.

End-to-end CUDA GDB validation on `sm_120`:

```text
$1 = Option::Some (8)
$2 = Result::Err (4294967305)
$3 = DebugDirectEnum::Wide (8589934603)
$4 = Option::Some (0x...)
```

and:

```text
debug-smoketest: PASS (source debugging + closure environments + Rust enums verified on sm_120)
```

## Out of scope

Debug declarations for source variables backed by MIR place projections are not included here and will be handled separately.

PR #1007, opened 2026-08-18
debug: expose enum payload projection bindings

### LINKED ISSUE #1006: Debug source bindings for enum payload projections
## Summary

Full device debug info can describe Rust enums themselves, including direct-tag and niche layouts, but source variables bound to enum payloads are still skipped when their MIR debug place contains an enum `Downcast`.

For example, a binding produced by:

```rust
match value {
    DebugDirectEnum::Wide(projected_enum_payload) => {
        // projected_enum_payload
    }
}
```

may be represented by rustc as a projected debug place equivalent to:

```text
Downcast(Wide) -> Field(0)
```

`static_debug_projection()` currently handles `Field` and forward `ConstantIndex` projections, but rejects `Downcast`, so the projected source binding is not emitted.

## Existing support

This builds on the debug infrastructure already landed in:

* #937: full Rust enum debug types, including direct-tag and niche layouts
* #939: static MIR place projection bindings using address-offset DWARF expressions

Enum payload addressing itself is already supported by the compiler. The missing piece is specifically source-level debug binding generation for enum payload projections.

## Proposed behavior

Extend static debug projection resolution to support:

```text
Downcast(variant) -> Field(field)
```

The field byte offset should come from rustc's physical enum layout rather than being inferred from the discriminant representation.

`enum_variant_field_offsets()` already provides the required per-variant physical field offsets.

The implementation should continue to reject unsupported projection kinds rather than approximating them.

Supported compositions should include:

```text
Downcast -> Field
Downcast -> Field -> Field
Downcast -> Field -> ConstantIndex
```

while dereferences, runtime indices, subslices, opaque casts, and composite fragments remain unsupported.

## Acceptance criteria

* A source binding represented as `Downcast -> Field` is emitted as a projected debug variable.
* The emitted binding uses the correct rustc-reported payload byte offset.
* Existing struct, tuple, and fixed-array static projection debugging remains unchanged.
* `cuda-gdb` can inspect an enum payload source binding inside the matching variant scope.
* A regression fixture verifies the complete path on a real GPU.
* Documentation is updated to list enum payload projections as supported.

## Example expected debugger behavior

At a breakpoint inside the `Wide` arm:

```text
(cuda-gdb) ptype projected_enum_payload
type = @local u64

(cuda-gdb) print projected_enum_payload
$1 = 8589934603
```

Related tracking issue: #226.

uurl:
Closes #1006

## Summary

Teach full-debug source binding generation to resolve enum payload projections represented by MIR as:

```text
Downcast(variant) -> Field(field)
```

This closes the remaining gap between the enum debug-type support from #937 and the static projected-variable support from #939.

## Implementation

`static_debug_projection()` now tracks an enum `Downcast` until the following `Field` projection.

For that field, the byte offset is obtained from:

```rust
crate::translator::layout::enum_variant_field_offsets(...)
```

using rustc's physical layout for the selected variant.

The resulting offset is accumulated into the existing projected-variable path, so no changes are required in `llvm-export` or `mir-lower`.

The implementation deliberately fails closed:

* a `Downcast` must be consumed by a following `Field`
* nested/unconsumed downcasts are rejected
* `ConstantIndex` cannot consume a pending enum downcast
* dereferences, runtime indices, subslices, opaque casts, and composite fragments remain unsupported

This also allows static compositions after the payload field, such as:

```text
Downcast -> Field -> Field
Downcast -> Field -> ConstantIndex
```

## Regression coverage

The `compiler_features` full-debug fixture now binds the payload of:

```rust
DebugDirectEnum::Wide(u64)
```

to a source variable named:

```text
projected_enum_payload
```

A dedicated `cuda-gdb` smoke-test pass breaks inside that variant arm and verifies:

```text
ptype projected_enum_payload
print projected_enum_payload
```

with the expected value:

```text
8589934603
```

This validates debugger consumption of the generated DWARF, not only LLVM metadata generation.

## Validation

Validated with:

```text
cargo fmt --all -- --check
cargo test -p mir-importer
cargo test -p llvm-export
scripts/debug-smoketest.sh compiler_features
just book
git diff --check
```

The device debug smoke test passes on `sm_120`, including the new enum payload projection check.

## Documentation

The supported-features matrix now documents static enum `Downcast -> Field` payload bindings as supported.

## Related

* #226
* #937
* #939


nihalpasham:
- Verified with cuda-gdb on sm_120: the projection breakpoint hits and the payload bindings print; the fixture's host check still passes.
- One nit for the record: the PR text says nested downcasts are rejected, but Downcast-Field-Downcast-Field is accepted, which is the correct behavior.

nihalpasham (approved):

PR #1011, opened 2026-08-18
test(codegen): validate spill-free three-slot WGMMA counted pipeline

### LINKED ISSUE #1010: Validate spill-free three-slot WGMMA counted pipeline on sm_90a
## Motivation

The current counted WGMMA pipeline has backend coverage for two accumulator slots using `wait_group<1>`.

Before extending the production counted-loop recognizer and carrier to support a three-slot pipeline with `wait_group<2>`, we should first validate that the larger tied-accumulator lifetime remains register-resident on `sm_90a`.

The main question is whether the backend can keep the following values live inside a single convergent inline-PTX counted loop without introducing stack or spill traffic:

- 96 tied `f32` accumulator values
- 6 descriptor bases
- 6 descriptor deltas
- 1 trip count
- 3 WGMMA groups per loop iteration
- `wait_group<2>` throttling
- final `wait_group<0>`

This deliberately isolates backend feasibility from production recognition and lowering changes.

## Proposed validation

Extend the existing counted-pipeline carrier test with a three-slot variant.

The generated PTX should contain:

- 1 `wgmma.fence.sync.aligned`
- 3 `wgmma.mma_async.sync.aligned.m64n64k16.f32.bf16.bf16`
- 3 `wgmma.commit_group.sync.aligned`
- 3 `wgmma.wait_group.sync.aligned 2`
- 1 final `wgmma.wait_group.sync.aligned 0`
- 6 descriptor recurrences

The `ptxas -arch=sm_90a --compile-only -v` result must report:

- 0 bytes stack frame
- 0 bytes spill stores
- 0 bytes spill loads

The probe should also verify that at least 96 registers remain live, matching the three 32-value accumulator slots.

## Scope

This issue is only for the backend feasibility probe.

It does not change:

- the production counted-loop recognizer
- the counted WGMMA carrier/op
- MIR lowering
- the currently supported two-slot `wait_group<1>` production path

A production implementation for three slots should be handled separately once this feasibility check succeeds.

uurl:
Closes #1010 

## Summary

Extend the existing counted WGMMA pipeline backend probe with a three-slot variant using `wait_group<2>`.

The test keeps the production-independent structure of the existing two-slot carrier and exercises a larger convergent inline-PTX lifetime containing:

- 96 tied `f32` accumulator values
- 6 descriptor bases
- 6 descriptor deltas
- 1 trip count
- 3 WGMMA groups per counted-loop iteration
- `wait_group<2>` throttling
- final `wait_group<0>`

This PR intentionally does not modify the production counted-loop recognizer or lowering path. Its purpose is to resolve the backend register-pressure question before adding three-slot production support.

## Results

On `sm_90a`, `ptxas` reports:

```text
two-slot baseline:
    0 bytes stack frame
    0 bytes spill stores
    0 bytes spill loads
    Used 90 registers

three-slot probe:
    0 bytes stack frame
    0 bytes spill stores
    0 bytes spill loads
    Used 122 registers
```

The three-slot carrier therefore remains fully register-resident despite keeping 96 tied accumulator values and 13 control values live across the counted PTX loop.

## Validation

```text
cargo test -p cuda-oxide-codegen \
  --test wgmma_counted_pipeline_carrier_ptx -- --nocapture

cargo fmt --all -- --check

git diff --check
```

Result:

```text
2 passed; 0 failed
```

The three-slot `ptxas` probe reports zero stack and zero spill traffic.

## Scope

This PR only extends:

```text
crates/cuda-oxide-codegen/tests/wgmma_counted_pipeline_carrier_ptx.rs
```

It does not change production recognition or lowering.

A follow-up PR can use this result to add production support for:

```text
2 slots + wait_group<1>
3 slots + wait_group<2>
```

### DIFF
--- a/crates/cuda-oxide-codegen/tests/wgmma_counted_pipeline_carrier_ptx.rs
+++ b/crates/cuda-oxide-codegen/tests/wgmma_counted_pipeline_carrier_ptx.rs
@@ -3,26 +3,36 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//! Proves that a two-slot BF16 WGMMA counted pipeline can keep sixty-four tied
-//! `f32` accumulator values spill-free while loop control and affine descriptor
+//! Proves that BF16 WGMMA counted pipelines can keep multiple independent
+//! accumulator slots spill-free while loop control and affine descriptor
 //! recurrences remain inside the same convergent inline-PTX lifetime.
 //!
 //! This is a backend feasibility probe. It intentionally bypasses the production
 //! counted-loop recognizer and combined carrier that a follow-up implementation
-//! may add. The test isolates the harder backend question first:
+//! may add. The file covers both the existing two-slot baseline and the next
+//! three-slot feasibility step:
 //!
 //! ```text
-//! 64 tied f32 accumulator values
-//! 4 descriptor bases
-//! 4 descriptor deltas
-//! 1 trip count
-//! one PTX loop
-//! wait_group<1> throttling
-//! final wait_group<0>
+//! baseline:
+//!   64 tied f32 accumulator values
+//!   4 descriptor bases
+//!   4 descriptor deltas
+//!   wait_group<1> throttling
+//!
+//! three-slot probe:
+//!   96 tied f32 accumulator values
+//!   6 descriptor bases
+//!   6 descriptor deltas
+//!   wait_group<2> throttling
+//!
+//! both:
+//!   1 trip count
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Reproduced locally with llc-21 + CUDA 13.x ptxas: 122 registers, 0 stack bytes, 0 spill bytes for the three-slot loop on sm_90a. That answers #1010's feasibility question.

nihalpasham (approved):

PR #1013, opened 2026-08-18
perf(mir-importer): promote packed struct array constants

### LINKED ISSUE #1012: Promote byte-faithful packed struct array constants to immutable device globals
## Problem

#925 added immutable-device-global promotion for scalar, tuple, and struct array constants, but intentionally excluded packed structs because LLVM's natural struct layout can diverge from rustc's recorded layout.

For example, a Rust layout equivalent to:

```rust
#[repr(C, packed)]
struct Packed {
    tag: u8,
    value: u32,
}
```

places `value` at byte offset 1 and has a stored size of 5 bytes. A natural LLVM struct would place the `u32` at offset 4, so promoting the Rust byte image through that representation would read the wrong bytes.

#941 subsequently added packed aggregate support by value and can select packed LLVM struct storage when that representation exactly reproduces rustc's layout.

This makes it possible to safely extend the immutable array-constant promotion path to packed structs when the selected LLVM representation is byte-faithful.

## Proposed behavior

Allow recursively promotable arrays containing packed structs to use the immutable-device-global path when:

* rustc provides a complete physical layout for the struct;
* the selected natural or packed LLVM representation reproduces every recorded field offset;
* the LLVM stored size exactly matches rustc's stored size;
* nested packed aggregates satisfy the same condition recursively; and
* the layout contains no overlapping or otherwise unrepresentable fields.

Layouts that cannot be represented exactly should continue to fail closed and must not enter the promotion path.

Pointer-to-array constants such as:

```rust
const TABLE: [Packed; N] = ...;
const TABLE_REF: &[Packed; N] = &TABLE;
```

should reuse the same immutable initializer rather than materializing a local table.

## Acceptance criteria

* `repr(packed)` struct arrays are promoted when their packed LLVM layout exactly matches rustc's byte layout.
* `repr(packed(N))` is accepted when its selected storage representation is byte-faithful.
* Nested packed structs are accepted only when every nested layout is representable exactly.
* Overlapping or otherwise non-representable layouts remain rejected.
* Bare array values and pointer-to-array constants deduplicate to the same immutable initializer.
* Generated LLVM uses packed aggregate storage rather than a natural-layout approximation.
* Existing non-packed promotion behavior remains unchanged.
* Runtime and code-shape regression coverage is added to `array_constants`.

## Context

Follow-up to #925, building on the packed aggregate lowering support from #941.

uurl:
Closes #1012

## Summary

Extend immutable-device-global promotion for array constants to packed structs when their selected LLVM storage representation is byte-faithful to rustc's recorded layout.

This follows #925, which deliberately excluded packed structs because LLVM's natural struct layout could place fields at different offsets, and builds on #941, which added packed aggregate lowering by value.

## What changed

- Extend the struct-array promotion gate to accept packed layouts when the selected natural or packed LLVM representation exactly reproduces:

  - rustc field offsets; and
  - rustc stored size.
- Apply the same check recursively to nested aggregate fields.
- Keep overlapping and otherwise non-representable layouts outside the promotion path.
- Preserve the semantic pointer type for promoted constants while using a byte-oriented validation view for byte-faithful packed storage.

  - `repr(C, packed)` struct arrays;
  - packed struct pointer-to-array constants;
  - packed layout byte images;
  - immutable-global reuse between value and reference forms; and
  - rejection of natural-layout storage for packed arrays.
- Update the supported-features documentation.

## Why

A packed Rust struct can have a physical layout that LLVM's natural struct layout cannot represent.

For example:

```rust
#[repr(C, packed)]
struct Packed {
    tag: u8,
    value: u32,
}
```

has offsets `[0, 1]` and size `5`.

Using a natural LLVM `{ i8, i32 }` layout would place the `i32` at offset `4`, which would make promotion of the original Rust byte image incorrect.

After #941, the lowering pipeline can represent such aggregates with packed LLVM struct storage. Promotion is therefore safe when that representation is proven to reproduce the rustc layout exactly.

## Safety / correctness

Promotion remains fail-closed.

A packed aggregate is admitted only when the importer can prove that one of the LLVM storage representations available to lowering reproduces rustc's physical layout exactly. This includes both field offsets and stored size.

Nested aggregates are checked recursively, and overlapping or otherwise unrepresentable layouts are not promoted.

The pointer-to-array path keeps the semantic aggregate type, so downstream typed GEP/load lowering continues to use the packed representation selected by the aggregate lowerer.

## Generated code

For a two-element packed table equivalent to:

```rust
[
    Packed { tag: 0xa5, value: 0x11223344 },
    Packed { tag: 0x5a, value: 0x99aabbcc },
]
```

the immutable initializer is the exact 10-byte Rust storage image:

```text
a5 44 33 22 11 5a cc bb aa 99
```

The value path copies from the read-only device global, while the pointer-to-array path references the same deduplicated initializer.

## Validation

Focused tests:

```text
promotable_array_element_tests:       3 passed
pointer_array_constant_type_tests:    2 passed
```

End-to-end:

```text
array_constants: PASS

array_constants: PASS

verify-code-shape.sh
```

Affected crate suites:

```text
mir-importer --all-targets: PASS
mir-lower --all-targets:    PASS
```

Lint / formatting:

```text
cargo fmt --all -- --check: PASS
cargo clippy -p mir-importer --all-targets -- -D warnings: PASS
cargo clippy -p mir-lower --all-targets -- -D warnings: PASS
git diff --check: PASS
```

Full repository smoketest:

```text
Pass: 215 / 215
Fail:   0 / 215
```

## Related

* Follow-up to #925
* Builds on #941


nihalpasham:
- Verified on the 5090: array_constants passes in both pipelines and verify-code-shape.sh is green.
- One commit layered on: promotion soundness rested on two layout walks in different crates (the importer's admission gate and build_struct_slot_map's packed selection) agreeing forever, with nothing coupling them. Exposed struct_value_lowering_is_byte_faithful from mir-lower and added an importer test asserting agreement over the promotion corpus: packed, gapped packed(2), natural, and overlapping.
- For the record: tightening the importer walk to reject gaps looked like the fix but is wrong; mir-lower inserts explicit [N x i8] padding slots, so gapped packed(2) layouts lower faithfully. The coupling test is the right guard.

nihalpasham (approved):

PR #1017, opened 2026-08-18
docs(examples): three READMEs state a minimum GPU below their own runtime gate

vyncint:
## Summary

Fourteen examples gate `main` on compute capability. Three of them also state a **Minimum GPU** in their README, and **all three state one lower than the gate they ship with** — so the README says the example runs on hardware where it prints `skipping:` and returns.

| Example | README said | Gate in `main()` | The gate's own message |
|:--|:--|:--|:--|
| `future_apis` | Volta (sm_70) | `major < 9` | "fence.proxy.async … requires sm_90+" |
| `coop_groups_demo` | Volta (sm_70+) | `major < 9` | "thread block clusters require sm_90+" |
| `barrier` | Volta (sm_70) | `major < 8` | "mbarrier requires sm_80+" |

`future_apis` is the one with history: `8615368f` corrected that gate to `major < 9` during the **#666** review — the requirement is `fence.proxy.async`, emitted by `ManagedBarrier::init_by` for every barrier kind, which ptxas rejects below sm_90, and the `TmaBarrier` typestate is inert `PhantomData` that detects nothing. The gate moved; the README did not.

In all three cases the sm_70 figure is not invented — it is the floor of *one* instruction family the example touches (mbarrier, or `match.any.sync`) rather than the floor for running the example. Each entry now states the real floor, cites the gate that enforces it, and keeps the sm_70 fact as the aside it always was, so the next reader does not have to reconcile them.

Every gate here is in `main()` and returns immediately, so this is the whole example's floor, not one test's.

## Testing

- [x] Swept all 215 example READMEs: these three were the **only** ones whose stated Minimum GPU did not match a gate in their own `main`, and after this change none do
- [x] `check-error-example-status` and `check-example-smoketest-contract` pass
- [ ] `cargo oxide run <example>` — not applicable; verifying the floors themselves would need Hopper and Ampere parts, which is why this takes each example's own gate as the authority rather than re-deriving it

### DIFF
--- a/crates/rustc-codegen-cuda/examples/barrier/README.md
+++ b/crates/rustc-codegen-cuda/examples/barrier/README.md
@@ -77,8 +77,9 @@ cargo oxide run barrier
 
 - **Minimum GPU**: Ampere (sm_80) or newer. `main` skips cleanly below that
   (`if major < 8`), matching the `mbarrier requires sm_80+` message it
-  prints. Some mbarrier instructions are sm_70+, but sm_80 is the floor for
-  running this example.
+  prints. If you have seen `cuda::barrier` work on sm_70, that is libcu++'s
+  software fallback; the PTX `mbarrier` instructions this example emits
+  need sm_80+.
 - **CUDA Driver**: 11.0+
 
 ## Mbarrier Functions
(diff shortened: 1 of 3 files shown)


nihalpasham:
- Verified all three corrected floors against the gate each example ships (barrier sm_80, coop_groups_demo and future_apis sm_90) and against the instructions in the checked-in PTX.
- One commit layered on: the new sm_70 mbarrier asides claimed the instructions exist below their PTX floor (libcu++'s sm_70 support is a software fallback; the emitted mbarrier needs sm_80+), and two passages unreachable behind the gates (the pre-Hopper combined-mode text and the Pre-Volta troubleshooting row) are gone.

PR #1021, opened 2026-08-18
feat(wgmma): support three-slot counted pipelines

### LINKED ISSUE #1020: Support three-slot WGMMA counted pipelines with wait_group<2>
## Motivation

The production counted WGMMA pipeline currently recognizes and lowers a fixed two-slot form using `wait_group<1>`.

A separate backend feasibility probe has shown that a three-slot counted pipeline remains spill-free on `sm_90a`, keeping 96 tied `f32` accumulator values live together with the corresponding descriptor recurrences and loop control.

The next step is to extend the production counted-loop path to support:

```text
2 slots + wait_group<1>
3 slots + wait_group<2>
```

while keeping the recognizer deliberately narrow and fail-closed.

## Proposed change

Generalize the existing counted WGMMA pipeline carrier, recognizer, and lowering so that the slot count is derived from the partial-wait depth:

```text
slot_count = max_pending_groups + 1
```

For this change, only the following production depths should be accepted:

```text
max_pending_groups = 1  ->  2 slots
max_pending_groups = 2  ->  3 slots
```

The counted-loop carrier should therefore support:

```text
2 slots:
  64 f32 accumulator values
  4 descriptor bases
  4 descriptor steps
  1 trip count

3 slots:
  96 f32 accumulator values
  6 descriptor bases
  6 descriptor steps
  1 trip count
```

The recognizer should require:

* one canonical accumulator per slot
* pairwise-distinct accumulator pointers
* one `MMA -> commit_group -> wait_group<N>` stage per slot
* the same `N` for every partial wait
* `N` in `{1, 2}`
* two descriptor recurrences per slot
* a final `wait_group<0>` in the loop exit

Unsupported or ambiguous shapes must remain unfused.

## Validation

Add positive coverage for the three-slot production path and negative coverage for malformed pipelines, including:

* mixed partial-wait depths
* `wait_group<2>` with only two slots
* reuse of the same accumulator across multiple slots
* unsupported counted-pipeline depths

The end-to-end codegen test should compile the three-slot production path with:

```text
ptxas -arch=sm_90a --compile-only -v
```

and require:

* 0 bytes stack frame
* 0 bytes spill stores
* 0 bytes spill loads
* at least 96 registers used

## Scope

This issue only extends the counted-loop production path from two to three accumulator slots.

It does not generalize counted pipelines to all `wait_group<N>` depths supported by the straight-line WGMMA pipeline.

uurl:
Closes #1020 

## Summary

Extend the production BF16 WGMMA counted-loop pipeline from the existing two-slot `wait_group<1>` form to also support a three-slot `wait_group<2>` form.

The counted carrier, recognizer, and lowering now derive the accumulator-slot count from the partial-wait depth while deliberately restricting the production contract to:

```text
2 slots + wait_group<1>
3 slots + wait_group<2>
```

The straight-line WGMMA pipeline continues to support its broader `wait_group<N>` range independently.

## Implementation

The counted-loop carrier now models a variable number of accumulator slots and carries the corresponding partial-wait depth.

The production layout is derived from:

```text
slot_count        = max_pending_groups + 1
accumulator_count = slot_count * 32
descriptor_count  = slot_count * 2
control_count     = slot_count * 4 + 1
```

This gives:

```text
wait_group<1>:
  2 slots
  64 accumulator values
  9 control values

wait_group<2>:
  3 slots
  96 accumulator values
  13 control values
```

The counted inline-PTX template now emits one `MMA -> commit_group -> wait_group<N>` stage per slot, advances all descriptor recurrences inside the same counted loop, and performs a final `wait_group<0>` before any accumulator result becomes visible to LLVM.

## Recognition

The counted-loop recognizer remains fail-closed.

It now requires:

* one canonical accumulator pointer per slot
* pairwise-distinct accumulator pointers
* homogeneous partial-wait depths
* exactly one `MMA -> commit_group -> wait_group<N>` stage per slot
* `N` equal to either 1 or 2
* two affine descriptor recurrences per slot
* one final `wait_group<0>`

Malformed counted pipelines are rejected rather than partially fused.

Negative coverage includes:

* mixed `wait_group<1>` / `wait_group<2>` stages
* `wait_group<2>` with only two slots
* reused accumulator slots

## Validation

```text
cargo test -p dialect-nvvm
cargo test -p mir-lower

cargo test -p cuda-oxide-codegen \
  --test wgmma_pipelined_counted_loop_ptx -- --nocapture

cargo fmt --all -- --check
git diff --check
```

Results:

```text
dialect-nvvm:
  50 passed

mir-lower unit tests:
  246 passed

mir-lower lowering tests:
  111 passed

counted-pipeline codegen:
  2 passed
```

`ptxas` results on `sm_90a`:

```text
two-slot production path:
  0 bytes stack frame
  0 bytes spill stores
  0 bytes spill loads
  Used 90 registers

three-slot production path:
  0 bytes stack frame
  0 bytes spill stores
  0 bytes spill loads
  Used 122 registers
```

The three-slot production path therefore keeps all 96 tied accumulator values register-resident without introducing stack or spill traffic.

## Scope

This PR modifies the counted WGMMA production path only.

It does not extend counted pipelines beyond `wait_group<2>` and does not change the existing broader straight-line WGMMA pipeline support.

### DIFF
--- a/crates/cuda-oxide-codegen/tests/wgmma_pipelined_counted_loop_ptx.rs
+++ b/crates/cuda-oxide-codegen/tests/wgmma_pipelined_counted_loop_ptx.rs
@@ -3,15 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//! End-to-end probe for the production two-slot BF16 WGMMA counted pipeline.
+//! End-to-end probes for production BF16 WGMMA counted pipelines.
 //!
-//! The test starts from the canonical pointer-form MIR shape recognized by
-//! `mir-lower`: two distinct `[[f32; 8]; 4]` accumulator slots, four affine
-//! descriptor recurrences, two `MMA -> commit_group -> wait_group<1>` stages in
-//! the unique latch, and a final `wait_group<0>` in the exit. The fusion pass
-//! must replace that CFG with one value-form carrier, which then lowers to one
-//! convergent inline-PTX counted pipeline and survives `ptxas -arch=sm_90a`
-//! without stack or spill traffic.
+//! The tests start from the canonical pointer-form MIR shape recognized by
+//! `mir-lower` and cover both supported production depths: two distinct
+//! `[[f32; 8]; 4]` accumulator slots with `wait_group<1>` and three slots with
+//! `wait_group<2>`. Each slot owns one affine descriptor pair. The fusion pass
+//! must replace the counted CFG with one value-form carrier, which then lowers
+//! to one convergent inline-PTX counted pipeline and survives
+//! `ptxas -arch=sm_90a` without stack or spill traffic.
 
 #![cfg(unix)]
 
@@ -47,11 +47,8 @@ use std::num::NonZeroUsize;
 
 const ACCUMULATOR_LEN: u32 = 32;
 const RESULT_COUNT: u32 = 64;
+const THREE_SLOT_RESULT_COUNT: u32 = 96;
 const TRIP_COUNT: u64 = 4;
-const DESC_A0_STEP: u64 = 16;
-const DESC_B0_STEP: u64 = 32;
-const DESC_A1_STEP: u64 = 48;
-const DESC_B1_STEP: u64 = 64;
 
 fn append_unsigned_constant(
     ctx: &mut Context,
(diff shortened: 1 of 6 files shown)


nihalpasham:
- Reproduced both ptxas probes locally: 90 registers for two-slot, 122 for three-slot, 0 spill bytes each on sm_90a. The wgmma smoketest category compiles on this non-sm_90a host.
- On merge this closes #1020 and #1010. #1011's probe is the value-carrier form, this one is the pointer form, so both tests stay.

nihalpasham (approved):

PR #1029, opened 2026-08-18
docs(book): the quick reference's DSMEM snippet cannot compile and would double-map

vyncint:
## Summary

`api-quick-reference.md`'s Cluster Programming block reads:

```rust
let remote_ptr = cluster::map_shared_rank(local_ptr, target_rank);
let val = cluster::dsmem_read_u32(remote_ptr);
```

Three things are wrong, and this block is a self-contained copy-paste reference with its own `use` line — not an excerpt from a larger `unsafe` scope.

**1. Arity.** The real signature takes two arguments:

```rust
pub unsafe fn dsmem_read_u32(local_ptr: *const u32, target_rank: u32) -> u32
```

**2. Unsafety.** Both functions are `unsafe fn` (`cuda-device/src/generated/cluster_memory.rs:20` and `:33`) and neither call is in an `unsafe` block.

**3. Composition.** This is the part a corrected arity alone would not fix. `dsmem_read_u32` takes the *local* pointer and does the rank mapping itself, so feeding it a pointer already returned by `map_shared_rank` maps twice. The two functions are alternatives, not a pipeline.

## The tree already shows the right form

The fix follows these rather than inventing one:

```rust
// examples/cluster/src/main.rs:162
unsafe { cluster::dsmem_read_u32(addr_of!(SHMEM) as *const u32, neighbor_rank) }

// book/advanced/cluster-programming.md
let remote_ptr = unsafe { cluster::map_shared_rank(DATA.as_ptr().add(tid), neighbor) };
let neighbor_val = unsafe { *remote_ptr };
```

The snippet now shows both paths separately — read directly with `dsmem_read_u32`, or map and dereference yourself — and says why a mapped pointer must not be handed to the reader.

Not caught by `check-book-api-names.sh` by design: that guard checks whether a name *resolves*, and both of these do. Arity, unsafety and composition are all outside its scope.

## Testing

Local, no GPU.

- [x] Swept the whole book for both shapes — module-qualified calls whose argument count differs from the declaration, and calls to `unsafe`-only functions with no `unsafe` in scope. Every other hit was a multi-line call my regex truncated or a fragment inside a wider `unsafe` block; **this page's block was the only real one**
- [x] `just book` (`sphinx-build -W --keep-going`) succeeds; `check-book-api-names.sh` passes

### DIFF
--- a/cuda-oxide-book/appendix/api-quick-reference.md
+++ b/cuda-oxide-book/appendix/api-quick-reference.md
@@ -457,15 +457,16 @@ let rank = cluster::block_rank();        // This block's rank in the cluster
 let size = cluster::cluster_size();      // Number of blocks in cluster
 cluster::cluster_sync();                 // Barrier across all cluster blocks
 
-// Distributed Shared Memory. Both are `unsafe fn`, and both take the *local*
-// pointer plus the target rank -- they do the rank mapping themselves, so a
-// pointer already returned by `map_shared_rank` must not be passed to
-// `dsmem_read_u32`.
+// Distributed Shared Memory. Both are `unsafe fn` and both take the *local*
+// pointer plus the target rank; they do the rank mapping themselves, so
+// never pass a pointer already returned by `map_shared_rank` back in.
+// Read a remote block's value with `dsmem_read_u32`:
 let val = unsafe { cluster::dsmem_read_u32(local_ptr, target_rank) };
 
-// `map_shared_rank` is for reading through the mapped pointer yourself.
+// `map_shared_rank` returns a cluster-address carrier you pass to
+// intrinsics, not a pointer you dereference: a plain deref compiles to a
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified the corrected signatures and the double-mapping claim against cuda-device and the lowering: `dsmem_read_u32` fuses `mapa` + `ld.shared::cluster` itself, so piping `map_shared_rank` into it maps twice.
- **One commit layered on:** the snippet's closing deref of the `map_shared_rank` pointer is the pattern the cluster example's troubleshooting table maps to `CUDA_ERROR_ILLEGAL_ADDRESS`. The reference now routes remote reads through `dsmem_read_u32` and calls the mapped pointer a carrier for intrinsics, not something to dereference. Filed #1034 for the chapter that still teaches the deref.

PR #1032, opened 2026-08-18
feat(intrinsics): add f32 redux.sync family

almightychang:
## Summary

- Add all eight `redux.sync` f32 min/max variants: min/max × optional `.abs` × optional `.NaN`.
- Carry the family end to end through the intrinsic catalog, ABI, device compatibility API, dialect/importer/lowering pipeline, generated target metadata, probes, and reference output.
- Restrict the family to Blackwell family 10x targets and add an `sm_100a` compile-only example lane that checks the emitted PTX.

## Target matrix rationale

Pinned LLVM defines `hasReduxSyncF32` as accelerator family `{100}` at PTX 8.6 or family `{100}` at PTX 8.8. Direct `llc` verification selects these intrinsics on `sm_100a`, `sm_100f`, `sm_103a`, and `sm_103f`, while `sm_110a` and `sm_120a` fail with `Cannot select`.

The catalog therefore records the PTX 8.6 accelerator instruction floor with the `sm_100a|sm_100f|sm_103a|sm_103f` target union. Per-target effective PTX floors are handled by `target.rs`, which raises family and later 10x targets as required by LLVM.

## Verification

- `cargo run -p cuda-intrinsics-gen -- check` — generated outputs are current.
- `cargo run -p cuda-intrinsics-gen -- probe --all` — all 1010 catalog routes pass.
- Candidate probes — all 8 f32 redux variants pass with pinned `llc` at `sm_100a +ptx86`.
- Full relevant generator, dialect, importer, lowering, intrinsic, and device test suites are green; `cargo test -p cuda-intrinsics-gen` reports 291 passing tests.
- The `sm100-compile` lane asserts `.version 8.6`, `.target sm_100a`, and all 8 total and unique `redux.sync` f32 forms in each example kernel.
- Root and example formatting checks pass.

## Deferred follow-ups

- The libNVVM route remains ungated, a pre-existing family-wide property shared with integer redux.
- Update the book documentation separately.
- GPU execution validation is pending B200 access. The example is compile-only in CI, matching the existing `blackwell-compile` posture.


nihalpasham:
- Rebased onto main (f2a6220e) and force-pushed; the only conflict was the import list in `dialect-nvvm/tests/ops_test.rs`, resolved by union with main's two new wgmma ops. Your four commits are intact on top.
- Verified after the rebase: `check` reports generated outputs current; `check-abi-history` shows 1002 base entries preserved with `i1003`-`i1010` appended; 291 generator + 51 dialect-nvvm tests pass; all 8 forms probe to the expected `redux.sync.{min,max}[.abs][.NaN].f32` for sm_100a; `smoketest.sh redux_f32` passes compile-only on this sm_120 host, which matches the blackwell-compile precedent for landing.
- GPU execution of the NaN-policy check stays outstanding until we get time on a B200. Book docs deferred per the description; tracking next to #808 so the family does not stay undocumented.

PR #1033, opened 2026-08-19
fix(nix): set CUDA toolkit env and cache config

kennethdsheridan:
## Summary\n- Set both CUDA_HOME and CUDA_TOOLKIT_PATH to the Nix-provided CUDA toolkit in this flake so cuda-bindings can find cuda.h without relying on /usr/local/cuda.\n- Add flake-level nixConfig to keep binary substitution on parrisisland and cache.nixos.org and avoid broken ac-sheridan cache entries.\n\n## Validation\n- nix flake show (successful with updated GitHub token and nixConfig)\n- nvcc/build flow now follows configured Nix cache settings

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -127,6 +127,7 @@
           ];
 
           CUDA_HOME = cudaSymlinked;
+          CUDA_TOOLKIT_PATH = cudaSymlinked;
           LIBCLANG_PATH = "${llvmPkgs.libclang.lib}/lib";
         };
 
@@ -218,6 +219,7 @@
 
           shellHook = ''
             export CUDA_HOME="${cudaSymlinked}"
+            export CUDA_TOOLKIT_PATH="${cudaSymlinked}"
             export LIBCLANG_PATH="${llvmPkgs.libclang.lib}/lib"
 
             # GPU driver setup borrowed from https://github.com/NVlabs/cutile-rs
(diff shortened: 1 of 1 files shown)


nihalpasham:
This is two unrelated changes in one diff. One we want, one we can't take in any form. Cleanest is a fresh PR with just the keeper.

- **Keep: the two `CUDA_TOOLKIT_PATH` exports.** They match the repo's documented discovery order (`CUDA_TOOLKIT_PATH` first, `CUDA_HOME` second) and fix the case where a stray cargo-config value shadows the Nix toolkit. Good change.

- **Remove: the `nixConfig` cache block.** Not fix, remove:
  - where it sits (inside `outputs`), Nix ignores it completely. It does nothing today. Whatever cache behavior you observed came from your local `nix.conf`, not this diff.
  - and if it were moved to the flake top level where it *would* work, it becomes something we can't ship:
    - every `nix develop` user would be prompted to trust a personal signing key and a plain-HTTP LAN address (`10.22.0.2:8097`) that's unreachable outside your network
    - it uses `substituters` / `trusted-public-keys`, which *replace* a user's existing settings. The `extra-*` forms exist so you don't clobber caches people already rely on
    - a project flake endorsing a personal, unvetted cache is a supply-chain risk regardless of intent
  - personal cache settings belong in your own `~/.config/nix/nix.conf` or `--option` flags, where they affect only you.

- If the project ever wants a shared cache, that's its own decision: project-owned, HTTPS, project-controlled key, `extra-*` forms. Separate discussion.

- **Practical ask:** open a new PR with just the env-var half (and a title to match, since "prioritize parrisisland cache" won't describe it anymore).

nihalpasham:
Closing. Please see above comment.

kennethdsheridan:
thanks, and this is an artifact of me getting my agent setup with this repo, sorry. Won't happen again :)

Issue #604, opened 2026-07-31
Canonical variant lists assign ABI IDs by position, which conflicts with the append-only ledger


honeyspoon:
## Summary

A family's canonical variant list assigns ABI IDs **sequentially by list position**, but `intrinsics/abi-v1.toml` is **globally append-only**. Once a later family consumes the ID that follows a family's block, that family can no longer be extended — the two invariants conflict and there is no way to satisfy both.

## Reproduction

Adding one verified-lowering variant to `scalar_math` (f16 `ex2.approx`):

1. `ScalarMathFormat` gains `F16` (it is `{F32, F64}` today), plus the six `match` arms the compiler identifies.
2. `canonical_scalar_math_variants()` gains `(F16, Ex2, Approx, Preserve)`.
3. `intrinsics/overlay/scalar_math.toml` gains the matching `[[scalar_math.variant]]`.

Then:

```
error: ex2_approx_f16 must reserve ABI ID i0822
```

`i0822` follows `i0821`, the last `scalar_math` variant. But it is already taken:

```toml
abi_id = "i0822"
catalog_id = "cvt_rn_satfinite_e4m3x2_f16x2"
operation_key = "packed.convert.f16x2.e4m3x2.nearest_even.satfinite"
```

`packed_conversion` claimed `i0822`–`i0829` afterwards, closing `scalar_math`'s block. Using the next globally free ID instead (`i0830`) fails the other check, since the expansion derives the expected ID from list position.

So the addition is blocked in both directions, and not by anything about the intrinsic itself: `llvm.nvvm.ex2.approx.f16` lowers cleanly to `ex2.approx.f16` under the pinned `llc` (LLVM 22.1.2-rust), verified by probe.

## Why it will keep happening

This is structural, not specific to `scalar_math`. Every family with a `canonical N variants` assertion has the same coupling, so any family whose ID block has been closed off by a later family is now frozen. The more families are added, the more get frozen.

It currently blocks at least these verified-lowering additions (each probed with `llc` on `nightly-2026-04-03`, instruction confirmed in the emitted PTX):

| intrinsic | emits |
|---|---|
| `llvm.nvvm.ex2.approx.f16` | `ex2.approx.f16` |
| `llvm.nvvm.bmsk.clamp` / `.wrap` | `bmsk.clamp.b32` / `bmsk.wrap.b32` |
| `llvm.nvvm.fns` | `fns.b32` |
[...]

PR #1038, opened 2026-08-19
fix(intrinsics): decouple ABI IDs from canonical variant order

uurl:
Partially addresses #604: this unblocks the compact-admission families (scalar_math, packed_conversion, tcgen05, TMA reduction). The register_mma float/FP8/sparse and cluster-sreg families still derive ABI IDs positionally, so #604 stays open for those.

## Summary

Canonical compact-family variant lists currently serve two independent roles: they define the closed semantic set and, in several families, implicitly assign ABI IDs from variant position.

This change separates those responsibilities.

Canonical lists continue to enforce semantic membership, completeness, and ordering, while compact admission entries carry the explicit ABI ID. The global append-only ABI ledger remains responsible for validating historical ABI identity and uniqueness.

## Changes

Remove positional ABI derivation from:

- TMA reduction variants
- tcgen05 copy variants
- tcgen05 load variants
- tcgen05 store variants
- tcgen05 offset load/store variants
- tcgen05 MMA variants
- scalar arithmetic
- scalar math
- extended min/max

Semantic recipe construction remains closed over the canonical variant sets, but recipes no longer derive or own ABI IDs from their position.

Admission expansion now propagates each variant's explicit `abi_id`.

Policy validation continues to validate the semantic identity independently of ABI numbering, while the global ABI ledger remains the authority for stable ABI identity.

## Behavior

Before:

```text
canonical semantic position
    -> family ABI base + position
    -> expected ABI ID
    -> admission ABI ID must match
```

After:

```text
canonical variant list
    -> semantic membership / completeness / ordering

admission variant.abi_id
    -> explicit ABI identity

intrinsics/abi-v1.toml
    -> global uniqueness / history / append-only validation
```

This allows a future family variant to use the next globally available ABI ID even when that ID is non-contiguous with previous members of the family.

## Tests

Regression coverage verifies that compact admissions still reject semantic reordering while accepting a valid non-contiguous ABI ID at the expansion layer.

Existing ABI protection remains enforced by the global ledger.

Validation performed:

```text
cargo fmt --all -- --check
cargo test -p cuda-intrinsics-gen
cargo run -p cuda-intrinsics-gen -- check-abi-history --base-ref nvlabs/main
git diff --check
```

Results:

* 290 tests passed
* generated intrinsic outputs remain current
* ABI history preserves all 1002 existing entries
* no ABI ledger entries were added, removed, or changed

### DIFF
--- a/crates/cuda-intrinsics-gen/src/resolve.rs
+++ b/crates/cuda-intrinsics-gen/src/resolve.rs
@@ -4449,7 +4449,6 @@ const TMA_OPERATIONS: [TmaOperation; 47] = [
     TmaOperation::PrefetchTileGather4TwoDimensionalCacheHint,
 ];
 
-const TMA_REDUCTION_FIRST_ABI_ID: u32 = 923;
 const TMA_REDUCTION_OPERATIONS: [TmaReductionOperation; 8] = [
     TmaReductionOperation::Add,
     TmaReductionOperation::And,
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified end to end in a worktree: 290/290 generator tests, `check` green with zero generated-file diff, `check-abi-history --base-ref origin/main` green with all 1002 ledger entries preserved. No ID value changes anywhere.
- Checked the surviving guard layer four ways for corruption (swapped `i0782`/`i0783`, bogus `i9999`, in-family duplicate, stealing packed_conversion's `i0822`): every one is rejected loudly by the same `check` command CI runs. The positional derivation was redundant armor over the ledger, not load-bearing.
- Edited the description to "Partially addresses #604": `register_mma` `float`/`FP8`/`sparse` and the cluster sregs still derive IDs positionally (a sixth Ampere float MMA variant would freeze at `i0525`), so #604 stays open for those families.

PR #1040, opened 2026-08-19
feat(mir): preserve Rust pointer and reference kind

### LINKED ISSUE #413: Preserve Rust reference kind and exclusivity through MIR import and lowering

uurl:
## Summary

Preserve Rust pointer/reference kind through MIR import instead of collapsing
references and raw pointers that share the same pointee, mutability, and address
space.

Fixes #413.

This introduces `MirPointerKind` with the following categories:

- `SharedRef` for `&T`
- `UniqueRef` for `&mut T`
- `RawConst` for `*const T`
- `RawMut` for `*mut T`
- `Erased` for compiler-generated/internal addresses

The kind is carried by both `MirPtrType` and `MirSliceType`.

### dialect-mir

- Added `MirPointerKind`.
- Added pointer kind to `MirPtrType`.
- Added pointer kind to `MirSliceType`.
- Added constructors that explicitly preserve source pointer kind.
- Existing synthetic/internal pointer constructors default to `Erased`.
- Added verification that concrete pointer kinds agree with the pointer
  mutability bit.
- Documented the distinction between mutability and Rust reference uniqueness.

### MIR importer

Rust types now map as follows:

| Rust type | MIR pointer kind |
|-----------|------------------|
| `&T` | `SharedRef` |
| `&mut T` | `UniqueRef` |
| `*const T` | `RawConst` |
| `*mut T` | `RawMut` |

Pointer kind is preserved through relevant importer paths, including:

- `Rvalue::Ref`
- `Rvalue::AddressOf`
- projected addresses
- slices and subslices
- pointer-like local stores
- static pointer materialization
- CUDA address-space alignment

Compiler-generated storage and temporary addresses remain `Erased`.

Pointer normalization is intentionally conservative. It may restore a concrete
Rust kind at an explicit Rust-typed semantic boundary, but generic coercion does
not silently convert one concrete Rust pointer category into another. For
example, it cannot synthesize:

- `SharedRef -> UniqueRef`
- `RawConst -> RawMut`
- raw pointer -> reference

### MIR lowering

`MirPointerKind` is semantic-only and is intentionally erased during MIR-to-LLVM
type conversion.

When a cast changes only MIR pointer kind and both sides lower to the same LLVM
type, lowering forwards the value directly instead of emitting an unnecessary
LLVM cast or memory round-trip.

This PR deliberately does not emit LLVM `noalias`, `readonly`,
`dereferenceable`, or related metadata.

`Retag` remains a codegen no-op; dynamic Stacked Borrows / Tree Borrows state is
outside the scope of this change.

## Tests

Added coverage for:

- distinguishing references from raw pointers
- preserving pointer kind in slices
- pointer-kind/mutability consistency
- preserving pointer kind during address-space alignment
- restoring fat-pointer kind at typed boundaries
- rejecting concrete pointer-kind changes during generic normalization
- erasing pointer-kind-only coercions without emitting LLVM instructions

Validation performed:

```text
cargo fmt --all -- --check
```

All checks pass.

### DIFF
--- a/crates/dialect-mir/README.md
+++ b/crates/dialect-mir/README.md
@@ -10,17 +10,17 @@ rustc MIR ──► mir-importer ──► dialect-mir ──► mir-lower ─�
 
 The dialect defines nine types that preserve Rust-level semantics:
 
-| Type                  | Description                                        | Example                               |
-|-----------------------|----------------------------------------------------|---------------------------------------|
-| `MirTupleType`        | Heterogeneous tuples                               | `mir.tuple<i32, f32, i64>`            |
-| `MirPtrType`          | Pointers with address space and mutability         | `mir.ptr<f32, mutable, addrspace: 3>` |
-| `MirSliceType`        | Fat pointers (`&[T]` = ptr + len)                  | `mir.slice<f32, addrspace: 1>`        |
-| `MirDisjointSliceType`| `DisjointSlice<T>` -- per-thread unique access     | `mir.disjoint_slice<f32, ...>`        |
-| `MirStructType`       | Named structs with layout metadata                 | `mir.struct<"Point", [f32, f32]>`     |
-| `MirUnionType`        | Rust unions -- each field a view of the same bytes | `mir.union<"Repr", [a, b], [i32, f32], 4, 4>` |
-| `MirEnumType`         | Rust enums with their exact rustc layout           | `mir.enum<"Ordering", i8, ...>`       |
-| `MirArrayType`        | Fixed-size arrays                                  | `mir.array<f32, 256>`                 |
-| `MirFP16Type`         | IEEE 754 binary16 -- Rust's `f16`                  | `mir.fp16`                            |
+| Type                  | Description                                                  | Example                                                               |
+|-----------------------|--------------------------------------------------------------|-----------------------------------------------------------------------|
+| `MirTupleType`        | Heterogeneous tuples                                         | `mir.tuple<i32, f32, i64>`                                            |
+| `MirPtrType`          | Thin pointers with address space, mutability, and source kind | `mir.ptr<f32, mutable: true, addrspace: 3, kind: UniqueRef>`           |
+| `MirSliceType`        | Fat pointers with retained source kind (ptr + len)           | `mir.slice<f32, kind: SharedRef>`                                     |
+| `MirDisjointSliceType`| `DisjointSlice<T>` -- per-thread unique access               | `mir.disjoint_slice<f32, ...>`                                        |
+| `MirStructType`       | Named structs with layout metadata                           | `mir.struct<"Point", [f32, f32]>`                                   |
+| `MirUnionType`        | Rust unions -- each field a view of the same bytes           | `mir.union<"Repr", [a, b], [i32, f32], 4, 4>`                       |
+| `MirEnumType`         | Rust enums with their exact rustc layout                     | `mir.enum<"Ordering", i8, ...>`                                     |
+| `MirArrayType`        | Fixed-size arrays                                            | `mir.array<f32, 256>`                                                 |
+| `MirFP16Type`         | IEEE 754 binary16 -- Rust's `f16`                            | `mir.fp16`                                                            |
 
 `MirEnumType` records the enum's layout the way rustc computed it: Direct,
 Niche, Single, or Empty; the physical integer/pointer carrier and absolute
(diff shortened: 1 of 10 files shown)


nihalpasham:
The direction is generally right, but this version breaks the compiler, so it can't merge. This needs a rethink deep enough that a fresh PR is the better route.

- **Where it fails:** the branch can't build the most basic example.

  ```text
  $ cargo oxide build vecadd
  Verification failed for 'vecadd':
  MirStoreOp value type must match pointer element type
    mir.ptr<fp32, ..., kind:UniqueRef>  vs  mir.ptr<fp32, ..., kind:RawMut>
  ```

- **Why:** the new rule says a pointer can never change from one concrete kind to another. But Rust does that constantly. It's called a reborrow:

  ```rust
  let p: *mut f32 = ...;
  let r: &mut f32 = unsafe { &mut *p };   // RawMut -> UniqueRef, and it's legal Rust
  ```

  - At that `&mut *` the kind must change (rustc's own MIR type says so), but `pointer_kind_retype_allowed` refuses, no cast gets inserted, and the store verifier rejects the module.
  - This pattern is everywhere: `DisjointSlice::get_mut`, the sysroot's `slice::index`, `split_at_mut`. Nearly every kernel hits one of them.

- **Why the tests didn't catch it:** all 1400+ unit tests pass, but none of them compiles an example end to end, and CI never ran on the fork branch. Any respin needs `vecadd` plus one SharedArray example in its validation, not just unit tests.

- **Two more problems to fix in the redo:**

  - `maybe_ptr_coerce` used to bridge any thin-pointer mismatch with a cast. The rewrite only casts when the pointees already match, so cases that worked on main (like `shared_alloc`'s pointer feeding a differently-typed local) now fail downstream.
  - the kind can be laundered. Each hop below is legal on its own:

    ```text
    SharedRef -> Erased -> UniqueRef
    ```

    Two normalizations later a shared ref claims to be unique. A future `noalias` pass keyed on `UniqueRef` would believe it.

**Suggested path:** let's agree on the transition table in #413 first, then a new PR. The core question is where kind changes are allowed: explicit `Rvalue::Ref` / `Rvalue::AddressOf` boundaries should trust rustc's declared type (reborrow, addr-of), while implicit normalizations stay locked down, and `Erased` needs a rule that can't launder. That policy deserves thought before code.

**Also:** drop "Fixes #413" from the description. #413 is a tracking issue; this covers the representation half only and stays open either way.

nihalpasham:
Closing. Please see above.

PR #1041, opened 2026-08-19
Preserve map_shared_rank results in cluster-shared addrspace(7)

### LINKED ISSUE #1034: Cluster chapter, map_shared_rank stub doc, and the mapa lowering disagree on whether the mapped pointer is dereferenceable
Three parts of the repo tell three different stories about what you can do with the pointer `cluster::map_shared_rank` returns. Only one of them matches what the GPU actually does.

**1. The book chapter says deref works.** `cuda-oxide-book/advanced/cluster-programming.md` presents "Method 1: map_shared_rank (pointer remapping)" (lines 101-130) with this ending:

```rust
let remote_ptr = unsafe {
    cluster::map_shared_rank(DATA.as_ptr().add(tid), neighbor)
};
let neighbor_val = unsafe { *remote_ptr };
```

The follow-up prose (lines 147-151) goes further and recommends this path for anything wider than u32: "for larger types, `map_shared_rank` with the appropriate pointer type works."

**2. The stub doc says deref is forbidden.** The generated Safety doc at `crates/cuda-device/src/generated/cluster_memory.rs:30` reads: "The result is a cluster address carrier. Do not use an ordinary pointer dereference for a remote load."

**3. The lowering makes deref a CTA-local load.** `MapaSharedClusterOp` in `crates/mir-lower/src/convert/generated_intrinsics.rs` emits `mapa.shared::cluster.u64` via inline asm and then inttoptr's the i64 result into an addrspace(3) pointer. A later Rust deref of that pointer is just a load from addrspace(3), which llc prints as a plain CTA-scoped `ld.shared.u32`, not `ld.shared::cluster.u32`. The mapped address names another block's shared memory, so the CTA-scoped load faults. The cluster example README already records the observed result in its troubleshooting table: `CUDA_ERROR_ILLEGAL_ADDRESS`, with "Use `dsmem_read_u32` instead" as the fix.

So the doc and the lowering agree with each other, and the chapter's Method 1 is a runtime crash written up as a tutorial. PR #1029 fixed the API quick reference, but the chapter still teaches the broken pattern.

**What resolving this could look like** (pick one story and make all three sources tell it):

- Option A: make the deref real. Return the mapa result as a pointer in the cluster shared address space (addrspace(7), `shared::cluster` in modern LLVM) so plain loads lower to `ld.shared::cluster`. The repo already floors at LLVM 21, which knows addrspace(7). Then Method 1 becomes true as written.
- Option B: declare deref unsupported. Rewrite Method 1 to route all remote reads through the dsmem intrinsics, add wider read variants (u64 and friends) since the chapter currently sends anything bigger than u32 to the deref path, and consider making `map_shared_rank` return an opaque carrier type instead of `*const T` so the signature stops inviting the deref.

Either way, the chapter, the stub Safety doc, and the emitted PTX should stop contradicting each other.

uurl:
Fixes #1034.

## Summary

Preserve the result of `map_shared_rank` and `map_shared_rank_mut` as a cluster-shared pointer (`addrspace(7)`) through MIR import and LLVM lowering.

This allows ordinary Rust pointer dereferences of mapped DSMEM addresses to select cluster-shared loads and stores instead of losing the mapped pointer semantics.

## Changes

- Add MIR address-space support for cluster-shared memory (`addrspace(7)`).
- Make `MapaSharedClusterOp` preserve pointee type and mutability while returning a cluster-shared pointer.
- Preserve `addrspace(7)` when `map_shared_rank` and `map_shared_rank_mut` results are stored into MIR locals.
- Lower `mapa.shared::cluster` results to LLVM `ptr addrspace(7)`.
- Update the generated probe and generated intrinsic documentation.
- Update the cluster example to validate:
  - mapped remote reads through ordinary Rust dereference;
  - mapped remote writes through ordinary Rust assignment.

## Result

The source pointer passed to `mapa.shared::cluster` remains CTA shared memory:

```llvm
ptr addrspace(3)
```

The mapped result is represented as:

```llvm
ptr addrspace(7)
```

An ordinary Rust read lowers to:

```llvm
%mapped = inttoptr i64 %mapped_integer to ptr addrspace(7)
%value = load i32, ptr addrspace(7) %mapped
```

and produces:

```ptx
mapa.shared::cluster.u64 ...
ld.shared::cluster.b32 ...
```

The mutable path similarly produces:

```llvm
store i32 %value, ptr addrspace(7) %mapped
```

and:

```ptx
mapa.shared::cluster.u64 ...
st.shared::cluster.b32 ...
```

## Validation

* Targeted crate/unit tests passed.
* `rustc-codegen-cuda` library tests: 23/23 passed.
* Workspace Clippy passed with warnings denied.
* Generated intrinsic outputs are current.
* Generated intrinsic probes: 1002/1002 passed.
* Intrinsic ABI history check passed.
* Cluster example build and pipeline passed.
* Mapped remote load passed on cluster-capable NVIDIA hardware.
* Mapped remote store passed on cluster-capable NVIDIA hardware.
* Compute Sanitizer memcheck: 0 errors.
* `git diff --check` passed.

### DIFF
--- a/crates/cuda-device/src/generated/cluster_memory.rs
+++ b/crates/cuda-device/src/generated/cluster_memory.rs
@@ -27,7 +27,7 @@ pub unsafe fn dsmem_read_u32(local_ptr: *const u32, target_rank: u32) -> u32 {
 ///
 /// # Safety
 /// `local_ptr` must point into CTA shared memory, and `target_rank` must name a rank in the same live cluster.
-/// The result is a cluster address carrier. Do not use an ordinary pointer dereference for a remote load.
+/// The result is a cluster-shared pointer in address space 7. Dereferencing it performs a remote DSMEM access; the target CTA must remain live and synchronization must make the access race-free.
 #[must_use]
 #[inline(never)]
 pub unsafe fn map_shared_rank<T>(local_ptr: *const T, target_rank: u32) -> *const T {
@@ -38,7 +38,7 @@ pub unsafe fn map_shared_rank<T>(local_ptr: *const T, target_rank: u32) -> *cons
 /// Maps a mutable CTA-shared address to another cluster rank.
 ///
 /// # Safety
-/// The mapping requirements above apply, and writes must not race.
+/// The mapping requirements above apply. The target CTA must remain live, and remote writes must be synchronized and race-free.
 #[must_use]
 #[inline(never)]
 pub unsafe fn map_shared_rank_mut<T>(local_ptr: *mut T, target_rank: u32) -> *mut T {
--- a/crates/cuda-intrinsics-gen/src/render.rs
+++ b/crates/cuda-intrinsics-gen/src/render.rs
@@ -5730,7 +5730,7 @@ fn render_compat_cluster_memory(catalog: &CatalogFile, hash: &str) -> String {
             ClusterMemoryOperation::MapSharedRank => {
                 output.push_str(
                     "/// `local_ptr` must point into CTA shared memory, and `target_rank` must name a rank in the same live cluster.\n\
-                     /// The result is a cluster address carrier. Do not use an ordinary pointer dereference for a remote load.\n",
+                     /// The result is a cluster-shared pointer in address space 7. Dereferencing it performs a remote DSMEM access; the target CTA must remain live and synchronization must make the access race-free.\n",
                 );
                 output.push_str("#[must_use]\n#[inline(never)]\n");
                 output.push_str(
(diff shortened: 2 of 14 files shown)


nihalpasham:
- Verified end to end: rebuilt the cluster example against this branch; the .ll loads and stores through `ptr addrspace(7)`, the PTX shows `mapa.shared::cluster.u64` + `ld.shared::cluster.b32` / `st.shared::cluster.b32`, and `cargo oxide run cluster` passes all five tests on an RTX 5090 (cc 12.0), including the new mapped remote store (ring values 3000-3003 as expected). Generator tests 289 pass, ledger check green.
- **Three commits layered on:**
  - the MapSharedRank doc string in `render_raw_abi` was truncated with a literal `...` while the generated file was hand-edited, so `cuda-intrinsics-gen check` failed on drift; completed the sentence and regenerated.
  - the book still referred to the old semantics: the quick reference's "carrier, don't deref" warning and the cluster chapter's "generic load path" framing are now false, and the PR's own replacement snippet passed the mapped pointer into `dsmem_read_u32` (it takes the local pointer plus a rank) and dropped `unsafe`. Both passages rewritten; deref is the primary method, `dsmem_read_u32` the fixed-width alternative. This is what actually lets #1034 close.
  - added `verify-code-shape.sh` for the cluster example (same presence-triggered pattern six other examples use) asserting `addrspace(7)` in the .ll and `ld/st.shared::cluster` in the PTX, so a regression to CTA-local loads cannot slip past compile-only CI.

nihalpasham (approved):

PR #1043, opened 2026-08-19
ci: cargo-deny reaches two of the three non-example workspace roots

vyncint:
`cargo deny check` stops at a `[workspace]` boundary, which is why this workflow already runs it twice: once at the root and once for `crates/rustc-codegen-cuda`. There is a **third** such root under version control that resolves third-party crates, and nothing judges it.

`crates/cuda-macros/tests/device-only` declares its own `[workspace]` on purpose — the fixture exists to prove a graph that does *not* contain `cuda-host` (#701/#702), so it must resolve independently rather than share the root lock. That same boundary puts it outside all three existing runs:

- the root `cargo deny check` stops at it,
- the backend run names a different manifest,
- `check-example-license-policy.sh` walks `crates/rustc-codegen-cuda/examples` only.

## The gap is not theoretical

Its lock against both governed locks:

| crate | fixture | root | backend | governed? |
|---|---|---|---|---|
| `proc-macro2` | 1.0.107 | 1.0.106 | 1.0.106 | **no** |
| `quote` | 1.0.47 | 1.0.45 | 1.0.44 | **no** |
| `syn` | 2.0.119 | 2.0.117, 3.0.3 | 2.0.114, 3.0.3 | **no** |
| `unicode-ident` | 1.0.24 | 1.0.24 | 1.0.22 | yes |

Three of its four third-party `(name, version)` pairs appear in neither governed lock, so no advisory, license, ban or source policy has ever been applied to them. `(name, version, source)` is this repo's own unit of policy — it is the grouping key `check-example-license-policy.sh` uses to argue its runs-for-workspaces equivalence — so a version present nowhere else is a distinct policy question, not a duplicate.

## What this does

Adds the third run with `--locked`, like the other two. A check that can rewrite its own input is not a check, and verified here: deleting a package block from that lock makes the run stop with `cannot update the lock file` rather than silently re-resolving.

`crates/fuzzer/rustlantis` is the one remaining `[workspace]` root and is **deliberately left out**: it is a vendored upstream project rather than first-party code, so bringing its graph under `deny.toml` is a policy call for a maintainer, not a gap to close silently. The header comment now records which roots are in scope and why.

Two present-tense counts in this file had also drifted and are corrected against the current tree:

- the backend workspace resolves **88** third-party crates, not 82 (`cargo metadata --manifest-path crates/rustc-codegen-cuda/Cargo.toml`, packages with a non-null `source`);
- the example-policy script is **31 runs over 217 lock files**, not 26 over 187 — 31 is the number the script itself announces.

## Verification

- The new invocation reports `advisories ok, bans ok, licenses ok, sources ok` and exits 0 on today's tree, so `main` stays green.
- Every workflow still parses.
- `just check-guards` carries the same third run, so the local mirror keeps predicting CI.
- All four counts re-measured against `f1e62fe2`.

### DIFF
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -6,9 +6,11 @@ name: cargo-deny
 # Concurrency for the CI path is governed by CI's own group.
 #
 # Four parts of the same concern. `check` enforces the license *policy* over
-# the whole resolved graph from deny.toml -- once for the root workspace and
-# once for crates/rustc-codegen-cuda, which has its own `[workspace]` and so is
-# a separate resolution the root run cannot reach. `license-manifest` enforces that
+# the whole resolved graph from deny.toml, once per `[workspace]` root that
+# resolves third-party crates and is not an example: the root workspace,
+# crates/rustc-codegen-cuda, and the cuda-macros device-only fixture. Each
+# declares its own `[workspace]`, so each is a separate resolution the others
+# cannot reach. `license-manifest` enforces that
 # the human-readable inventory in dependency-licenses.csv still lists what the
 # workspace declares, which nothing checked before: CONTRIBUTING asks
 # contributors to update that file by hand, and it had fallen behind by eight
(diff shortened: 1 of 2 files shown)


nihalpasham:
- Confirmed the gap was real: three of the fixture's four external deps are locked at versions that appear in no other checked lockfile, so no advisory or license rule had ever run on those exact versions.
- Reproduced the new CI command locally: `cargo deny --manifest-path crates/cuda-macros/tests/device-only/Cargo.toml --locked check` passes (advisories, bans, licenses, sources all ok), and `just check-guards` runs the same third scope end to end, 34 verdicts total.
- Leaving the vendored fuzzing project out on purpose, with the comment saying why, is the right call.

nihalpasham (approved):

PR #1044, opened 2026-08-19
ci: the example-policy guard's parse self-test cannot detect the rot it guards

vyncint:
`check-example-license-policy.sh` collapses every example lock file into groups keyed by their third-party `(name, version, source)` triples and runs cargo-deny once per distinct group. The comment above its second self-test states the hazard exactly:

```
# Mirrors the inventory guard: if the lock-file regexes silently rotted, the
# groups would quietly collapse and a single run would vouch for everything.
```

**That test cannot see it.** `seen` incremented once per `[[package]]` string split, *before any regex ran*, so it counted blocks rather than parses. Simulating a rot of `^name` over the real tree:

```
healthy          locks=217  seen=13871  groups=31
regexes rotted   locks=217  seen=13871  groups=1      <- both floors pass
```

Every triple fails to match, every lock parses as having zero third-party crates, all 217 collapse into the single empty group, and the guard runs cargo-deny **once, over one example, and exits 0** — vouching for everything, which is the outcome the comment says it prevents. Verified for `^name`, `^version` and `^source` independently; each rots to the same result.

## Three changes

1. **`seen` becomes `parsed`,** counted only after `name` and `version` match. Every locked package carries both, so the tally now goes to zero on a rot instead of staying at 13871. Its floor is scaled to the locks found rather than fixed, so it cannot go stale as the tree grows.

2. **A per-lock assertion is the actual detector.** Every example reaches crates.io through cuda-core/cuda-device/cuda-host, so a lock resolving *no* third-party crate means the parse failed. Measured range across the tree is **46 to 142** per lock, so the margin is wide — and a rot takes every lock to zero.

3. **The `len(locks) < 20` floor is replaced by a cross-check against an independent enumeration.** A count cannot tell a narrowed glob from a smaller tree — `[a-m]*` still finds 137 of 217 locks and clears any floor loose enough to survive ordinary growth — whereas every example directory with a `Cargo.toml` has a `Cargo.lock` beside it (215 of each today), so comparing the two sets names exactly what the glob stopped reaching.

## Mutations

Each fails with a specific message where the old code passed silently:

| mutation | result |
|---|---|
| rot `^name` | `parse self-test failed: no third-party crates in .../abi_hmm/Cargo.lock` |
| rot `^version` | same |
| rot `^source` | same |
| glob → `[a-m]*` | `found no Cargo.lock for 78 example(s), so deny.toml cannot be enforced over them: ne_bytes_transmute, nested_index_assignment, …` |

## Scope note

The scope sentence is also corrected: it described 187 lock files (186 examples plus **one** nested sub-workspace) collapsing to 26 runs. Today's tree is **217** locks — 215 examples plus **two** nested sub-workspaces, `cutile_inter_kernel/simt` and `interop_cubin_identity/device` — collapsing to the **31** runs the script itself announces.

The `# Measured when this guard was written:` line above it is a timestamped measurement and is left exactly as written.

## Verification

The guard passes unchanged on today's tree: 31 representatives, every one reporting `advisories ok, bans ok, licenses ok, sources ok`, exit 0. All counts re-measured against `f1e62fe2`.

### DIFF
--- a/scripts/check-example-license-policy.sh
+++ b/scripts/check-example-license-policy.sh
@@ -21,8 +21,8 @@
 #   Every example depends on cuda-core/cuda-device/cuda-host by path, so each
 #   lock file re-lists the root workspace's own transitive crates.  Grouping the
 #   lock files by their exact set of third-party (name, version, source) triples
-#   collapses 187 lock files (186 example workspaces plus one nested
-#   sub-workspace) to 26 distinct sets and one cargo-deny run each.  That is an
+#   collapses 217 lock files (215 example workspaces plus two nested
+#   sub-workspaces) to 31 distinct sets and one cargo-deny run each.  That is an
 #   equivalence, not a sample: license, source and advisory verdicts are
 #   per-crate properties, so two workspaces resolving the identical crate set
 #   get the identical verdict.  `bans.multiple-versions` is a graph property,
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified the hardened self-test both ways: clean tree passes (31 grouped cargo-deny runs), and with a deliberately dead name pattern the guard now fails on the first lock with `parse self-test failed: no third-party crates in crates/rustc-codegen-cuda/examples/abi_hmm/Cargo.lock`. The old count-before-parse check stayed green under the same breakage, which is the hole this closes.
- The glob-vs-disk cross-check naming missed examples beats the fixed "at least 20" floor. No changes needed.

nihalpasham (approved):

PR #1046, opened 2026-08-19
docs(cuda-async): the README's key-methods table omits both non-blocking paths

vyncint:
`DeviceOperation` has five scheduling entry points. The README's table lists three: `.await`, `.sync()` and `.sync_on(&stream)`. The two it leaves out are `schedule` and `async_on` — which is to say, **both of the ones that do not block**, under a table whose own column is "Blocks thread?". Every row reads `Yes` or `No (suspends)` because the rows that would read otherwise are absent.

## `async_on` is the one that matters

It is the trait's only `unsafe` scheduling method, and it differs from `sync_on` by exactly one line:

```rust
fn sync_on(...)         { let res = unsafe { self.execute(&ctx) };
                          finish_sync(res, stream.synchronize()) }
unsafe fn async_on(...) { unsafe { self.execute(&ctx) } }
```

So it returns while GPU work may still be in flight, and the caller owns every buffer the operation touches until the stream is synchronized some other way. A reader of this README currently learns nothing about it.

#1028 made the two in-source tables say this, after they had listed `async_on` beside the four safe methods. This README is the **third** table and still did not mention it at all. This adds both missing rows, adds the `unsafe?` column the in-source table gained, and states the consequence in a sentence beneath, so the three tables agree.

`execute` is the trait's other `unsafe fn` and stays out of the table, noted parenthetically instead: it is the primitive the four safe methods are built from rather than an entry point to call.

## Verification

Cross-checked against the source rather than the docs:

- the trait declares exactly `execute`, `schedule`, `sync`, `sync_on` and `async_on` as scheduling methods (`and_then`, `and_then_with_context`, `apply` and `arc` are the combinators the next README line already lists);
- `execute` and `async_on` are the only two `unsafe`;
- `async_on`'s body contains no `synchronize` call and `sync_on`'s does — extracted by brace matching, not a fixed window;
- `schedule` returns a `DeviceFuture`.

All re-measured against `f1e62fe2`.

### DIFF
--- a/crates/cuda-async/README.md
+++ b/crates/cuda-async/README.md
@@ -35,15 +35,21 @@ A lazy, composable unit of GPU work. Implements `Send + Sized + IntoFuture`. Key
 | `.sync_on(&stream)`   | The explicit stream you provide       | Yes            | No        |
 | `.async_on(&stream)`  | The explicit stream you provide       | No             | **Yes**   |
 
-`.async_on` is the only one of the five that is `unsafe`, and it is the only
-one that returns while GPU work may still be in flight: it executes and does
-not synchronize, so the caller owns keeping every buffer the operation touches
-alive and unread until the stream is synchronized by other means. Each of the
-other four ties its result to completion: `.sync` and `.sync_on` block,
+`.async_on` is the only `unsafe` method here, and the only one that returns
+while the GPU may still be working. It launches the work and never
+synchronizes. Until you synchronize the stream yourself, every buffer the
+operation touches must stay alive, and its outputs must not be read. Each of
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified all five rows against `device_operation.rs`: `execute` and `async_on` are the only `unsafe` ones, `.schedule` returns a `DeviceFuture`, and `async_on` has no synchronize. The README table now matches the two in-source tables from #1028.
- **One commit layered on:** split the `async_on` safety sentence into three short ones and added a usage snippet with the `// SAFETY:` comment, same as the README's other unsafe call sites.

PR #1047, opened 2026-08-19
fix(cargo-oxide): `cargo oxide fmt` misses three scopes the fmt gate checks

vyncint:
CONTRIBUTING tells contributors to run `cargo oxide fmt` rather than a bare `cargo fmt` for one stated reason: otherwise the gate *"will fail on code you never had a chance to format."* That only holds while the command covers what the gate covers, and **since #1027 it does not**.

`.github/workflows/fmt.yml` checks four scopes: the root workspace, the codegen backend, `crates/cuda-macros/tests/device-only`, and every `Cargo.toml` under `examples/` via a `**` glob — 225 manifests, nested ones included. `format_all` checked three: root, backend, and `cargo fmt --all` in each of the 215 top-level example directories. That misses exactly three manifests:

- **the device-only fixture** — not under `examples/` at all, and it had no scope of its own here (the gate gained one in #1027);
- **`examples/cutile_inter_kernel/simt`**;
- **`examples/interop_cubin_identity/device`**.

The last two are the only nested example manifests declaring their own `[workspace]`, so the parent's `cargo fmt --all` stops before reaching them. The other eight nested manifests are path-dependency members and were already covered.

## Measured, not inferred

With one unformatted function appended to a file in each of the three:

| what | result |
|---|---|
| `cargo fmt --all` in `cutile_inter_kernel` | **exit 0** — missed |
| `cargo fmt --all` in `interop_cubin_identity` | **exit 0** — missed |
| `cargo oxide fmt --check` (old) | walked 185 scopes including **both** parents, reported no failure, and never printed a fixture scope at all |
| the gate's fixture step | exit 1 — caught |
| the gate's `--manifest-path` on each nested manifest | exit 1 — caught |

## The fix

The examples loop becomes one `--manifest-path` run per manifest found, which is what the gate does, for the two reasons the gate's own comment gives: `--all` stops at a nested `[workspace]`, and `--all` also formats an example's path dependencies, so running it per example re-formats the large shared workspaces 215 times.

**That second point was expensive.** The old implementation had not finished 185 of 217 scopes after roughly 25 minutes. The new one covers all 228 in 19–41s depending on cache state.

The walk skips `target` and dot directories. The gate's glob does not need to — a fresh checkout has no build directories — but a working tree does, and a packaged or vendored manifest under one is not a crate this repository formats. A unit test covers the walk: nested manifests found, workspace members found, `target/` and dot directories skipped.

## Three stale documentation sites

All three stated the old three-scope shape and are corrected: the function's own doc comment, CONTRIBUTING's formatting section (whose clippy paragraph also said "the same three scopes" — clippy has more, since #1026), and the `cargo oxide fmt` comment in the book's CLI block. I grepped for other sites stating a scope count; these are the only ones.

## Verification

- **228 scopes visited**, matching the gate's four exactly (root + backend + fixture + 225 manifests).
- With the three probe files unformatted: reports exactly those three diffs, nothing else, exit 1.
- On the clean tree: exit 0.
- `cargo test -p cargo-oxide --all-targets` — 177 pass; `clippy --all-targets -- -D warnings` clean; `cargo fmt` clean.
- `check-cli-doc-coverage.sh` and `check-device-only-build.sh` pass; book builds under `sphinx-build -W`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -146,16 +146,21 @@ hand even so.
 
 - Run `cargo oxide fmt` before submitting. All code must be formatted with
   `rustfmt`. Use `cargo oxide fmt` rather than a bare `cargo fmt`: the codegen
-  backend and every example are their own workspaces, so `cargo fmt` at the
-  repository root reaches none of them, while the `fmt` CI job checks all three
-  scopes and will fail on code you never had a chance to format.
-- Run clippy and address any warnings where reasonable. It has the same three
-  scopes, and there is no single command covering them:
+  backend, every example and the `cuda-macros` device-only test fixture are
+  each their own workspace, so `cargo fmt` at the repository root reaches none
+  of them, while the `fmt` CI job checks all four scopes and will fail on code
+  you never had a chance to format. `cargo oxide fmt` mirrors that job, nested
+  example workspaces included.
+- Run clippy and address any warnings where reasonable. There is no single
+  command covering its scopes, and it has more of them than `fmt`: the two
+  workspaces below, plus one run per example, plus the nested example
+  workspaces that their parent does not list as members, plus the device-only
+  fixture. `.github/workflows/clippy.yml` is the full list; the two worth
+  running by hand are:
 
   ```bash
   cargo clippy --workspace --all-targets -- -D warnings
   (cd crates/rustc-codegen-cuda && cargo clippy --all-targets -- -D warnings)
-  # and per example, as .github/workflows/clippy.yml does
   ```
 - Follow existing code patterns and conventions in the crate you are
   modifying.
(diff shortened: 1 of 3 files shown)


nihalpasham:
- Reproduced the gap and the fix locally: `cargo test -p cargo-oxide` passes 177, and `oxide fmt --check` exits 0 covering 228 scopes (root + backend + device-only fixture + 225 example manifests), matching the gate's four scopes.
- Probe test: dropped an unformatted fn into the fixture and both nested example workspaces; the new walk flags all three where the old command exited 0.
- Worth a follow-up issue: nothing automated compares fmt.yml's glob against the command's walk, so the next gate change could drift again. Filing separately.

nihalpasham (approved):

PR #1050, opened 2026-08-19
cargo-oxide: hold the fmt gate and `cargo oxide fmt` to one scope list

### LINKED ISSUE #1048: fmt gate and cargo oxide fmt have no automated parity check
The fmt CI gate (`.github/workflows/fmt.yml`) and `cargo oxide fmt` are two implementations of the same scope list that mirror each other by convention only. That is how the drift fixed in #1047 happened: the gate grew a fourth scope and the command kept three, so contributors could format everything locally and still fail CI.

#1047 restores parity and documents the mirroring invariant on both sides, but nothing enforces it. Two options, either closes this:

- a small check that parses fmt.yml's scope steps and compares them against `collect_example_manifests` plus the fixed scopes, failing when the sets diverge;
- have the fmt job run `cargo oxide fmt --check` directly so there is only one implementation. The catch today is the job runs on a bare checkout with a 10-minute timeout, and compiling cargo-oxide first costs more than the check itself, so this needs a cached or prebuilt binary to be viable.

Until one of these lands, any PR touching fmt.yml scopes must remember to touch `format_all` in `crates/cargo-oxide/src/commands.rs` too.

NgoQuocViet2001:
Closes #1048, taking the first of the two options there — the check that compares the gate's scopes against the command's — since the second needs a cached backend binary the fmt job does not have.

## What changed

`format_all` hand-unrolled its three fixed scopes as three blocks. They are now one list:

```rust
pub(crate) const FIXED_FMT_SCOPES: &[(&str, &str)] = &[
    (".", "root workspace"),
    ("crates/rustc-codegen-cuda", "rustc-codegen-cuda"),
    ("crates/cuda-macros/tests/device-only", "cuda-macros device-only fixture"),
];
```

and `format_all` iterates it. The `Cargo.toml` existence check that guarded only the device-only fixture now guards every scope, which is what the fixture's block was doing anyway.

`fmt_gate_and_cargo_oxide_fmt_cover_the_same_scopes` reads `.github/workflows/fmt.yml` and takes the directory from each whole-directory `cargo fmt` step — the `working-directory:` that follows it, or the workspace root when there is none — then compares that set against `FIXED_FMT_SCOPES`.

The examples walk is held the same way: `FMT_EXAMPLES_DIR` now names the directory `resolve_context` builds `examples_dir` from, and the test asserts the gate globs `{FMT_EXAMPLES_DIR}/**/Cargo.toml`. So the constant is load-bearing rather than a second copy of the path.

The test also asserts every listed scope really has a `Cargo.toml`, so a rename cannot leave the command silently formatting nothing while the gate still checks it.

## Test plan

- Ran: `cargo test -p cargo-oxide` — 178 passed, 0 failed (177 pre-existing + 1 new).
- Ran: `cargo clippy -p cargo-oxide --all-targets` — 0 warnings.
- Ran: `cargo fmt --all -- --check` — clean.
- Checked: reproduced the #1047 drift by adding a fourth `working-directory:` step to `fmt.yml` and leaving the command alone. The check fails with the diff spelled out:

```
assertion `left == right` failed: fmt.yml and FIXED_FMT_SCOPES disagree;
a scope added to one needs the other
  left:  [".", "crates/cuda-host", "crates/cuda-macros/tests/device-only", "crates/rustc-codegen-cuda"]
  right: [".", "crates/cuda-macros/tests/device-only", "crates/rustc-codegen-cuda"]
```

## Scope

Deliberately text-level rather than a YAML dependency: the shape it has to recognise is "a step running `cargo fmt --all -- --check`, optionally with a `working-directory:`", which is what the gate writes today and what a new scope would be written as. If the gate ever moves to a matrix or a composite action this needs revisiting, and the assertion message says which two places disagree so that is a visible failure rather than a silent pass.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -438,7 +438,7 @@ pub fn resolve_context() -> Context {
 pub fn resolve_passive_context() -> Context {
     if let Some(workspace_root) = backend::find_workspace_root() {
         let codegen_crate = workspace_root.join("crates/rustc-codegen-cuda");
-        let examples_dir = codegen_crate.join("examples");
+        let examples_dir = workspace_root.join(FMT_EXAMPLES_DIR);
         let config = load_oxide_config_lenient(&workspace_root);
         let backend_so = backend::backend_so_candidate(&workspace_root, config.backend.as_deref());
         return Context {
@@ -4021,16 +4021,6 @@ pub fn codegen_debug(
 // Fmt command
 // =============================================================================
 
-/// Format (or check formatting of) every scope the `fmt` CI gate checks.
-///
-/// `.github/workflows/fmt.yml` checks four: the root workspace, the codegen
-/// backend crate, the cuda-macros device-only fixture, and every `Cargo.toml`
-/// under `examples/`, nested ones included. This mirrors that set on purpose --
-/// the reason CONTRIBUTING tells contributors to prefer this command over a
-/// bare `cargo fmt` is so the gate cannot fail on code they had no way to
-/// format, which only holds while the two cover the same ground.
-///
-/// In `check` mode, reports which files need formatting without modifying them.
 /// Directories `cargo oxide fmt` formats in their own right, relative to the
 /// workspace root, with the label each is announced under.
 ///
@@ -4060,6 +4050,16 @@ pub(crate) const FIXED_FMT_SCOPES: &[(&str, &str)] = &[
 /// with it.
 pub(crate) const FMT_EXAMPLES_DIR: &str = "crates/rustc-codegen-cuda/examples";
 
+/// Format (or check formatting of) every scope the `fmt` CI gate checks.
+///
+/// `.github/workflows/fmt.yml` checks four: the root workspace, the codegen
+/// backend crate, the cuda-macros device-only fixture, and every `Cargo.toml`
+/// under `examples/`, nested ones included. This mirrors that set on purpose --
+/// the reason CONTRIBUTING tells contributors to prefer this command over a
+/// bare `cargo fmt` is so the gate cannot fail on code they had no way to
+/// format, which only holds while the two cover the same ground.
+///
+/// In `check` mode, reports which files need formatting without modifying them.
 pub fn format_all(ctx: &Context, check: bool) {
     let mode = if check { "Checking" } else { "Formatting" };
     let mut failed = false;
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -388,7 +388,7 @@ pub struct Context {
 pub fn resolve_context() -> Context {
     if let Some(workspace_root) = backend::find_workspace_root() {
         let codegen_crate = workspace_root.join("crates/rustc-codegen-cuda");
-        let examples_dir = codegen_crate.join("examples");
+        let examples_dir = workspace_root.join(FMT_EXAMPLES_DIR);
         let config = load_oxide_config(&workspace_root);
         let backend_so = backend::find_or_build_backend(&workspace_root, config.backend.as_deref());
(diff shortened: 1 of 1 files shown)


nihalpasham (changes requested):
Verified against the tree: the parity test parses the real fmt.yml and matches its scope set, and the branch correctly builds on #1047 (your base already has it). This closes #1048 via its option 1.

Two small changes needed to merge, both in commands.rs:

- The doc comment moved: `FIXED_FMT_SCOPES` was inserted right after `format_all`'s doc block, so the whole function doc (including the check-mode sentence) now documents the const, and `pub fn format_all` has no rustdoc. Keep the scope-list prose on the const and move the function doc back onto the function.
- One more hardcoded scope: `resolve_passive_context` still builds its examples dir by hand. Derive it from `FMT_EXAMPLES_DIR` so there's truly one list:

```diff
-        let examples_dir = codegen_crate.join("examples");
+        let examples_dir = workspace_root.join(FMT_EXAMPLES_DIR);
```

With both applied, `cargo test -p cargo-oxide --all-targets` passes 178 including your new parity test, and clippy `-D warnings` plus fmt stay clean; we ran all three on your branch with the changes in place.

To land it:

- Push the two changes, or tick "Allow edits by maintainers" and we'll push the validated commit.

Worth a separate issue so it doesn't get lost, not needed here: also assert every `working-directory:` in fmt.yml maps to a known scope, so a future step written with a different run spelling still trips the test.

NgoQuocViet2001:
Both applied, pushed.

**Doc placement.** You're right, and it was worse than misplaced prose: `pub fn format_all` had no rustdoc at all. Dropping the const in immediately after the doc block silently handed the whole thing — including the check-mode sentence, which describes an argument the const does not have — to `FIXED_FMT_SCOPES`. The scope-list prose stays on the const; the function doc is back on the function.

**Second hardcoded scope.** Fixed — `resolve_passive_context` now derives from `FMT_EXAMPLES_DIR`. That one undercut the whole point of the PR, since the path still lived in two places.

One I deliberately left: `codegen_crate.join("examples")` also appears in `workspace_clean_removes_root_backend_and_example_targets`. That one builds a synthetic temp tree to clean, rather than resolving the real scope, so pointing it at the const would couple a fixture to production layout without buying a single source of truth. Say the word if you'd rather it move too.

**Verification.** `cargo test -p cargo-oxide --all-targets` → **177 passed, 1 failed**, and the failure is not from this change: `backend::tests::backend_build_command_isolates_target_dir_and_forces_the_host` compares `C:/tmp/codegen\target` against `/tmp/codegen/target`, i.e. Windows path separators. I confirmed it fails identically on unmodified `main`, so it is my machine, not the branch — which matches your 178/178 on Linux. `rustfmt --check` on `commands.rs` is clean.

**On the follow-up:** agreed that asserting every `working-directory:` in `fmt.yml` maps to a known scope is the stronger invariant — it would catch a step written with a different run spelling, which this test would sail past. Happy to open that issue and take it, unless you'd rather file it yourself so the framing is yours.

nihalpasham:
Closing this PR because the production scope refactor is useful, but its parity gate can still report success when CI does not enforce the same scope.

- Adding `if: false` or `continue-on-error: true` to a real fmt step remains green.
- Moving an extra format invocation to the wrong working directory remains green.
- Leaving the manifest glob only in a comment still satisfies the substring check.

The test scans lines and substrings, so its happy path repeats the same assumption instead of proving PR-reachable, failure-propagating enforcement.

Under the [acceptance criteria shared in Discord](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775), fixes need to address the root cause across the full path and include negative and unaffected controls. A replacement should use one executable scope source or structural, fail-closed validation with these mutations. That is a redesign rather than a small maintainer correction.

The constant and production refactor can still inform a narrower replacement.

PR #1025, opened 2026-08-18
feat(wgmma): add value-threaded F16 full-drain lowering

uurl:
Closes #1024 

## Summary

Adds value-threaded lowering for Hopper

`wgmma.mma_async.sync.aligned.m64n64k16.f32.f16.f16`

in the canonical linear full-drain case.

F16 now follows the existing BF16 register-resident strategy for:

```text
wgmma.fence
mma+
commit_group
wait_group<0>
```

The 32 `f32` accumulator values remain tied through one convergent inline-PTX region, preventing LLVM from introducing a spill boundary while the asynchronous WGMMA group is in flight.

### Device API / importer

* Enables importer handling for `wgmma_mma_m64n64k16_f32_f16`.
* Keeps TF32 unsupported.
* Documents the supported F16 lowering shape.

### NVVM dialect

Adds:

* `WgmmaMmaM64N64K16F32F16Op`
* `WgmmaMmaGroupValuesM64N64K16F32F16Op`

The F16 value carrier has the same 32 × `f32` accumulator interface as the existing BF16 carrier.

No F16 counted-loop or pipeline carrier ops are added.

### MIR lowering

* Generalizes the linear value-group PTX template over BF16/F16 input kind.

* Emits:

  ```text
  wgmma.mma_async.sync.aligned.m64n64k16.f32.f16.f16
  ```

  for F16.

* Recognizes canonical pointer-form F16 linear full-drain regions.

* Converts them to the value-threaded carrier.

* Requires homogeneous BF16 or F16 MMAs within a single linear full-drain region.

* Rejects non-canonical F16 accumulators instead of using the BF16 deferred
  pointer fallback.

Existing counted-loop and partial-wait recognizers remain BF16-only.

### Tests

Adds coverage for:

* direct F16 value-carrier lowering
* F16 pointer-form full-drain fusion
* non-canonical F16 rejection
* mixed BF16/F16 rejection
* F16 partial-wait rejection
* F16 NVVM op verification
* F16 importer support

The existing `wgmma_value_carrier_ptx` end-to-end test is also extended to F16.

## Validation

Validated locally with CUDA 13.2 / `ptxas` targeting `sm_90a`.

```text
cargo fmt --all -- --check
git diff --check
cargo check -p dialect-nvvm -p mir-importer -p mir-lower -p cuda-device
cargo clippy -p dialect-nvvm -p mir-importer -p mir-lower -p cuda-device \
  --all-targets -- -D warnings
```

WGMMA lowering tests:

```text
[...]

PR #1052, opened 2026-08-19
docs(book): the dialect-nvvm counts predate the catalog, by the page's own test

vyncint:
`mlir-dialects.md` anchors its op counts to a catalog stamp and says how to tell when they have rotted:

> At catalog SHA-256 `20440c06` (the stamp in every `ops/generated/` file header), the dialect holds 560 operations across 42 modules … **If the header stamp no longer starts with `20440c06`, the counts on this page predate the catalog you are reading.**

Every generated file now stamps `26a9cc13`, so that condition is true.

| claim | page | tree |
|---|--:|--:|
| dialect total | 560 | **570** |
| hand-written ops | 19 | **21** |
| hand-written `wgmma` row | 7 | **9** |
| generated ops | 541 | **549** |
| catalog entries | 1002 | **1010** |
| generated "Warp-level" row | 31 | **39** |

Both deltas are traceable. #1021 and #1025 added two hand-written `wgmma` ops (19 → 21). #1032 added the eight `redux.sync` f32 variants, which moves the generated total, the catalog count, and the one area row that owns `redux` (541 → 549, 1002 → 1010, 31 → 39).

## The description moved with the count, not just the number

The `wgmma` row read "the m64n64k16 **bf16** shapes". #1025 added `WgmmaMmaM64N64K16F32F16Op` and `WgmmaMmaGroupValuesM64N64K16F32F16Op`, so the module holds f16 shapes too — the row now says `bf16/f16`. A count fix that left that alone would have replaced one wrong statement with another.

Module counts did not change: 7 hand-written plus 35 generated is still 42.

## Verification

Measured against the tree, not against the merge descriptions:

- `^pub struct \w+Op;` counts **21** under `src/ops/` and **549** under `src/ops/generated/` — **570** together.
- `intrinsics/catalog.json` holds **1010** entries.
- All 35 generated files carry exactly one distinct stamp, `26a9cc13`.
- Both tables sum to their new totals (7 rows → 21, 11 rows → 549), and each still lists every module on disk with nothing in it that is not a module.
- Book builds clean under `sphinx-build -W`.

### DIFF
--- a/cuda-oxide-book/compiler/mlir-dialects.md
+++ b/cuda-oxide-book/compiler/mlir-dialects.md
@@ -296,16 +296,16 @@ they become `call` instructions to `@llvm.nvvm.*` intrinsics.
 
 ### Architecture Coverage
 
-At catalog SHA-256 `20440c06` (the stamp in every `ops/generated/` file
-header), the dialect holds 560 operations across 42 modules, and they come
+At catalog SHA-256 `26a9cc13` (the stamp in every `ops/generated/` file
+header), the dialect holds 570 operations across 42 modules, and they come
 from two different places. The split is the first thing to know about it,
 because it decides where -- and whether -- you would add one. If the header
-stamp no longer starts with `20440c06`, the counts on this page predate the
+stamp no longer starts with `26a9cc13`, the counts on this page predate the
 catalog you are reading.
 
 **Hand-written**, directly under `crates/dialect-nvvm/src/ops/`. These are the
 ops with bespoke verification or lowering that the intrinsic catalog does not
-describe. There are seven modules and 19 operations:
+describe. There are seven modules and 21 operations:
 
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Re-derived every number from the tree: 570 total ops (21 hand-written + 549 generated from 1010 catalog entries), wgmma 9, warp-level 39, and catalog.json hashes to `26a9cc13`. All six changed counts and the ten unchanged rows are correct; `sphinx-build -W` passes on the head.
- The bf16 to bf16/f16 wording matches the two F16 ops #1025 added (the body also credits #1021, which added lowering only, no op structs).
- Merge-order note: the counts re-stale the moment another op-adding PR lands, and no CI gate ties the page's stamp to the catalog hash. Filing a follow-up.

nihalpasham (approved):

PR #1053, opened 2026-08-19
docs(book): the warp chapter says the f32 redux.sync form does not exist

vyncint:
`warp-level-programming.md` closes its `redux.sync` section with:

> - **Integers only.** There is no `f32`/`f64` form before `sm_100`, so float reductions keep the butterfly.

#1032 shipped that form. `cuda-device` now exports eight `f32` entry points the chapter mentions nowhere:

```
redux_sync_min_f32          redux_sync_max_f32
redux_sync_min_abs_f32      redux_sync_max_abs_f32
redux_sync_min_nan_f32      redux_sync_max_nan_f32
redux_sync_min_abs_nan_f32  redux_sync_max_abs_nan_f32
```

A reader following this page today writes the shuffle butterfly for an `f32` min/max reduction the hardware does in one instruction, because the book told them the instruction does not exist. `f64` still has no form, so that half of the claim stands and is kept.

## Three things the new subsection is careful about

**It is not one function per operation.** The integer family is, and the page says so. For `f32`, `.abs` and `.NaN` are independent instruction modifiers, so each of the four combinations is its own entry point.

**The `.NaN` distinction is stated in the catalog's own terms.** The plain and `.abs` forms ignore NaN inputs; the `.NaN` forms propagate NaN when any lane contributes one. That is the difference that decides which to call, and it comes from the catalog summaries rather than from my own gloss on IEEE semantics.

**The floor is narrower than "sm_100+".** The catalog records PTX 8.6 with a target union of `sm_100a|sm_100f|sm_103a|sm_103f`, so the intrinsic is not selected on `sm_110a` or `sm_120a` — consumer Blackwell included. Saying just "Blackwell" would send RTX-50 users down a path that cannot lower.

## No performance claim

The integer section's 4.74x is a measurement on an A10G. The f32 family needs a B200/GB200, `redux_f32` sits in `SM100_COMPILE_EXAMPLES` so CI checks its emitted PTX rather than running it, and its `main` skips unless the device reports exactly `sm_100`. The subsection says that instead of implying a number.

## Verification

- All eight names exist as `pub fn` in `crates/cuda-device/src/warp.rs`.
- Arch union, PTX version and NaN/abs semantics all come from `intrinsics/catalog.json`.
- Every one of the eight is written module-qualified, so `check-book-api-names.sh` now covers all of them: it checks **39** qualified names where it checked 35, and renaming one to `redux_sync_max_abs_nan_f64` makes it fail.
- Book builds clean under `sphinx-build -W`.

### DIFF
--- a/cuda-oxide-book/advanced/warp-level-programming.md
+++ b/cuda-oxide-book/advanced/warp-level-programming.md
@@ -183,18 +183,22 @@ own entry point.
 
 The `.NaN` forms matter for reductions that must not lose a NaN. The plain and
 `.abs` forms *ignore* NaN inputs; the `.NaN` forms propagate NaN when any lane
-contributes one. Pick deliberately -- the two differ only when a lane holds a
-NaN, which is exactly when it matters.
+contributes one. Pick deliberately. The two differ only when a lane holds a
+NaN, and that is when the choice matters.
 
-**This is Blackwell-only, and narrower than "sm_100+".** The catalog records
-the family at PTX 8.6 with a target union of
-`sm_100a|sm_100f|sm_103a|sm_103f`, so it is not selected on `sm_110a` or
-`sm_120a` -- consumer Blackwell included. Build for one of those four targets
-or the intrinsic will not lower.
+Say one lane holds `NaN` and the smallest ordinary value in the warp is `-1.0`:
 
-The `redux_f32` example is the working reference. It pins `sm_100a`, so CI
-compiles it and checks the emitted PTX rather than running it, and its `main`
-skips unless the device reports exactly `sm_100`.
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified every claim against the tree: the four targets and PTX 8.6 match the generated ABI, `redux_sync_min_f32` / `redux_sync_min_nan_f32` exist with the stated NaN semantics, and the example's `main` skips on anything but `(10, 0)`. Book build and the api-names guard pass.
- **One commit layered on:** the availability paragraph now speaks the chapter's language ("exist only on...", "compilation fails") instead of compiler internals ("catalog", "target union", "will not lower"), and a two-value worked example shows NaN-ignore vs NaN-propagate, the one difference that decides which form to call.

nihalpasham (approved):

PR #1054, opened 2026-08-19
docs(book): the feature appendix has no row for tensor cores, and calls FP8 Planned

vyncint:
The page opens by promising "every compiler capability, runtime API, and hardware feature along with its current support status." Three gaps against that, in increasing size.

## 1. FP8 is listed as not implemented

Under **Not Yet Implemented**:

```
| FP8 / MX Data Types | **Planned** | Roadmap item for Blackwell.
                                      No architectural limitation. |
```

The catalog holds **158** entries for these formats — 12 conversions and 146 matrix ops — surfacing as `pub unsafe fn` in `cuda_device::convert`, `cuda_device::wmma` and `cuda_device::tcgen05`, and the `mma_mxf8f6f4` example compiles them. By the page's own legend that is **Partial**, not Planned.

The gap is real and worth naming rather than deleting the row: there is **no arithmetic** on these formats. Grouping the 158 by kind gives 12 `cvt` and 146 `mma`, and nothing matching add/mul/sub/fma/min/max/abs. So a value is carried as a packed bit pattern and converted before use, unlike `f16`/`bf16`, which have their own arithmetic modules. That is what the new row says.

## 2. The Warp section has no `redux.sync` row

`cuda_device::warp` exports sixteen of them; the section lists shuffle, vote and lane/warp ID. The new row carries both halves and their different floors: integers on sm_80+, and `f32` min/max with optional `.abs`/`.NaN` on `sm_100a`/`sm_100f`/`sm_103a`/`sm_103f` (#1032). `f64` has no form, which the row also says.

## 3. There is no matrix or tensor-core section at all

`grep -i` for "tensor core", "wgmma", "tcgen05", "wmma", "mma" or "TMEM" across the whole appendix returns **nothing**. No row for warp-level `mma.sync`, sparse MMA, Hopper warpgroup MMA, Blackwell tcgen05/TMEM, or the accumulator fragment algebra — in the feature matrix of a compiler whose headline example is a Blackwell GEMM. This adds a section for it, and moves the FP8 row there, since the matrix path is where those formats are used.

## Calibration and sourcing

Statuses follow the existing TMA section, the other Hopper+ hardware feature, which marks compile-only-in-CI capabilities **Full** with the architecture stated in the description.

Every row is sourced from the tree: module summaries from each module's own `//!` header (`wgmma` states Hopper `sm_90a`, `tcgen05` Blackwell sm_100+), sparse MMA from the `mma.sp.sync.aligned.*` strings in `generated/sparse_mma.rs`, and the redux names from `pub fn` declarations in `warp.rs`. Book builds clean under `sphinx-build -W`; `check-book-api-names.sh` passes.

### DIFF
--- a/cuda-oxide-book/appendix/supported-features.md
+++ b/cuda-oxide-book/appendix/supported-features.md
@@ -217,6 +217,7 @@ Anonymous promoted allocations remain unsupported.
 | Warp Shuffle Operations | **Full** | `shuffle`, `shuffle_xor`, `shuffle_down`, `shuffle_up`. Unsuffixed forms take `u32`; `_f32`, `_u64`, `_f64` variants and a `_sync` form of each. |
 | Warp Vote Operations | **Full** | `all(pred)`, `any(pred)`, `ballot(pred)` → bitmask. |
 | Lane/Warp ID | **Full** | `lane_id()` (0–31), `warp_id()`. Direct register reads. |
+| Warp Reduction (`redux.sync`) | **Full** | One-instruction full-warp reduction. Integers on sm_80+: `redux_sync_add`, `_min_{u32,i32}`, `_max_{u32,i32}`, `_and`, `_or`, `_xor`. `f32` min/max with optional `.abs` and `.NaN` on `sm_100a`/`sm_100f`/`sm_103a`/`sm_103f`. No `f64` form. |
 
 ## Runtime Library: Cooperative Groups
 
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified the new rows against the tree and found one real error: the wgmma row claimed "bf16/f16/tf32 accumulate" under Full, but tf32 is a compatibility stub the importer rejects (pending #1076), bf16/f16 are inputs with an f32 accumulator, and the lowering covers specific proven loop patterns (f16 straight-line only, per the wgmma.rs module header). A reader following the row would write a kernel the compiler rejects, the failure mode this PR was filed to prevent.
- **One commit layered on:** wgmma is now Partial with the gap named in the row, matching the page's own legend and this PR's FP8 row; plus sentence-level touch-ups on the FP8 and sparse MMA rows. Book build and api-names guard pass.
- Everything else checked out, including the deliberately narrow f32 redux arch list and stmatrix under tcgen05.

nihalpasham (approved):

PR #1055, opened 2026-08-19
ci: four tooling comments keep tree sizes by hand, and all four drifted in a day

vyncint:
#1032 added one example (`redux_f32`) and eight catalog entries. That single merge falsified numbers in four places:

| site | said | now |
|---|--:|--:|
| `check-example-license-policy.sh` (×3) | 217 locks | 218 |
| `check-example-license-policy.sh` (×2) | 215 examples | 216 |
| `check-example-license-policy.sh` | 13871 packages, floor 2170 | 13934, 2180 |
| `cargo-deny.yml` | 31 runs instead of 217 | … of 218 |
| `Justfile` | 986 catalog entries | 1010 |
| `examples-compile.yml` | 210 example crates | 216 |

**Six of those transcriptions are mine, from #1043 and #1044, written the day before.** Refreshing them buys one day. None is load-bearing — each is there to make a structural point — so state the structure and let the code report the numbers.

## What changes

- The license-policy guard already announced its representative count. It now announces the lock count beside it, so the grouping ratio is observable output rather than a comment to keep in step:

  ```
  Checking deny.toml over 31 representative example workspaces,
  grouped from 218 example lock files.
  ```

- Its scope comment and enumeration cross-check now say "one lock per example directory, plus one per nested sub-workspace" instead of naming today's totals.

- The parsed-package floor is computed as ten per lock, so the comment says that rather than quoting the current product.

- `cargo-deny.yml` says one run per distinct dependency set instead of one per lock file, and points at the script's own output.

- The `Justfile` comment's point was that the generated path dominates, which **35 generated op modules against 7 hand-written** makes without an entry count. Module counts are structural and did not move through any of these merges.

- `examples-compile.yml`'s number only justified a timeout, so "every example crate" does the same job.

## One number kept, and re-labelled

The `[a-m]*` glob illustration in the cross-check comment now reads "measured once … 137 of the 217 locks **then present**", matching the timestamped-measurement idiom the same file already uses two paragraphs above. The argument needs a concrete ratio; making it explicitly historical stops it being a live claim.

## Verification

The guard passes unchanged — 31 representatives grouped from 218 lock files, `deny.toml holds over every example workspace outside POLICY_EXEMPT_EXAMPLES`, exit 0. `bash -n` clean, `just --list` parses, all eleven workflows parse, and dialect-nvvm really does hold 35 generated and 7 hand-written op modules.

### DIFF
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -101,7 +101,8 @@ jobs:
 
       # Resolves each example workspace's own graph, so it needs the network but
       # no CUDA toolkit and no LLVM. One run per distinct dependency set rather
-      # than per workspace: 31 runs instead of 217, covering the same crates.
+      # than per workspace: one run per distinct dependency set instead of one
+      # per lock file, covering the same crates. The script prints both counts.
       - name: Run cargo-deny over the example workspaces
         run: bash scripts/check-example-license-policy.sh
 
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -101,8 +101,8 @@ jobs:
 
       # Resolves each example workspace's own graph, so it needs the network but
       # no CUDA toolkit and no LLVM. One run per distinct dependency set rather
-      # than per workspace: one run per distinct dependency set instead of one
-      # per lock file, covering the same crates. The script prints both counts.
+      # than one per lock file, covering the same crates. The script prints
+      # both counts.
       - name: Run cargo-deny over the example workspaces
         run: bash scripts/check-example-license-policy.sh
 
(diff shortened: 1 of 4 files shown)


nihalpasham:
- Verified the enumeration read-only: 31 representatives from 218 locks, both workflow YAMLs parse, and the Justfile / examples-compile hunks are correct as-is.
- One rewritten comment was false: "two orders of magnitude above ten packages per lock". The script's own counts (13934 packages over 218 locks) put it at about 64 per lock, roughly 6x the floor.
- **One commit layered on:** corrects that figure, removes the duplicated "one run per distinct dependency set" in cargo-deny.yml, reconnects "That is an equivalence" to the grouping it describes, and the echo now says "across N example lock files" since the two exempt locks are counted but never grouped.

nihalpasham (approved):

PR #1056, opened 2026-08-19
docs(cuda-device): the re-export table lists 10 of the 13 macros re-exported

vyncint:
`src/lib.rs` re-exports thirteen macros from `cuda-macros`:

```rust
pub use cuda_macros::{
    cluster_launch, constant, convergent, cooperative_launch, cuda_module,
    device, gpu_printf, kernel, launch_bounds, launch_contract, ptx_asm,
    pure, readonly,
};
```

The README's "Proc-Macro Re-exports" table lists ten. Missing are `#[cuda_module]`, `#[launch_contract]` and `#[constant]` — the entry point for typed module loading, the attribute that unlocks a safe launch, and the one that puts a static in constant memory.

Not obscure corners: **this same README already relies on two of the three.** Its `DisjointSlice` section uses `#[launch_contract(domain = 1, ...)]` in prose, and its module table documents the `constant` module. So the crate's own README uses macros it does not list.

Add the three rows, and tie the table's lead-in to the `pub use` list it mirrors so the next macro added there has an obvious place to land.

Descriptions are taken verbatim from the book's API quick reference, which already carries all three — so the two documents now say the same thing rather than two things.

**Verified:** the thirteen names in the `pub use` and the thirteen backticked rows in the table are the same set, with nothing in the table that is not re-exported.

### DIFF
--- a/crates/cuda-device/README.md
+++ b/crates/cuda-device/README.md
@@ -210,21 +210,21 @@ These are defined in `cuda-macros` and re-exported from `cuda-device` for
 convenience. The table is the whole `pub use cuda_macros::{...}` list in
 `src/lib.rs`:
 
-| Attribute                | Purpose                                       |
-|--------------------------|-----------------------------------------------|
-| `#[kernel]`              | Mark a function as a GPU kernel entry point   |
-| `#[device]`              | Mark a helper function or extern block        |
-| `#[cuda_module]`         | Collect a module's kernels into a typed host module with `load` and launchers |
-| `#[launch_bounds]`       | Set max threads / min blocks per SM           |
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified the table against `src/lib.rs`: with the three added rows it is now the whole 13-entry `pub use cuda_macros::{...}` list, and each new description matches the macro's behavior.
- **One commit layered on:** "launch shape" is now "launch requirements" (the contract also carries `dynamic_shared` and `min_compute_capability`, so it is more than dimensions), the dangling "with a host `set_<name>`" now says it is a setter, and the raw-text pipes are re-aligned.

nihalpasham (approved):

PR #1057, opened 2026-08-19
docs(cuda-macros): the README omits `#[constant]`, and its only derive

vyncint:
`cuda-macros` exports sixteen macros. Its 500-line README documents fourteen. The two it never mentions are `#[constant]` and `#[derive(DeviceCopy)]` — and the second is the crate's only derive macro.

## The omissions are connected

The "Source Layout" tree lists three files under `src/`:

```text
├── lib.rs       # All proc-macro definitions (kernel, device, launch, etc.)
├── printf.rs    # gpu_printf! implementation
└── ptx_asm.rs   # ptx_asm! implementation
```

There are four. **`device_copy.rs` — the derive's whole implementation — is not in the tree either**, so nothing in the README points at the macro or at the file that defines it.

## What each section says

`#[constant]` gets the three restrictions that decide whether it works, from its own rustdoc: the initializer must be `ConstantMemory::UNINIT` (non-zero initializers are not honored yet), it must appear inside a `#[cuda_module]` — elsewhere it silently produces an unreachable symbol with no setter, which is the failure mode worth warning about — and the identifier must not start with the reserved prefix. Both generated setters are named, with the stream-ordered one marked as the one to prefer.

`#[derive(DeviceCopy)]` gets what makes it more than `Copy`: the trait is `unsafe` and promises every bit pattern is valid, which `DeviceBuffer` relies on when it turns raw device bytes back into values. It also records that structs and unions are accepted and enums are rejected with a diagnostic, since that is the first thing someone tries.

## Verification

Against the tree:

- The sixteen exported macros — eleven `#[proc_macro_attribute]`, four `#[proc_macro]`, one `#[proc_macro_derive]` — are now all named in the README, and the four files under `src/` are all in the tree.
- The `#[constant]` snippet is the shape `examples/constant_memory_coeffs/src/main.rs` already uses, line for line.
- `DeviceCopy`'s `Copy` bound, its `bool`/`char`/`NonZeroU32` rationale and its enum rejection come from `cuda-core/src/device_buffer.rs` and `cuda-macros/src/device_copy.rs`; the derive example matches the `#[derive(Copy, Clone, DeviceCopy)]` form in `cuda-core`'s trybuild fixtures.

Documentation only — no source file changes, so `#[constant]`'s and the derive's behaviour is untouched.

### DIFF
--- a/crates/cuda-macros/README.md
+++ b/crates/cuda-macros/README.md
@@ -235,6 +235,37 @@ pub fn grid_sync_kernel(mut out: DisjointSlice<u32>) {
 
 Semantic markers for the codegen backend (pass-through -- no code transformation).
 
+### `#[constant]` -- Constant-Memory Statics
+
+Marks a module-scope `static` as a CUDA constant-memory global. The static must
+be typed `ConstantMemory<T>`, whose `UnsafeCell<T>` semantics on the device stop
+the compiler constant-folding the initializer, so the host's `cuMemcpyHtoD`
+writes stay observable from kernels.
+
+```rust
+#[cuda_module]
+mod kernels {
+    #[constant]
+    static COEFFS: ConstantMemory<[f32; 4]> = ConstantMemory::UNINIT;
+}
+```
+
+The macro attaches a reserved `export_name` so the PTX symbol is resolvable via
+`cuModuleGetGlobal`, and the host-side `#[cuda_module]` expansion generates the
+setters:
+
+- `module.set_<name>(&stream, &value)` -- stream-ordered async write, which
+  orders correctly against surrounding launches. Prefer this one.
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified the new sections against the macro's impl: the two setters, the `#[cuda_module]` requirement, the prefix check, and the enum rejection are all as documented.
- **One commit layered on:** the three restrictions are a bullet list instead of one 60-word sentence, and the first now matches lib.rs (any all-zeros initializer works, not `UNINIT` alone); the "product type / discriminated union" rationale is in plain language; "byte pattern" and "bit pattern" are one term; the serde lead-in reads directly.

nihalpasham (approved):

PR #1059, opened 2026-08-19
fix(mir-importer): materialize relocation-free pointer/integer unions

### LINKED ISSUE #1058: Support relocation-free pointer/integer union constants via byte images
## Summary

Pointer/integer union constants are currently rejected because a byte-only
representation cannot preserve pointer provenance when the evaluated allocation
contains a pointer relocation.

There is, however, a narrower case where provenance is not involved: a
pointer/integer union whose evaluated storage contains no overlapping
relocations already has a complete byte representation. One common example is
a union initialized through its full-width integer alternative.

In that case, rustc's evaluated allocation already contains the complete storage
representation, so the union can be materialized using the existing byte-image
path.

## Proposed scope

Support relocation-free mixed pointer/integer unions when:

- the union storage is exactly one pointer word,
- the union has natural pointer alignment,
- all pointer alternatives are compatible thin pointers,
- all integer alternatives are full pointer-width integers, and
- no relocation overlaps the union storage.

For example:

```rust
#[repr(C)]
union PointerOrBits {
    pointer: *const u32,
    bits: u64,
}

const VALUE: PointerOrBits = PointerOrBits {
    bits: 0x1122_3344_5566_7788,
};
```

On a 64-bit target, this constant has no pointer relocation. Its evaluated byte
image is therefore sufficient to reconstruct the union storage exactly.

## Non-goals

This should remain fail-closed for:

* pointer/integer unions containing an actual pointer relocation,
* fat pointers,
* nested pointer aggregates,
* partial-width integer alternatives,
* mixed pointer address spaces,
* padded or over-aligned pointer unions, and
* mixed pointer/integer device-global union initializers.

Thin-pointer-only unions with relocations should continue to use the existing
provenance-preserving pointer carrier path.

## Implementation direction

Keep the existing provenance-aware union classifier unchanged.

Before invoking it during allocation-backed union constant translation, detect
the relocation-free pointer/integer shape described above and route it through
the existing byte-image materializer.

This keeps relocation-bearing behavior unchanged while admitting the case where
the evaluated bytes are the complete storage truth.

## Regression coverage

Add coverage for:

* direct relocation-free pointer/integer union constants,
* `[U; N]` relocation-free pointer/integer union constants,
* optimized and non-optimized GPU execution,
* rejection of pointer-width mismatches and partial-width integer alternatives,
* LLVM code shape requiring byte-image materialization, and
* absence of `inttoptr` reconstruction.

This is a follow-up to the pointer-union constant work in #927.

uurl:
Closes #1058 

## Summary

Allow a narrow class of pointer/integer union constants to use the existing byte-image materialization path when their evaluated allocation contains no relocation provenance.

Previously, `classify_union_constant_storage` rejected all pointer/integer overlap because a byte image cannot represent both arbitrary integer bits and pointer provenance.

That remains correct for relocation-bearing constants. However, when the evaluated union storage contains no relocations—for example, when initialized through a full-width integer alternative—rustc's evaluated byte image is already the complete storage truth.

## What changed

`translate_union_constant_from_alloc` now recognizes relocation-free mixed pointer/integer unions before invoking the provenance-aware classifier.

The byte-image exception is intentionally limited to unions where:

- storage size is exactly one target pointer word,
- ABI alignment is exactly one target pointer word,
- non-ZST pointer alternatives are thin pointers,
- pointer alternatives use the same address space,
- integer alternatives are exactly pointer-width, and
- no relocation overlaps the union storage.

The existing `classify_union_constant_storage` behavior is unchanged.

This is intentional: relocation-bearing pointer/integer unions and the device-static union validation path continue to use the existing strict provenance rules.

## Still rejected

This change does not admit:

- relocation-bearing pointer/integer unions,
- fat pointers,
- nested pointer aggregates,
- partial-width integer alternatives,
- mixed pointer address spaces,
- padded or over-aligned pointer unions, or
- mixed pointer/integer device-global union initializers.

Thin-pointer-only unions carrying relocations continue to use the existing typed pointer carrier.

## Tests

Extended `array_constants` with direct and runtime-indexed `[U; N]` pointer/integer union constants initialized through their integer alternatives.

The LLVM code-shape regression verifies that these constants:

- materialize through an `[8 x i8]` byte image on the current 64-bit target, and
- do not reconstruct a pointer using `inttoptr`.

Existing pointer-only union checks continue to verify the provenance-preserving representation.

Validated locally with:

```text
cargo fmt --all -- --check
cargo test -p mir-importer --lib
scripts/smoketest.sh --only '^array_constants$' --no-color --keep-logs
bash crates/rustc-codegen-cuda/examples/array_constants/verify-code-shape.sh
git diff --check
```

`mir-importer` result:

```text
1101 passed; 0 failed
```

The optimized and non-optimized `array_constants` runs, focused smoketest, and LLVM code-shape verification all pass.

## Documentation

Update `supported-features.md` to distinguish:

* relocation-free pointer/integer unions, which may use a byte image,
* relocation-bearing pointer/integer unions, which remain unsupported, and
* thin-pointer-only device-global unions, which may preserve the supported single relocation at byte zero.

Follow-up to #927.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -8417,6 +8417,74 @@ fn classify_union_constant_storage(
     })
 }
 
+/// Whether a mixed thin-pointer/integer union can use its evaluated byte image
+/// when the allocation carries no relocation provenance.
+///
+/// Keep this deliberately narrower than [`classify_union_constant_storage`]:
+/// the ordinary classifier remains the provenance-preserving authority used by
+/// pointer-bearing constants and device-static union initializers. This helper
+/// only admits one naturally aligned pointer word whose non-ZST alternatives
+/// are compatible thin pointers or full-width integers.
+fn relocation_free_pointer_integer_union_uses_byte_image(
+    ctx: &Context,
+    union_ty: TypeHandle,
+    pointer_width: usize,
+) -> bool {
+    let (union_size, union_align, field_types) = {
+        let ty_ref = union_ty.deref(ctx);
+        let Some(union_ty) = ty_ref.downcast_ref::<dialect_mir::types::MirUnionType>() else {
+            return false;
+        };
+        (
+            union_ty.total_size(),
+            union_ty.abi_align(),
+            union_ty.field_types().to_vec(),
+        )
+    };
+
+    let pointer_width_u64 = pointer_width as u64;
+    if union_size != pointer_width_u64 || union_align != pointer_width_u64 {
+        return false;
+    }
+
+    let integer_width = (pointer_width * 8) as u32;
+    let mut pointer_address_space = None;
+    let mut saw_pointer = false;
+    let mut saw_integer = false;
+
+    for field_ty in field_types {
+        if types::is_zst_type(ctx, field_ty) {
+            continue;
+        }
+
+        let field_ref = field_ty.deref(ctx);
+        if let Some(pointer) = field_ref.downcast_ref::<dialect_mir::types::MirPtrType>() {
+            match pointer_address_space {
+                None => pointer_address_space = Some(pointer.address_space),
+                Some(address_space) if address_space == pointer.address_space => {}
+                Some(_) => return false,
+            }
+            saw_pointer = true;
+            continue;
+        }
+
+        if field_ref
+            .downcast_ref::<IntegerType>()
+            .is_some_and(|integer| integer.width() == integer_width)
+        {
+            saw_integer = true;
+            continue;
+        }
+
+        // Fat pointers, nested pointer aggregates, partial-width integers, and
+        // unrelated scalar/aggregate alternatives keep the existing fail-closed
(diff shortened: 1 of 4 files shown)


nihalpasham:
- Verified end to end: mir-importer 1101 tests green, `cargo oxide run array_constants` passes on an RTX 5090 in both the default and `CUDA_OXIDE_NO_OPT=1` pipelines, `verify-code-shape.sh` passes, per-example fmt/clippy clean, 95 workspace suites green, full compile-only smoketest 216/216.
- The gate is fail-closed on every axis attacked: it sits after the range-scoped relocation scan and refuses any union whose storage is not one naturally aligned full-width pointer word, so pointer-only and pointer-bearing unions cannot be rerouted through the byte image, and relocations straddling into the union's byte range are caught by the overlap predicate.
- The book paragraph matches the implemented boundary.

nihalpasham (approved):

Issue #883, opened 2026-08-14
mir-lower: fat-pointer casts can round-trip raw shared-memory bytes (emit_pointer_cast lacks the transmute guard)


nihalpasham:
## The gap

The transmute path refuses to let shared-memory (`addrspace(3)`) pointer
bytes travel through the stack-slot round trip, for good reason. But
`emit_pointer_cast` (`crates/mir-lower/src/convert/ops/cast.rs`, the
dispatches around lines 484 and 501) sends non-transmute casts of
aggregate-valued pointers down that same road with no shared-pointer check.
The reachable case is fat pointers, because a fat pointer is a two-field
value `(data ptr, len)` and two-field values take the memory path:

```rust
// s points into shared memory
let a: *mut [f32] = s;
let b = a as *mut [u32];   // (data ptr, len) round-trips through a
                           // scratch stack slot, raw addrspace(3)
                           // bytes and all
```

## Why those raw bytes are dangerous

Two concrete hazards, both the same ones the transmute guard exists for:

```text
hazard 1: physical width depends on output mode
  llc / legacy NVVM:    shared ptr = 8 bytes
  modern NVVM (p3:32):  shared ptr = 4 bytes, while Rust sizes every
                        pointer at 8, so the store/load bytes disagree

hazard 2: the raw bytes are an offset, not an address
  first shared allocation:  raw shared bytes = 0x0000  (looks like null!)
  what Rust code should see: the 64-bit generic address (never 0)
```

Unlike the DST gaps, this one is silent: it produces bytes, not an error.
Silent-wrong is the worst category, even though the trigger pattern is
unusual. Pre-existing on main; surfaced during the #879 review.

## What to fix

Give `emit_pointer_cast`'s memory-round-trip dispatch the same policy the
transmute path has: if either side contains an `addrspace(3)` pointer
(`llvm_type_contains_pointer_in_address_space` already answers this,
recursing through structs and arrays), reject loudly with the same message
style, naming both types and the remedy (convert the scalar pointer; it
bridges through its 64-bit generic address). If a sound bridging exists for
the fat case (addrspacecast the data-pointer half, copy the metadata half),
that is a fine follow-up, but loud rejection closes the silent-wrong hole
today.

## Acceptance criteria

- The fat-pointer repro above fails at compile time with the located,
  actionable diagnostic instead of producing bytes.
- Unit tests next to the existing transmute rejection tests, covering
[...]

PR #1061, opened 2026-08-20
fix(mir-lower): legalize address-space 3 slice fat-pointer casts in SSA

### LINKED ISSUE #1060: mir-lower: legalize AS3 slice fat-pointer casts in SSA
## Context

#883 identifies a correctness gap in `emit_pointer_cast`: aggregate-valued
pointer casts can currently fall through to the memory transmute path, which
may round-trip raw `addrspace(3)` pointer bytes through a stack slot.

For slice fat pointers, this can be legalized without observing the physical
representation of the shared pointer.

A slice pointer has the semantic shape:

```text
{ data pointer, length }
```

The data pointer can be converted explicitly between address spaces while the
metadata is preserved unchanged.

## Proposed lowering

For semantic slice fat-pointer casts:

```text
{ src_data, len }
    |
    | extractvalue
    v
src_data
    |
    | addrspacecast, when address spaces differ
    v
dst_data

result = { dst_data, len }
```

The aggregate should be reconstructed directly in SSA using
`extractvalue` / `insertvalue`.

The lowering must not use:

```text
alloca
store
load
ptrtoint
inttoptr
```

for the fat-pointer representation conversion.

## Scope

Keep this legalization intentionally narrow:

* semantic pointer casts only
* slice-shaped fat pointers only
* exactly two fields
* field 0 is a pointer
* field 1 is integer metadata
* source and destination metadata types must match
* preserve existing behavior for arbitrary aggregates
* do not extend aggregate `Transmute`
* do not generalize this to cluster-shared `addrspace(7)`

Trait-object-like `{ ptr, ptr }` values must not enter this path.

## Acceptance criteria

The lowering should:

* legalize shared-to-generic slice fat-pointer casts
* legalize generic-to-shared slice fat-pointer casts
* handle same-address-space slice casts without a memory round-trip
* preserve slice metadata exactly
* emit an `addrspacecast` only for the data-pointer field when required
* reconstruct the destination aggregate in SSA
* emit no temporary `alloca` / `store` / `load` sequence

Add lowering tests covering the SSA shape and extend the
`shared_ptr_transmute` GPU regression to validate that:

* the slice length is preserved
* the resulting data pointer denotes the original shared allocation
* stores through the recast slice are visible through the original shared storage

Related: #883

uurl:
Closes #1060 

## Summary

Legalize semantic slice fat-pointer casts in SSA, including casts involving shared-memory `addrspace(3)` pointers, without round-tripping the aggregate through memory.

Instead of treating the fat pointer as raw bytes, the lowering now handles its two semantic components independently:

```text
{ data pointer, metadata }
        |
        +-- extract data pointer
        |       |
        |       +-- addrspacecast when required
        |
        +-- extract metadata unchanged
        |
        +-- reconstruct destination aggregate in SSA
```

This implements the sound fat-pointer bridge described in #883.

## Why

The generic aggregate cast fallback uses an `alloca` + `store` + `load` round-trip.

That is unsafe for aggregates containing shared-memory pointers because the physical representation of an `addrspace(3)` pointer is target-mode dependent and its raw shared-memory value is not the Rust-visible generic address.

A slice fat pointer does not require byte reinterpretation. Its data pointer can be converted semantically with `addrspacecast`, while its length metadata can be copied unchanged.

## Implementation

`emit_pointer_cast` now recognizes the narrow LLVM representation used by slice fat pointers:

```text
{ pointer, integer metadata }
```

When both source and destination match that shape and their metadata types match, the lowering:

1. extracts the data pointer
2. extracts the metadata
3. emits `addrspacecast` for the data pointer when the address spaces differ
4. reconstructs the destination aggregate with `insertvalue`

No temporary memory representation is created.

Other aggregate casts continue to use the existing fallback behavior.

The legalization is intentionally restricted to the slice-shaped `{ pointer, integer metadata }` representation within the semantic pointer-cast path.

It does not extend to:

* other aggregate shapes
* trait-object-like `{ ptr, ptr }` aggregates
* aggregate `Transmute`
* cluster-shared `addrspace(7)`

## Tests

Added `mir-lower` coverage for:

* shared/generic slice fat-pointer casts
* generic/generic slice fat-pointer casts
* metadata preservation
* SSA reconstruction
* absence of `alloca`, `store`, and `load`
* required `addrspacecast` behavior

Extended `shared_ptr_transmute` with an end-to-end GPU regression that checks:

* slice length preservation
* shared data-pointer identity
* write-through visibility after the fat-pointer cast

Validation performed:

```text
cargo oxide fmt --check
    PASS

git diff --check
    PASS

cargo test -p mir-lower
    255 unit tests passed
    117 lowering tests passed

    PASS

cargo oxide run shared_ptr_transmute
    PASS
    sum=528
    scalar_readback=0xC0DE
    slice_readback=0x1234

cargo oxide sanitize shared_ptr_transmute --tool memcheck
    0 errors

cargo oxide sanitize shared_ptr_transmute --tool racecheck
    0 hazards
```

Refs #883

### DIFF
--- a/crates/mir-lower/src/convert/ops/cast.rs
+++ b/crates/mir-lower/src/convert/ops/cast.rs
@@ -39,7 +39,8 @@
 //! | ptr → struct (thin→fat, no niche)              | `insertvalue` into undef                    |
 //! | ptr → integer                                  | genericize named-space pointers, `ptrtoint` |
 //! | integer → ptr                                  | `inttoptr`                                  |
-//! | struct → struct (transmute)                    | `alloca` + `store` + `load`                 |
+//! | slice-shaped struct → slice-shaped struct     | fieldwise SSA; `addrspacecast` data as needed |
+//! | other struct → struct                          | `alloca` + `store` + `load`                 |
 //! | ptr → ptr (diff addrspace)                     | `addrspacecast`                             |
 //! | struct → integer, equal size                   | `alloca` + `store` + `load`                 |
 //! | struct → integer, mismatched size              | cuda-oxide error (see issue #21)            |
@@ -58,7 +59,7 @@ use crate::convert::types::{
 use crate::helpers;
 use dialect_mir::attributes::MirCastKindAttr;
 use dialect_mir::ops::MirCastOp;
-use dialect_mir::types::{MirArrayType, MirPtrType};
+use dialect_mir::types::{MirArrayType, MirPtrType, address_space};
 use llvm_export::op_interfaces::{CastOpInterface, CastOpWithNNegInterface};
 use llvm_export::ops as llvm;
 use llvm_export::types::{FuncType, PointerType, PointerTypeExt};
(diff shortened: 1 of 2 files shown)


nihalpasham:
- Verified end to end: mir-lower 255+117 tests green, clippy clean, `cargo oxide run shared_ptr_transmute` passes on an RTX 5090 (sum=528, scalar_readback=0xC0DE, slice_readback=0x1234), per-example fmt/clippy clean.
- Reachability holds: `Transmute` still routes to `emit_transmute` with its own AS3 rejections, so the `{ ptr, integer }` recognizer only fires for PtrToPtr/coercion/Subtype fat-pointer casts with identical metadata, and AS7 bails out. No way to mis-fire it on a struct-shaped non-fat value.

nihalpasham (approved):

PR #1064, opened 2026-08-20
fix(mir-lower): support recursive packed address-space 3 internal ABI

### LINKED ISSUE #1063: Support recursive and multi-leaf packed AS3 aggregates across the internal device ABI
## Problem

The packed-AS3 internal device ABI currently supports only a narrow shape: one direct shared-memory pointer leaf inside a byte-faithful packed struct.

The target-stable storage machinery is already recursive and can represent broader aggregate shapes by replacing AS3 pointer leaves with generic pointers at the physical ABI boundary. However, the admission logic still rejects:

- multiple direct AS3 pointer leaves;
- AS3 pointer leaves nested inside structs;
- AS3 pointer leaves nested inside tuples.

This prevents otherwise representable packed aggregates from being returned by value from internal device functions, which also blocks source-level call/return flows that bounce such values through device helpers.

## Proposed scope

Extend the packed-AS3 internal ABI admission logic to support:

- multiple direct AS3 pointer leaves;
- recursively nested AS3 leaves through structs;
- recursively nested AS3 leaves through tuples.

The semantic value inside the function should continue to use the original packed aggregate with AS3 pointers. Only the physical internal ABI carrier should recursively replace AS3 leaves with generic pointers.

## Out of scope

Keep the following fail-closed:

- arrays containing AS3 pointer leaves;
- vectors containing AS3 pointer leaves;
- kernel-entry ABI changes;
- whole-value packed AS3 loads/stores;
- packed-AS3 field projections that require local materialization;
- enums and other unrelated aggregate kinds.

Arrays should be handled separately with an explicit bounded expansion policy.

## Acceptance criteria

- A packed struct with two or more direct AS3 pointer leaves can be returned by value from an internal device function.
- A packed struct containing a nested struct with AS3 leaves can be returned by value from an internal device function.
- A packed struct containing a nested tuple with AS3 leaves can be returned by value from an internal device function.
- The physical ABI carrier contains generic-pointer replacements for every admitted AS3 leaf.
- The semantic SSA aggregate continues to contain AS3 pointers.
- Arrays and vectors remain rejected.
- Existing packed-AS3 and shared-memory regressions continue to pass.

uurl:
Closes #1063 

## Summary

Generalize the packed-AS3 internal device ABI support introduced by #1001 from a single direct shared-pointer leaf to recursive and multi-leaf aggregate shapes.

The internal ABI can now carry packed structs containing:

- multiple direct AS3 pointer leaves;
- AS3 pointer leaves nested inside structs;
- AS3 pointer leaves nested inside tuples.

The function body still operates on the semantic packed aggregate containing AS3 pointers. Internal device arguments continue to use the existing flattened argument ABI. At the return boundary, the existing target-stable storage machinery recursively replaces every admitted AS3 leaf with a generic pointer.

## Implementation

This change widens `packed_shared_internal_abi_info` without duplicating the existing recursive storage/coercion machinery.

The classifier now:

- admits scalar and pointer leaves;
- recursively traverses MIR structs and tuples;
- accepts one or more AS3 leaves;
- continues to require a byte-faithful packed root struct;
- rejects arrays and vectors for this lane.

`target_stable_storage_type` remains responsible for the actual recursive AS3 -> generic storage rewrite.

The packed aggregate construction path continues to keep semantic AS3 pointers in SSA and uses the widened classifier only as the admission gate.

## Tests

Added coverage for:

- multiple direct AS3 leaves;
- nested struct AS3 leaves;
- nested tuple AS3 leaves;
- arrays remaining out of scope;
- semantic packed SSA construction with multiple direct leaves;
- semantic packed SSA construction with nested leaves;
- end-to-end device call/return coverage for multi-leaf and nested packed aggregates.

Validation performed:

- `git diff --check`
- `cargo test -p mir-lower packed_shared_internal_abi -- --nocapture`
- `cargo test -p mir-lower packed_struct_construction_with -- --nocapture`
- `cargo test -p mir-lower --all-targets`
- `cargo oxide build packed_aggregate_abi`
- `cargo oxide run packed_aggregate_abi`
- `cargo oxide run addressof_sharedarray`

`mir-lower` passes 257 unit tests and 117 lowering tests.

The end-to-end `packed_aggregate_abi` regression passes on `sm_120a`, including recursive and multi-leaf packed shared internal ABI coverage.

## Out of scope

- AS3 pointer arrays;
- AS3 pointer vectors;
- whole-value packed AS3 memory loads/stores;
- packed-AS3 field projections requiring local materialization;
- kernel-entry ABI changes.

Those remain separate follow-up work.

### DIFF
--- a/crates/mir-lower/src/convert/ops/aggregate.rs
+++ b/crates/mir-lower/src/convert/ops/aggregate.rs
@@ -464,9 +464,10 @@ pub(crate) fn convert_construct_struct(
     }
 
     // A packed struct containing AS3 remains target-dependent as a physical
-    // memory image. The narrow internal-call ABI exception is safe to construct
-    // in SSA because its return boundary rebuilds the value into a target-stable
-    // generic-pointer carrier before the aggregate crosses the function ABI.
+    // memory image. The internal-call ABI exception is safe to construct in SSA
+    // because its return boundary recursively rebuilds every supported AS3 leaf
+    // into a target-stable generic-pointer carrier before the aggregate crosses
+    // the function ABI.
     if llvm_packed_struct_contains_pointer_in_address_space(
         ctx,
         map.llvm_struct_ty,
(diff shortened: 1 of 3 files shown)


nihalpasham:
- Verified end to end: mir-lower 262 lib + 117 integration tests green, clippy/fmt clean, `cargo oxide run packed_aggregate_abi` passes on an RTX 5090 (cc 12.0), full compile-only smoketest 216/216. Also exercised the modern NVVM path this PR had not: built with `--emit-nvvm-ir --arch=sm_120` and compiled the result through libNVVM + nvJitLink to a cubin, confirming the multi-leaf/nested carrier is accepted on the data layout where the 32-bit AS3 vs 64-bit generic divergence actually exists (internal call returns `<{ i8, ptr, ptr }>` while the semantic type keeps `ptr addrspace(3)`).
- **Two commits pushed to this branch:**
  - the new MIR-shape predicate dropped the ZST skip the old post-conversion field scan had, so a packed struct carrying a zero-sized non-struct field (e.g. a `[u8; 0]` marker) fell out of the ABI lane and hit the fail-closed construction error. Added a MIR-side zero-sized check mirroring `is_zero_sized_type` plus a classifier test.
  - rustfmt pass on the example.

nihalpasham (approved):

PR #1065, opened 2026-08-20
feat: add schedule perturbation campaign

drbh:
this pr adds a schedule perturbation campaign for existing cuda-oxide examples.

the campaign builds an example once, inserts deterministic `nanosleep.u32` delays at structural synchronization sites in the generated PTX (atomics, mbarriers, fences, ordered loads/stores, warp collectives, loop back-edges), patches a copy of the executable's embedded artifact, and runs each variant behind a watchdog. the code and inputs never change; only warp timing does, and every mutation is a pure function of the seed so a finding replays exactly.

```bash
GEMM_SOL_PHASE=4b-correctness cargo oxide fuzz-schedule gemm_sol \
  --arch sm_100a --seeds 56..57 --focus mbarrier --confirm-runs 3 --timeout-secs 20
```

outputs

```text
schedule-fuzz: baseline Pass
schedule-fuzz: FINDING seed=56 - SCHEDULE-SENSITIVE CORRECTNESS FAILURE
  confirmation: 3/3 reproductions (CONFIRMED)
RESULT: FOUND 1 CANDIDATE(S)
```

that finding is a real race: on 1x B200 (`sm_100a`), sweeping seeds 0..140 with `--focus mbarrier` surfaced a `TILE_INFO` mailbox race in the gemm_sol phase 4B kernel: the producer warp can republish tile coordinates before all five reader warps have copied them, leaving 264 of 1024 output tiles unwritten. seeds 56 and 135 each reproduce it 3/3 against a passing pristine baseline. the fix (and full failure output) is in #1066

findings are reported as schedule-sensitive outcomes rather than proof of a data race or deadlock: `CORRECTNESS FAILURE`, `OUTPUT CHANGE` (with `--compare-output`), `TIMEOUT`, `CRASH`, or `GPU WEDGE` candidates. a finding is rerun up to `--confirm-runs` times, and a timeout is rechecked with pristine PTX so a wedged device is not misreported. mutated PTX, decisions, logs, `replay.sh`, `sites.json`, and `summary.json` land under `crates/fuzzer/artifacts/schedule/<example>/`; the standalone `ptx-schedule` binary can also `--list-sites` or rewrite a PTX file directly. details are in `crates/fuzzer/README.md`.

`cargo test -p ptx-schedule` passes (7 tests) and `cargo fmt --all -- --check` is clean.

please let me know if I should make any changes or adjust the approach. Thanks!


nihalpasham:
- Ran the campaign end to end on an RTX 5090: vecadd and cuda_module_in_lib baselines pass with zero findings, barrier gets real injections (7 then 11 of its 14 sites) and passes, per-seed artifacts plus sites.json and summary.json all written, and the objcopy-patched executables load and run.
- **One commit layered on:** wires the four repo gates the branch missed (SPDX header on campaign.rs, both crate inventories, the CLI doc table, the unit-tests matrix plus Justfile parity); rejects `--materialize-cubin` for fuzz-schedule since it was accepted but never forwarded, and a cubin-materialized bundle has no PTX payload to patch; drains child stdout/stderr concurrently in run_binary (output past the pipe buffer wedged the child into a false Hang); records per-seed perturb/patch errors without killing the campaign or summary.json; and makes find_executable search all packages' bin targets, since `packages.first()` picked the kernel lib in every lib+bin example workspace and cuda_module_in_lib errored before running. Tests grow 7 to 9; clippy, fmt, gate scripts, and the book build are all clean.
- One question: the lib.rs test comment references a "recovered Python classifier" that exists nowhere in the repo. Out-of-tree prototype? A reword or a pointer would help the next reader.
- Noted for follow-up, not blocking: two classified sites inside one inline-asm block would map to the same insertion offset, and EditScript treats same-offset inserts as conflicts (latent today; no checked-in PTX trips it); replay.sh hardcodes the GEMM_SOL_PHASE env capture.

nihalpasham (approved):

PR #1066, opened 2026-08-20
fix: gemm sol phase4b 

drbh:
this pr fixes a tile-metadata race in the gemm_sol phase 4B kernel (`gemm_sol_clc`), found by the schedule perturbation campaign in this pr #1065

**note: this change set looks bigger than it is because it includes the the commit from #1065, if the schedule perturbation campaign commit is merged, this pr only updates a single file (the gemm_sol file) as seen in my forks version of this pr https://github.com/drbh/cuda-oxide/pull/3

the producer warp publishes `{tile_m, tile_n, has_work}` through the shared `TILE_INFO` mailbox, signals `TILE_READY`, and then reuses the mailbox for the next stolen tile. the MMA warp and four epilogue warps wait on `TILE_READY` and read the mailbox, but nothing acknowledged that all five readers had copied the current values before the producer overwrote them, so a different warp schedule could pair epilogue output with the wrong tile coordinates.

the fix adds a `TILE_INFO_FREE` mbarrier with an arrival count of 5:

- each reader warp copies the mailbox after its `TILE_READY` wait, syncs its lanes, and lane 0 arrives on `TILE_INFO_FREE`
- the producer waits on `TILE_INFO_FREE` parity before republishing `TILE_INFO` (including the final `has_work = 0` publish)

### before and after

on 1x B200 (`sm_100a`), with `GEMM_SOL_PHASE=4b-correctness` isolating the 4096x4096x4096 phase 4B validator, the campaign surfaced the race on the unfixed kernel at seeds 56 and 135 of a 0..140 `--focus mbarrier` sweep, each reproduced 3/3:

```bash
GEMM_SOL_PHASE=4b-correctness cargo oxide fuzz-schedule gemm_sol \
  --arch sm_100a --seeds 56..57 --focus mbarrier --confirm-runs 3 --timeout-secs 20
```

on the unfixed kernel this outputs

```text
schedule-fuzz: baseline Pass
schedule-fuzz: FINDING seed=56 - SCHEDULE-SENSITIVE CORRECTNESS FAILURE
  confirmation: 3/3 reproductions (CONFIRMED)
RESULT: FOUND 1 CANDIDATE(S)
```

and the example's own validator reports, identically on all three executions:

```text
  C[2048,2048] =          0  (expected       4096)  FAIL
  Zero tiles: 264 / 1024
  Row 0 sum: 37748736 (expected 75497472) FAIL
FAILED: gemm_sol_clc 4096x4096x4096
```

264 of 1024 output tiles were never written: epilogue warps woke from `TILE_READY`, were delayed past the producer republishing `TILE_INFO`, and drained their TMEM stage onto another tile's coordinates. the pristine baseline passes; only the injected timing differs.

with the fix, the same command passes and the full sweep is clean:

```text
schedule-fuzz: baseline Pass
schedule-fuzz: seed=56 PASS sites=262/502

=== schedule-fuzz result ===
example: gemm_sol
baseline: PASS
variants: 140
RESULT: no schedule-sensitive failures found
```

### performance

the added wait is once per tile, off the K-loop, so `GEMM_SOL_PHASE=4b` throughput is unchanged within noise: 502.8 -> 501.5 TFLOPS at 4096, 497.3 -> 497.3 at 8192, 484.2 -> 481.8 at 16384.


nihalpasham (changes requested):
- The fix itself checks out line by line: TILE_INFO_FREE's 5 arrivals match the 5 reader warps (MMA plus four epilogue), the first publish is deliberately unguarded, the terminal has_work=0 publish waits without toggling, and cleanup invalidation sits behind the full block+cluster sync. The pre-fix race is real: the epilogue's coordinate reads lag the producer by a full tile with no acknowledgment.
- One structural blocker: this branch carries #1065's commit byte-for-byte, so merging it as-is would land the whole ptx-schedule crate under a fix title. Once #1065 merges, rebase (or say the word and we'll rebase for you) so the diff shrinks to the single gemm_sol file, then this goes in.
- Our host compile-gates sm_100a only, so the B200 evidence (seeds 56 and 135 fixed, TFLOPS within noise) stays author-attested; acceptable for a correctness fix.
- Worth a follow-up: phases 4A, 4C, and 4D use the identical mailbox pattern and were not swept.

nihalpasham:
- Superseded by the formal [changes-requested review](https://github.com/NVlabs/cuda-oxide/pull/1066#pullrequestreview-5001746251) on the current head.

nihalpasham (changes requested):
- **Blocker:** the five-reader handshake fixes the Phase 4B TILE_INFO reuse, but the same producer/consumer race is still present in Phases 4A, 4C, and 4D.
- **Where:** Phase 4A publishes at `1624–1630` and reads at `1724–1727` / `1839–1849`; Phase 4C publishes at `2763–2766` and `2907–2920` and reads at `3031–3034` / `3141–3150`; Phase 4D publishes at `3460–3463` and `3573–3602` and reads at `3696–3699` / `3848–3857`.
- **Required fix:** give every phase the same explicit publish/read/ack ownership rule, then run per-phase B200 sweeps that exercise wraparound and the terminal tile. Compile-only checks cannot rule out a parity deadlock.
- **DCO:** amend the contributor commit with a `Signed-off-by` trailer.
- **Local validation:** fmt, clippy, compile-only smoke, LLVM inspection, and `ptxas -arch sm_100a` pass for Phase 4B. No suitable GPU is available here, so the complete synchronization claim remains unverified.

drbh:
thanks for re-reviewing, i've updated the pr to address all of the stages and resolved the other items below!

> Blocker: the five-reader handshake fixes the Phase 4B TILE_INFO reuse, but the same producer/consumer race is still present in Phases 4A, 4C, and 4D.

done - 4A, 4C, and 4D now use the same `TILE_INFO_FREE` publish/read/ack rule. 4A gates the wait on `tile_seq` (its single publish site covers both the first tile and the terminal `has_work = 0`), and there are new `4a`/`4a-correctness` arms for isolating it.

> Required fix: give every phase the same explicit publish/read/ack ownership rule, then run per-phase B200 sweeps that exercise wraparound and the terminal tile. Compile-only checks cannot rule out a parity deadlock.

ran the per-phase sweeps on a B200. they surfaced two deadlocks in the extension, both fixed: a producer readback race in 4A (`warp::sync_mask` between lane 0's publish and the warp's readback), and a lane-stranding hazard on the pipeline barriers (a `warp::sync_mask` now sits between every pipeline-barrier wait and the lane 0 commit/load that re-arms it, so a barrier can never advance past a lane still waiting on it).

```bash
for p in 4a 4b 4c 4d; do
  GEMM_SOL_PHASE=${p}-correctness cargo oxide fuzz-schedule gemm_sol \
    --arch sm_100a --seeds 0..140 --focus mbarrier --confirm-runs 3 --timeout-secs 20
done
```

outputs

```text
4a: baseline PASS, 140/140 -- no schedule-sensitive failures found
4b: baseline PASS, 140/140 -- no schedule-sensitive failures found
4c: baseline PASS, 140/140 -- no schedule-sensitive failures found
4d: baseline PASS, 140/140 -- no schedule-sensitive failures found
```

perf is within ~1% of the pre-fix kernels (interleaved A/B, means of 3: 4B 505.1 -> 500.1 TFLOPS at 4096, 497.0 -> 492.5 at 8192, 483.0 -> 478.8 at 16384) - the residual is the per-iteration warp sync, which is the minimum needed for the no stranding guarantee.

> DCO: amend the contributor commit with a Signed-off-by trailer.

done

> No suitable GPU is available here, so the complete synchronization claim remains unverified.

everything above ran on 1x B200 (`sm_100a`); fmt, clippy, the compile-only build, and `ptxas -arch sm_100a` still pass on the final code.


please let me know if there are any other adjustments/changes that should be made! thanks again

nihalpasham (changes requested):
Thanks, @drbh — the latest push is an improvement.

Though a few full-path blockers remain:

- Both active `gemm_sol_final` kernels still use the unprotected `TILE_INFO` / `TILE_READY` mailbox and the old wait/rearm sequence. Fixing only `gemm_sol` leaves the same race in the canonical final example.
- Phase 4D multicast uses CTA-scope `mbarrier` arrive/wait without a per-generation cluster rendezvous or async-proxy release/acquire fences. Cluster Launch Control multicast is a cluster operation; Phase 4C already shows the required repository pattern. Please use that complete protocol in 4D as well.
- The current oracle is tile-periodic: inputs depend on row/column modulo 8 while tile strides are multiples of 8. A whole-tile permutation can therefore produce zero mismatches; a modeled 4096x4096 tile rotation did exactly that.
- There is still no committed regression that distinguishes the fix from a plausible wrong handshake. Changing one `TILE_INFO_FREE` arrival count from 5 to 4 still passes compile-only smoke and `ptxas`.

Please extend the fix across both final kernels, correct the Phase 4D cluster protocol, and add a tile-distinct full-output oracle plus a repeatable B200 fail-before/pass-after regression, an unaffected control, terminal/wraparound cases, and sanitizer evidence. Also update the title/body from Phase 4B to the actual scope.

That evidence is required by the [acceptance criteria shared in Discord](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

nihalpasham:
Did another full pass on this as part of a queue-wide review, including an independent re-derivation of the race and the handshake. Consolidating where it stands so there's one current list.

**Confirmed good, independently re-checked**

- the race is real and your fix has the right structure: TILE_INFO_FREE with arrival count 5 turns the one-way "ready" signal into a proper full/empty pair, which is the same protocol CUTLASS's pipeline classes use for reusable slots
- the parity sequencing, the `tile_seq > 0` first-publish guard, terminal-publish handling, and the arrive-after-warp-sync ordering all check out in all four phases
- no conflicts: the branch merges clean onto current main, so nothing to rebase

**Still blocking (these need your B200; we can only compile-gate sm_100a here)**

- **gemm_sol_final has the identical unprotected mailbox.** Both kernels publish and read TILE_INFO with no ack barrier (`kernels.rs`: publish ~195-198 / ~813-816, read ~485-494 / ~1140-1149, zero hits for TILE_INFO_FREE in the file). The demonstrated race survives in the canonical final example unless the same handshake lands there.

- **Phase 4D's protocol question is still open.** 4D wraps a cluster multicast (CTA_MASK_PAIR, PEER_BIT_MASK aliasing) in a CTA-scope handshake. 4C already shows the cluster-safe pattern (MCAST_BAR/CLC_READY rendezvous). Either adopt it or write down why the aliasing scheme is safe across the CTA pair.

- **The committed oracle can't see the failure it exists to catch.** Expected values depend only on `row%8` and `col%8` while tiles are 128 wide, so a whole-tile permutation (what a 4-arrival wrong handshake would produce) passes with zero mismatches:

  ```
  tile (3,7) results written to tile (3,8)'s address:
    oracle checks (row%8+1)*(col%8+1)*k  ->  identical in both tiles  ->  green
  ```

  Make values depend on absolute row/col and the same bug turns the whole tile red.

- **Commit the fail-before/pass-after evidence.** Right now the B200 runs live only in the PR description. A repeatable regression (plus an unaffected control) is what lets us re-verify after any future edit to these kernels.

- Small one: retitle from "phase4b", the head touches 4A through 4D.

The fuzz sweep caveat also matters for confidence: your 140 clean seeds only perturbed instructions ptx-schedule could classify, and the tcgen05 commit/wait pairs were unclassified at the time (#1113). That classification gap is being fixed in #1195, so a re-run after it lands will actually shake the async pipeline.

nihalpasham:
A couple of points: 

- This PR only fixes the `gemm_sol` example, 
- However as both `gemm_sol_final` and `gemm_sol` share some part of their implementation and we were able to reproduce the same bugs with gemm_sol_final, I'm documenting them here.
- Although the same fixes apply to `gemm_sol_final` - the fix costs 1.5-2% on the base kernel.

Details:

**Bug 1,  the tile-info mailbox race:**
  - the TMA warp republishes `TILE_INFO` (tile coords + has_work) with no confirmation that the 5 reader warps finished copying the previous values. 
  - A reader that's still mid-copy can see half old / half new:
    
```text
    producer:  write(tile A) ──── write(tile B)   ← too early
    readers:        copy A ... still copying ⚠
    result:    epilogue stores tile-A data at tile-B coordinates
```

**Bug 2 — the stranded-lane hang:**
  - `TMA_BAR`/`MMA_BAR` are reused every 4 K-laps and store 1 bit of phase.
  - All 32 lanes wait on them but only lane 0 acts. A lane the scheduler stalls through one full reuse cycle loses the phase and waits forever:
```text
    lap N:    32 lanes wait ─▶ tick ─▶ 31 leave, 1 slow lane stays ⚠
    lap N+4:  same barrier tocks — the slow lane missed both
              → polls a phase that already passed → hang
```

**Evidence:** 
- schedule-fuzz on a B200 (sm_100a), 
- mbarrier-focused delays,
- seeds 0..30, 3/3 confirmation: 17/30 seeds fail — 1 seed corrupts ~11.8M/16.7M outputs (bug 1), 16 seeds hang (bug 2), both kernels.

Plain runs pass by the hundreds: natural schedules don't open the race window, and delays can't break correctly-synchronized code — so these are real orderings bugs, not fuzzer artifacts.

Repro: 

```sh
GEMM_SOL_MODE=validate cargo oxide fuzz-schedule gemm_sol_final \
         --arch sm_100a --seeds 0..30 --focus mbarrier --confirm-runs 3
```

The fix pattern (TILE_INFO_FREE handshake + post-wait warp syncs) is in `gemm_sol` — see PR #1066 and `scripts/regress-gemm-sol-schedule.sh`.

PR #1067, opened 2026-08-20
fix(intrinsics-gen): an unescaped pipe drops a column from three reference rows

vyncint:
`intrinsics/generated-reference.md`'s main table has six columns. Three of its 1010 rows render with **seven** cells, and a renderer drops the extra one — so those three lose their **Backend evidence** column entirely.

The cause is a `|` in cell data. GFM splits a table row on pipes *before* parsing inline spans, so a pipe splits the cell even inside a code span; the spec's own rule is that a literal pipe must be written `\|` wherever it appears. Three catalog entries write a PTX destination pair that way:

```
elect.sync <register|predicate>, <register-or-immediate>;
match.all.sync.b32 <register|predicate>, <register-or-immediate>, ...
match.all.sync.b64 <register|predicate>, <register-or-immediate>, ...
```

`render_reference` interpolated `expected_ptx` straight into the row, so `elect_sync`, `match_all_sync` and `match_all_i64_sync` are the three intrinsics whose evidence column is unreachable in the rendered table.

## What changes

Escape the pipe. The helper is applied to every catalog-derived value in that row rather than to `expected_ptx` alone: the other fields carry no pipe today (measured — 1007 of 1010 rows have exactly seven delimiters, and no row carries a target union), so it is a no-op for them, and the next PTX pattern or LLVM symbol that grows one lands in an already-escaped cell. `availability` is left alone because the renderer builds it, markdown link included, rather than taking it from the catalog.

A unit test renders the full reference and asserts every row has exactly seven cell delimiters after escaped pipes are removed, **plus** that at least three rows needed escaping at all — so it cannot pass by the patterns quietly disappearing. Reverting the escape fails it with the offending row printed.

## Verification

- 1010 rows before and after; the three affected rows now read `<register\|predicate>` and no other row changed — 6 lines of diff.
- markdownlint, with the repo's own `.markdownlint.json`, drops from 21 violations on this file to 12: the 3 `MD056/table-column-count` ("Too many cells, extra data will be missing") and the 6 `MD038/no-space-in-code` they caused are gone.
- All three `generated-intrinsics` CI steps pass: `check` reports outputs current, `probe --all --skip-terminal` probes all 1010 routes, `check-abi-history` preserves all 1010 base entries.
- `intrinsics/catalog.json` is **untouched**, so the catalog SHA-256 the generated headers stamp is unchanged.
- 293 tests pass; fmt and `clippy --all-targets -- -D warnings` clean.

## Out of scope, deliberately

The remaining 12 `MD037` violations on this file have a different source: an evidence `detail` string writes `Literal _p64 and _p32` without backticks, so markdown consumes the underscores it is documenting. That prose lives in `intrinsics/evidence/`, and editing it would move the catalog hash, so it belongs in its own change.


nihalpasham:
- Reproduced the fix end to end: `cargo test -p cuda-intrinsics-gen` passes 293 including the new full-catalog render test, fmt is clean, and the reference diff is only the three escaped rows.
- Escaping at the cell boundary is the right placement: catalog values stay raw and the escape happens where GFM needs it. The unescaped availability column is covered by the new test's per-row delimiter count, so a future pipe there fails CI loudly instead of corrupting the page.
- **One commit layered on:** doc-comment wording only (names the literal pipe, splits the dash aside and the semicolon chain).

nihalpasham (approved):

PR #1068, opened 2026-08-20
fix(docs): an example README names `warp::shfl`, which has never existed

vyncint:
`hashmap_v3/README.md` contrasts v3's typed cooperative-groups API against what v2 did:

> - **Raw `warp::ballot` / `warp::shfl` find kernel.** v3 uses only the typed cooperative-groups API; …

`cuda-device` exports **32** `shuffle*` functions from `warp` and **no** `shfl` at all. `shfl` is the cooperative-groups *method* name, which is what v3 itself calls (`.shfl(`, twice) — so the sentence describes the raw path using the typed path's vocabulary. v2's source settles the right name: it calls `warp::shuffle` five times, alongside `warp::ballot` six times and `warp::lane_id` twice.

## Nothing would have caught it

`check-book-api-names.sh` exists precisely because a name the tree does not have *"renders, it builds, and a reader finds out by pasting it"* — but it looked only at `cuda-oxide-book`. Crate and example READMEs carry 122 qualified device names, are the first thing a reader of an example opens, and rot the same way. This extends the guard's file scope to them, minus the vendored `rustlantis` subtree, whose files keep their upstream form and document another project's API.

## Extending it exposed a second limitation

The family-shorthand expander was end-anchored (`^(.*)_\{...\}$`), so `warp::reduce_{sum,max,min}_f32` — real notation in `warp_reduce/README.md` — parsed as the family `reduce_{sum,max,min}` with the `_f32` dropped, and the guard reported three functions missing that were never named. The pattern now allows the group anywhere in the name, so `foo_{a,b}_bar` expands to `foo_a_bar` and `foo_b_bar` while `threadIdx_{x,y,z}` keeps working.

## Verification

- The guard passes on this tree: **42** device-API calls in **465** Rust blocks across the book and **159** READMEs, and **38** qualified prose names, all resolving against cuda-device's 1190 exported functions — up from 34 calls in 277 blocks and 31 names.
- Putting `warp::shfl` back fails it, naming the file.
- Planting `warp::ballot_nonexistent` in a README fails it, naming the file.
- Before the expander change the guard reported `warp::reduce_sum`, `warp::reduce_max` and `warp::reduce_min` missing; after it, none.
- Both new self-tests fire on an empty result: fewer than 20 book pages, or fewer than 50 READMEs, stops the run rather than reporting a clean tree.

The script keeps its name and its `status-guard.yml` wiring; only its scope and its two messages change.

### DIFF
--- a/.github/workflows/status-guard.yml
+++ b/.github/workflows/status-guard.yml
@@ -93,10 +93,12 @@ jobs:
       - name: Verify the book documents every cargo oxide subcommand
         run: bash scripts/check-cli-doc-coverage.sh
 
-  # A Rust code block in the book is something a reader pastes. When a device
-  # function is renamed the book keeps compiling and rendering, so the stale name
+  # A Rust code block in the book or a README is something a reader pastes. When
+  # a device function is renamed the docs still render, so the stale name
   # survives until someone tries it -- #797 and the `warp::shuffle_xor_i32` entry
-  # in the API quick reference were both found by hand, months apart.
+  # in the API quick reference were both found by hand, months apart. An example
+  # README naming `warp::shfl`, which has never existed, widened the scope to
+  # the crate and example READMEs.
   book-api-names:
     name: book device-API names
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


nihalpasham:
- Ran the guard on the branch: 42 device-API calls across 465 Rust blocks in the book and 159 READMEs, all resolving; confirmed the header's claim (32 `shuffle*` exports, no `shfl`). The family expander and the `rustlantis` vendored-docs exclusion both check out (`rustlantis-smoke` stays in scope).
- **One commit layered on:** finishes the book-to-docs rename the wider scope implies: the calls-pass error heading and the self-test messages no longer blame "the book" for scans that now cover READMEs, and the status-guard.yml comments mention READMEs (job and check names untouched).

nihalpasham (approved):

PR #1069, opened 2026-08-20
docs(book): the fuzzing chapter says the adapter rejects five types it accepts

vyncint:
`fuzzing-and-differential-testing.md` documents the trace API's type support as nine scalars and then rules out five more:

> Today the trace supports:
> ```
> bool, i8, i16, i32, i64, u8, u16, u32, u64
> ```
> It does not yet support `u128`, `i128`, `usize`, `isize`, or `char`.

`SCALAR_TRACE_TYPES` in `crates/fuzzer/tools/mir_generator.py` holds **sixteen**, including all five of those plus `f32` and `f64`, and `supported_trace_type` gates the adapter on exactly that set. Importing the module and calling the real predicate:

```
u128 True   i128 True   usize True   isize True   char True   f32 True   f64 True
[u128; 4] True     (u32, char) True
*const i8 False    &i32 False    (i8,i8,i8,i8,i8,i8) False
```

So the paragraph is wrong about every type it names, and understates the rest: support is not set membership but a recursive question, because an array or tuple folds by its leaves up to arity 5 (#792) — and `f32`/`f64` have been in since #484.

`crates/fuzzer/README.md` **already states all of this correctly**, listing the same sixteen scalars and the folding rule. The book page is the one outlier, so this brings it in line with the README and the code rather than inventing a third description.

## Two consequences

- The sample `UNSUPPORTED [adapter]` line used `u128` as its rejected type, which the adapter now accepts, so that example cannot be produced. It becomes `*const i8`, which the predicate does reject (measured above) and which the fuzzer README already uses as its own example of an adapter refusal.

- The widening plan listed "Add trace support for more scalar types (`u128`, `i128`, `usize`, `isize`)" and "Add arrays, tuples, and eventually structs/enums" as future work. The first is done and the second is half done, so the plan now records what landed and keeps structs and enums as what remains.

## Left alone deliberately

The `RigidTy(Char)` backend classification in the same sample output. `char` behaviour is nuanced — translation maps it to `u32` (`mir-importer/src/translator/types.rs:543`), while whether `#[device]` extern signatures accept it is open (#875) — and I did not reproduce that line end to end, so it is not mine to rewrite here.

**Verified:** the predicate results above come from importing `mir_generator.py` and calling `supported_trace_type`, not from reading it. Book builds clean under `sphinx-build -W`; `check-book-api-names.sh` passes.

### DIFF
--- a/cuda-oxide-book/compiler/fuzzing-and-differential-testing.md
+++ b/cuda-oxide-book/compiler/fuzzing-and-differential-testing.md
@@ -158,7 +158,7 @@ The runner prints one line per seed and then a full summary:
 
 ```text
 results:
-  seed 0: UNSUPPORTED [adapter] unsupported dumped type for Stage 2 adapter: u128 (...)
+  seed 0: UNSUPPORTED [adapter] unsupported dumped type for Stage 2 adapter: *const i8 (...)
   seed 1: COMPILE_FAIL [backend] Unsupported construct: Type translation not yet implemented for: RigidTy(Char) (...)
 summary: COMPILE_FAIL=1, UNSUPPORTED=1
 ```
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Verified the accept/reject boundary against the adapter: the corrected scalar list, arrays/tuples to any depth, the 5-element tuple cap, and the pointer refusal all check out.
- Two errors fixed in the layered commit: both `*const i8` lines carried the dump-site message, which that type can never reach since rustlantis keeps pointers out of `dump_var`; the producible refusal is "unsupported return type for return-value tracing" (mir_generator.py:406, the one seed 1436 emits per the fuzzer README). And "the harness has not grown up yet" contradicted the README's stance that a pointer is a permanent refusal (CPU oracle and GPU hold different addresses), so the chapter now separates permanent exclusions from widenable gaps.
- Same commit: plain wording for the folding paragraph, an accepted-vs-rejected type block, and attribution corrected to #484 (last scalars) plus #792 (aggregates). Book build and the api-names guard pass.

nihalpasham (approved):

Issue #1036, opened 2026-08-19
Support field projections of packed aggregates containing address-space 3 pointers


nihalpasham:
## Problem

#1001 lets a packed aggregate with a shared-memory pointer cross internal device calls and returns by coercing to a generic-pointer carrier at the boundary. But the value is only usable as a whole: any field projection (`value.tag`, `value.ptr`) routes through the local's stack slot, which is a whole-value store of the packed AS3 aggregate, and that store fails closed:

```text
storing a packed aggregate containing a shared-memory pointer by value is
target-mode dependent and is not yet supported
```

Concretely, this compiles after #1001:

```rust
let value = bounce_packed_shared(PackedShared { tag: 0x21, ptr: raw });
consume_packed_shared(value, raw, out);   // whole-value use only
```

but reading `value.ptr` inside the callee (or the caller) does not. It also means no runtime test can observe the round-tripped pointer; #1001's coercion is proven structurally by unit tests instead.

## Proposed approach

Reuse the target_stable_storage machinery #1001 generalized: give the local slot the carrier type (`<{ scalar, p0 }>`), coerce semantic-to-carrier on store and back on load, and keep field GEPs by-slot so the scalar fields stay byte-identical. The AS3 pointer field loads as p0 plus an addrspacecast, the same conversion the call boundary already does.

## Scope

* local-slot (alloca) storage and field projections only;
* same aggregate restrictions as #1000 (packed struct, one direct AS3 pointer leaf, remaining non-ZST fields scalar);
* strengthen the packed_aggregate_abi example so consume_packed_shared writes through the round-tripped `value.ptr` and reads back through the raw pointer, making a caller-side reconstruction bug observable at runtime.

## Out of scope

Everything #1000 already excludes: kernel parameters, device globals, arrays/vectors/enums with AS3 leaves, multiple AS3 pointer leaves.

uurl:
Following up on the design question from #1094, here is the narrow representation contract I have in mind for #1036 on top of #1186.

### Problem

The semantic MIR type and the physical local-storage type cannot be treated as the same thing for a packed aggregate containing an AS3 pointer.

For the narrow #1036 shape:

```text
semantic value:  <{ u8, p3 }>
local storage:   <{ u8, p0 }>
```

[...]

PR #1073, opened 2026-08-20
fix(mir-lower): support bounded packed AS3 arrays

### LINKED ISSUE #1072: Support bounded AS3 pointer arrays in the packed internal device ABI
## Problem

The packed-AS3 internal device ABI can recursively handle direct shared-memory pointer leaves and leaves nested through structs and tuples.

Fixed arrays remain excluded, even though the target-stable storage machinery already supports recursively rebuilding arrays and replacing AS3 pointer leaves with generic pointers at the physical return boundary.

Allowing arbitrary arrays without a bound would be undesirable: a compact source type such as `[&shared T; N]` expands into one extract/cast/insert sequence per shared-pointer element when the return carrier is rebuilt.

## Proposed scope

Extend the packed-AS3 internal ABI admission logic to support fixed arrays containing AS3 pointer leaves, subject to an explicit array-expansion budget.

The semantic value inside the function should continue to use the original packed aggregate with AS3 pointers.

Internal device arguments should retain the existing flattened argument ABI. At the return boundary, the target-stable storage machinery should recursively rebuild admitted arrays and replace their AS3 pointer leaves with generic pointers.

Use a maximum of 16 array-expanded AS3 pointer leaves per packed return carrier.

Direct AS3 leaves and struct/tuple nesting remain outside this array-specific budget because their code shape is proportional to the source structure.

## Out of scope

Keep the following fail-closed:

- fixed arrays requiring more than 16 array-expanded AS3 pointer conversions;
- vectors containing AS3 pointer leaves;
- kernel-entry ABI changes;
- whole-value packed AS3 loads/stores;
- packed-AS3 field projections that require local materialization;
- enums and other unrelated aggregate kinds.

## Acceptance criteria

- A packed struct containing a fixed array of AS3 pointer leaves can be returned by value from an internal device function when the array-expanded AS3 leaf count is within the supported bound.
- Arrays nested through otherwise supported struct/tuple shapes are handled recursively.
- A return carrier requiring exactly 16 array-expanded AS3 pointer conversions is accepted.
- A return carrier requiring more than 16 array-expanded AS3 pointer conversions is rejected.
- AS3 pointer vectors remain rejected.
- The physical return carrier contains generic-pointer replacements for every admitted AS3 array element.
- The semantic SSA aggregate continues to contain AS3 pointers.
- Internal device argument flattening remains unchanged.
- Existing recursive/multi-leaf packed-AS3 regressions continue to pass.

uurl:
Closes #1072

Depends on #1064 

This PR is stacked on #1064. Until #1064 merges, GitHub's diff against `main` also includes the recursive/multi-leaf packed-AS3 changes from that PR. 
The incremental change in this PR is the bounded fixed-array support.

## Summary

Extend the recursive packed-AS3 internal device ABI support with bounded fixed arrays of shared-memory pointer leaves.

Packed structs can now be returned from internal device functions when their supported recursive shape contains fixed AS3 pointer arrays whose total array-expanded shared-pointer leaf count does not exceed 16.

The function body continues to operate on the semantic packed aggregate containing AS3 pointers. Internal device arguments retain the existing flattened argument ABI. At the return boundary, the existing target-stable storage machinery recursively rebuilds admitted arrays and replaces their AS3 leaves with generic pointers.

## Implementation

This change widens `packed_shared_internal_abi_info` without duplicating the existing recursive storage or value-coercion machinery.

The classifier now:

- recursively traverses MIR structs and tuples as before;
- recursively admits fixed arrays;
- accepts up to 16 array-expanded AS3 pointer leaves per packed return carrier;
- continues to accept direct and recursively nested non-array AS3 leaves without charging them against the array-specific budget;
- continues to require a byte-faithful packed root struct;
- keeps vectors and unrelated aggregate kinds fail-closed.

`target_stable_storage_type` remains responsible for the recursive AS3 -> generic storage rewrite and already reconstructs arrays element by element at the physical value boundary.

The packed aggregate construction path continues to keep semantic AS3 pointers in SSA. No address-space conversion is introduced during semantic aggregate construction.

## Tests

Added coverage for:

- a bounded fixed AS3 pointer array using the target-stable return carrier;
- acceptance at the exact 16-leaf array rewrite bound;
- rejection above the 16-leaf array rewrite bound;
- explicit rejection of AS3 pointer vectors;
- semantic packed SSA construction containing an AS3 pointer array;
- end-to-end device call/return flow exercising the bounded array return carrier.

Validation performed:

- `git diff --check`
- `cargo test -p mir-lower packed_shared_internal_abi -- --nocapture`
- `cargo test -p mir-lower packed_struct_construction_with -- --nocapture`
- `cargo test -p mir-lower --all-targets`
- `cargo oxide build packed_aggregate_abi`
- `cargo oxide run packed_aggregate_abi`
- `cargo oxide run addressof_sharedarray`

The packed-AS3 classifier coverage passes with 8 tests, including the explicit vector rejection and both sides of the array rewrite bound.

The `mir-lower` suite passes 260 unit tests and 117 lowering tests.

The end-to-end `packed_aggregate_abi` regression passes on `sm_120a`, including the bounded-array packed shared internal ABI path.

## Out of scope

- AS3 pointer arrays above the explicit 16-leaf rewrite budget;
- AS3 pointer vectors;
- whole-value packed AS3 memory loads/stores;
- packed-AS3 field projections requiring local materialization;
- kernel-entry ABI changes.

Those remain separate work.


nihalpasham:
- Rebased this branch onto the updated #1064 head (force-push; the only conflict was the doc comment on the shape predicate). After #1064 merges, the diff collapses to the array delta: the `MAX_..._ARRAY_REWRITE_LEAVES` budget, the `MirArrayType` arm, and the array tests/kernels.
- Verified end to end on the rebased head: mir-lower 262+117 tests green including the four new array tests, `cargo oxide run packed_aggregate_abi` passes on an RTX 5090 with the three bounded-array kernels, the array-carrier NVVM IR compiles through libNVVM to a sm_120 cubin, full compile-only smoketest 216/216.
- The multiplicative `checked_mul` budget through nested arrays and the fail-closed vector rejection both check out; #1036 stays open for the projection gap.

nihalpasham (approved):

Issue #968, opened 2026-08-17
Feature request: borrowed `DeviceBuffer` subviews for typed kernel launches


MeccBai:
**What problem does this solve?**
One or two sentences on the gap this fills.

**Proposed API / behavior**
Show the Rust you would like to write (no need to get # Feature request: borrowed `DeviceBuffer` subviews for typed kernel launches

## Summary

Could `cuda-core` provide non-owning immutable and mutable views over contiguous
subranges of a `DeviceBuffer<T>`, with direct support in the generated typed
launcher API?

Possible names are:

```rust
DeviceSlice<'a, T>
DeviceSliceMut<'a, T>
```

This request is limited to host-side borrowing and kernel argument lowering. It
does not require strided views, device-side slice projection, or ownership of an
interior allocation pointer.

## Motivation

A larger device allocation often contains several logical vectors, matrix rows,
or temporary regions. A kernel should be able to operate directly on one
contiguous region without this sequence:

1. allocate another `DeviceBuffer`;
2. copy the region into it;
3. launch the kernel;
4. copy the result back.

Constructing an owning `DeviceBuffer` from an interior pointer is not a valid
workaround. The resulting value could free an interior address or cause the
original allocation to be freed twice.

The desired usage is approximately:

```rust
let mut buffer = DeviceBuffer::<f32>::zeroed(&stream, rows * cols)?;
let mut row = buffer.slice_mut(start..start + cols)?;

module.scale(
    &stream,
    &prepared,
    &mut row,
    2.0,
)?;
```

The subview must borrow the original allocation, carry its adjusted device
pointer and element count, and never free memory.

## Current downstream solution

We currently implement this outside CUDA-Oxide with two borrowing wrappers and
two ABI descriptors.

The host-side wrappers retain the borrow with `PhantomData`:

```rust
pub struct DeviceSpan<'a, T> {
    ptr: *const T,
    len: usize,
    _borrow: PhantomData<&'a [T]>,
}

pub struct DeviceSpanMut<'a, T> {
    descriptor: DeviceSliceMutDescriptor<T>,
    _borrow: PhantomData<&'a mut [T]>,
}
```

Kernel arguments use explicit `#[repr(C)]` descriptors:

```rust
#[repr(C)]
pub struct DeviceSliceDescriptor<T> {
    pub ptr: *const T,
    pub len: usize,
}

impl<T> Copy for DeviceSliceDescriptor<T> {}
[...]

PR #1081, opened 2026-08-20
Add borrowed DeviceBuffer subviews for typed kernel launches

uurl:
## What this adds

Borrowed `DeviceBuffer` subviews let a typed kernel operate on part of an allocation while the parent buffer retains ownership. Partially addresses #968.

## Shape of the change

- Add immutable and mutable views with checked ranges and consuming mutable splits.
- Pass each view through the existing pointer/length launch packet.
- Extend generated synchronous launchers to accept the new slice argument traits.

```text
parent allocation: [ a b | c d e | f ]
borrowed view:            c d e
kernel packet:           pointer + length 3
```

## Validation

- Focused runtime, host and macro checks passed on the submitted head.
- Borrowing controls cover parent lifetimes and disjoint mutable views.

## Remaining work

The runtime API must first land in its current owner, cutile-rs. Main now consumes `cuda-core` 0.3.1 from that project; the local runtime crates changed here were removed. Publish the view API there, update the shared dependency pins, then port the launcher integration to current main. GPU validation remains pending that integration.

### DIFF
--- a/crates/cuda-core/src/device_buffer.rs
+++ b/crates/cuda-core/src/device_buffer.rs
@@ -31,6 +31,7 @@
 use std::marker::PhantomData;
 use std::mem::MaybeUninit;
 use std::num::Wrapping;
+use std::ops::{Bound, RangeBounds};
 use std::sync::Arc;
 
 use cuda_bindings::CUdeviceptr;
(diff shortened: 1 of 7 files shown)


nihalpasham:
- Putting this on hold, not for code reasons: the host crates (cuda-core, cuda-host, cuda-async) are effectively frozen while we move to shared, published host crates ahead of the mono-repo consolidation. The substantive new surface here (`DeviceSlice`, `DeviceSliceMut`, `slice`/`slice_mut` on `DeviceBuffer`) lands in cuda-core, which is the crate being replaced, so merging now means re-porting the API immediately after.
- The work itself held up under review: all offset arithmetic is checked, mutable views have no Copy/Clone and split by value so overlapping mutable views cannot exist, the generated-launcher change to `impl KernelSliceArg` broke no example in the full corpus, and the GPU-backed subview tests pass locally (RTX 5090), with `cuda_module_api` green with and without `--features async`.
- Once the shared crates are in place we would take a re-targeted version of this; the borrowck-enforced disjoint-halves launch model is the right design for #968.

nihalpasham:
This PR remains on hold for the host-API migration described above; this is not a code-quality hold and is separate from #413.

- Review can resume after the shared host crates stabilize and this API is retargeted onto them.
- The refreshed branch should include a real middle-subview/async runtime case and compile-fail lifetime coverage on the final API.
- Until then, merging would create an API that immediately needs to be ported again.

When the work is retargeted, please keep the scope and before/after evidence aligned with the [current acceptance criteria](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

nihalpasham (changes requested):
Reviewed `5bfb68b5`: **changes requested**.

- Land `DeviceSlice` / `DeviceSliceMut` in their current owner, cutile-rs, and publish the shared-crate API first.
- Current main uses cuda-core 0.3.1 and has removed the local crates this patch edits. Then update the shared pins and port the typed launchers to the current macro structure.
- Preserve the borrowing contract and pointer/length ABI; rebase and sign the revised commits.

```text
cutile-rs views + release -> shared version pins -> oxide launch adapters
```

The focused borrowing/compile-fail controls pass. The ownership prerequisite prevents a bounded in-repository repair. #968 stays open.

uurl:
Understood. I’ll land the borrowed `DeviceSlice` / `DeviceSliceMut` API in `cutile-rs` first and wait for the shared-crate release. After that I’ll update the shared pins here, port the launcher integration to the current macro structure, rebase this PR, and keep the revised commits signed off.

nihalpasham (dismissed):

uurl:
Understood. I’ll move the borrowed view API and its lifetime/disjoint-range coverage to `cutile-rs` first. Once that shared runtime API is merged and published, I’ll update the dependency pins here, rebase this branch onto current main, and port the typed-launch integration and regressions to the current launcher/macro structure with signed commits.

nihalpasham:
Closing this as we'll anyway need to move this over post mono-repo migration

PR #1089, opened 2026-08-20
perf(llvm): emit proven Rust reference parameter attrs

### LINKED ISSUE #1088: Emit LLVM `noalias` / `readonly` for Rust references with proven semantics
## Motivation

Part of #413.

After #1062 preserves Rust pointer/reference kind through dialect MIR, the compiler can distinguish:

- `&T` → `SharedRef`
- `&mut T` → `UniqueRef`
- `*const T` → `RawConst`
- `*mut T` → `RawMut`
- compiler-generated / erased pointers → `Erased`

This makes it possible to attach LLVM parameter attributes when the corresponding Rust semantics are actually proven, instead of inferring them from the lowered pointer type.

This matters for optimization. In particular, alias information can enable substantially better load/store optimization and vectorization for reference-based kernels.

## Proposed semantics

Follow rustc's conservative reference rules:

| Rust parameter | Condition | LLVM attributes |
| --- | --- | --- |
| `&T`, `&[T]` | pointee is `Freeze` | `noalias readonly` |
| `&mut T`, `&mut [T]` | pointee is `Unpin` | `noalias` |
| `*const T` | — | none |
| `*mut T` | — | none |
| erased / compiler-generated pointer | — | none |

In particular:

- `&UnsafeCell<T>` must not receive `readonly`
- shared references should not receive aliasing attributes unless the `Freeze` requirement is satisfied
- mutable references should not receive `noalias` unless the `Unpin` requirement is satisfied
- raw pointers must not receive reference-derived attributes
- `DisjointSlice` should remain uninterpreted for this change

## Implementation outline

1. During MIR import, inspect the original rustc parameter type while the semantic information is still available.
2. Record only proven per-source-argument facts on `MirFuncOp`.
3. Preserve those facts through MIR lowering.
4. Remap source arguments to their physical LLVM parameters after ABI flattening.
   - For slices, attributes apply only to the pointer component, not the length field.
5. Make `llvm-export` mechanically serialize the already-proven facts.
   - The exporter must not infer aliasing from pointer type, mutability, address space, or `DisjointSlice`.

## Acceptance criteria

- `&T` / `&[T]` with `Freeze` emit `noalias readonly`
- `&UnsafeCell<T>` emits neither attribute
- `&mut T` / `&mut [T]` with `Unpin` emit `noalias`
- `&mut !Unpin` emits no reference-derived attribute
- raw pointers emit no reference-derived attributes
- slice metadata parameters are never annotated
- invalid attribute combinations are rejected
- legacy typed-pointer LLVM export continues to work
- existing CUDA examples continue to build
- end-to-end generated LLVM IR contains the expected attributes

## Out of scope

- `dereferenceable`
- recursive alias analysis
- deriving attributes from `DisjointSlice`
- deriving attributes from raw pointers
- alias inference based on LLVM-level pointer shapes

uurl:
Closes #1088

This PR is stacked on top of #1062; its actual delta is one commit touching 7 files. Once #1062 lands, I’ll rebase this PR onto `main` so the review diff contains only the LLVM attribute-emission changes.

## Summary

Part of #413.

This change emits LLVM `noalias` / `readonly` parameter attributes from Rust reference semantics that are proven during MIR import.

#1062 preserves source pointer/reference kind through dialect MIR. This PR builds on that information and captures the additional rustc semantic conditions required to safely emit LLVM parameter attributes.

The policy follows rustc conservatively:

| MIR pointer kind | Additional condition | LLVM attributes |
| --- | --- | --- |
| `SharedRef` | pointee is `Freeze` | `noalias readonly` |
| `UniqueRef` | pointee is `Unpin` | `noalias` |
| `RawConst` | — | none |
| `RawMut` | — | none |
| `Erased` | — | none |

## Design

The semantic decision is made at the MIR import boundary, where the original rustc type is still available.

For each source-level parameter, `mir-importer` records only the facts that rustc semantics justify.

These facts are then:

1. stored on `MirFuncOp`
2. propagated through MIR lowering
3. remapped from source arguments to physical LLVM arguments after ABI flattening
4. serialized mechanically by `llvm-export`

`llvm-export` deliberately performs no Rust-semantic inference.

For slices, only the physical pointer component receives the attributes. The length parameter remains unannotated.

## Safety / conservative cases

This PR intentionally does not emit attributes for:

- raw pointers
- erased/compiler-generated pointers
- `&UnsafeCell<T>` and other non-`Freeze` shared references
- non-`Unpin` mutable references
- `DisjointSlice` based solely on its lowered representation

This avoids deriving Rust aliasing semantics from LLVM pointer shape or mutability alone.

## Validation

Focused tests cover:

- valid `SharedRef` and `UniqueRef` attributes
- rejection of attributes on raw pointers
- rejection of invalid `readonly` combinations
- rustc `Freeze` / `Unpin` behavior
- source-to-physical LLVM argument remapping
- slice pointer vs. metadata handling
- LLVM textual export
- legacy typed-pointer export
- rejection of reference attributes on non-pointer LLVM parameters

Full validation:

```text
cargo fmt --all -- --check

cargo test -p llvm-export --all-targets

cargo oxide build vecadd
cargo oxide build addressof_sharedarray

git diff --check
```

End-to-end pipeline validation also confirms that the attributes reach generated LLVM IR.

For example, `vecadd` now lowers to:

```llvm
define ptx_kernel void @vecadd(
    ptr noalias readonly %v0,
    i64 %v1,
    ptr noalias readonly %v2,
    i64 %v3,
    ptr %v4,
    i64 %v5
)
```

This verifies that:

* shared `Freeze` references receive `noalias readonly`
* slice metadata remains unannotated
* raw/non-proven pointers remain bare
* the source-to-physical LLVM argument mapping is correct

`addressof_sharedarray` also confirms that shared-memory/raw pointer parameters do not acquire reference-derived attributes accidentally.

## Out of scope

* `dereferenceable`
* `DisjointSlice` alias inference
* recursive provenance analysis
* raw-pointer alias inference

### DIFF
--- a/crates/dialect-mir/README.md
+++ b/crates/dialect-mir/README.md
@@ -87,16 +87,31 @@ declared by rustc.
 but does not attempt to model dynamic Stacked Borrows / Tree Borrows tags or
 retag epochs.
 
-MIR-to-LLVM lowering deliberately erases `MirPointerKind`. All four Rust source
-categories keep the same LLVM pointer/fat-pointer representation, and this
-change does not emit `noalias`, `readonly`, `dereferenceable`, or related
-metadata. Any future alias metadata requires a separate audited policy; it must
-not be inferred from `is_mutable` alone.
+MIR-to-LLVM type conversion still erases `MirPointerKind`; all four Rust source
+categories keep the same physical LLVM pointer/fat-pointer representation.
+Function arguments carry separate, audited proof markers when rustc's own type
+semantics justify LLVM parameter attributes:
 
-For GPU-specific abstractions such as `SharedArray`, the Rust reference kind is
-still retained when the source type is a reference, while the CUDA address
-space remains an independent property. The compiler does not currently turn a
-`UniqueRef` to shared memory into cross-thread LLVM alias guarantees.
+| Source argument | Additional proof | LLVM parameter attributes |
+|-----------------|------------------|---------------------------|
+| `&T` / `&[T]` | pointee is `Freeze` | `noalias readonly` |
+| `&mut T` / `&mut [T]` | pointee is `Unpin` | `noalias` |
+| raw pointer | none | none |
+| erased/compiler pointer | none | none |
+
+The importer computes these facts while the original rustc `Ty` and trait
+queries are available and stores them on `mir.func` by source argument index.
+`mir-lower` remaps those indices after ABI flattening, so a slice applies the
+attributes only to its pointer component and never to its length. The LLVM
+exporter serializes the resulting proof bits without inferring anything from
+mutability, address space, or pointer kind on its own. In particular, a shared
+reference containing `UnsafeCell` is not `readonly`, and raw `*mut T` never
+acquires `noalias` merely because it is writable.
+
+`dereferenceable` and other provenance-sensitive attributes remain outside this
+policy. For GPU-specific abstractions such as `SharedArray`, CUDA address space
+remains independent from Rust reference kind and from these function-boundary
+proofs.
 
 ### Address Spaces
 
--- a/crates/dialect-mir/README.md
+++ b/crates/dialect-mir/README.md
@@ -10,17 +10,17 @@ rustc MIR ──► mir-importer ──► dialect-mir ──► mir-lower ─�
 
 The dialect defines nine types that preserve Rust-level semantics:
 
-| Type                  | Description                                        | Example                               |
-|-----------------------|----------------------------------------------------|---------------------------------------|
-| `MirTupleType`        | Heterogeneous tuples                               | `mir.tuple<i32, f32, i64>`            |
-| `MirPtrType`          | Pointers with address space and mutability         | `mir.ptr<f32, mutable, addrspace: 3>` |
-| `MirSliceType`        | Fat pointers (`&[T]` = ptr + len)                  | `mir.slice<f32, addrspace: 1>`        |
-| `MirDisjointSliceType`| `DisjointSlice<T>` -- per-thread unique access     | `mir.disjoint_slice<f32, ...>`        |
-| `MirStructType`       | Named structs with layout metadata                 | `mir.struct<"Point", [f32, f32]>`     |
-| `MirUnionType`        | Rust unions -- each field a view of the same bytes | `mir.union<"Repr", [a, b], [i32, f32], 4, 4>` |
-| `MirEnumType`         | Rust enums with their exact rustc layout           | `mir.enum<"Ordering", i8, ...>`       |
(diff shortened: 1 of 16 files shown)


nihalpasham:
- Not a merge call yet, same `soundness-audit` tag and reason as #1062, which this stacks on: wrong `noalias` miscompiles quietly, so I'm walking the chain from `is_freeze`/`is_unpin` to the printed attribute myself before signing off. Blocked on #1062 either way.
- The policy table matches rustc's own (`&T` + `Freeze` pointee -> `noalias readonly`, `&mut T` + `Unpin` -> `noalias`, raw pointers and `Erased` -> nothing), the proof runs where the real `Ty` is available, and the `MirFuncOp` verifier plus the fail-closed argument remap reject anything malformed. Two changes were needed on top, done and held with the #1062 fixes.
- The one smoketest failure is the optimization working, and the test couldn't see it. `aligned_field_loads` asserts fusion width in PTX; with `noalias readonly` on the parameter, llc routes the reads through the read-only cache:

  ```text
  before:  ld.global.v2.b32     {%r8, %r9}, [%rd16];
  after:   ld.global.nc.v2.b32  {%r8, %r9}, [%rd16];   <- __ldg fast path
  ```

  Fusion held in both directions: align(8) kernels fuse to v2, the natural-align-4 controls stay scalar, and every read additionally became `.nc`. The shape check now accepts the optional `nc.` qualifier so it checks fusion width, not the full instruction spelling.
- The behavior change has to be said where raw launches are documented: overlapping read/write buffers through a raw launch used to be latent UB and is now a real miscompile (loads hoisted or cached across stores). Typed launches already rule it out since writable parameters take `&mut DeviceBuffer`. Added that to the `cuda_launch!` module docs and the launching-kernels book page.
- Spot-checks so far, evidence rather than sign-off: tiled_gemm's PTX keeps strict `st.shared -> bar.sync -> ld.shared -> bar.sync -> st` ordering, attrs land only on pointer halves (slice lengths, raw pointers, DisjointSlice stay bare), and `&UnsafeCell<u32>` / `&mut PhantomPinned` get nothing per the fixture test. With the fixes: 95 workspace suites, 216/216 compile smoketest, GPU execution on tiled_gemm, barrier, and aligned_field_loads, book builds under `-W`.

nihalpasham:
Closing this PR because it is explicitly stacked on #1062 and derives `noalias` / `readonly` from the pointer-kind model in that branch.

#1062 is now closed because that model is unsound as a proof source: it cannot establish, end to end, that each pointer label came from the exact Rust fact that justifies it. Emitting optimizer attributes from those labels could turn a classification mistake into an LLVM aliasing miscompile.

The replacement will keep pointer construction and alias proof separate. This work can be reconsidered after the new proof exists end to end.

#1088 remains open.

PR #1092, opened 2026-08-20
fix(mir-lower): support packed AS3 field projections through carrier locals

### LINKED ISSUE #1036: Support field projections of packed aggregates containing address-space 3 pointers

uurl:
Closes #1036

## Summary

Support local materialization and field projections for the deliberately narrow packed-AS3 shape introduced by the internal device ABI work.

Eligible packed locals now use a target-stable carrier representation:

```text
semantic <{ scalar, p3 }>
        |
        | store coercion
        v
carrier  <{ scalar, p0 }>
```

Loads reconstruct the semantic value in the opposite direction.

Field projections from those locals are addressed directly through the carrier slots. An AS3 pointer field therefore physically loads/stores as p0 and is explicitly converted between p0 and p3 at the semantic boundary.

Arbitrary packed-AS3 memory remains fail-closed.

### Local carrier admission

Add a dedicated `packed_shared_local_storage_info` gate for local storage.

The local-storage gate is intentionally independent of the internal ABI classifier:

* packed struct root;
* exactly one direct AS3/shared pointer field;
* all remaining non-ZST fields scalar;
* no recursive aggregates;
* no arrays or vectors;
* no multiple shared-pointer leaves.

Keeping this gate independent prevents future widening of the internal device ABI classifier from implicitly widening local-storage support.

### Local allocation

`mir.alloca` uses the target-stable carrier type for eligible packed-AS3 locals:

```text
<{ scalar, p3 }>  semantic type
<{ scalar, p0 }>  physical local storage
```

### Whole-value local stores and loads

Stores into compiler-owned carrier locals coerce the semantic packed value to the target-stable storage representation before emitting `llvm.store`.

Loads use the physical carrier type and reconstruct the semantic packed value after `llvm.load`.

The existing fail-closed path is preserved for arbitrary packed-AS3 memory.

### Field projections

`mir.field_addr` recognizes an eligible carrier-backed local and emits a typed GEP through the carrier struct.

This keeps field addressing by-slot while preserving the existing rustc byte-offset fallback for ordinary packed aggregates.

For the AS3 field:

```text
projected load:   p0 -> addrspacecast -> p3
projected store:  p3 -> addrspacecast -> p0
```

The projected load/store path uses operand type history to distinguish the semantic MIR pointee from the physical carrier pointee.

### End-to-end regression

Strengthen `packed_aggregate_abi` so `consume_packed_shared` actually projects both fields from the returned packed value.

The test now:

* reads `value.tag`;
* reads `value.ptr`;
* writes shared memory through the round-tripped `value.ptr`;
* observes the mutation through the independently passed raw shared pointer.

This makes broken carrier-to-semantic reconstruction observable at runtime, while the lowering tests independently cover the inverse projected-store conversion.

## Tests

Added lowering coverage for:

* carrier allocation for an eligible packed-AS3 local;
* whole-value semantic -> carrier store;
* projected pointer-field p0 -> p3 load;
* projected pointer-field p3 -> p0 store;
* carrier pointer storage remaining in generic address space.

Validation performed:

```text
cargo fmt --all -- --check
git diff --check

```

Results:

```text
mir-lower unit tests: 254 passed
lowering tests:       117 passed
clippy -D warnings:   pass
packed_aggregate_abi: pass on sm_120a
```

## Out of scope

* kernel-entry ABI changes;
* device-global packed AS3 storage;
* arbitrary packed-AS3 memory;
* recursive or nested AS3 local projections;
* multiple AS3 pointer leaves;
* AS3 pointer arrays;
* AS3 pointer vectors;
* AS3-containing enums.

Those remain separate work.

### DIFF
--- a/crates/mir-lower/src/convert/ops/aggregate.rs
+++ b/crates/mir-lower/src/convert/ops/aggregate.rs
@@ -42,7 +42,7 @@ use crate::convert::types::{
     build_union_storage_type, convert_type, is_zero_sized_type, llvm_byte_faithful_twin,
     llvm_packed_struct_contains_pointer_in_address_space, llvm_type_contains_i1,
     llvm_type_size_align, make_slice_struct, mir_element_stride, mir_type_abi_align,
-    packed_shared_internal_abi_info,
+    packed_shared_internal_abi_info, packed_shared_local_storage_info,
 };
 use dialect_mir::ops::{
     MirConstantOp, MirConstructEnumOp, MirEnumPayloadOp, MirExtractFieldOp, MirFieldAddrOp,
@@ -54,7 +54,7 @@ use dialect_mir::types::{
 };
 use llvm_export::attributes::{ICmpPredicateAttr, IntegerOverflowFlagsAttr};
 use llvm_export::op_interfaces::{
-    CastOpInterface, CastOpWithNNegInterface, IntBinArithOpWithOverflowFlag,
+    CastOpInterface, CastOpWithNNegInterface, IntBinArithOpWithOverflowFlag, PointerTypeResult,
 };
 use llvm_export::ops as llvm;
 use llvm_export::types as llvm_types;
@@ -1797,6 +1797,23 @@ pub(crate) fn convert_field_addr(
 
     let map = build_struct_slot_map(ctx, &layout).map_err(anyhow_to_pliron)?;
 
+    // A compiler-owned packed-AS3 local may have a target-stable carrier
+    // allocation even though the MIR pointer still names the semantic struct.
+    // Use the carrier as the typed GEP source so the pointer field addresses
+    // generic-pointer storage. Arbitrary pointers keep the semantic layout and
+    // remain subject to the existing fail-closed whole-value memory rules.
+    let local_storage_ty = if let Some(info) =
+        packed_shared_local_storage_info(ctx, mir_ptr_pointee).map_err(anyhow_to_pliron)?
+    {
+        ptr_operand
+            .defining_op()
+            .and_then(|def| Operation::get_op::<llvm::AllocaOp>(def, ctx))
+            .filter(|alloca| alloca.result_pointee_type(ctx) == info.storage_ty)
+            .map(|_| info.storage_ty)
+    } else {
+        None
+    };
+
     let slot = match map.decl_to_llvm.get(field_index) {
         Some(Some(slot)) => *slot,
         Some(None) => {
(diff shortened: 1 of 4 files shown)


nihalpasham:
Closing this PR because the packed-AS3 carrier rule is incomplete and can generate wrong code.

- The change rewrites eligible AS3 storage to an AS0 carrier, but recovers that carrier identity only from a small set of defining operations.
- Passing the pointer through a call or block argument means carrier recognition can no longer be recovered. A later field projection then treats AS0 carrier storage as AS3 storage.
- The active #413 pointer-contract work supplies the semantic type and layout foundation a replacement must use; it does not itself implement this packed-AS3 fix.

A replacement needs to address the full pointer path and include call, block-argument, cast, and projection controls, consistent with the [acceptance criteria announced on Discord](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

PR #1093, opened 2026-08-20
fix(intrinsics): decouple remaining ABI IDs from variant order

### LINKED ISSUE #604: Canonical variant lists assign ABI IDs by position, which conflicts with the append-only ledger

uurl:
Closes #604 

## Summary

Finish decoupling compact-family ABI assignment from canonical variant
position.

The remaining register MMA, sparse MMA, and cluster-sreg expansions now leave
generated ABI IDs unbound and rely on the existing ABI ledger binding by
catalog identity.

Canonical admission checks continue to enforce semantic membership, ordering,
and completeness, but no longer derive ABI identity from variant position.

This completes the remaining positional ABI-ID coupling described in #604,
following #1038.

## Validation

- `cargo test -p cuda-intrinsics-gen` (293 passed)
- `cargo run -p cuda-intrinsics-gen -- check-abi-history --base-ref upstream/main`
- `git diff --check upstream/main..HEAD`
- `intrinsics/abi-v1.toml` unchanged
- generated intrinsic outputs unchanged

### DIFF
--- a/crates/cuda-intrinsics-gen/src/model.rs
+++ b/crates/cuda-intrinsics-gen/src/model.rs
@@ -825,7 +825,10 @@ pub struct RegisterMmaF8F6F4Admission {
     pub llvm_evidence_profile: String,
     pub libnvvm_evidence_profile: String,
     pub runtime_validation: RuntimeValidation,
-    pub first_abi_id: String,
+    /// Legacy shard metadata retained only so older overlays continue to parse.
+    /// ABI identity is bound from the append-only ledger by catalog ID.
+    #[serde(default, rename = "first_abi_id")]
+    pub _legacy_first_abi_id: Option<String>,
     pub a_elements: Vec<RegisterMmaElement>,
     pub b_elements: Vec<RegisterMmaElement>,
     pub product_count: usize,
@@ -839,7 +842,10 @@ pub struct RegisterMmaFp8Admission {
     pub llvm_evidence_profile: String,
     pub libnvvm_evidence_profile: String,
     pub runtime_validation: RuntimeValidation,
-    pub first_abi_id: String,
+    /// Legacy shard metadata retained only so older overlays continue to parse.
+    /// ABI identity is bound from the append-only ledger by catalog ID.
+    #[serde(default, rename = "first_abi_id")]
+    pub _legacy_first_abi_id: Option<String>,
     pub shapes: Vec<RegisterMmaShape>,
     pub accumulators: Vec<RegisterMmaAccumulator>,
     pub a_elements: Vec<RegisterMmaElement>,
@@ -854,7 +860,10 @@ pub struct RegisterMmaAmpereFloatAdmission {
     pub llvm_evidence_profile: String,
     pub libnvvm_evidence_profile: String,
     pub runtime_validation: RuntimeValidation,
-    pub first_abi_id: String,
+    /// Legacy shard metadata retained only so older overlays continue to parse.
+    /// ABI identity is bound from the append-only ledger by catalog ID.
+    #[serde(default, rename = "first_abi_id")]
+    pub _legacy_first_abi_id: Option<String>,
     pub product_count: usize,
     #[serde(rename = "variant")]
     pub variants: Vec<RegisterMmaAmpereFloatVariant>,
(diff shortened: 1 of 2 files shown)


nihalpasham:
Verified complete:

- crate-wide, zero positional ABI derivation left in production code (the remaining `i{:04}` + index sites are test fixtures)
- the binder fails closed when a variant has no ledger entry, and the ledger identity checks (catalog ID, operation key, signature) still guard every record
- #1104 landing ex2.approx.f16 through the sibling mechanism is real-world proof the binder handles family growth

This is the #1038 remainder and closes #604.

Sequencing, important: #1139 reintroduces this pattern (`first_abi_id = "i1017"`, IDs as i1017 + index) in the same files. This PR lands first; #1139 adopts blank-ID + `bind_generated_abi_ids` (noted there too).

On our side at landing: the byte-identical claim was made on a 2026-08-20 base and main has merged #1104/#1085 since (both touched resolve.rs), so we rerun `check`, `check-abi-history`, and the crate tests on the merged tree. Post-merge tidy-ups we'll take: delete the now-inert `first_abi_id` lines from the six overlay shards, and make `RegisterMmaRecipe.abi_id` optional so the five-ID skip list can go.

nihalpasham (approved):
Re-proven on the merged tree (with #1104/#1085/#1186/#1105 in): outputs byte-identical, ABI history preserves all 1016 entries, zero positional derivation outside test fixtures; CI green. Merging ahead of #1139 per the ledger sequencing.

PR #1104, opened 2026-08-21
feat(intrinsics): add ex2 approximate f16

almightychang:
## Summary

Add catalog-owned `ex2.approx.f16` as ABI entry `i1016`, using the existing scalar-f16 `u16` API / `i16` dialect carriers, with a GPU validation example.

PTX documents a PTX 7.0 / sm_75 floor. Pinned LLVM 22 selects `llvm.nvvm.ex2.approx.f16` at `sm_75 +ptx70` and rejects sm_70 / PTX 6.5. LLVM 22 also rejects `.ftz.f16`; `.ftz.bf16` is a separate PTX 7.8 / sm_90 variant.

## Evidence

The LLVM route emits `ex2.approx.f16` at the floor. The libNVVM route uses a dedicated CUDA 12.0 evidence profile, following #946 because CUDA 13.3 was unavailable on the measurement host. CUDA 12 libNVVM produced LTOIR that cuda-oxide linked into the recorded 4,264-byte sm_75 cubin (`4fd44330...bfe`); disassembly contains `MUFU.EX2.F16`. A per-variant `libnvvm_evidence_profile` selects that profile only for this record; the other 40 scalar-math variants retain the family default and every variant still resolves exactly two routes.

## Dependency

Depends on #1103, which lets the materializer load Debian CUDA 12's `__nvJitLink*_12_0` exports. That binding fix is not included here and should merge first.

## Example and validation

`examples/ex2_approx_f16` launches the intrinsic and checks six inputs, including a result in the binary16 subnormal range, against the PTX maximum relative-error bound of 2^-9.9. On an RTX 5080 (`sm_120`) it produced:

```text
input=0x0000 output=0x3c00 expected=0x3c00
input=0x3c00 output=0x4000 expected=0x4000
input=0xbc00 output=0x3800 expected=0x3800
input=0x4000 output=0x4400 expected=0x4400
input=0x3800 output=0x3da8 expected=0x3da8
input=0xcb80 output=0x0200 expected=0x0200
PASS: ex2.approx.f16 results are within the PTX 2^-9.9 relative-error bound
```

For GPUs below sm_75, smoketest builds device code for sm_75 and lets the example's host capability check skip before module loading.

Validation performed:

```text
cargo run -p cuda-intrinsics-gen -- check
cargo run -p cuda-intrinsics-gen -- probe --all --skip-terminal
cargo run -p cuda-intrinsics-gen -- check-abi-history --base-ref origin/main
cargo test -p cuda-intrinsics-gen
[...]

PR #1105, opened 2026-08-21
feat(intrinsics): probe every declared target

almightychang:
## Problem

`probe --all` lowered each intrinsic record against one target: the one its
recorded evidence selected. Every other target the record claims was taken
on faith.

    record:  clc_try_cancel_multicast
    claims:  sm_100a  sm_101a  ...
    probed:  sm_100a  (evidence target only)

- 395 of 1,016 records declare more than one target
- 1,349 record/target combinations never touched llc
- a bad target claim only surfaced if driver ptxas happened to reject it at
  runtime; the near-miss was #1032's first draft of `redux.sync.f32`, which
  listed sm_120a/sm_121a even though LLVM's `hasReduxSyncF32()` limits the
  instruction to the Blackwell 10x family

The PTX-floor table needed to check those targets was duplicated between
`cuda-oxide-codegen` and `cuda-intrinsics-gen`, with nothing keeping either
copy honest. Measuring both against the pinned backend found one real error:
sm_90a recorded a PTX 7.8 floor (inherited from sm_90), but
`llc -mcpu=sm_90a -mattr=+ptx78` aborts with "Minimum required PTX version
is 8.0".

## Fix

One owner for target policy, then exhaustive probing on top of it:

- new dependency-free `cuda-target-spec` crate owns the target grammar
  (`CudaArch`) and the PTX floors recorded against the pinned LLVM 22
  backend; `cuda-oxide-codegen`, `cuda-intrinsics-gen`, and `libnvvm-sys`
  all resolve targets through it, so the tables cannot drift apart again
- codegen fails closed: every PTX requirement, including `Default`, must
  resolve to a recorded floor before compilation proceeds; unknown targets
  are rejected instead of silently inheriting llc defaults
- `probe --per-target` (CI now runs `probe --all --skip-terminal
  --per-target`) lowers every record against every declared target with
  that target's derived floor
- derived targets are LLVM-selection-only; terminal ptxas evidence stays
  with the one selected evidence target, and the evidence schema is
  unchanged
- the recorded floors are tested against the backend itself: each recorded
  target is lowered with no `-mattr` and the emitted `.version` must match
  the table (uses the Rust toolchain's bundled llc; requires LLVM 22, since
  LLVM 21 rejects several of these targets outright)
- sm_90a's floor is corrected to PTX 8.0

## Probe coverage

                 llc runs    target claims checked
    before          1,016    1,016 / 2,365   (43%)
    after           2,365    2,365 / 2,365  (100%)

The exhaustive run found no live target-selection bug in the catalog; the
sm_90a floor was the only disagreement. Warm local timing for the probe
step: 7.4s before, 13.8s after.

## Validation

- `cargo test -p cuda-target-spec -p cuda-oxide-codegen -p libnvvm-sys
  -p cuda-intrinsics-gen` (all green; the floor-derivation tests ran
  against the toolchain's LLVM 22.1.2 llc, not skipped)
- `cargo run -p cuda-intrinsics-gen -- check` (generated outputs current)
- `cargo run -p cuda-intrinsics-gen -- probe --all --skip-terminal
  --per-target` (2,365 target configurations across 1,016 routes, no
  failures)
- `cargo oxide fmt --check` (root, codegen, and every example workspace)
- status/naming/license guard scripts and all three cargo-deny workspaces
- `cargo oxide run vecadd` end to end on an RTX 5090 (sm_120a), plus a
  `CUDA_OXIDE_TARGET=sm_90a` compile-only build emitting `.target sm_90a`
  PTX


nihalpasham:
- Superseded by the formal [changes-requested review](https://github.com/NVlabs/cuda-oxide/pull/1105#pullrequestreview-5001746277) on the current head.

nihalpasham (changes requested):
- **Benefit:** probing every declared target closes a real CI-correctness gap: an unsupported target/PTX pair can otherwise pass on one valid selected target and fail only for users of another target. That prevention benefit is worth adding; it is not a runtime-performance change.
- **Blocker:** the probe copies production's target → PTX policy, creating two authorities. The copy is already wrong for `sm_90a`: pinned LLVM rejects `+ptx78` and requires PTX 8.0, so the new check can report a false green.
- **Required design:** move the policy into one small dependency-free component used by both production codegen and the probe. Unknown or malformed targets must fail closed; neither consumer should keep a private fallback table.
- **Required tests:** derive every supported target's default/minimum against the pinned LLVM binary, including `sm_90a` default / `+ptx80` pass / `+ptx78` reject and the existing `sm_103a` / PTX 8.6 near miss.
- Keep derived per-target checks explicitly LLVM-selection-only; only the selected evidence target may claim terminal `ptxas` validation. Re-run the full 2,364-configuration probe and all inventory, license, CI-matrix, clippy, and docs guards after the redesign.

almightychang:
@nihalpasham 

Addressed the changes requested in the review:

- **Single policy component:** `cuda-target-spec` is now the dependency-free owner of CUDA target parsing, exact LLVM 22 target-to-PTX floors, and supported PTX feature spellings; production codegen and the probe both consume it (`f1955637`, `38f872dc`).
- **Fail closed / no private fallbacks:** malformed and unrecorded targets, including `Default` requirements, now fail before codegen, and the remaining consumer-local floor/feature tables and format/reparse detours were removed (`787e256c`, `aed7ebda`).
- **Backend-derived coverage:** leaf tests check every recorded target's default/minimum against the pinned LLVM 22.1.2 `llc`, including `sm_90a` default 8.0 / `+ptx80` pass / `+ptx78` reject and `sm_103a` default 8.8 / `+ptx86` reject / `+ptx88` pass (`f1955637`, with the original correction in `5055ecae`).
- **Evidence attribution:** derived per-target checks remain explicitly LLVM-selection-only; only the selected evidence target claims terminal `ptxas` validation (`38f872dc`).
- **Verification:** reran all crate tests (including 300 `cuda-intrinsics-gen` tests), the full 2,364-configuration probe across 1,015 routes, generated-output checking, clippy with warnings denied, formatting/docs checks, and the inventory, SPDX, dependency-license, and CI-matrix guards; all passed.

Local caveat: CUDA 12.0 `ptxas` predates `sm_120`, so the two pre-existing `libdevice_linking` integration tests for `sm_120` cannot pass locally; this branch does not touch them, and CI covers them with a current toolkit.

nihalpasham:
The redesign lands what the review asked for, verified value for value:

- `cuda-target-spec` is the one owner of grammar + PTX floors, used by production codegen, the probe, the resolver, and libnvvm-sys
- floor table matches the deleted copy except the intended sm_90a 7.8 -> 8.0 fix; known-target set unchanged; unknown targets fail closed
- backend-derived tests check every floor against the pinned LLVM 22 llc (sm_90a reject/pass pair, sm_103a near miss)
- production llc invocations behave the same for every reachable target
- per-target probing covers all declared configurations (2,365 across 1,016 routes on the merged tree; #1104 added a route since the PR body was written)

The rest I handled at landing:

- refreshed the PR description (it still described the superseded first draft, and it becomes the squash body)
- rebased onto current main and regenerated the example lockfiles
- full validation: per-target probe, generated-output check, ABI history, guard scripts, workspace tests

Sequencing: #1184 stacks on this and currently bundles it. This lands first; #1184 then rebases down to its own diff. Dropping the stale needs-implementation label.

nihalpasham (approved):
Redesign verified value for value on the merged tree (2,365-config per-target probe, llc-derived floors, end-to-end 5090 run); supersedes the earlier changes-requested review. Merging.

PR #1085, opened 2026-08-20
feat(intrinsics): add standard narrow tcgen05 MMA aliases

uurl:
Part of #339
Closes #1084 

## Summary

Add typed standard non-warp-specialized tcgen05 MMA convenience aliases for the five homogeneous narrow floating-point formats:

- `tcgen05_mma_e4m3`
- `tcgen05_mma_e5m2`
- `tcgen05_mma_e2m3`
- `tcgen05_mma_e3m2`
- `tcgen05_mma_e2m1`

The generic standard tcgen05 MMA path could already express these operations through `.kind::f8f6f4`; this change adds the generated typed surface analogous to the existing convenience APIs.

## Semantics

All five aliases select:

```text
tcgen05.mma.cta_group::1.kind::f8f6f4.collector::a::discard
```

with the compatibility signature:

```rust
pub unsafe fn tcgen05_mma_<format>(
    d_tmem: u32,
    a_desc: u64,
    b_desc: u64,
    idesc: u32,
    enable_d: bool,
)
```

## Changes

* Add five standard `tcgen05.mma` variants to the tcgen05 overlay.
* Append ABI entries `i1011` through `i1015`.
* Extend compact tcgen05 MMA admission from 19 to 24 generated variants:

  * 14 generic forms
  * 5 existing warp-specialized narrow aliases
  * 5 new standard narrow aliases
* Generate the public `cuda-device` APIs.
* Generate dialect, MIR importer, and lowering routes for the five aliases.
* Generate codegen target matching for the fixed standard selectors.
* Add five intrinsic probes.
* Extend both tcgen05 MMA evidence profiles to cover the new aliases.
* Preserve existing generic and warp-specialized behavior.

## Generator details

The standard aliases reuse the existing tcgen05 MMA selector representation with:

```text
kind = f8f6f4
cta_group = 1
collector_a = discard
```

The generated target matcher now imports the corresponding CTA-group and collector-A dialect attributes required by the new standard fixed-selector routes.

## ABI

The append-only ABI ledger grows from 1010 to 1015 entries.

Validation:

```text
intrinsics/abi-v1.toml preserves 1010 base entries and has 1015 total entries
```

No existing ABI entry was removed, reordered, or changed.

## Evidence

Both existing tcgen05 MMA evidence profiles now cover 24 variants.

The five new generated probes pass against the pinned LLVM/NVPTX evidence route and lower to the expected standard narrow MMA form:

```text
[...]

PR #1079, opened 2026-08-20
feat(wgmma): add m64n128 BF16 full-drain lowering

uurl:
Closes #1078 
Part of #338

## Summary

Adds value-threaded lowering for:

`wgmma.mma_async.sync.aligned.m64n128k16.f32.bf16.bf16`

for canonical linear full-drain regions.

The new path carries a canonical `[[f32; 8]; 8]` accumulator as 64 tied SSA `f32` values through one convergent inline-PTX scope.

## What changed

- Added the public `wgmma_mma_m64n128k16_f32_bf16` API.
- Added `Acc64x128 = [[f32; 8]; 8]`.
- Added pointer-form and 64-result value-form NVVM WGMMA ops.
- Added MIR importer dispatch for the new intrinsic.
- Generalized value-form WGMMA lowering from a fixed 32-value carrier to shape-specific 32/64-value carriers.
- Added m64n128 BF16 recognition to deferred-accumulator fusion.
- Added canonical `[[f32; 8]; 8]` full-drain lowering.
- Recomputed accumulator element addresses after the fused m64n128 inline-PTX scope rather than keeping all addresses live across it.
- Added verifier, lowering, codegen, integration, example, and documentation coverage.

## Supported shape

```text
wgmma.fence
one or more homogeneous m64n128k16.f32.bf16.bf16 MMAs
wgmma.commit_group
wgmma.wait_group<0>
```

The accumulator is:

```rust
[[f32; 8]; 8]
```

and is represented by 64 tied `f32` values inside the fused asm region.

## Intentionally unsupported

This PR does not add:

* m64n128 F16;
* m64n256;
* TF32 support;
* counted m64n128 loops;
* m64n128 partial waits or pipelines;
* pointer fallback for non-canonical m64n128 accumulators;
* mixed m64n64/m64n128 full-drain regions.

Those cases remain fail-closed.

## Backend validation

The dedicated `sm_90a` carrier tests compile real m64n128 WGMMA PTX through `ptxas`.

For both one MMA and two chained MMAs under one commit:

```text
0 bytes stack frame
0 bytes spill stores
0 bytes spill loads
Used 90 registers
```

The existing m64n64 BF16/F16 carriers remain spill-free at 58 registers.

## Validation

Passed locally:

```text
cargo fmt --all -- --check
git diff --check

cargo check \
  -p cuda-device \
  -p dialect-nvvm \
  -p mir-importer \
  -p mir-lower \
  -p cuda-oxide-codegen

cargo clippy \
[...]

PR #1077, opened 2026-08-20
feat(wgmma): add TF32 m64n64k8 value-threaded full-drain lowering

uurl:
Closes #1076 

## Summary

Adds value-threaded lowering for Hopper TF32 WGMMA

```text
wgmma.mma_async.sync.aligned.m64n64k8.f32.tf32.tf32
```

in the canonical linear full-drain case.

TF32 now follows the existing register-resident strategy for:

```text
wgmma.fence
mma+
commit_group
wait_group<0>
```

The 32 `f32` accumulator values remain tied through one convergent inline-PTX region, preventing LLVM from introducing a spill boundary while the
asynchronous WGMMA group is in flight.

TF32 differs from the existing BF16/F16 forms in two relevant ways:

* the hardware shape uses `K=8`;
* the PTX instruction has no transpose immediates.

The lowering therefore models TF32 as a distinct instruction form instead of
only changing the input-type suffix.

### Device API / importer

Adds:

```rust
wgmma_mma_m64n64k8_f32_tf32
```

for the canonical `[[f32; 8]; 4]` accumulator.

The importer now emits a dedicated TF32 `m64n64k8` pointer-form NVVM operation.

The existing legacy compatibility entry

```text
wgmma_mma_m64n64k16_f32_tf32
```

remains explicitly unsupported and is not repurposed.

### NVVM dialect

Adds:

* `WgmmaMmaM64N64K8F32Tf32Op`
* `WgmmaMmaGroupValuesM64N64K8F32Tf32Op`

The TF32 value carrier uses the same 32 × `f32` accumulator interface as the
existing BF16/F16 `m64n64` carriers.

No TF32 counted-loop or pipeline carrier ops are added.

### MIR lowering

Extends canonical linear full-drain recognition to TF32.

A region of the form

```text
wgmma.fence
tf32 mma+
wgmma.commit_group
wgmma.wait_group<0>
```

is rewritten from the pointer-form intrinsic to the 32-value SSA carrier.

The generated PTX is:

```text
wgmma.mma_async.sync.aligned.m64n64k8.f32.tf32.tf32
```

with:

```text
accumulators, desc_a, desc_b, 1, 1, 1
```

and no F16/BF16 transpose operands.

Linear regions remain homogeneous. Mixing TF32 with BF16 or F16 is rejected.

TF32 has no deferred pointer fallback.

Existing counted-loop and partial-wait recognizers remain BF16-only.

### Tests

Adds coverage for:

* TF32 NVVM pointer-form verification
[...]

PR #1075, opened 2026-08-20
feat(wgmma): add register-resident F16 counted K-loop lowering

uurl:
Closes #1074 

## Summary

Extend the existing register-resident WGMMA counted K-loop lowering to `m64n64k16.f32.f16.f16`.

The new path reuses the same counted-loop control and accumulator model already used by BF16:

```text
fence
counted K-loop:
    mma
    descriptor recurrence
commit_group
wait_group<0>
```

All 32 per-thread `f32` accumulator values remain tied across one convergent
inline-PTX statement, so LLVM never observes an in-flight asynchronous WGMMA accumulator between loop iterations.

## What changed

* Add an F16 counted-loop value carrier:
  `WgmmaMmaLoopValuesM64N64K16F32F16Op`.
* Extend the single-slot counted-loop recognizer to accept canonical F16 loops.
* Generalize the counted-loop PTX template over the existing WGMMA input kind.
* Add MIR-to-LLVM conversion for the F16 counted-loop carrier.
* Add verifier and lowering coverage for the new carrier.
* Extend the real `llc` / `ptxas` counted-loop probe to cover both BF16 and F16.
* Update the public WGMMA documentation to describe F16 counted-loop support.

The existing counted-pipeline path remains BF16-only.

## Scope

Supported F16 shape:

* canonical `[[f32; 8]; 4]` accumulator;
* one MMA per counted-loop iteration;
* compile-time positive trip count;
* affine `u64` descriptor recurrences;
* `fence -> counted loop -> commit_group -> wait_group<0>`.

Still unsupported:

* F16 `wait_group<N>` for `N > 0`;
* F16 multi-slot pipelines;
* F16 counted pipelines;
* non-canonical F16 accumulator fallback;
* dynamic trip counts;
* arbitrary loop CFGs;
* TF32 WGMMA.

## Validation

```text
cargo fmt --all -- --check
cargo check -p dialect-nvvm -p mir-lower -p cuda-device -p cuda-oxide-codegen
cargo test -p dialect-nvvm --test ops_test
cargo test -p mir-lower --test lowering_test
cargo test -p cuda-oxide-codegen --test wgmma_counted_loop_ptx -- --nocapture
cargo clippy -p dialect-nvvm -p mir-lower -p cuda-device -p cuda-oxide-codegen --all-targets -- -D warnings
```

Results:

* `dialect-nvvm`: 51 passed
* `mir-lower lowering_test`: 117 passed
[...]

PR #1106, opened 2026-08-21
ci(book): gate the dialect page's catalog stamp against catalog.json

### LINKED ISSUE #1090: ci(book): nothing ties the dialect-nvvm page's catalog stamp to catalog.json
#1052 refreshed the dialect-nvvm book page after two op-adding PRs rotated the catalog stamp, and the review found the page has no automated freshness check: the page prints the catalog SHA and its own rule says a mismatch means every count is stale, but nothing in CI compares the printed stamp against `intrinsics/catalog.json`. The page failed its own test for weeks (560 printed vs 570 in the tree) and only a hand audit caught it.

A small gate closes this permanently:

- hash `intrinsics/catalog.json` (first 8 hex of SHA-256, the format the page prints),
- grep the stamp out of `cuda-oxide-book/compiler/mlir-dialects.md`,
- fail with a message pointing at the page when they differ.

That turns "the counts are stale" from a silent rot into a red check on the op-adding PR itself, the same pattern the generated-reference `check` gate uses for `intrinsics/generated-reference.md`. Counts themselves stay hand-written; only the stamp needs machine checking, since the page's own protocol derives staleness from it.

vyncint:
Closes #1090.

`mlir-dialects.md` states its own freshness rule:

> At catalog SHA-256 `26a9cc13` (the stamp in every `ops/generated/` file header), the dialect holds 570 operations across 42 modules …
> If the header stamp no longer starts with `26a9cc13`, the counts on this page predate the catalog you are reading.

Nothing enforced that rule, so the rule that would expose stale counts was itself unchecked. The page failed its own test for weeks before #1052 (560 printed against 570 in the tree), and **it is failing again right now**: the catalog hashes to `00372dfe`, five hand-written ops and five catalog entries after that refresh landed. Two hand audits in two batches is the pattern this closes.

## The gate

One file hashed, one page grepped — no cargo, no build, no network. The hash is the one `cuda-intrinsics-gen` stamps into every generated file (`sha256_bytes(catalog_json.as_bytes())` over the file as committed), so a plain `sha256sum` reproduces it, and the page prints its first eight digits.

Three checks rather than one, so each failure says which thing moved:

1. **The generated sources must stamp the hash `catalog.json` actually has.** Failing here means the generated outputs are stale, and the message says to run `cuda-intrinsics-gen -- generate` — not to edit the book.
2. **The page must print exactly one distinct stamp.** It names it twice, once stating it and once in the rule explaining what a mismatch means, so updating one and not the other leaves the rule contradicting the statement above it.
3. **The stamps must agree**, and the failure prints both sides plus the two commands that re-derive the counts the stamp guards.

It refuses to report success from a check that found nothing: a page reworded so no stamp survives fails with that stated. That path needed `|| true` on the *capture* rather than on a later pipeline — under `pipefail` a grep matching nothing aborted the script before the diagnostic, turning a reworded page into a silent non-zero exit. I only found that because the mutation produced no message.

## The refresh that makes it pass

| claim | was | now |
|---|--:|--:|
| catalog stamp (×2) | `26a9cc13` | `00372dfe` |
| total operations | 570 | 575 |
| hand-written operations | 21 | 26 |
| hand-written `wgmma` row | 9 | 14 |
| catalog entries | 1010 | 1015 |

Generated stays 549 across 35 modules and **every area row is unchanged**, because #1085's five new catalog entries are aliases onto existing structural ops — exactly what the page's "several intrinsics can share one structural op" sentence predicts. The +5 hand-written are all `wgmma`, from #1075, #1077 and #1079.

The `wgmma` row's description moves with its count, as in #1052: it said "the m64n64k16 bf16/f16 shapes", and the module now holds four shape/type pairs — `m64n64k16` bf16 and f16, `m64n128k16` bf16 (#1079), `m64n64k8` tf32 (#1077).

## Verification

- **Four mutations, each failing with its own message:** a stale stamp; only one of the two stamps updated; the page reworded so none is found; an edited `catalog.json`. The healthy tree passes.
- Every count re-derived from the tree: 26 + 549 = 575, both tables sum to their stated totals (7 rows → 26, 11 rows → 549), and each still lists every module on disk.
- All eleven guard scripts pass, `check-spdx-headers.sh` accepts the new file, `cuda-intrinsics-gen -- check` reports outputs current, the workflow parses with all nine jobs on `ubuntu-latest`, and the book builds clean under `sphinx-build -W`.

## One deferral

`just check-guards` should list this script too, one line beside the other ten. Its recipe is being restructured in #1102, so that line is left for a follow-up rather than conflicting. No guard enforces that list, so CI is unaffected either way.

### DIFF
--- a/.github/workflows/status-guard.yml
+++ b/.github/workflows/status-guard.yml
@@ -112,6 +112,26 @@ jobs:
       - name: Verify every device call the book shows resolves
         run: bash scripts/check-book-api-names.sh
 
+  # The dialect-nvvm book page prints the catalog SHA-256 and states its own
+  # freshness rule: if the stamp no longer matches, every count on the page
+  # predates the catalog being read. Nothing enforced that rule, so the page
+  # failed its own test for weeks -- 560 operations printed against 570 in the
+  # tree -- and only a hand audit caught it, twice. Machine-checking the stamp
+  # turns "the counts are stale" into a red check on the pull request that
+  # rotated the catalog. The counts stay hand-written; only the stamp is gated,
+  # because the page derives staleness from it.
+  book-catalog-stamp:
+    name: book catalog stamp
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      # Hashes one file and greps one page: no cargo, no build, no network.
+      - name: Verify the dialect page's catalog stamp is current
+        run: bash scripts/check-book-catalog-stamp.sh
+
   # `unit-tests.yml`'s matrix is the only thing deciding which crates get
   # `cargo test`, and a member missing from it fails nothing: the workflow is
   # green, `just check` is green, and the tests never ran. #971 found four
--- a/cuda-oxide-book/compiler/mlir-dialects.md
+++ b/cuda-oxide-book/compiler/mlir-dialects.md
@@ -296,16 +296,16 @@ they become `call` instructions to `@llvm.nvvm.*` intrinsics.
 
 ### Architecture Coverage
 
-At catalog SHA-256 `26a9cc13` (the stamp in every `ops/generated/` file
-header), the dialect holds 570 operations across 42 modules, and they come
+At catalog SHA-256 `00372dfe` (the stamp in every `ops/generated/` file
+header), the dialect holds 575 operations across 42 modules, and they come
 from two different places. The split is the first thing to know about it,
 because it decides where -- and whether -- you would add one. If the header
-stamp no longer starts with `26a9cc13`, the counts on this page predate the
+stamp no longer starts with `00372dfe`, the counts on this page predate the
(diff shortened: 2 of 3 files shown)


nihalpasham:
- Ran the guard both ways: passes on this branch, fails on main's stale page naming both hashes; all four mutation modes (stale page, stale generated header, zero stamps found, multiple distinct stamps) fail closed with their own messages.
- Every refreshed number re-derives from the tree: 549 generated ops across 35 modules plus 26 hand-written (wgmma now 14) is 575; the catalog holds 1015 entries; stamp `00372dfe` matches `sha256sum` of the committed catalog and every generated header.
- Placement matches the other eight status guards (same job shape, same script conventions, no cargo/network), and the generated-header cross-check gives a stale generator run its own failure instead of blaming the book. Closes #1090.

nihalpasham (approved):

PR #1107, opened 2026-08-21
fix(intrinsics-gen): an overlay file left out of overlay.toml is silently ignored

vyncint:
`intrinsics/overlay.toml` is an explicit index: `read_overlay` reads the shards it names and nothing else. So a family file added under `intrinsics/overlay/` and left out of `shards` is **never read at all** — its intrinsics simply do not exist — and every gate agrees, because the file was never an input and the resolved catalog therefore matches the committed one.

Verified on `main` by adding an unlisted `intrinsics/overlay/zz_probe.toml`:

```
$ cargo run -p cuda-intrinsics-gen -- check
generated intrinsic outputs are current
```

**The opposite mistake is already caught.** Dropping a listed shard fails with `active ABI ledger entry i0252 (prmt) has no overlay record`, because the ledger entry outlives the record. Only the unlisted-file direction is silent — and it is the likelier slip: creating the file is the visible half of the work, editing the index the forgettable half.

`read_overlay` already rejects a duplicate shard, an unsorted list, an inline record and an empty list. This is the one case it cannot see, because it never looks at the directory.

## Where the check lives

In `generate.rs`, beside `stale_generated_outputs`, whose job is the same reconciliation in the other direction: that function walks the managed output roots looking for files the manifest no longer expects; this one walks the overlay directory looking for files the manifest never named. Both need a directory listing rather than manifest parsing, so both belong in the pass that owns repo-wide consistency rather than in the manifest reader.

It runs on `generate` and on `check`, so the existing `generated-intrinsics` CI job and `just check-intrinsics` both pick it up with **no new wiring**.

## Refusing to answer from nothing

Two guards, because the failure mode worth surviving is "silently compares empty sets":

- an index that parsed to no shards would make every file on disk look unindexed, so that stops the run instead;
- a directory read that found no `.toml` would report a clean reconciliation from an empty comparison, so that stops it too.

## Verification

- On `main` the index and the directory agree exactly — **64 files, 64 shards**, neither side extra — so this guards against future drift rather than fixing a present break: `check` still reports outputs current.
- An unlisted file now fails both `check` and `generate`, with the file named and the remedy stated.
- Three unit tests: an unindexed file is reported, a fully indexed directory reports nothing, and each self-test refuses rather than answering. Replacing the membership test with `if false` fails the first one.
- 296 tests pass; fmt and `clippy --all-targets -- -D warnings` clean; `check-spdx-headers.sh` accepts the file; `probe --all --skip-terminal` and `check-abi-history` both still pass over all 1015 entries.


nihalpasham:
- Reproduced the silent ignore on main: an unlisted probe under `intrinsics/overlay/` is never read and `check` stays green; on this branch both `check` and `generate` fail with the file named. 298 tests, fmt and clippy clean.
- Placement is right: beside `stale_generated_outputs`, the same reconciliation in the other direction, riding the existing CI job with no new wiring; both refusal guards kill the compares-empty-sets failure mode.
- **One commit layered on:** the manifest grammar (`validate_overlay_shard_path`) accepts nested shard paths, but the scan was a flat `read_dir`, so an unlisted file in a subdirectory was still silently ignored, and a fully nested but indexed layout would trip the empty-directory refusal on a valid tree. The walk now recurses, mirroring `collect_stale_generated`, with tests in both directions.

nihalpasham (approved):

PR #1108, opened 2026-08-21
docs(ptx-schedule): the one published crate with no README

vyncint:
`ptx-schedule` arrived with #1065 and is the only publishable crate in the workspace with neither a `readme` key nor a `README.md` on disk. Checked across `cargo metadata`: every other publishable member has both. crates.io renders that file as the package page, so this one would publish with an empty body.

Same gap #981 closed for `ptx-parse`, and this README follows that one's shape.

## Content is taken from the crate, not the commit that added it

- The purpose and the *"owns neither CUDA execution nor an input generator"* boundary come from `lib.rs`'s module header.
- The seven `SiteKind` variants and the eight campaign `RunKind` verdicts are each enumerated **in full** — a partial list of a closed enum is exactly the thing that rots (#978, #1045, #1056).
- `InjectionOptions`' four fields, `DEFAULT_MAX_SLEEP_NS` (64 µs), the `nanosleep.u32` insertion, and the `confirm_runs` re-run before a finding is reported are all as implemented.
- `cargo oxide fuzz-schedule` is named as the user-facing entry point, with the crate's own binary shown for single-file work; the flags in that example come from `main.rs`'s usage string.

## One detail carried from #981's review

The License link is absolute rather than `../../LICENSE`, because crates.io rewrites relative README links against the repository root and the relative form 404s there. `ptx-parse`'s README was corrected the same way.

## Verification

- `cargo metadata` now reports `readme: README.md` for the crate and the file resolves.
- All seven `SiteKind` and all eight `RunKind` names appear; the four `InjectionOptions` fields appear; no relative `../` links remain.
- `check-crate-inventory.sh`, `check-spdx-headers.sh` and `check-book-api-names.sh` pass; `cargo fmt --all --check` clean; `RUSTDOCFLAGS="-D warnings" cargo doc -p ptx-schedule` clean.

### DIFF
--- a/crates/ptx-schedule/Cargo.toml
+++ b/crates/ptx-schedule/Cargo.toml
@@ -9,6 +9,7 @@ authors.workspace = true
 license.workspace = true
 repository.workspace = true
 description = "Structural PTX schedule analysis and deterministic perturbation"
+readme = "README.md"
 
 [dependencies]
 dialect-ptx = { workspace = true }
--- a/crates/ptx-schedule/README.md
+++ b/crates/ptx-schedule/README.md
@@ -5,17 +5,17 @@ Structural PTX schedule analysis and deterministic perturbation.
 A kernel that races only under one interleaving passes every run until it does
 not. This crate makes that interleaving reachable on purpose: it finds the
 points in a PTX module where thread progress is observable, then inserts a
-seeded `nanosleep.u32` at a subset of them so the same seed always produces the
-same schedule.
+seeded `nanosleep.u32` at a subset of them so the same seed always produces
+the same rewrite.
 
-## What it deliberately does not do
+## What the analyzer deliberately does not do
 
-It owns neither CUDA execution nor an input generator. It reads PTX text and
-writes PTX text. Building, launching, watchdogging and verdict assignment are
-the campaign driver's job, and the kernel inputs stay whatever the example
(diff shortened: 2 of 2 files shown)


nihalpasham:
- Confirmed the gap: of 29 publishable crates, ptx-schedule was the only one with neither a README nor a `readme` key, so it would publish with an empty package page. The site table, verdict table and CLI examples check out against the source.
- **One commit layered on:** six statements did not match the code. The execution disclaimer covered the whole crate while `campaign.rs` builds, launches and watchdogs two sections down (now scoped to the analyzer); `intensity` is a probability dial (0.75x per site, and it also scales sleep length), not a fraction of sites; `focus` re-weights by substring over opcode and instruction text (0.95x hits, 0.15x misses), it never restricts to one callable; `Skipped` requires a printed skip marker (a zero-injection seed still runs and reports like the baseline); findings carry confirmation counts rather than being gated on them; a seed pins the rewrite, not the schedule. Same commit: `mbarrier.*` in the Barrier row, the ordering qualifiers OrderedMemory actually matches, the conservative same-or-earlier-block back-edge test, and the block-index field called out as back-edge-only.
- The `redux` entry in the WarpCollective row becomes true when #1111 lands, so that one merges first.

nihalpasham (approved):

PR #1110, opened 2026-08-21
fix(backend): the forbidden-crate diagnostic panics on a non-ASCII path

vyncint:
## What is wrong

The collector renders `CUDA-OXIDE: FORBIDDEN CRATE IN DEVICE CODE` in a boxed error. Fitting the function path into that box was:

```rust
// Truncate fn_path if too long (max 48 chars to fit in box)
let fn_display = if fn_path.len() > 48 {
    format!("{}...", &fn_path[..45])
} else {
    fn_path.clone()
};
```

The comment says characters, and the box pads the field with `{:<48}`, which counts characters. Both operations here are on bytes, and each is wrong in its own way.

**`&fn_path[..45]` aborts the compiler.** It is a byte index, so when byte 45 falls inside a multi-byte character it panics. Rust identifiers may be non-ASCII (stable since 1.53), so a path can reach here in that shape — and the result is an abort *while emitting a user error*. The diagnostic this box exists to print is replaced by an ICE. Reproduced against the old expression:

```
path = "é".repeat(60)   // 120 bytes, 60 chars, byte 45 mid-character
byte index 45 is not a char boundary; it is inside 'é' (bytes 44..46 of string)
```

**`fn_path.len() > 48` measures the wrong thing.** A 48-character path spelled partly with multi-byte characters is 51 bytes, so it was truncated even though it fits the box:

```
path = "a" * 45 + "é" * 3   // 48 chars, 51 bytes, byte 45 is a boundary
old -> truncated to 45 chars + "..."      new -> left alone
```

## The change

Both go through one character-based helper, extracted so it is testable. `truncate_path_for_box` returns the path unchanged when it fits the character budget, and otherwise keeps `width - 3` characters and appends an ellipsis, so the result never exceeds `width` characters and never splits one.

The other `CollectDecision::Forbidden` arm formats the full path with no slicing and is unaffected.

## Scope

I swept the tree for the same shape — a format width specifier in a file that also guards on `.len() > N` — and `collector.rs` is the only hit. `local_memory_diagnostic.rs:464` (`&full[..40]`) is inside a `#[test]` over ASCII input; `ptx_asm.rs:403` (`&value[1..]`) indexes a fixed constraint set.

## Tests

Four, in `collector`'s existing `mod tests`:

| test | pins |
|---|---|
| `truncating_a_path_for_the_box_respects_char_boundaries` | the panic. Asserts the fixture really does straddle byte 45 (`!path.is_char_boundary(45)`), so the test cannot silently stop covering the bug |
| `a_path_that_fits_is_left_alone` | short paths untouched |
| `a_multibyte_path_within_the_char_budget_is_not_truncated` | the bytes-vs-chars half: 30 chars / 60 bytes must not truncate |
| `truncation_never_exceeds_the_width` | 49/60/200-char inputs all come out at exactly 48 |

## Verification

All local, no GPU needed.

- `cargo test --lib` in `crates/rustc-codegen-cuda`: **27 passed, 0 failed**. CI runs this same command in its `cargo test --lib (rustc-codegen-cuda)` job.
- `cargo fmt --all --check`: clean
- `cargo clippy --all-targets -- -D warnings`: clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps`: clean

One file, `crates/rustc-codegen-cuda/src/collector.rs`. I checked every open PR's file list via `gh api --paginate .../pulls/N/files` — no open PR touches it.

### DIFF
--- a/crates/rustc-codegen-cuda/src/collector.rs
+++ b/crates/rustc-codegen-cuda/src/collector.rs
@@ -559,6 +559,32 @@ pub fn is_fully_monomorphized<'tcx>(tcx: TyCtxt<'tcx>, instance: Instance<'tcx>)
 /// are not part of `core_float_math`; `sin`/`cos` are listed defensively in
 /// case a build ever takes the `std` route too. Only `std::`-prefixed names
 /// belong here; `core`-based shims are already allowed.
+/// Fit a function path into the forbidden-crate diagnostic box.
+///
+/// The box pads this field with `{:<48}`, which counts characters, so the
+/// budget is a character budget. Two things follow, and the previous
+/// `if fn_path.len() > 48 { &fn_path[..45] }` got both wrong for a path that
+/// is not pure ASCII:
+///
+/// * `len()` is bytes. A 30-character path spelled with multi-byte characters
+///   can exceed 48 bytes and be truncated even though it would have fit.
+/// * `&fn_path[..45]` is a byte index. When byte 45 lands inside a multi-byte
+///   character it panics with "byte index 45 is not a char boundary" -- so
+///   emitting the diagnostic for a forbidden call would abort the compiler
+///   instead of printing the error the box exists to print. Rust identifiers
+///   may be non-ASCII, so a path can reach here in that shape.
+///
+/// Truncation keeps `width - 3` characters and appends an ellipsis, so the
+/// result never exceeds `width` characters.
+fn truncate_path_for_box(fn_path: &str, width: usize) -> String {
+    debug_assert!(width > 3, "the ellipsis needs room");
+    if fn_path.chars().count() <= width {
+        return fn_path.to_owned();
+    }
+    let kept: String = fn_path.chars().take(width - 3).collect();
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Confirmed the mismatch: the box pads with `{:<48}` (characters) while the truncation used byte length and a byte index, so a multi-byte path could panic mid-character while printing an error, and a multi-byte path that fit got truncated anyway. The helper plus tests lock both directions; tests, fmt, clippy clean.
- Worth noting: `Forbidden` is built at one site, only for `std` def-paths, which are pure ASCII today, so the ICE is latent; this is hardening of objectively wrong code rather than a user-visible crash fix.
- **One commit layered on:** the helper and its docs were spliced inside `is_intrinsic_lowered_cmath_shim`'s doc block, so the fused comment attached to the helper and the cmath allowlist lost its documentation, including the float_math.rs keep-in-sync note. Moved the helper above the block; no code change.

nihalpasham (approved):

PR #1128, opened 2026-08-21
debug: describe AS1 device globals

### LINKED ISSUE #1120: Full debug: describe ordinary Rust device globals (AS1)

Ordinary Rust device statics are materialized as generated AS1 LLVM globals,
but cuda-oxide emits no `DIGlobalVariable` / `DIGlobalVariableExpression`, no
global `!dbg` attachment, and no compile-unit globals list. cuda-gdb therefore
cannot resolve source globals such as `GLOBAL_COUNTER`.

This is a child of #226 and should establish reusable global-debug plumbing for
the separate AS3 shared-memory task.

**Proposed behavior**

- Preserve the source leaf name, structured Rust scope path, definition span,
  visibility/locality, generated linkage name, and semantic Rust type while
  stable MIR still exposes them. Join/debug-deduplicate through the opaque
  mangled static identity, never the non-injective display path.
- Carry that record through `mir.global_alloc` and LLVM lowering.
- Emit a `DIGlobalVariableExpression`, attach it to the generated AS1 global,
  and retain it in the compile unit's globals list.
- Describe the semantic source type even when initialized physical storage is
  represented as `[N x i8]`.

**Expected debugger behavior**

- A global such as `debuginfo::GLOBAL_COUNTER` is readable/writable by its
  canonical qualified Rust name and has the correct Rust type.
- Same-leaf statics in different modules remain distinct and resolve by their
  qualified Rust names.
- Debugger lookup never silently hides one same-leaf static behind another.

**Acceptance criteria**

- Exact LLVM metadata includes source name, qualified scope, physical
  `linkageName`, declaration file/line, semantic type, alignment, correct
  locality, global attachment, and CU retention.
- Duplicate references to one static deduplicate to one global/debug entry.
- Two same-leaf statics in distinct module/function/block scopes retain
  distinct physical storage, relocation targets, DIEs, values, and addresses;
  repeated references to one definition deduplicate.
- Tests cover initialized byte-array storage and aggregate semantic types.
- Full debug emits variable metadata; line tables/debug off do not.
- AS3 shared and AS4 constant memory remain explicitly out of scope.
- LLVM verifier, llc/ptxas, and live cuda-gdb checks pass.

**Out of scope / follow-up**

Unqualified `print GLOBAL_COUNTER` is contextual in cuda-gdb: it only resolved
in controls where the stopped `DISubprogram` carried the exact same namespace,
and it did not resolve from a child namespace to an ancestor. Until function
scopes use the same structured namespace graph (or an imported-entity/alias
mechanism is validated), the canonical T28 acceptance command is
`print debuginfo::GLOBAL_COUNTER`.

`StaticDef::name()` is not injective: two opposite-block `static X`
declarations in one function can have the same display path. The AS1 physical
storage key, initializer relocation targets, and debug side table must all use
the same exact mangled identity; merely rejecting the debug collision still
leaves physical storage coalesced.

**Current evidence**

On baseline `6c5458fe`, the AS1 allocation exists as `@__device_global_N`, and
its semantic MIR type survives long enough to preserve, but no global DWARF is
emitted. The cuda-devtools T28 assessment reports no source symbol.

**Environment**

- Baseline: `6c5458fe991bbde32c5bee74d87822aef1b5a691`
- CUDA/cuda-gdb: 13.3
- GPU: RTX 5090 (sm_120)

nihalpasham:
## Summary

Ordinary Rust device statics now keep their source name and Rust type all the
way to the generated AS1 storage, so cuda-gdb can inspect them by their
qualified name.

```text
Before
debuginfo::GLOBAL_COUNTER ──> @__device_global_0 ──> no DWARF name/type
cuda-gdb: No symbol

After
debuginfo::GLOBAL_COUNTER ──> @__device_global_0 + DIGlobalVariable
cuda-gdb: @global u64 = 0
```

Closes #1120.

## Changes

- Carry the source leaf, namespace, definition span, locality, and semantic
  Rust type through the full-debug pipeline.
- Use an exact tagged mangled identity for storage, relocations, and debug joins
  so same-leaf statics and promoted constants cannot collide.
- Emit retained AS1 global DWARF with the generated symbol as `linkageName`.
- Fail closed per unsupported type without removing the real global.

Qualified lookup is the correctness boundary. Unqualified lookup is contextual
in cuda-gdb; AS3 shared and AS4 constant-memory debug metadata stay separate.

## Testing

- Importer, lowering, exporter, backend, strict Clippy, LLVM-verifier, and
  debug-mode checks passed on `ac0282c4`.
- Final head `a57bc934` only formats the device-global example; the exact
  per-example formatter gate now passes.
- CUDA 13.3 / RTX 5090: `debuginfo::GLOBAL_COUNTER` is `@global u64 = 0`.
- Same-leaf and opposite-block controls keep distinct values, addresses,
  relocations, and DIEs; repeated references deduplicate.

- [ ] `just check` passes (affected suites and validators pass)
- [x] Live GPU debugger regression exercised
- [x] New example added (existing device-global example extended)

## Checklist

---

### DIFF
--- a/crates/dialect-mir/src/ops/memory.rs
+++ b/crates/dialect-mir/src/ops/memory.rs
@@ -1151,6 +1151,39 @@ impl Verify for MirSharedAllocOp {
 // MirGlobalAllocOp
 // ============================================================================
 
+const RUST_STATIC_GLOBAL_KEY_PREFIX: &str = "cuda-oxide:rust-static:v1:";
+const PROMOTED_GLOBAL_KEY_PREFIX: &str = "cuda-oxide:promoted:v1:";
+
+fn encode_device_global_key(prefix: &str, payload: &str) -> String {
+    format!("{prefix}{}:{payload}", payload.len())
+}
+
+/// Encode rustc's opaque static symbol as a domain-tagged lowering key.
+///
+/// The tag keeps a user-controlled `#[export_name]` from aliasing a
+/// compiler-made promoted allocation with the same string fingerprint. The
+/// length prefix makes the symbol payload unambiguous even when it contains
+/// punctuation.
+pub fn encode_rust_static_global_key(symbol: &str) -> String {
+    encode_device_global_key(RUST_STATIC_GLOBAL_KEY_PREFIX, symbol)
+}
+
+/// Encode one compiler-made promoted allocation fingerprint.
+pub fn encode_promoted_global_key(fingerprint: &str) -> String {
+    encode_device_global_key(PROMOTED_GLOBAL_KEY_PREFIX, fingerprint)
+}
+
+/// Recover the exact rustc symbol payload from a tagged static key.
+///
+/// Constant-memory globals use the payload as their externally visible
+/// linkage name; the tagged key itself remains internal to lowering and
+/// relocation lookup.
+pub fn rust_static_symbol_from_global_key(key: &str) -> Option<&str> {
+    let encoded = key.strip_prefix(RUST_STATIC_GLOBAL_KEY_PREFIX)?;
+    let (length, symbol) = encoded.split_once(':')?;
+    (length.parse::<usize>().ok()? == symbol.len()).then_some(symbol)
+}
+
 /// MIR device-global address operation.
 ///
 /// Represents the address of an ordinary Rust `static` / `static mut` reachable
@@ -1165,7 +1198,7 @@ impl Verify for MirSharedAllocOp {
 /// | Name            | Type        | Description                      |
 /// |-----------------|-------------|----------------------------------|
 /// | `global_type`   | TypeAttr    | Type stored in the global        |
-/// | `global_key`    | StringAttr  | Stable key for deduplication     |
+/// | `global_key`    | StringAttr  | Domain-tagged deduplication key  |
 /// | `global_alignment` | IntegerAttr | Optional alignment            |
 /// | `global_immutable` | UnitAttr | Storage is never written         |
 /// ```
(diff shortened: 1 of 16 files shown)


nihalpasham:
Notes from the rebase onto post-#1186 main, for the record. The rework went beyond conflict resolution in three ways:

**1. Latent #1186 regression found and fixed here.** Full-debug builds of this PR's own `device_global` example failed on main before this landed:

```text
union RelocatedPointerUnion { word: &'static u32, byte: &'static u8 }  // documented example

release build:      compiles (MIR opts fold the load away)
CUDA_OXIDE_DEBUG=full: "initializer cannot preserve pointer provenance:
                        ... mixes reference pointee types"   <- #1186's classifier
```

The rejection is load-bearing on the SSA-reconstruction path (it mints a value of the carrier field's reference type) but was firing on the physical-storage path too, where no typed reference is ever created: the static lowers to raw bytes plus one integer relocation slot, and provenance rides the relocation symbolically (`ptrtoint (addrspacecast (getelementptr @target, addend)))`). The classifier now takes a usage flag: reconstruction keeps the rejection, storage does not. Verified adversarially down to the emitted IR/PTX; negative tests cover both paths.

**2. Structure folded into #1186 instead of duplicating it.** #1186 independently landed `DeviceGlobalRecord`/`DeviceGlobalDeclaration` with equality-based cache validation, so this PR's parallel record and validator were dropped; debug identity is now a `debug_info` field on main's declaration and rides the existing equality check. The promoted-key fingerprint wrapper follows main's new alignment-carrying format.

**3. Small #1126 catch-up:** `type_is_bounded` in the debug-info encoder gained the `TypedPointer` arm (recurses into the pointee), required to compile against the new debug-type node.

Follow-ups noted, not blocking: (a) the mixed-union rejection on the SSA path is covered at unit level only; an end-to-end negative driving `translate_union_constant_from_alloc` would keep a future flag-flip honest. (b) Pre-existing and unrelated: GPU-less target detection falls back to `sm_80`, which can't hold this example's pointer initializers (needs PTX ISA 7.1), so `CUDA_OXIDE_TARGET` matters on GPU-less machines.

PR #1129, opened 2026-08-21
debug: describe AS3 shared statics

### LINKED ISSUE #1121: Full debug: describe static shared-memory objects (AS3)

A function-local `static mut TILE: SharedArray<T, N>` materializes real AS3
backing storage, but its source name reaches LLVM only as a diagnostic comment.
No global DWARF describes the object, so cuda-gdb reports `No symbol 'TILE'`.

This is a child of #226 and should build on the common global-DI plumbing from
the AS1 device-global task.

**Proposed behavior**

- Join a named shared static to its allocation using the exact opaque mangled
  static identity, while carrying its human name/scope/span separately. Do not
  use `StaticDef::name()`: two block-local same-leaf statics can share that
  spelling.
- Carry the source identity, owning device function, declaration span, and
  physical layout to the generated `@__shared_mem_N` allocation.
- Scope the global DIE to the owning function's single `DISubprogram`. A flat
  or namespace-only scope does not make `TILE` visible to `info locals`.
- Emit and retain an AS3 `DIGlobalVariableExpression` with NVPTX's explicit
  shared address-class expression:
  `DIExpression(DW_OP_constu, 8, DW_OP_swap, DW_OP_xderef)`.
- Describe the compiler-materialized `[N x T]` backing allocation, not the
  Rust `SharedArray<T, N>` marker's zero-sized `PhantomData` layout.

**Expected debugger behavior**

After the relevant barrier, cuda-gdb includes `TILE` in `info locals`, resolves
bare `whatis TILE` / `print TILE[i]`, and reports the correct physical array
shape. Values retain block-local shared-memory semantics. The final DWARF and
cuda-gdb's C++ language view must identify address class 8 / `@shared`; the
Rust formatter may omit that textual qualifier. A namespace-only DIE makes a
qualified spelling work but fails the existing T27 local-variable contract, so
qualified lookup is not a substitute for function scope.

**Acceptance criteria**

- LLVM AS3 global has a retained debug attachment and CU global entry, and the
  DIE reuses the exact `DISubprogram` node emitted for its owning function.
- Metadata preserves source name `TILE`, generated linkage name, declaration
  file/line, physical `[N x T]` backing type, size, and alignment; the
  independently synthesized debug type must exactly match the materialized
  element/count/alignment.
- LLVM contains the explicit class-8 `constu/swap/xderef` expression; cubin
  DWARF and cuda-gdb's C++ view report shared address class 8 / `@shared`, and
  the Rust view reports `[T; N]` with the correct byte size/address.
- `info locals`, `whatis TILE`, and bare element reads succeed at the T27 stop.
- Repeated uses of one static deduplicate; two same-leaf statics in different
  blocks of one function retain distinct exact identities, linkage names,
  declaration lines, storage, and DIEs.
- Live tests switch between blocks and confirm different per-block values at
  the same shared offset.
- Unsupported element-type graphs fail closed without suppressing the physical
  shared allocation or breaking full-debug compilation.
- `Barrier` allocations either receive independently validated metadata in
  this task or are explicitly deferred with a negative regression test.
- Full debug emits variable metadata; line tables/debug off do not.
- LLVM verifier, llc/ptxas, and live cuda-gdb checks pass.

**Current evidence**

On baseline `6c5458fe`, `MirSharedAlloc` lowers to an AS3
`@__shared_mem_N`. The source path is explicitly diagnostic-only and exporter
output contains no `DIGlobalVariable*`. The cuda-devtools T27 assessment reports
no `TILE` symbol.

A reviewed implementation is available as stacked branch
`feat/debug-as3-shared-statics` at
`d516d6fb982f61877a7a70219decb9ec3690ca37`, based on the repaired AS1
prerequisite `a57bc934b39bf1710671da9c1602afdc99aec3aa`. The AS3 patch ID and
parent-to-child binary diff are unchanged from the independently reviewed
pre-restack commit.

**Environment**

- Baseline: `6c5458fe991bbde32c5bee74d87822aef1b5a691`
- CUDA/cuda-gdb: 13.3
- GPU: RTX 5090 (sm_120)

nihalpasham:
## Summary

Function-local `SharedArray` and `Barrier` statics now appear as real shared
objects in cuda-gdb, with the physical array shape and NVPTX shared address
class.

```text
Rust source                         Debug metadata / cuda-gdb
static mut TILE: SharedArray<i32, 32> = SharedArray::UNINIT;
          │
          └─> @__shared_mem_0 ──> [i32; 32], address class 8
                                  info locals: TILE = [...]
                                  print TILE[1] = 1 (block 0)
                                                  101 (block 1)
```

Implements #1121. This is stacked on #1128, so GitHub cannot close #1121 yet;
change this to `Closes #1121` after retargeting to `main`.

## Changes

- Join named shared statics to AS3 storage with their exact identity, layout,
  declaration, and owning function.
- Describe the materialized `[T; N]` backing array, not `SharedArray`'s
  zero-sized Rust marker.
- Reuse the owner's single `DISubprogram` and emit the explicit NVPTX shared
  address-class expression.
- Fail closed on layout/owner/type conflicts; keep physical storage intact.
- Add a persistent `shared_debug` example and verifier.

The Rust formatter shows `[T; N]` but may omit the text `@shared`; cubin
DWARF and cuda-gdb's C++ view confirm address class 8. Multi-owner inlined
materialization remains a follow-up.

## Testing

- Final head `d516d6fb` is the identical reviewed AS3 patch restacked on
  #1128's fmt-only repair; focused/full exporter, backend, and exact formatter
  gates passed.
- LLVM 21/22 → `llc` → `ptxas` → cubin checks retain six class-8 DIEs.
- CUDA 13.3 / RTX 5090: bare `TILE`, `info locals`, size 128, and per-block
  values all work; Off/LineTables output stays byte-identical to the AS1 base.

- [ ] `just check` passes (affected suites and toolkit gates pass)
- [x] Live GPU debugger regression exercised
- [x] New example added (`shared_debug`)

## Checklist

---


nihalpasham:
Notes on what changed during the rebase onto post-#1186/#1128 main, beyond plain conflict resolution:

**1. The branch didn't compile on new main: two `get_type_size` calls replaced.** #1186 deleted that helper (it guessed 8 bytes for unknown types). `shared_debug_type_matches_physical` now uses `llvm_type_size_align` and treats "size unknown" as "debug type doesn't match", so the metadata is dropped instead of guessed:

```text
before:  size = get_type_size(elem)        // 8 on unknown, silent
after:   llvm_type_size_align(elem)?        // exact, or no debug info
```

Side effect: padded-struct elements now compare with their real (padded) size.

**2. Cache validation folded into #1186's structure instead of a parallel walk.** Same move as #1128's rework: `SharedGlobalDeclaration` gained `debug_info` + `debug_owner_function` fields, and the existing equality check covers them. The PR's `validate_shared_cache_hit` module walk is gone.

**3. Owner conflict is now fail-open (review item).** One function-local shared static reaching two functions used to hard-error in full-debug builds. Debug metadata should never fail a build that release accepts, so:

```text
fn a() { static S: SharedArray<...> }  ─┐   physical storage matches:
fn b() { /* same S via inlining */   } ─┤   one global, debug attachment
                                        └─► DROPPED, build compiles
   (physical MISMATCH on the same key is still a hard error)
```

New test covers the drop directly.

**4. Catch-ups forced by the PRs that landed underneath:**
- #1127 (source vs linkage names): the shared-scope prepass and `debug_subprogram_for_function` now key the subprogram cache by the exported linkage name, so the reserved owner `DISubprogram` and the definition export meet at one node; verify script regexes updated for `linkageName:`.
- #1126 (typed pointers): `SharedArray<*mut u32>` and struct-of-typed-pointer now get real debug types instead of the old blanket rejection; a new `*mut ContainsPointer` case keeps the fail-closed coverage.

**Validation on the final head:** full workspace suites, workspace clippy `-D warnings`, per-example lint, `CUDA_OXIDE_DEBUG=full` build + verify script (llc → ptxas → llvm-dwarfdump cubin loop, 6 DIEs with `DW_AT_address_class 0x08`), and live GPU runs of both `shared_debug` and #1125's `debug` example on the RTX 5090, the first time all four debug features ran together.

PR #1131, opened 2026-08-21
fix(codegen): support char in #[device] extern signatures

### LINKED ISSUE #875: `char` is rejected in `#[device]` extern signatures

arpanpathak:
Closes #875

## What

`TyKind::Char` is now mapped through the same `unsigned_integer(32)` helper as `u32` in `rustc_ty_to_device_extern_type`. A `#[device] extern "C"` function can now pass and return `char` instead of failing with the old "Rust `char` is not supported in device extern signatures" error.

Rust `char` is a 4-byte Unicode scalar value, so it is ABI-compatible with `u32`. It crosses the boundary as a plain i32 with no signext/zeroext in parameter, result, and pointee positions.

## Test

Extends `device_ffi_test` with a `char_identity` round-trip:

- CUDA C++ side: `extern "C" __device__ unsigned int char_identity(unsigned int c)`
- Rust side: `unsafe extern "C" { fn char_identity(c: char) -> char; }` with `#[allow(improper_ctypes)]`
- Kernel passes `'A'` and verifies it comes back as `'A'` (65)

Generated NVVM IR:

```
declare i32 @char_identity(i32)
call i32 @char_identity(i32 65)
```

## Local validation (Jetson Orin, sm_87)

Ran the full example on a Jetson Orin with CUDA 12.6:

```
--- Test 5: test_char_ffi ---
    Function: char_identity (Rust `char` <-> C `unsigned int`)
    ✓ PASSED (char_identity('A') = 'A')

=== Summary ===
Passed: 5
Failed: 0

✓ All tests PASSED!
```

The upstream example's LTOIR tools and CCCL wrappers assume CUDA 13 APIs (`NVJITLINK_ERROR_*` enum values and `cuda::minimum` / `cuda::maximum`). To build on this CUDA 12.6 host I used temporary local shims that are not part of this PR. CI on CUDA 13 should exercise the same path.

## Docs

Documents the ABI contract in the `device_ffi_test` README:

- A value from C outside the Unicode scalar range (above 0x10FFFF, or a surrogate in 0xD800..=0xDFFF) is UB on the Rust side, the same caveat family as bool's 0/1 rule.
- rustc's `improper_ctypes` lint still fires on `char` externs; callers may `#[allow(improper_ctypes)]`.

## Note on duplicate PR #1049

PR #1049 also addresses #875. This PR is based on the latest `main` and includes the two changes requested in the review of #1049: `DeviceBuffer::zeroed` instead of `stream.alloc_zeros`, and `#[allow(improper_ctypes)]` on the char extern. It also includes the two rustfmt reflow hunks in `device_codegen.rs` required by `cargo fmt --check`.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/device_ffi_test/README.md
+++ b/crates/rustc-codegen-cuda/examples/device_ffi_test/README.md
@@ -59,6 +59,12 @@ Device: NVIDIA GeForce RTX 5090
 --- Test 3: test_mixed_attrs ---
     ✓ PASSED (outputs are finite)
 
+--- Test 4: test_smem_alignment_cross_module ---
+    ✓ PASSED (same shared memory, values correct)
+
+--- Test 5: test_char_ffi ---
+    ✓ PASSED (char_identity('A') = 'A')
+
 ✓ All tests PASSED!
 ```
 
@@ -177,6 +183,7 @@ extern "C" __device__ float warp_reduce_sum(float val) {
 | `warp_ballot`        | Warp ballot             |
 | `simple_add`         | Simple a + b            |
 | `clamp_value`        | Clamp to range          |
+| `char_identity`      | Rust `char` <-> `unsigned int` round-trip |
 
 ### From `extern-libs/cccl_wrappers.cu` (CUB)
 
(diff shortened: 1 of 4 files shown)


nihalpasham:
- Main still rejects `char`; this behavior has not already landed.
- This duplicates the earlier #1049 implementation across the same four compiler/test/doc paths. #1049 also carries the stronger end-to-end test, so keeping that branch preserves the earlier contributor's implementation and credit.
- #1049 still needs three repairs before merge: zero-initialize `DeviceBuffer`, allow the intentional `improper_ctypes` boundary, and add the contributor DCO trailer.
- Local validation of this branch passes fmt, clippy, all 27 codegen tests, the compile smoke test, and cubin materialization.
- I labeled this PR `duplicate`; #875 stays open until the surviving implementation lands.

nihalpasham:
Closing as duplicate.

PR #1132, opened 2026-08-22
ci(licenses): the guard's own workspace-root list is a claim nothing checks

vyncint:
Follow-up to #1114, which your review asked for: *"This drift has now happened twice (#662, here), so a self-check comparing the pass list against on-disk `[workspace]` manifests is follow-up material."*

## What is wrong

`check-dependency-licenses.sh` reaches each first-party workspace with one `declared_crates` pass per root, because `cargo metadata` stops at a `[workspace]` boundary. **Which roots exist was hard-coded, and nothing compared that list against the tree.** That is how the guard has failed twice:

- **#662** found `crates/rustc-codegen-cuda` unread. The backend — the largest first-party crate in the tree — had no CSV row while every other member did.
- **#1114** found `crates/cuda-macros/tests/device-only` unread, and its member `device-only-kernels` likewise unrecorded.

Both times the guard printed OK. A short list is indistinguishable from a complete one, a count cannot tell them apart, and the comment reading "Both first-party workspaces" was itself the wrong number for months.

## The change

The list becomes two named arrays, compared against `git ls-files -- '*Cargo.toml'` filtered to manifests declaring `[workspace]`, in both directions:

- a manifest on disk that no pass reads fails, naming it, and says to add it as first-party or as vendored;
- a name in the arrays that is no longer such a manifest fails as stale, since a renamed root left behind is a permanent hole.

`crates/fuzzer/rustlantis` is the one vendored root, listed separately with the reason: third-party code attributed in THIRD_PARTY_NOTICES whose dependencies are not ours to record, excluded by `check-spdx-headers.sh` on the same grounds. Roots under `crates/rustc-codegen-cuda/examples` stay out of scope here — the second half of the script already resolves those, including the two nested sub-workspaces, from their committed lock files.

## Verification

All local, no GPU.

| mutation | result |
|---|---|
| drop the device-only root from the list | fails, naming that manifest |
| add a name that is not a manifest | fails as stale, naming it |
| a new tracked `[workspace]` manifest appears (`git add -N`) | fails, naming it |
| the `[workspace]` scan pattern rots | refuses: *"the scan broke, so a clean result means nothing"* |

Unmutated: `OK: dependency-licenses.csv records all 59 declared crates`, example half unchanged, `bash -n` clean. `git` joins the `require_tool` list next to `cargo` and `python3`.

One file. Checked against every open PR's file list via `gh api --paginate .../pulls/N/files` — no overlap.

## Maintainer repair

- Cargo now resolves every tracked manifest: 264 manifests -> 222 workspace roots.
- The check compares four named non-example roots and 218 example root/lock pairs in both directions.
- Missing, orphaned, stale, malformed, and untracked-lock mutations all fail closed.
- Full license guard, `bash -n`, and diff checks pass.

### DIFF
--- a/scripts/check-dependency-licenses.sh
+++ b/scripts/check-dependency-licenses.sh
@@ -39,6 +39,7 @@ require_tool() {
 }
 require_tool cargo
 require_tool python3
+require_tool git
 
 test -s "${CSV}"
 
@@ -96,8 +97,9 @@ print("\n".join(sorted(names)))
 
 # Every first-party workspace root, not just the root one.  `cargo metadata`
 # stops at a `[workspace]` boundary, so one pass per root is the only way to
-# reach them all, and the tree has three under version control outside
-# `examples/` (the second half below covers those):
+# reach them all.  The list is no longer a claim: the self-check below compares
+# it against every tracked `[workspace]` manifest outside `examples/`, which the
+# second half covers.  Today that is three, and how each got here matters:
 #
 #   1. Cargo.toml -- the root workspace.
 #   2. crates/rustc-codegen-cuda/Cargo.toml -- its own `[workspace]` for the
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Reproduced two false greens: valid `[ workspace ]` spelling was missed, and an example workspace could rely on an untracked lock file.
- I layered one commit that asks Cargo to resolve all 264 tracked manifests, then checks four non-example roots and 218 example root/lock pairs in both directions.
- Missing, orphaned, stale, malformed, and untracked-lock mutations fail. The license guard, shell check, and diff check pass; hosted CI is queued and waiting for explicit fork-run approval.
- This is the follow-up requested in #1114; no issue closes on merge.

nihalpasham (approved):

PR #1134, opened 2026-08-22
fix(ptx-schedule): the marker predicates fold case only if the caller does

vyncint:
Follow-up to #1112, whose review named this residual: *"`has_skip_marker`'s case-insensitivity relies on callers pre-lowercasing."*

## What is wrong

`has_skip_marker` and `has_mismatch_marker` implement a case-blind contract — the smoketest greps the same markers with `grep -i` — but neither folded case. They worked because their one caller lowercased first:

```rust
let combined = format!("{stdout}\n{stderr}").to_ascii_lowercase();
```

That is a precondition living at a call site, in a crate whose one job is classifying process output. Called with a raw stream, both predicates silently disagree with their own names — measured on this tree:

```
has_skip_marker("Skipping: needs sm_90")            -> false
has_skip_marker("PASS (skipped): requires sm_75+")  -> false
has_mismatch_marker("MISMATCH at index 3")          -> false
```

Every one of those spellings is real. `Skipping:` and `PASS (skipped):` are what examples print — `generated_ldmatrix` prints the second verbatim — and `MISMATCH` in capitals is the conventional spelling for the finding, the one `crates/fuzzer` uses in its own status vocabulary. A caller reaching for these predicates without knowing the rule would classify a declined run as a pass and a failed comparison as a pass: the exact misclassification #1112 fixed.

## The change

Fold case in the predicates instead of documenting the precondition. `has_skip_marker` compares the line's leading bytes with `eq_ignore_ascii_case`; `has_mismatch_marker` lowercases once inside itself. `run_binary` then passes the raw combined stream, so the precondition is **gone rather than moved** — and the one allocation moves from the call site into the predicate that needs it. Behaviour on the existing path is unchanged: the input was already lowercase there.

## Verification

One test over the spellings that matter, both directions: five mixed-case skip declines and five mixed-case failure markers must match, and folding case must not widen what counts as a marker — a bare `pass`/`PASS`, `pass: 1024 elements verified`, `SUCCESS`, a mid-line `no skipping: here`, and `all checks passed` all stay negative. It fails with the folding reverted.

15 tests pass, fmt clean, `clippy --all-targets -- -D warnings` clean, `cargo doc --document-private-items` clean.

One file. No open PR touches it.

## Maintainer repair

- Classification is now `timeout -> nonzero exit -> mismatch -> skip -> pass`.
- Case folding stays inside the marker predicates.
- Mixed-marker tests pin the stronger verdict.
- 16 tests, fmt, strict clippy, rustdoc, and diff checks pass.

### DIFF
--- a/crates/ptx-schedule/src/campaign.rs
+++ b/crates/ptx-schedule/src/campaign.rs
@@ -730,7 +730,9 @@ fn run_binary(executable: &Path, cwd: &Path, timeout: Duration) -> RunResult {
         Ok(status) => (status.code(), status.success(), stderr),
         Err(error) => (None, false, error.to_string()),
     };
-    let combined = format!("{stdout}\n{stderr}").to_ascii_lowercase();
+    // Both marker predicates below fold case themselves, so this is the raw
+    // combined stream rather than a lowercased copy.
+    let combined = format!("{stdout}\n{stderr}");
     let kind = if timed_out {
         RunKind::Hang
     } else if has_skip_marker(&combined) {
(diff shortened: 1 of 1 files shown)


nihalpasham:
- Reproduced the case bug and the stronger failure being hidden by skip text.
- I layered one commit that folds case inside each predicate and fixes the order to timeout → nonzero exit → mismatch → skip → pass.
- Mixed-marker tests pin that order. All 16 tests, fmt, strict clippy, and rustdoc pass; hosted CI is queued and waiting for explicit fork-run approval.
- This follows #1112; no issue closes on merge.

nihalpasham (approved):

PR #1135, opened 2026-08-22
fix(ptx-schedule): an example that declines to run is reported as a failure

vyncint:
## What is wrong

`RunKind::Skipped` became reachable in #1112, when `has_skip_marker` learned the second spelling examples use to decline. **Nothing consumes it distinctly.** Both places that read the baseline test for `Pass` and treat everything else alike:

```rust
// campaign.rs
if !matches!(baseline.kind, RunKind::Pass) {
    ... println!("schedule-fuzz: BASELINE FAILURE: {:?}; ...")

// cargo-oxide/src/main.rs
Ok(summary) if !matches!(summary.baseline.kind, RunKind::Pass) => {
    eprintln!("Error: baseline example did not pass; ...");
    std::process::exit(1);
}
```

So `cargo oxide fuzz-schedule redux_f32` on anything below sm_100a prints `BASELINE FAILURE: Skipped`, then `Error: baseline example did not pass`, and exits 1.

The example did not fail. It ran, declared it cannot run on this device, and exited 0 — which is precisely what `scripts/smoketest.sh` reports as `PASS (skipped)`. **#665 exists because reporting a decline as anything else made an arch-gated example indistinguishable from a broken one**, and this is the same conflation in the campaign: sweeping a fleet, "below this example's arch floor" comes back red alongside a genuinely wedged device.

## The change

Rather than repeat a `matches!` in two crates, the distinction becomes one typed definition. `RunKind::baseline_verdict()` returns `Usable`, `Declined` or `Broken`, and the campaign and the CLI both consult it. The match is exhaustive, so a new `RunKind` cannot join a class by accident — adding one now fails with `E0004: non-exhaustive patterns` instead of silently reading as broken.

Behaviour:

| baseline | before | after |
|---|---|---|
| `Skipped` | `BASELINE FAILURE: Skipped`, exit **1** | `BASELINE DECLINED`, a `Note:` on stderr, exit **0** |
| any other non-`Pass` | `BASELINE FAILURE: {kind}`, exit 1 | unchanged, and the CLI message now names the kind |

`--fail-on-finding` is untouched and remains the switch for failing on real findings. **If you would rather a decline carry a distinct nonzero code, that is a one-line change in the new arm** — I picked 0 to match `verdict_standard`, but it is your call.

## Verification

Two tests: the three classes pinned exhaustively over all eight `RunKind` variants, and that every *finding* kind is a `Broken` baseline — a mismatch is a finding in a variant, but as the baseline it means there is no ground truth to compare against. The first fails with `Skipped` moved back into the broken class; adding a ninth variant fails to compile.

16 ptx-schedule tests and 177 cargo-oxide tests pass, fmt clean, `clippy --all-targets -- -D warnings` clean on both crates, `cargo doc --document-private-items` clean.

Two files. No open PR touches either. It shares `campaign.rs` with my #1134 and #1137; the three touch disjoint regions and I verified all six of this batch merge to a byte-identical tree in either order.

## Maintainer repair

- Strong failures now beat skip text.
- The CLI policy is explicit: declined exits 0 with zero seeds, broken exits 1, and usable findings obey `--fail-on-finding`.
- 17 ptx-schedule and 178 cargo-oxide tests pass, with fmt, strict clippy, rustdoc, and diff checks.

### DIFF
--- a/crates/cargo-oxide/src/main.rs
+++ b/crates/cargo-oxide/src/main.rs
@@ -610,6 +610,31 @@ fn validate_materialization_cli(cli: &Cli) -> Result<(), String> {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum FuzzScheduleDisposition {
+    Success,
+    Declined,
+    BrokenBaseline,
+    Findings(usize),
+}
+
+fn fuzz_schedule_disposition(
+    baseline: ptx_schedule::campaign::BaselineVerdict,
+    finding_count: usize,
+    fail_on_finding: bool,
+) -> FuzzScheduleDisposition {
+    use ptx_schedule::campaign::BaselineVerdict;
+
+    match baseline {
+        BaselineVerdict::Declined => FuzzScheduleDisposition::Declined,
+        BaselineVerdict::Broken => FuzzScheduleDisposition::BrokenBaseline,
+        BaselineVerdict::Usable if fail_on_finding && finding_count > 0 => {
+            FuzzScheduleDisposition::Findings(finding_count)
+        }
+        BaselineVerdict::Usable => FuzzScheduleDisposition::Success,
+    }
+}
+
 fn main() {
     // Handle both invocation methods:
     // 1. Cargo subcommand: `cargo oxide run vecadd` → argv = ["cargo-oxide", "oxide", "run", "vecadd"]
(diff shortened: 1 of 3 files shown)


nihalpasham:
- A skipped baseline had its own verdict, but skip text could still hide a nonzero exit or mismatch, and the CLI exit policy was not tested.
- I layered one commit that fixes the shared classifier and pins the CLI behavior: declined runs execute zero seeds and exit 0; broken baselines exit 1; usable findings obey `--fail-on-finding`.
- All 17 `ptx-schedule` tests, 178 `cargo-oxide` tests, fmt, strict clippy, and rustdoc pass. Hosted CI is queued and waiting for explicit fork-run approval.
- This builds on #1112; no issue closes on merge.

nihalpasham (approved):
- Declined baselines now stay distinct from broken baselines and real findings.
- Failure precedence and CLI exit behavior are pinned by regression and control tests.
- All seven workflows pass on exact head `327e8d64`; both commits have valid signatures and DCO trailers.

PR #1136, opened 2026-08-22
fix(ptx-schedule): the usage line brackets an option the CLI requires

vyncint:
Follow-up to #1109, whose review flagged it: *"the brackets mark the flag optional, but outside `--list-sites` the output path is required (main.rs exits with the usage message when it is missing). Candidate for a later polish pass."*

## What is wrong

```
usage: ptx-schedule INPUT.ptx [--list-sites] ... [-o|--output OUTPUT] ...
```

One bracketed list cannot state this CLI's shape: `--output` is required to perturb and meaningless when listing, so bracketing it is wrong in one mode and misleading in the other.

It becomes one line per invocation form, which is what `cuda-intrinsics-gen`'s usage text already does for the same reason:

```
usage:
  ptx-schedule INPUT.ptx --list-sites
        print the schedule-sensitive sites as JSON; the options below are ignored
  ptx-schedule INPUT.ptx -o|--output OUTPUT [--seed N] [--intensity F]
        [--max-sleep-ns N] [--focus TEXT] [--decisions-json FILE]
        write a perturbed copy of INPUT.ptx to OUTPUT
```

## The second half of the same problem

Every way to get the command line wrong printed the identical block with nothing said about which way. `usage_because` prefixes the reason, and the six copies of `args.next().unwrap_or_else(|| usage())` — each meaning "this flag needs a value", none saying so — become one `next_value` helper that names the flag. Before, all four of these printed the same wall of text:

```
ptx-schedule                   error: an input PTX file is required
ptx-schedule x.ptx --seed      error: --seed needs a value
ptx-schedule x.ptx --bogus     error: unrecognized argument "--bogus"
ptx-schedule x.ptx --seed 7    error: -o|--output is required unless --list-sites is given
```

Exit code stays 2 in every case. Both working forms are unchanged, verified against a real module: `--list-sites` prints the JSON, and `--seed 7 -o out.ptx` reports `sites=2 injected=2`.

## Verification

Two tests. The first turns #1109's hand-checked parity into an assertion: the usage text names exactly the long flags the parser accepts, both directions, reading the arms out of the source — an arm can name two spellings (`"-o" | "--output"`), so it takes every quoted flag on the line. The second pins the shape this PR is about: `--output` is not bracketed on the form that requires it, and does not appear on the listing form. They fail respectively when a flag is dropped from the text and when the brackets are put back.

The README needs no change — its two examples are already the two forms, one with `-o` and one without.

16 tests pass, fmt clean, `clippy --all-targets -- -D warnings` clean, `cargo doc --document-private-items` clean. One file; no open PR touches it.

## Maintainer repair

- Clap models exactly two forms: list sites, or rewrite to an output file.
- Malformed, missing, and conflicting arguments exit 2 before file access.
- 14 crate tests and 3 CLI integration tests pass, with fmt, strict clippy, rustdoc, and diff checks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1086,6 +1086,7 @@ version = "0.1.0"
 name = "ptx-schedule"
 version = "0.1.0"
 dependencies = [
+ "clap",
  "dialect-ptx",
  "oxide-artifacts",
  "ptx-parse",
--- a/crates/ptx-schedule/Cargo.toml
+++ b/crates/ptx-schedule/Cargo.toml
@@ -12,6 +12,7 @@ description = "Structural PTX schedule analysis and deterministic perturbation"
 readme = "README.md"
 
 [dependencies]
+clap = { version = "4", features = ["derive"] }
 dialect-ptx = { workspace = true }
 oxide-artifacts = { workspace = true, features = ["object-read"] }
 ptx-parse = { workspace = true }
--- a/crates/ptx-schedule/src/main.rs
+++ b/crates/ptx-schedule/src/main.rs
@@ -8,16 +8,44 @@ use std::env;
 use std::fs;
 use std::path::PathBuf;
 
+/// The two invocation forms, one per line.
+///
+/// A single bracketed list cannot state this CLI's shape: `--output` is
+/// required to perturb and meaningless to list, and bracketing it read as
+/// optional in both. One line per form is what `cuda-intrinsics-gen`'s usage
+/// text does for the same reason.
+const USAGE: &str = "usage:
+  ptx-schedule INPUT.ptx --list-sites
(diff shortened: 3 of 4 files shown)


nihalpasham:
- Reproduced the parser failures: another flag could be consumed as a value, parse errors returned 1, and list mode accepted rewrite-only options.
- I layered one commit that models the two forms with Clap and rejects invalid combinations before file access.
- Integration tests cover missing values, flag-as-value, invalid and unknown flags, conflicting modes, no-output guarantees, and both valid forms. All 14 crate tests, 3 CLI tests, fmt, strict clippy, and rustdoc pass; hosted CI is queued and waiting for explicit fork-run approval.
- This is the follow-up requested in #1109; no issue closes on merge.

nihalpasham (approved):

PR #1143, opened 2026-08-23
feat(codegen): forward assert-bounded result-bundle indexing

### LINKED ISSUE #1142: Forward assert-bounded compiler result-bundle indexing
## Problem

Compiler-owned result bundles can currently forward terminal dynamic array
projections when the index is explicitly bounded as an unsigned `value % C`.

A common equivalent shape produced by Rust bounds checks is instead:

```text
assert(index < C);
bundle[index]
```

When `C` is a small constant within the bundle width, this also gives a finite
candidate set, but `forward_compiler_result_bundles` currently fails closed and
keeps the aggregate memory path.

## Proposed scope

Extend compiler result-bundle forwarding to recognize the same narrow
assert-derived bound already handled by borrowed aggregate read scalarization.

Accept only an unsigned runtime index with an exact preceding assertion of the
form:

```text
assert(index < C)
```

where:

* the assertion is the unique predecessor of the load block;
* the assertion's success edge targets that load block;
* the comparison uses the exact same index value;
* `C` is an integer constant with the same type as the index;
* `C > 0`;
* `C <= array length`;
* `C <= MAX_SCALARIZED_CANDIDATES`.

For an accepted assertion-derived bound, canonicalize the index to:

```text
index % C
```

before creating `mir.extract_array_element`, allowing the existing bounded
array-extraction lowering to remain unchanged.

## Non-goals

This should not introduce general range analysis or recognize broader forms such
as:

* `index <= C`;
* reversed comparisons;
* nested boolean conditions;
* assumptions;
* switch-derived ranges;
* loop induction ranges;
* bit-mask bounds.

Unsupported forms should continue to fail closed.

## Acceptance criteria

* Exact `assert(index < C)` bounds can forward compiler-owned result bundles.
* Existing `% C` forwarding behavior remains unchanged.
* Invalid, ambiguous, signed, dynamic, oversized, or unrelated assertions fail
  closed.
* The generated `ldmatrix` example exercises the assertion-derived path through
  the rustc-to-PTX pipeline.
* The resulting PTX does not require a local-memory round trip for the selected
  result bundle.

uurl:
Closes #1142 

## Summary

Extend compiler result-bundle forwarding to support terminal dynamic array indexing whose candidate set is proven by an exact bounds assertion:

```text
assert(index < C);
bundle[index]
```

This complements the existing unsigned `index % C` path added in #1005.

## Implementation

`forward_compiler_result_bundles` now accepts an assertion-derived dynamic index only when all of the following are proven:

* the index has an unsigned integer type;
* the load block has exactly one predecessor;
* that predecessor terminates with `mir.assert`;
* the assertion's success edge targets the load block;
* the asserted condition is exactly `mir.lt(index, C)`;
* the comparison uses the exact same index value;
* `C` has the same type as the index;
* `C` is a constant;
* `0 < C <= array length`;
* `C <= MAX_SCALARIZED_CANDIDATES`.

For this case the pass materializes:

```text
bounded_index = mir.rem(index, C)
result = mir.extract_array_element(array, bounded_index)
```

This reuses the existing bounded `mir.extract_array_element` lowering rather than adding a second scalarization mechanism.

The proof intentionally mirrors the narrow assertion shape already used by borrowed aggregate read scalarization from #650.

## Fail-closed behavior

The pass still rejects unsupported forms, including:

* unrelated assertions;
* dynamic assertion bounds;
* signed indices;
* bounds larger than the array;
* candidate sets larger than `MAX_SCALARIZED_CANDIDATES`;
* unbounded dynamic indices.

No general range analysis is introduced.

## End-to-end coverage

The `generated_ldmatrix` example now exercises both bounded dynamic forms:

* the existing remainder-bounded selection;
* an assertion-bounded selection using a runtime kernel argument.

Both paths select one register from the compiler-owned `ldmatrix.m8n8.x4.b16` result bundle and verify the result on the host.

## Validation

```text
cargo fmt --all
cargo fmt --all -- --check
git diff --check

cargo test -p mir-transforms
cargo test -p mir-lower

cargo clippy \
  -p mir-transforms \
  -p mir-lower \
  -p cuda-oxide-codegen \
  --all-targets -- \
  -D warnings

cargo oxide run generated_ldmatrix

CUDA_OXIDE_NO_OPT=1 \
  cargo oxide run generated_ldmatrix
```

Results:

* `mir-transforms`: all tests passed, including the new assertion-bound positive and fail-closed cases.
* `mir-lower`: all tests passed.
* `cuda-oxide-codegen`: all tests passed.
* clippy passed with `-D warnings`.
* `generated_ldmatrix` passed on `sm_120` with normal optimization.
* `generated_ldmatrix` passed on `sm_120` with `CUDA_OXIDE_NO_OPT=1`.

Runtime output included:

```text
PASS: all 32 remainder-bounded dynamic ldmatrix selections matched on sm_120
PASS: all 32 assert-bounded dynamic ldmatrix selections matched on sm_120
```
PTX inspection also reports no local-memory round trip for the generated
result-bundle selection:

```text
st.local: 0
ld.local: 0
```

## Related

* #1005: bounded compiler result-bundle indexing via unsigned `% C`
* #650: exact assertion-derived bounds for borrowed aggregate reads

### DIFF
--- a/crates/mir-transforms/src/forward_compiler_result_bundles.rs
+++ b/crates/mir-transforms/src/forward_compiler_result_bundles.rs
@@ -22,7 +22,8 @@
 //! * the alloca has no other stores, copies, escapes, or unknown pointer users;
 //! * every read is a non-volatile load through constant projections, or through
 //!   one terminal array projection whose unsigned runtime index is proven to be
-//!   `value % C` with a small constant candidate set;
+//!   `value % C` or guarded by an exact `assert(index < C)` with a small constant
+//!   candidate set;
 //! * the producer/store dominate every rewritten load.
 //!
 //! Any failed proof leaves the IR untouched.  Ordinary Rust aggregates are
@@ -31,9 +32,9 @@
 use dialect_mir::{
     attributes::{COMPILER_RESULT_BUNDLE_ATTR_KEY, CompilerResultBundleAttr},
     ops::{
-        MAX_SCALARIZED_CANDIDATES, MirAllocaOp, MirArrayElementAddrOp, MirConstantOp,
+        MAX_SCALARIZED_CANDIDATES, MirAllocaOp, MirArrayElementAddrOp, MirAssertOp, MirConstantOp,
         MirConstructArrayOp, MirConstructStructOp, MirConstructTupleOp, MirExtractArrayElementOp,
-        MirFieldAddrOp, MirLoadOp, MirRemOp, MirStoreOp,
+        MirFieldAddrOp, MirLoadOp, MirLtOp, MirRemOp, MirStoreOp,
     },
     types::MirArrayType,
 };
@@ -41,7 +42,7 @@ use pliron::{
     basic_block::BasicBlock,
     builtin::types::{IntegerType, Signedness},
     context::{Context, Ptr},
-    graph::dominance::DomInfo,
+    graph::{ControlFlowGraph, dominance::DomInfo},
     irbuild::{
         listener::Recorder,
         rewriter::{IRRewriter, Rewriter},
@@ -58,11 +59,17 @@ use pliron::{
 use rustc_hash::FxHashSet;
 
 #[derive(Clone, Copy)]
+enum BoundedIndex {
+    Direct(Value),
+    Asserted { index: Value, bound: Value },
+}
+
+#[derive(Clone, Copy)]
 enum LoadReplacement {
     Direct(Value),
     BoundedArrayElement {
         array: Value,
-        index: Value,
+        index: BoundedIndex,
         result_type: TypeHandle,
     },
 }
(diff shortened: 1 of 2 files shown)


nihalpasham:
Verified sound, meets every #1142 criterion:

- the proof mirrors merged #650 with a stricter same-block rule
- assert polarity is normalized by the importer, so the single accepted form is what rustc emits
- `MirAssertOp` keeps trap-on-false, so canonicalizing to `mir.rem(index, C)` is value-preserving on every executing path
- anything unprovable fails closed, and the marker gate keeps user aggregates out (#399 stays open, correctly)

The win:

```text
let frag = ldmatrix_x4(...);   // 4-register bundle
let v = frag[i];               // i is a runtime kernel arg

before: bundle spilled, st.local + ld.local on the hot path
after:  assert(i < 4) proves the bound -> select chain, no .local
```

Follow-ups we'll take post-merge: an entry-block guard for the hand-built-IR corner (same exposure exists in #650; unreachable from rustc MIR), and consolidating the duplicated assert-proof helper. At landing we rerun tests, the per-example lint gates for generated_ldmatrix, and the 5090 run with and without `CUDA_OXIDE_NO_OPT=1`, since the recorded evidence predates #1186.

nihalpasham (approved):
Re-validated on the merged tree: 4,269 workspace tests green, generated_ldmatrix passes on the 5090 with and without CUDA_OXIDE_NO_OPT=1, zero .local in the PTX in both variants; CI green. Merging.

PR #1153, opened 2026-08-23
fix(intrinsics-gen): fail closed when an unsafe family has no safety renderer

### LINKED ISSUE #1152: cuda-intrinsics-gen: unsafe families without a safety renderer inherit the packed-atomic contract
### Summary

`validate_renderable` says every unsafe record has a dedicated family safety renderer, but it only
checks a hand-maintained family-name allowlist (`crates/cuda-intrinsics-gen/src/render.rs:445-484`).
It does not verify that `render_raw_abi` has a matching safety-dispatch arm. An allowed family
without an arm reaches the terminal `else` at `render.rs:5173-5178` and inherits the packed-atomic
contract:

```rust
} else {
    output.push_str(
        "/// `addr` must designate four writable bytes in global memory and be naturally aligned to four bytes.\n\
         /// Do not overlap this operation with a whole-word atomic or any non-atomic access to either 16-bit lane.\n\
         /// Racing atomics must use scopes that include each other; this relaxed GPU-scope operation is not atomic with host/system access.\n",
    );
}
```

That text is correct for the two `packed_atomic` records. It is also emitted for six CLC records in
`crates/cuda-intrinsics/src/generated/abi_v1.rs:641-731`:

| ABI | Catalog ID | Signature |
|---|---|---|
| i0322 | `clc_try_cancel` | `(*mut u8, *mut u64) -> ()` |
| i0323 | `clc_try_cancel_multicast` | `(*mut u8, *mut u64) -> ()` |
| i0324 | `clc_query_is_canceled` | `(u64, u64) -> u32` |
| i0325 | `clc_query_get_first_ctaid_x` | `(u64, u64) -> u32` |
| i0326 | `clc_query_get_first_ctaid_y` | `(u64, u64) -> u32` |
| i0327 | `clc_query_get_first_ctaid_z` | `(u64, u64) -> u32` |

The four query functions have no pointer or `addr`. The two request functions write a naturally
aligned 16-byte response in `.shared::cta` and signal an mbarrier, not a four-byte global atomic.

### Reproduction

Commit `18cc1888` on `safety-guard-red` adds two invariants over rendered raw-ABI safety blocks.
Against the unfixed renderer, 298 tests pass and exactly these two fail:

```text
unsafe safety docs reference `addr` without a pointer argument: clc_query_get_first_ctaid_x, clc_query_get_first_ctaid_y, clc_query_get_first_ctaid_z, clc_query_is_canceled
unsafe safety blocks shared across families: [{"clc", "packed_atomic"}]
```

The generated reference independently contradicts the inherited prose:
`intrinsics/generated-reference.md:62-65` records all four `clc_query_*` operations as
`pure; memory none`.

### Why this is a guard defect

Editing the allowlist is sufficient to pass validation; no corresponding renderer arm is required.
`render_raw_abi`, the guard, the fallback, the CLC records, and the CLC allowlist entry arrived
together in `6a9a6373` (PR #406). The same failure mode surfaced during review of PR #1138 when
`barrier_reduction` was added to the allowlist without a safety arm.

### Impact and next step

The impact is bounded: `cuda_device::clc` has separate family-specific prose and is the recommended
path. However, `intrinsics/generated-reference.md:1227-1250` calls `cuda_intrinsics::clc::*` the
public source path, and `crates/cuda-intrinsics/src/generated/mod.rs:93-100` publicly re-exports it.

A two-commit RED/GREEN fix is ready on `safety-guard-red`: `18cc1888` adds the failing invariants;
`101b1a01` makes rendering fail closed and corrects the CLC contracts.

The remaining maintainer decision is whether those contracts should also spell out object lifetime
through asynchronous completion and proxy-fence ordering when a response is reused across loop
iterations.

No release tag contains `6a9a6373`; I did not find a published `cuda-intrinsics` crate containing
this code.

almightychang:
## Summary

Fixes #1152. The unsafe-family guard checked a hand-maintained family allowlist rather than whether `render_raw_abi` had a matching safety renderer, allowing CLC intrinsics to inherit the terminal packed-atomic contract. This change makes raw-ABI safety rendering fail closed and gives CLC shared safety contracts across the raw and compatibility renderers.

## Changes

- `b0b57459` adds two rendered-safety invariants. At that commit they are RED: 298 existing tests pass and exactly the two new tests expose the CLC/packed-atomic collision.
- `50a4d0b1` makes the tests GREEN by giving `packed_atomic` an explicit arm, returning a generation error for any unhandled unsafe family, removing the redundant mirrored allowlist, and rendering reviewed CLC contracts through one shared helper.
- Regenerated changes in `abi_v1.rs` and `clc.rs` are doc comments only; comparison confirmed that no non-`///` line changed. No catalog, ABI ledger, PTX, or lowering file is touched, so runtime/GPU validation is not applicable.

## Testing

- [x] `just check` passes (the local mirror of CI: fmt, clippy, tests, CUDA-linked tests, guards, generated-intrinsic checks, rustdoc, and doctests). Exit 0 on the commits in this PR, with CUDA 12.9.86 and the pinned toolchain; 90 test binaries pass and none fail.
  - A fork-internal CI dry run of these identical commits passed all 50 applicable checks, including examples-compile, CodeQL, and the book gates; only `deploy` was skipped, as it is on every pull request.
- [ ] `cargo oxide run <example>` passes, or `scripts/smoketest.sh -o '^<example>$'` (not applicable: documentation-only generated changes; no runtime, PTX, or lowering change)
- [ ] New example added (not applicable)

## Checklist

- [x] SPDX headers on new source files (no new source files)

### DIFF
--- a/crates/cuda-device/src/generated/clc.rs
+++ b/crates/cuda-device/src/generated/clc.rs
@@ -12,7 +12,8 @@
 /// Returns the X coordinate from a successful cancellation response.
 ///
 /// # Safety
-/// The response halves must come from a completed CLC request.
+/// `resp_lo` and `resp_hi` must be the two halves of an opaque response observed complete from a prior CLC try-cancel request.
+/// The result is meaningful only when `clc_query_is_canceled` returned one for this response.
 #[inline(never)]
 pub unsafe fn clc_query_get_first_ctaid_x(resp_lo: u64, resp_hi: u64) -> u32 {
     let _ = (resp_lo, resp_hi);
@@ -22,7 +23,8 @@ pub unsafe fn clc_query_get_first_ctaid_x(resp_lo: u64, resp_hi: u64) -> u32 {
 /// Returns the Y coordinate from a successful cancellation response.
 ///
 /// # Safety
-/// The response halves must come from a completed CLC request.
+/// `resp_lo` and `resp_hi` must be the two halves of an opaque response observed complete from a prior CLC try-cancel request.
+/// The result is meaningful only when `clc_query_is_canceled` returned one for this response.
 #[inline(never)]
 pub unsafe fn clc_query_get_first_ctaid_y(resp_lo: u64, resp_hi: u64) -> u32 {
     let _ = (resp_lo, resp_hi);
@@ -32,7 +34,8 @@ pub unsafe fn clc_query_get_first_ctaid_y(resp_lo: u64, resp_hi: u64) -> u32 {
 /// Returns the Z coordinate from a successful cancellation response.
 ///
 /// # Safety
-/// The response halves must come from a completed CLC request.
+/// `resp_lo` and `resp_hi` must be the two halves of an opaque response observed complete from a prior CLC try-cancel request.
+/// The result is meaningful only when `clc_query_is_canceled` returned one for this response.
 #[inline(never)]
 pub unsafe fn clc_query_get_first_ctaid_z(resp_lo: u64, resp_hi: u64) -> u32 {
     let _ = (resp_lo, resp_hi);
@@ -42,7 +45,7 @@ pub unsafe fn clc_query_get_first_ctaid_z(resp_lo: u64, resp_hi: u64) -> u32 {
 /// Returns whether the Cluster Launch Control request was canceled.
 ///
 /// # Safety
-/// The response halves must come from a completed CLC request.
+/// `resp_lo` and `resp_hi` must be the two halves of an opaque response observed complete from a prior CLC try-cancel request.
 #[inline(never)]
 pub unsafe fn clc_query_is_canceled(resp_lo: u64, resp_hi: u64) -> u32 {
     let _ = (resp_lo, resp_hi);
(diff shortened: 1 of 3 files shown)


nihalpasham:
Verified end to end. On main only packed_atomic and clc reach the silent catch-all, so the six CLC intrinsics really did ship an atomic contract they never perform:

```text
before: clc record -> no arm -> catch-all -> packed-atomic "# Safety" text
after:  clc record -> clc arm -> reviewed CLC contract
        unknown unsafe family -> hard error naming the record
```

- allowlist deleted, dispatch is the single source of truth; the checks that made it redundant stay in `validate_renderable`
- every line of the new CLC prose matches the PTX ISA clusterlaunchcontrol section
- the two catalog-wide tests fail on main with the CLC collision and pass here

Two notes, neither blocking:

- the duplicate-block test forbids identical text across families even when legitimately shared; a future shared contract needs distinct wording or a test tweak
- no direct negative test hits the new bail path; the catalog invariants cover recurrence

CI is approved on the newest head. The #1152 contract-depth question stays a separate decision. Labels added.

nihalpasham (approved):
Verified on the merged tree with #1105 in: generator tests green, generated outputs byte-current; CI green. Merging.

PR #1157, opened 2026-08-23
refactor(rustlantis): clarify ProjectionIter traversal invariant

### LINKED ISSUE #1156: Rustlantis: clarify ProjectionIter reference-cycle invariant
## Problem

`ProjectionIter` currently contains the following FIXME:

```rust
/// FIXME: this breaks if there's a reference cycle in the graph
```

However, the current traversal does not recursively follow arbitrary graph
edges.

`ProjectionIter::new()` may start from any outgoing projection of the root,
including `Deref`, but subsequent traversal follows only
`PlaceNode::subfield_edges`.

Those edges are populated by `add_place()` with structural projections such as
fields, tuple fields, array indices, and enum variant fields. Reference edges
created by `set_ref()` are added to the graph but are not inserted into
`subfield_edges`.

As a result, a reference cycle in `PlaceGraph` cannot currently be re-entered
by `ProjectionIter`.

## Expected behavior

Document this traversal invariant explicitly and guard the assumption relied on
by `ProjectionIter`, so that a future change to `subfield_edges` cannot silently
make the stale FIXME become a real bug.

## Acceptance criteria

* Remove the stale reference-cycle FIXME.
* Document why reference cycles cannot currently be re-entered by
  `ProjectionIter`.
* Assert that descendant `subfield_edges` do not contain `Deref` projections.
* Preserve the existing traversal behavior.
* Keep the existing pointer traversal coverage passing.

uurl:
Closes #1156

## Summary

Replace a stale `ProjectionIter` reference-cycle FIXME with documentation of
the traversal invariant that currently prevents the problem.

`ProjectionIter::new()` may start from any outgoing projection of the root,
including `Deref`, but subsequent traversal follows only
`PlaceNode::subfield_edges`.

`subfield_edges` contains structural projections created by `add_place()`.
Reference edges created by `set_ref()` are added to the graph but are not added
to `subfield_edges`.

Therefore, reference cycles in `PlaceGraph` cannot currently be re-entered by
`ProjectionIter`.

## Changes

- Replace the stale reference-cycle FIXME with documentation of the actual
  traversal invariant.
- Add a debug assertion that descendant `subfield_edges` do not contain
  `Deref` projections.
- Preserve the existing traversal behavior.

## Testing

```text
cargo fmt --all -- --check
cargo test -p generate
cargo test --workspace
git diff --check
```

All Rustlantis workspace tests pass.

### DIFF
--- a/crates/fuzzer/rustlantis/generate/src/pgraph.rs
+++ b/crates/fuzzer/rustlantis/generate/src/pgraph.rs
@@ -1353,8 +1353,12 @@ impl PlacePath {
     }
 }
 
-/// A depth-first iterator over all reachable projections from a local variable
-/// FIXME: this breaks if there's a reference cycle in the graph
+/// A depth-first iterator over all reachable projection paths from a local variable.
+///
+/// The root may be followed through any outgoing projection, including `Deref`.
+/// Subsequent traversal only follows `subfield_edges`, which contain structural
+/// projections created by `add_place`. Reference cycles therefore cannot be
+/// re-entered by this iterator.
 #[derive(Clone)]
 pub struct ProjectionIter<'pt> {
     pt: &'pt PlaceGraph,
@@ -1419,6 +1423,11 @@ impl<'pt> Iterator for ProjectionIter<'pt> {
             let new_edges = self.pt.places[target].subfield_edges.iter();
             self.to_visit.extend(new_edges.filter_map(|&eidx| {
                 let e = self.pt.places[eidx];
+                debug_assert!(
+                    !e.is_deref(),
+                    "subfield_edges must not contain Deref projections"
+                );
+
                 // Only downcast to current variants
                 if let ProjectionElem::DowncastField(vid, _, _) = e
                     && self.pt.known_variant(target) != Some(vid)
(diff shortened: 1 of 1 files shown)


nihalpasham:
This PR is not blocked by a dependency or by the pointer-contract refactor; it remains open in normal review.

- The change is limited to the `ProjectionIter` documentation and a debug assertion, and no semantic counterexample was found.
- The PR body reports formatting, generator-package tests, workspace tests, and a clean diff. Those author-reported results still need independent reproduction on the exact head and current base, together with strict lint and hosted CI checks.
- Until that verification is complete, the PR is not merge-ready.

The exact-head evidence requirements are described in the [Discord acceptance criteria](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

nihalpasham:
Verified the documented invariant on main, and it holds:

- `subfield_edges` is written only by `add_place`'s four structural arms
- `set_ref` puts its Deref edge in the petgraph graph, never in `subfield_edges`
- `ProjectionIter` follows only `subfield_edges` after the first hop

So cycles can't be re-entered, the FIXME was stale, and the debug_assert sits at the single point any future violating writer must pass. Upstream rustlantis still carries the same FIXME, so documenting it in the vendored copy is right. Closes #1156.

One thing I added at landing: a `#[should_panic]` test that injects a Deref into `subfield_edges` and proves the guard fires. CI is approved; we also rerun the nested fuzzer workspace tests, since root `--workspace` doesn't reach them.

nihalpasham (approved):
Invariant verified on main, guard test added and neuter-checked; CI green. Merging.

PR #1159, opened 2026-08-23
fix(rustlantis): invalidate nested pointer edges on uninit

### LINKED ISSUE #1158: rustlantis: make uninit reference invalidation transitive
## Problem

`PlaceGraph::mark_place_uninit` currently removes the outgoing `Deref` edge only when the place being marked uninitialized is itself a pointer:

```rust
// If this is a pointer, we have to remove the Deref edge, but not for other projections
// FIXME: this should be transitive
```

This leaves stale `Deref` edges when an aggregate containing references or raw pointers is marked uninitialized.

For example, if a tuple contains nested pointer values, marking the tuple uninitialized clears its storage but does not invalidate the pointer relationships represented by those nested nodes.

## Expected behavior

When a place is marked uninitialized, outgoing `Deref` edges belonging to pointer/reference values contained transitively in that place should also be removed.

Incoming `Deref` edges from pointer/reference values outside the place should
not be removed by this operation. Those pointers retain their graph relationship
to the place, while their validity reflects the pointee becoming uninitialized.

## Suggested approach

Traverse the structural subfields of the place independently from the existing memory-initialization traversal and remove outgoing `Deref` edges from every pointer/reference node encountered.

A separate traversal is needed because the current memory traversal stops once it reaches a node backed by a `RunPointer`.

## Acceptance criteria

* Marking a nested aggregate uninitialized removes outgoing `Deref` edges from contained references.
* The same applies to contained raw pointers.
* Incoming `Deref` edges from pointer/reference values outside the aggregate are preserved.
* Existing initialization/uninitialization behavior remains unchanged.
* A regression test covers the transitive case.

uurl:
Closes #1158 

## Summary

Make `PlaceGraph::mark_place_uninit` remove outgoing `Deref` edges from pointer/reference values contained transitively in the place being marked uninitialized.

Previously, `mark_place_uninit` only removed the edge when the root place itself was a pointer, leaving stale graph edges for pointers nested inside aggregates.

## Implementation

The change performs a structural traversal over the place and its subfields before marking the underlying storage uninitialized.

For each pointer/reference node encountered, its outgoing `Deref` edge is removed.

This traversal is intentionally separate from the existing memory invalidation traversal. The latter stops at nodes backed by a `RunPointer`, which is correct for clearing bytes but would otherwise prevent discovery of nested pointer nodes.

Incoming `Deref` edges are not removed. References outside the value that point into its storage therefore retain their graph relationship.

## Tests

Added a regression test covering an aggregate containing nested:

- shared references
- raw pointers
- an external reference pointing to the aggregate

The test verifies that:

- nested outgoing `Deref` edges are removed after `mark_place_uninit`
- the aggregate becomes uninitialized
- the external incoming reference is preserved

## Validation

From `crates/fuzzer/rustlantis`:

```text
cargo fmt --all -- --check                         PASS
cargo test -p generate \
    uninit_removes_nested_deref_edges_but_preserves_incoming_refs
                                                     PASS
cargo test -p generate                             PASS (18/18)
cargo test --workspace                             PASS
```

From the repository root:

```text
git diff --check                                   PASS
```

### DIFF
--- a/crates/fuzzer/rustlantis/generate/src/pgraph.rs
+++ b/crates/fuzzer/rustlantis/generate/src/pgraph.rs
@@ -1503,12 +1503,13 @@ mod tests {
     use std::rc::Rc;
 
     use config::TyConfig;
+    use index_vec::IndexVec;
     use mir::{
         syntax::{
-            BinOp, FieldIdx, Literal, Local, Mutability, Operand, Place, ProjectionElem, Rvalue,
-            TyId, TyKind, UintTy,
+            Adt, BinOp, FieldIdx, Literal, Local, Mutability, Operand, Place, ProjectionElem,
+            Rvalue, TyId, TyKind, UintTy, VariantDef, VariantIdx,
         },
-        tyctxt::TyCtxt,
+        tyctxt::{AdtMeta, TyCtxt},
     };
 
     use crate::{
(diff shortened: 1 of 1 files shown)


nihalpasham:
This PR is not blocked by a dependency or by the pointer-contract refactor; it remains open for normal correctness review.

- The added regression already covers nested shared-reference and raw-pointer edges and preserves an external incoming-reference control.
- We still need to reproduce that test failing on the base and passing on this head, then exercise array, enum, and move cases around the same traversal boundary.
- Fuzz/Miri coverage and exact-head workspace, lint, and hosted checks remain outstanding.

Those regression, negative, and control checks follow the [Discord acceptance criteria](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

nihalpasham:
Fix verified sound:

- `mark_place_uninit` now walks structural subfields and cuts each nested pointer's outgoing Deref edge
- it sits at the one choke point all uninit paths use (Deinit, moves, intrinsic re-init)
- keeping incoming edges matches the #1158 spec
- the shift is strictly toward soundness: stale edges made dead pointers look live

One test gap, which I closed at landing: the tuple-only test can't tell this fix from the plausible wrong one (cutting edges inside the run-pointer-bounded walk), because tuples have no RunPointer:

```text
(&x, *mut y)        tuple: both fixes pass
[*mut x, *mut y]    array: run sits on the array node,
                    only the subfield walk reaches the elements
```

I added the array case, an enum-variant pointer case, a `mark_place_moved` case, and recorded the fail-on-base run. Labels already right.

nihalpasham (approved):
Fix plus the added adversarial tests verified, fail-on-base recorded; CI green. Merging.

PR #1161, opened 2026-08-23
fix(rustlantis): validate projection chains at construction

### LINKED ISSUE #1160: rustlantis: validate projection chains when constructing places
## Problem

`mir::syntax::Place` currently accepts projection chains without validating them when they are constructed.

Both `Place::from_projected` and `Place::project` can therefore create a `Place` containing an invalid sequence of `ProjectionElem`s. Invalid projections can remain unchecked until `Place::ty()`
calls `TyId::projected_ty()`, which may panic while interpreting the chain.

This delays detection of malformed MIR state and makes the source of the invalid projection harder to identify.

## Proposed change

Validate each projection as it is appended to a `Place`.

Projection validation should cover:

- dereferencing only pointer/reference types
- tuple field bounds
- struct field bounds
- array projection source types and constant-index bounds
- enum downcast variant and field bounds
- consistency between the type stored in `DowncastField` and the actual field type

The projection type semantics should remain centralized rather than duplicating the `ProjectionElem` matching logic between construction and later type lookup.

## Expected behavior

After this change, invalid projection chains should fail at the point where the `Place` is constructed or extended, rather than later during `Place::ty()`.

Existing valid projection paths generated by Rustlantis should continue to behave unchanged.

uurl:
Closes #1160 

## Summary

Validate Rustlantis `Place` projection chains when they are constructed instead of relying on `Place::ty()` to detect invalid projections later.

This updates both `Place::from_projected` and `Place::project` so malformed projection chains fail immediately.

The projection semantics are centralized in the type projection logic and reused by both construction-time validation and `TyId::projected_ty()`.

Validation now covers:

- `Deref` on pointer/reference types
- tuple field bounds
- struct field bounds
- array projection source types and constant-index bounds
- enum variant and field bounds
- `DowncastField` type consistency

Call sites that construct projected places were updated to provide the base type and type context required for validation.

## Tests

Added regression coverage for:

- valid projection-chain construction
- out-of-bounds projections
- invalid projections rejected before mutating a `Place`
- struct field bounds
- enum downcast validation
- mismatched `DowncastField` field types

Validation performed from `crates/fuzzer/rustlantis`:

```text
cargo fmt --all -- --check
cargo test -p mir
cargo test -p generate
cargo check --workspace --all-targets
cargo test --workspace
```

Additionally, from the repository root:

```text
git diff --check
```

Results:

```text
mir tests:                  7 passed
generate tests:            17 passed
Rustlantis workspace tests: passed
Rustlantis workspace check: passed
```

### DIFF
--- a/crates/fuzzer/rustlantis/generate/src/generation/mod.rs
+++ b/crates/fuzzer/rustlantis/generate/src/generation/mod.rs
@@ -343,7 +343,12 @@ impl GenerationCtx {
             .except(lhs);
         if let Some(_) = self.pt.pointee(lhs.to_place_index(&self.pt).unwrap()) {
             // The MIR linter doesn't like it if we do x = &(*x)
-            selector = selector.except(lhs.clone().project(ProjectionElem::Deref));
+            let local_ty = self.current_decls()[lhs.local()].ty;
+            selector = selector.except(lhs.clone().project(
+                local_ty,
+                ProjectionElem::Deref,
+                &self.tcx,
+            ));
         }
         let (candidates, weights) = selector
             .into_weighted(&self.pt)
--- a/crates/fuzzer/rustlantis/generate/src/pgraph.rs
+++ b/crates/fuzzer/rustlantis/generate/src/pgraph.rs
@@ -1322,7 +1322,9 @@ impl PlacePath {
             .collect();
         Place::from_projected(
             pt.current_frame().get_by_index(self.source).unwrap(),
+            pt.ty(self.source),
             &projs,
+            &pt.tcx,
         )
     }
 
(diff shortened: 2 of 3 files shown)


nihalpasham:
Closing this PR because the proposed projection validation still accepts invalid `Place` construction.

- `Place::from_projected` and `Place::project` trust a caller-supplied `local_ty` instead of the authoritative local declaration.
- `ProjectionElem::Index` does not verify that the index local exists or has `usize` type.
- This needs a typed construction API based on `LocalDecls`, with invalid-local and invalid-index regression controls.

That redesign is larger than a maintainer follow-up. It should return only after meeting the root-cause and full-path requirements in the [Discord announcement](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

PR #1163, opened 2026-08-23
feat(fuzzer): generate Static and StaticMut operands

### LINKED ISSUE #1162: Rustlantis: support Static and StaticMut operands in custom MIR
## Description

Rustlantis currently supports `Copy`, `Move`, and `Constant` operands, but custom MIR also provides `Static` and `StaticMut` for creating pointers to static allocations.

The corresponding TODO is already present in `mir/src/syntax.rs`:

```rust
// define!("mir_static", fn Static<T>(s: T) -> &'static T);
// define!("mir_static_mut", fn StaticMut<T>(s: T) -> *mut T);
```

Supporting these operands would extend Rustlantis coverage to MIR programs that access static and mutable static allocations.

## Expected behavior

Rustlantis should be able to generate and serialize:

```rust
Static(S)
StaticMut(T)
```

with corresponding top-level static declarations.

The generated pointers should integrate with the existing place graph and memory model so that:

* static allocations remain alive independently of function frames
* `Static` produces a shared reference to the static allocation
* `StaticMut` produces a mutable raw pointer to the static allocation
* subsequent dereferences and assignments use the existing pointer and alias tracking

## Initial scope

A minimal implementation can start with scalar static initializers and existing pointer/reference types.

More complex cases such as aggregate initializers, statics containing pointers, TLS, or arbitrary constant expressions can remain out of scope initially.

uurl:
Closes #1162 

## Summary

Add support for generating custom MIR `Static` and `StaticMut` operands in Rustlantis.

This extends the existing operand model beyond `Copy`, `Move`, and `Constant` and allows generated MIR programs to reference static allocations.

## Changes

- add static declarations to the MIR syntax model
- add `Static` and `StaticMut` operand variants
- serialize top-level `static` and `static mut` declarations
- serialize `Static(...)` and `StaticMut(...)` custom MIR operands
- generate scalar static allocations for compatible reference and raw pointer types
- keep static allocations alive independently of function frames
- connect static-derived pointers to the existing place graph and aliasing model
- add tests for static operand serialization and static allocation lifetime

## Scope

The implementation currently focuses on scalar static initializers.

More complex static initializers and additional static-related MIR constructs are intentionally left out of scope.

## Validation

The following checks pass:

- `cargo test -p mir`
- `cargo test -p generate`
- `cargo check --workspace`
- `cargo test --workspace`

Generated programs were also verified to contain both:

```text
Static(...)
StaticMut(...)
```

End-to-end differential testing was performed with two LLVM configurations:

* `-Zmir-opt-level=0`
* `-Zmir-opt-level=4 -Copt-level=3 -Ctarget-cpu=native`

100 generated seeds passed with:

```text
Passed: 100
Failed: 0
```

### DIFF
--- a/crates/fuzzer/rustlantis/generate/src/generation/mod.rs
+++ b/crates/fuzzer/rustlantis/generate/src/generation/mod.rs
@@ -10,7 +10,7 @@ use mir::serialize::Serialize;
 use mir::syntax::{
     AggregateKind, BasicBlock, BasicBlockData, BinOp, Body, Callee, Function, IntTy, Literal,
     Local, LocalDecls, Mutability, Operand, Place, Program, ProjectionElem, Rvalue, Statement,
-    SwitchTargets, Terminator, TyId, TyKind, UnOp, VariantIdx,
+    StaticDecl, SwitchTargets, Terminator, TyId, TyKind, UnOp, VariantIdx,
 };
 use mir::tyctxt::TyCtxt;
 use rand::{Rng, RngCore, SeedableRng, seq::IteratorRandom};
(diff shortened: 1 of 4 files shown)


nihalpasham:
Closing this PR because generated `Static` and `StaticMut` operands do not survive cuda-oxide's production adapter.

- Static declarations are emitted before `#[custom_mir]`, but `extract_first_custom_mir_fn` discards that prefix.
- The adapted function retains references such as `Static(static0)` without defining `static0`, causing compilation failure.
- The tests check declarations and operands separately; they never pass a referenced static through the production adapter and compiler.

A replacement needs an end-to-end before/after regression on the real host and CUDA paths, as required by the [Discord acceptance criteria](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

PR #1165, opened 2026-08-23
fix(mir-importer): support chained projected call destinations

### LINKED ISSUE #1164: Support chained projected destinations for call and intrinsic results
## Problem

Call and intrinsic result lowering currently does not consistently support projected destinations beyond the limited shapes handled directly by the result-store path.

For example, MIR destinations such as:

```text
RET.1[index]
(*ptr).field
array[index].field
```

can already be represented as writable addresses by the existing place-address lowering, but call/intrinsic result storage maintains separate projection handling.

This creates two problems:

1. valid projection chains supported elsewhere by MIR place lowering can be rejected for call results;
2. result storage maintains an independent projection grammar that can diverge from normal projected-place semantics.

## Expected behavior

A call or intrinsic result should be writable to any projected destination that the existing `translate_place_address` machinery can represent as a writable address.

Result-producing call/intrinsic paths should not inspect the projection chain to decide which projection shapes they support.

Conceptually:

```text
result-producing call/intrinsic
        |
        v
store_result_to_place
        |
        v
translate_place_address
        |
        v
writable projected place
```

If the shared place-address walker cannot represent the destination, result storage should fail closed rather than falling back to the base local.

## Acceptance criteria

* A bare local destination continues to use the existing local-slot fast path.
* Projected destinations are materialized through the existing writable place-address walker.
* Multi-element chains such as `Field -> Index`, `Deref -> Field`, and `Index -> Field` are supported when the walker can represent them.
* Generated intrinsic result paths use the same result-store abstraction.
* No independent per-projection dispatcher is introduced for call/intrinsic results.
* Unsupported places fail closed rather than storing into `destination.local`.
* Existing projected-destination behavior remains covered by end-to-end GPU tests.

## Non-goals

* Extending the projection forms supported by `translate_place_address`.
* Changing pointer-kind semantics.
* Addressing the existing MIR fidelity difference where rustc evaluates the call destination address before the call while this lowering materializes it after the call operation.

uurl:
Closes #1164 

## Summary

Unify call and intrinsic result storage with the existing projected-place address lowering.

Previously, `store_result_to_place` maintained its own projection dispatcher and only supported a limited set of single-element destinations. Other result paths also stored directly into `destination.local` or rejected projected destinations.

This change routes projected result destinations through `translate_place_address`, so result storage now supports the same writable projection chains as the existing place-address walker.

## What changed

- Keep the existing fast path for bare local destinations.
- Route projected call/intrinsic destinations through `translate_place_address`.
- Remove the result-store-specific handling for individual `Deref`, `Field`, `Index`, and `ConstantIndex` projections.
- Route shared intrinsic epilogues, generated NVVM result paths, unit-result paths, ordinary function calls, closure calls, float-math results, and layout intrinsic results through `store_result_to_place`.
- Make place-address materialization independent of `mir::Body`.
  - Runtime `Index` projections now load their index local directly from `ValueMap`, matching the previous bare-local load path.
- Fail closed when the shared walker cannot materialize a writable address.

Generated intrinsic sources are not modified.

## Regression coverage

The projected-call-destination example now exercises:

```text
Deref
Field
Index
Field -> Index
Deref -> Field
Index -> Field
float result -> Field
float result -> Index
function result -> Field
sincos projected result
generated catalog intrinsic -> Field -> Index
```

The generated-catalog regression calls `cuda_intrinsics::sreg::block_dim_x()` directly with a `Field -> Index` destination, exercising the real generated intrinsic dispatch and result-store path on the GPU.

## Validation

```text
cargo fmt --all -- --check
git diff --check

scripts/smoketest.sh \
  --compile-only \
  -o '^mir_projected_call_destination$'

cargo oxide run mir_projected_call_destination

cargo test --workspace
```

`mir-importer`:

```text
1110 passed; 0 failed
```

Projected-destination GPU regression:

```text
PASS: device and host expectation agree on all eleven projected destinations
```

## Scope

This does not add new MIR projection semantics. It reuses the projection semantics already implemented by `translate_place_address`.

A projected result destination is supported if that existing walker can materialize it as a writable address; otherwise lowering fails closed.

## Known limitation

rustc evaluates a call destination address before performing the call. This importer currently materializes the projected destination address after emitting the call operation.

That pre-call vs post-call evaluation difference predates this change and is intentionally left out of scope.

### DIFF
--- a/crates/mir-importer/src/translator/payload_store.rs
+++ b/crates/mir-importer/src/translator/payload_store.rs
@@ -124,7 +124,6 @@ pub(crate) fn classify(
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn rebuild_and_store(
     ctx: &mut Context,
-    body: &mir::Body,
     value_map: &ValueMap,
     store: &CanonicalPayloadStore,
     new_value: Value,
@@ -136,7 +135,6 @@ pub(crate) fn rebuild_and_store(
 
     let Some((enum_ptr, prev)) = rvalue::translate_place_address(
         ctx,
-        body,
         value_map,
         &store.enum_place,
         /* is_mutable */ true,
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -815,7 +815,6 @@ pub fn translate_rvalue(
             // through the borrow mutate the borrowed place.
             if let Some((result_val, last_inserted)) = translate_place_address(
                 ctx,
-                body,
                 value_map,
                 place,
                 is_mutable,
@@ -909,7 +908,6 @@ pub fn translate_rvalue(
             // composes field + element addresses, ...).
             if let Some((result_val, last_inserted)) = translate_place_address(
                 ctx,
-                body,
                 value_map,
                 place,
                 is_mutable,
@@ -2883,7 +2881,6 @@ pub fn translate_place(
         PlaceReadStrategy::Address => {
             if let Some((value, last_op)) = translate_place_load_from_address(
                 ctx,
-                body,
                 place,
                 value_map,
                 block_ptr,
@@ -3157,7 +3154,6 @@ fn classify_place_read_strategy(
 #[allow(clippy::too_many_arguments)]
 fn translate_place_load_from_address(
     ctx: &mut Context,
-    body: &mir::Body,
     place: &mir::Place,
     value_map: &ValueMap,
     block_ptr: Ptr<BasicBlock>,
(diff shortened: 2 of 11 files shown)


nihalpasham:
Closing this PR because projected call destinations are materialized after the callee executes, contrary to Rust MIR semantics.

- For custom MIR `Call(array[i] = mutate(&mut i), ...)`, the MIR destination is fixed before the callee; this implementation reloads `i` afterward and can store elsewhere.
- The projected store path also bypasses canonical enum-payload handling and pointer coercion.
- This requires one typed destination-write abstraction that computes the destination before the call.

That is a substantial semantic redesign, not a small review correction. A replacement should meet the full-path and adversarial-test requirements in the [Discord announcement](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

Issue #1071, opened 2026-08-20
Tracking: device-side CUDA runtime API (cudadevrt) support in the Rust device compiler


lucifer1004:
## Background

The Rust→NVVM IR→nvJitLink pipeline currently cannot call CUDA device-runtime APIs from kernels, because those symbols are not defined in any Rust crate and must be resolved at the final device link. We have a local, not-yet-PR'd commit that adds exactly one such API — `cudaGraphSetConditional` — to support explicit CUDA Graph conditional nodes. We would like to agree on the general design before the PR series grows.

This issue tracks the design discussion for device-side CUDA runtime API support as a whole.

## What we have working today

A single-symbol vertical slice, ~370 lines across 16 files:

- `cuda-device`: `unsafe extern "C" fn cudaGraphSetConditional(handle: u64, value: u32)` declaration
- collector: leaves this exact symbol unresolved during monomorphization collection
- mir-lower: declares it at the call site, like `__nv_*` libdevice externs
- codegen pipeline: detects the call in typed MIR, sets `requires_cuda_device_runtime`, and forces the NVVM IR / external-link route
- finalizer: discovers `libcudadevrt.a`, fingerprints it (SHA-256 handshake with the cargo-oxide driver), and feeds it to nvJitLink as an `InputType::Library` alongside the LTOIR inputs

This produces a working `sm_120a` cubin whose selector kernel drives a real conditional node, loaded and exercised on hardware.

## Empirical findings that should shape the design

All observed with the CUDA 13.3 toolchain (libNVVM 2.0 / nvJitLink 13.3), `sm_120a`:

1. **`cudaGraphSetConditional` does not need the archive.** nvJitLink recognizes it and lowers the call to the driver service ABI (`CALL.ABS.NOINC` via `c[0x4][0x28]`); with and without `libcudadevrt.a` on the link, the kernel's SASS is instruction-identical. This reserved-symbol behavior appears to be **undocumented** — the nvJitLink user guide never mentions cudadevrt or reserved symbols.
2. **`cudaMalloc` and `cudaGraphLaunch` do need the archive.** Without it, nvJitLink **returns success** but silently drops every kernel of the offending module, producing a valid but empty cubin (~2.2 KB). With the archive, real device code from `cuda_device_runtime.cu` is linked in and all entries survive. (Consistent with `--keep-system-libraries` docs: the device linker GCs unreferenced system-library code.)
[...]

PR #1167, opened 2026-08-23
fix(finalizer): reject links that drop expected kernels

### LINKED ISSUE #1166: finalizer: reject successful links that drop expected kernels
## Problem

A successful `nvJitLink` invocation does not necessarily mean that the resulting
PTX or cubin still contains the kernels that the caller expects to launch.

This can produce a particularly bad failure mode: finalization succeeds and a
structurally valid artifact is emitted, but one or more expected kernel entry
points have been removed during linking.

This was exposed by the empty-cubin behavior discussed around #1071 and is also
related to the regression coverage added in #784.

## Proposed invariant

Finalization should enforce:

> `nvJitLink` success + missing expected kernel = error

The finalizer should therefore receive the expected kernel names, keep those
kernels rooted during the final link, inspect the resulting artifact, and reject
the result if any expected entry point is missing.

## Proposed scope

- Pass expected kernel names into the finalization path.
- Forward them to `nvJitLink` using repeated `-kernels-used=<name>` options.
- Include the normalized expected-kernel set in finalization artifact identity.
- Inspect final PTX kernel entries structurally.
- Inspect final cubin kernel entries from ELF symbol metadata.
- Return a specific error when:
  - an expected kernel name is invalid,
  - the final kernel inventory cannot be determined, or
  - an expected kernel is missing.
- Thread kernel metadata through build-time and deferred embedded-artifact
  finalization.
- Add unit and live regression coverage.

## Non-goals

This issue does not add the complete `libcudadevrt` link contract and does not
implement CUDA Graph APIs.

Those remain separate follow-up work under #1071.

uurl:
Closes #1166 
Part of #1071.

## Summary

This adds a final-link safety invariant for expected CUDA kernels.

A successful `nvJitLink` call is no longer considered sufficient when the
caller knows which kernel entry points must survive. The finalizer now roots
those kernels during linking and verifies that they are present in the emitted
PTX or cubin.

## What changed

- Added expected-kernel-aware finalization APIs while preserving the existing
  public entry points.
- Passed expected kernel names to `nvJitLink` as repeated
  `-kernels-used=<name>` options.
- Canonicalized the expected-kernel set for deterministic linker options and
  artifact digests.
- Included expected kernels in LTOIR finalization artifact identity.
- Added structural PTX entry-point inventory rather than substring matching.
- Added cubin kernel inventory by inspecting ELF symbol tables and CUDA entry
  symbols.
- Added explicit errors for:
  - invalid expected kernel names,
  - unavailable final kernel inventory,
  - missing expected kernels.
- Propagated known kernel metadata through:
  - build-time NVVM IR/LTOIR materialization,
  - embedded deferred NVVM IR/LTOIR finalization.
- Kept existing raw/file-backed finalization APIs compatible when no expected
  kernel metadata is available.

The enforced invariant is:

> `nvJitLink` success + missing expected kernel = error

## Validation

The following pass locally:

```text
cargo fmt --all -- --check
cargo test -p cuda-artifact-finalizer --all-targets
cargo test -p cuda-host --all-targets
cargo clippy -p cuda-artifact-finalizer --all-targets -- -D warnings
cargo clippy -p cuda-host --all-targets -- -D warnings
```

The CUDA Toolkit live regressions also pass:

```text
live_tests::live_pipeline_accepts_toolkit_cubins_and_emits_ptx_for_both_fma_policies
live_tests::live_link_report_parses_ptxas_resource_usage_for_a_spilling_kernel
```

And the broader validation passes:

```text
cargo test --workspace --all-targets
```

`rustc-codegen-cuda` was also validated independently with formatting, unit
tests, and Clippy.

## Scope

This is intentionally limited to the final-link safety portion of #1071.

It does not add `libcudadevrt` discovery/linking and does not implement the
public CUDA Graph API. Those remain separate follow-up changes.

### DIFF
--- a/crates/cuda-artifact-finalizer/src/lib.rs
+++ b/crates/cuda-artifact-finalizer/src/lib.rs
@@ -130,6 +130,18 @@ pub enum FinalizerError {
     #[error("nvJitLink returned an empty PTX artifact")]
     EmptyPtx,
 
+    /// A kernel root cannot be represented safely as an nvJitLink option.
+    #[error("expected CUDA kernel name is empty or contains an interior NUL byte: {name:?}")]
+    InvalidExpectedKernelName { name: String },
+
+    /// The linked image could not provide a structural kernel-entry inventory.
+    #[error("could not inspect linked {output} kernel entry inventory")]
+    LinkedKernelInventoryUnavailable { output: &'static str },
+
+    /// nvJitLink succeeded but one or more required kernel entry points vanished.
+    #[error("nvJitLink output is missing expected CUDA kernels: {missing:?}")]
+    MissingExpectedKernels { missing: Vec<String> },
+
     /// A pinned CUDA compilation tool changed around an operation. Its output can
     /// no longer be attributed to the provenance used by Cargo or a cache key.
     #[error(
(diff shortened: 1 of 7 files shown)


nihalpasham:
Closing this PR because expected-kernel validation is not consistently derived or applied.

- The implementation hardcodes wire tag `1` instead of using `ArtifactEntryKind` as the shared source of truth.
- The cargo-oxide compatibility finalization path remains unguarded.
- There is no real negative nvJitLink regression showing a kernel dropped before the fix and preserved afterward.

The underlying benefit is clear, but this needs cross-path design work rather than a small maintainer correction. A replacement should follow the existing abstractions and evidence requirements in the [Discord announcement](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

PR #1169, opened 2026-08-23
fix(ptx-schedule): classify async issue and wait sites

### LINKED ISSUE #1113: ptx-schedule: the async proxy pipeline (cp.async / wgmma / tcgen05 commit-wait pairs) is not classified, so 227 ordering instructions are invisible to a campaign
## Summary

`ptx-schedule`'s `classify_instruction` covers the synchronous ordering primitives — atomics, `red.*`, `bar.*`/`barrier.*`/`mbarrier.*`, `membar.*`/`fence.*`, the warp collectives, and ordered `ld`/`st`. It does not cover the **asynchronous proxy pipeline**: the `commit`/`wait` pairs and bulk-copy issues that carry ordering on Hopper and Blackwell.

An opcode that classifies as nothing is not a site of any kind, so it gets no `nanosleep.u32` and never appears in a campaign report. On the arch-gated examples that means the pipeline a Blackwell kernel is actually built around is the part the perturbation engine cannot reach.

## Measured

Counted over the 168 `.ptx` artifacts in `crates/rustc-codegen-cuda/examples/`, using the classifier's own rules:

| unclassified opcode | occurrences | `.ptx` files |
|---|---|---|
| `cp.async.bulk.tensor.2d.shared` | 64 | 5 |
| `tcgen05.ld.sync.aligned.16x256b.x1.b32` | 48 | 3 |
| `tcgen05.wait` | 48 | 3 |
| `tcgen05.commit.cta_group` | 47 | 3 |
| `tcgen05.dealloc.cta_group` | 11 | 3 |
| `tcgen05.relinquish_alloc_permit.cta_group` | 3 | 2 |
| `wgmma.fence.sync.aligned` | 1 | 1 |
| `wgmma.commit_group.sync.aligned` | 1 | 1 |
| `wgmma.wait_group.sync.aligned` | 1 | 1 |
| `cp.async.ca.shared.global` | 1 | 1 |
| `cp.async.commit_group` | 1 | 1 |
| `cp.async.wait_all` | 1 | 1 |

**227 occurrences across 7 files.** For contrast, the *fences* of the same pipeline are already sites: `fence.proxy.async.shared` (29 occurrences) matches the `fence.` prefix, and every `mbarrier.*` matches `mbarrier.`. So the async pipeline is covered in part — its fences and mbarriers are sites, its commit/wait pairs and bulk issues are not.

## Two separable pieces

**1. `wgmma.fence.sync.aligned` looks like a plain miss.** It is a fence, `SiteKind::Fence` exists for exactly that, and the only reason it is skipped is that the check is `head.starts_with("fence.")` while this mnemonic spells the fence in the second position. If you agree, that is a one-prefix fix and I am happy to send it.

**2. The commit/wait pairs are a design call, which is why I am asking rather than sending a patch.** `tcgen05.commit`/`tcgen05.wait`, `cp.async.commit_group`/`cp.async.wait_all`, `wgmma.commit_group`/`wgmma.wait_group` do not map cleanly onto the existing seven kinds:

- a `wait` blocks until a group completes, which is barrier-like — but folding it into `Barrier` loses the distinction from `bar.sync` in campaign reports, and the report groups by kind (`format_site_kinds`);
- a `commit` is an ordering point with no blocking, closest to `Fence` but not a memory fence;
- `cp.async.bulk.tensor...` and `tcgen05.ld` are *issues*, not ordering ops at all — their schedule sensitivity is that a delay between issue and wait changes overlap, which is arguably the most interesting perturbation of the three and fits none of the kinds.

So: fold into the existing kinds, or add one (`AsyncPipeline`) or two (`AsyncIssue` / `AsyncWait`)? Adding a variant to `SiteKind` is a `Serialize`/`Deserialize` change for `summary.json` and for `--decisions-json`, so it is worth deciding once rather than drifting into it.

## Context

Found while fixing a neighbouring gap in the same function: `redux.sync` classified as nothing either, which made the `redux_sum` and `redux_minmax` kernels report zero sites (#1111 — that one was unambiguous, since the `WarpCollective` list already enumerated every one of `redux`'s siblings).

I am glad to implement whichever shape you prefer, including just piece 1 if piece 2 should wait.

uurl:
Fixes #1113

## Summary

Classify async proxy pipeline instructions in `ptx-schedule` so they become visible
to schedule analysis and perturbation campaigns.

## Changes

- Add `SiteKind::AsyncIssue`.
- Add `SiteKind::AsyncWait`.
- Classify the async issue/commit forms identified in #1113 as `AsyncIssue`.
- Classify the corresponding completion/wait forms as `AsyncWait`.
- Classify `wgmma.fence.*` as the existing `Fence` kind.
- Keep `tcgen05` resource-lifecycle operations outside the async pipeline taxonomy.
- Add unit coverage for classification and perturbation.
- Add CLI integration coverage for the serialized `AsyncIssue` / `AsyncWait` names.
- Update the README `SiteKind` inventory.

The campaign reporting path requires no additional changes because site counts are
derived dynamically from the `SiteKind` debug names.

## Validation

```text
cargo fmt --all -- --check
cargo test -p ptx-schedule --all-targets
RUSTDOCFLAGS="-D warnings" cargo doc -p ptx-schedule --no-deps --document-private-items
git diff --check
```

Results:

* 19 unit tests passed.
* 4 CLI integration tests passed.
* Clippy passed with warnings denied.
* Rustdoc passed with warnings denied.
* `git diff --check` passed.

Manual CLI validation also confirms that both `--list-sites` and
`--decisions-json` serialize the new kinds as:

```text
AsyncIssue
AsyncWait
```

### DIFF
--- a/crates/ptx-schedule/README.md
+++ b/crates/ptx-schedule/README.md
@@ -27,7 +27,9 @@ places where another thread's progress can be observed:
 | `Atomic` | an atomic read-modify-write |
 | `Reduction` | a `red.*` reduction with no returned value |
 | `Barrier` | `bar.*` / `barrier.*` / `mbarrier.*` synchronization |
-| `Fence` | `fence.*` / `membar.*` ordering |
+| `Fence` | `fence.*` / `membar.*` / `wgmma.fence.*` ordering |
+| `AsyncIssue` | asynchronous work issue or commit points (`cp.async`, `wgmma.commit_group`, `tcgen05.ld` / `tcgen05.commit`) |
+| `AsyncWait` | waits for asynchronous pipeline progress (`cp.async`, `wgmma`, `tcgen05`) |
 | `OrderedMemory` | a load or store carrying an explicit ordering qualifier (`.volatile`, `.acquire`, `.relaxed`, ...) |
 | `WarpCollective` | `shfl`, `vote`, `match`, `redux` and friends |
 | `Backedge` | a `bra` back to the same or an earlier block -- a conservative loop detector |
--- a/crates/ptx-schedule/src/lib.rs
+++ b/crates/ptx-schedule/src/lib.rs
@@ -39,6 +39,8 @@ pub enum SiteKind {
     Reduction,
     Barrier,
     Fence,
+    AsyncIssue,
+    AsyncWait,
     OrderedMemory,
     WarpCollective,
     Backedge,
@@ -291,7 +293,10 @@ fn injection_points(source: &str, site: &ScheduleSite) -> (usize, usize) {
 }
 
 fn classify_instruction(instruction: &Instruction<'_>) -> Option<SiteKind> {
-    let head = instruction.head();
+    classify_head(instruction.head())
+}
+
+fn classify_head(head: &str) -> Option<SiteKind> {
     if head.starts_with("atom.") {
         return Some(SiteKind::Atomic);
     }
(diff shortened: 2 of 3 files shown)


nihalpasham:
Closing this PR because the async-instruction table is incomplete and can silently leave valid async PTX unclassified.

- It omits `cp.reduce.async.bulk`, `wgmma.mma_async`, and relevant `tcgen05.mma`, `.cp`, `.shift`, and `.st` families.
- The tests repeat the same partial table, so they cannot detect these omissions.
- Classification needs one authoritative semantic source and coverage for every supported async family.

A replacement needs to address the full instruction set rather than selected shapes, with negative and unaffected controls as required by the [Discord acceptance criteria](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

PR #1171, opened 2026-08-23
docs(book): document the catalog-generated intrinsics workflow

### LINKED ISSUE #808: book: the intrinsics guide covers only hand-written ops; the cuda-intrinsics-gen catalog workflow is undocumented

uurl:
Closes #808

## Summary

Document the contributor workflow for catalog-generated intrinsics and route the existing intrinsics guide to that workflow before the hand-written path.

This closes the remaining book-level documentation gap around `cuda-intrinsics-gen`: contributors can now see which files are authoritative, which files are generated, how ABI/evidence changes fit into the workflow, and which validation command mirrors the generated-intrinsics CI checks.

## Changes

- add `compiler/catalog-generated-intrinsics.md`;
- document the catalog source-of-truth pipeline:
  - `upstream.lock`
  - `imported.json`
  - `overlay.toml` / overlay shards
  - `abi-v1.toml`
  - lowering evidence
  - generated `catalog.json`;
- document the normal existing-family workflow;
- document the append-only ABI requirement;
- document when candidate probe evidence is needed;
- document that new overlay shards must be indexed in `overlay.toml`;
- clarify that `extract` is for updating imported LLVM facts, not for normal intrinsic additions;
- clarify that generated outputs must not be edited by hand;
- update `compiler/adding-new-intrinsics.md` to distinguish the catalog-generated and hand-written paths;
- remove stale guidance suggesting direct edits to `intrinsics/catalog.json`;
- add the new page to the book toctree.

## Validation

```text
just book
bash scripts/check-book-api-names.sh
just check-intrinsics upstream/main
git diff --check
```

`just check-intrinsics upstream/main` completed successfully, including all 1016 generated intrinsic routes and ABI-history validation.


nihalpasham:
Putting this PR on hold because it defines maintainer-owned policy tracked in #808; it is not a pointer-refactor hold.

- The documentation is mechanically coherent, but the policy itself must be agreed by maintainers before it becomes repository guidance.
- Review can resume after that decision, with the text adjusted to the approved policy and the documentation checks run on the exact head.
- Until then, implementation detail cannot substitute for the unresolved ownership decision.

This follows the existing-design and focused-scope requirements in the [Discord acceptance criteria](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

nihalpasham:
Fresh review pass over this one. It stays on hold, and to be fair: the blocker is on my side, not yours. #808 reserved the contributor-path policy for maintainers to define, and until I ratify it there, the normative sections here (ABI-ID rebase discipline, `just check-intrinsics` as the sanctioned interface, extraction policy) are inferred policy, however reasonable.

That said, the deep review found real accuracy issues you can fix while the policy question is pending:

- **The `schema` gloss conflates two different constants.** The chapter says overlay.toml's `schema` "identifies the serialized catalog format expected by the generator". It doesn't. There are two numbers:

  ```
  intrinsics/overlay.toml      schema = 44   (OVERLAY_SCHEMA: overlay input format)
            |
            v  cuda-intrinsics-gen
  intrinsics/catalog.json      "schema": 46  (CATALOG_SCHEMA: serialized catalog format)
  ```

  Both constants live in `crates/cuda-intrinsics-gen/src/resolve/overlay.rs`.

- **#808 asked about catalog.json's header, the chapter documents overlay.toml's.** The issue explicitly listed catalog.json's `schema` / `catalog_version` / `intrinsic_abi` / `generator_version`, and `generator_version` never gets documented at all. Cover both headers, or catalog.json's at minimum.

- **"Mirrors the generated-intrinsics CI job" oversells.** CI runs `probe --all --skip-terminal --per-target`; the Justfile recipe omits `--per-target`, so a local pass can be weaker than CI. Either align the Justfile (probably the right fix, separate small PR welcome) or soften the sentence.

- **Two environment facts from #808 are missing:** `check-abi-history` needs `fetch-depth: 0`, and the generator shells out to rustfmt (which is why the toolchain pin carries that component). Add them or note they're deliberately out of scope.

Everything else I spot-checked against the tree holds up (do-not-edit targets, fail-closed probe flags, append-only ledger rules, the three CI gates). Once #808's policy call lands, this needs one conformance pass against it and a rebase, then it can go in.

nihalpasham:
Off hold: the policy question this was parked on is settled on #808, and the path your chapter inferred is now the official one. So the normative sections stay as written; we finished the rest ourselves rather than round-tripping. Your commit is the base of the new head, untouched.

**What we stacked on top** (one signed commit):

- **The schema gloss was conflating two constants.** The chapter said overlay.toml's `schema` identifies the serialized catalog format. There are two numbers:

  ```
  intrinsics/overlay.toml   schema = 44   (OVERLAY_SCHEMA: the input format
          |                                the generator accepts)
          v  cuda-intrinsics-gen
  intrinsics/catalog.json   "schema": 46  (CATALOG_SCHEMA: the serialized
                                           catalog format it emits)
  ```

  Both live in `resolve/overlay.rs`; the chapter now shows this sketch and anchors each number to its constant.

- **catalog.json's own header is now documented** (`schema` / `catalog_version` / `intrinsic_abi` / `generator_version`), which is what #808 asked about by name. `generator_version` is the generator's crate version and had no docs anywhere.

- **The Justfile now really mirrors CI.** `check-intrinsics` was missing `--per-target`, so a local pass could be weaker than the hosted gate. Added it and verified: the recipe now probes 2374 target configurations across all 1025 routes, same as CI.

- **The two environment facts from #808**: `check-abi-history` needs full git history (CI clones with `fetch-depth: 0`), and the generator shells out to rustfmt (why the toolchain pin carries that component).

- Small path fix found while re-verifying every claim against the tree: `resolve.rs`/`render.rs`/`model.rs` are directories since the module split; the chapter now says so.

**Validation**: book builds with zero warnings, every path and command the chapter names resolves in today's tree, all nine repo guards pass, and a real `just check-intrinsics origin/main` run with the new recipe exits 0. Dropping the hold labels; CI + landing next, which closes #808.

nihalpasham (approved):

Issue #966, opened 2026-08-17
tma_multicast's README still says consumer Blackwell cannot run it, which #668 measured and corrected everywhere else


vyncint:
## Summary

#668 ("docs(examples): tma_multicast does run on consumer Blackwell", merged 2026-08-06) corrected the `tma_multicast` example on the strength of your RTX 5090 measurement:

> the sm_100a `tma_multicast` PTX JIT-loads and passes on an RTX 5090, so the "consumer Blackwell does not support TMA multicast" fallback text is outdated

It fixed the two places it found — the module header and the runtime `println!` — and **missed `README.md` entirely**. That file still carries four separate assertions of the claim #668 removed, so the example's own directory now contradicts itself. My PR, my miss; filing rather than fixing blind because the second half below needs a call I cannot make.

## The four stale passages

All in `crates/rustc-codegen-cuda/examples/tma_multicast/README.md`:

**1. L116 — heading groups sm_120 with the failing case**
```
### On Consumer Blackwell (sm_120) or Hopper (sm_90):
```

**2. L119-124 — a sample transcript showing a failure you measured as a success**
```text
GPU Compute Capability: sm_120

✗ PTX load failed: DriverError(CUDA_ERROR_INVALID_PTX, ...)

  TMA multicast requires sm_100a (Blackwell datacenter: B100/B200/GB200).
  Consumer Blackwell (sm_120) does NOT support multicast.
```
This is the worst of the four: a reader on a 5090 compares their real output against a documented failure and concludes something is wrong with their setup.

**3. L131 — the Hardware Requirements list**
```
- **NOT supported**: Consumer Blackwell (sm_120), Hopper (sm_90/sm_90a)
```

**4. L158-164 — an entire section explaining a mechanism for behaviour that does not occur**
```
## Why sm_100a Only?
... TMA multicast uses the L2 multicast fabric present only in datacenter
Blackwell GPUs. Consumer Blackwell (sm_120) has the same base ISA (sm_100)
but lacks this fabric, so `sm_100a` PTX cannot be JIT-compiled on sm_120.
```
Not just a wrong claim but a wrong *explanation*, which is stickier — it gives a reader a reason to disbelieve their own working run.

Against `src/main.rs:16-17`, which #668 rewrote:
```rust
//! - Blackwell (sm_100a or later): datacenter B100/B200/GB200, and reported to
//!   load and run on consumer sm_120 as well
```

[...]

PR #1172, opened 2026-08-23
docs(examples): correct stale tma_multicast consumer Blackwell claims

uurl:
## Summary

Corrects stale consumer Blackwell claims in the `tma_multicast` README.

#668 established through runtime measurement that the example's PTX JIT-loads and executes successfully on an RTX 5090 (sm_120), but the README still documented consumer Blackwell as unsupported and included an obsolete explanation for that behavior.

This PR:

- removes the stale sm_120 failure example;
- documents the measured consumer Blackwell result from #668;
- removes the incorrect claim that TMA multicast requires datacenter Blackwell-specific hardware;
- keeps Hopper runtime support explicitly unclaimed, since physical Hopper validation is still pending and the current host example continues to skip devices below compute capability 10.

No runtime or compiler behavior is changed.

## Validation

- `git diff --check`
- `cargo oxide fmt --check`

Addresses the consumer-Blackwell documentation portion of #966.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/tma_multicast/README.md
+++ b/crates/rustc-codegen-cuda/examples/tma_multicast/README.md
@@ -92,14 +92,13 @@ cargo oxide run tma_multicast
 
 ## Expected Output
 
-### On Blackwell Datacenter (sm_100a):
+### On Blackwell Datacenter (sm_100a)
 
 ```text
 === TMA Multicast Example (sm_100a) ===
 
 GPU Compute Capability: sm_100
-Loading PTX from: tma_multicast.ptx
-✓ PTX loaded successfully
+✓ embedded module loaded successfully
 
 --- TMA Multicast (tma_multicast_test) ---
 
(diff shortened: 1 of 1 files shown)


nihalpasham:
Facts all check out:

- PTX ISA: `cp.async.bulk.tensor` needs sm_90+, and `.multicast::cluster` is a performance advisory, not a datacenter restriction
- the deleted "L2 multicast fabric" section was fiction; #668's RTX 5090 run stands
- covers all four passages #966 names, plus two it missed

What was left is wording, which I fixed at landing:

- a few passages narrated the review instead of documenting support:

  ```text
  before: "consumer Blackwell must not be documented as unsupported"
  after:  "Runs on consumer Blackwell (sm_120): the sm_100a PTX
           JIT-loads and passes (RTX 5090, #668)"
  ```

- the sm_120 "Expected Output" section had lost its transcript; I captured a real one on our 5090
- the README hedged Hopper while main.rs still says "Not supported on Hopper (sm_90) or earlier" (they agreed before this diff, and #966's whole complaint is the directory contradicting itself); I aligned the two

nihalpasham (approved):
Docs-only; transcript captured on real hardware; CI green. Merging.

PR #1175, opened 2026-08-23
fix(mir-importer): remove Debug parsing from non-pointer constants

### LINKED ISSUE #1174: Remove remaining Debug parsing for non-pointer MIR constants
## Problem

The MIR importer still relies on rustc `Debug` output for several non-pointer constant categories.

The float constant path was already moved to structured MIR/allocation APIs in #527, but a few non-floating paths still depend on textual representations:

- integer constants have a `Debug`-parsing fallback when byte extraction fails
- enum discriminants fall back to parsing the formatted constant representation
- `SharedArray` const generic values are extracted from `TyConst` debug output

These paths depend on unstable formatting details rather than rustc's structured constant APIs.

The integer fallback is also particularly risky because a parse failure can silently produce `0`.

## Proposed change

Remove `Debug` parsing from the remaining non-pointer constant paths and use structured rustc APIs instead:

- decode integer constants from evaluated allocation bytes
- evaluate enum discriminants through structured constant APIs
- evaluate const generic values through `TyConst::eval_target_usize()`
- fail closed when a constant cannot be decoded instead of silently manufacturing a value

Atomic ordering extraction should preserve its conservative `SeqCst` fallback when the discriminant cannot be decoded.

## Scope

This issue intentionally does **not** cover pointer constants.

The existing pointer/no-provenance path still contains legacy `Debug` parsing and interacts with pointer provenance, relocations, and related MIR pointer work. That should remain a separate change.

## Acceptance criteria

- non-pointer integer constants do not parse rustc `Debug` output
- enum discriminants do not parse rustc `Debug` output
- const generic values do not parse rustc `Debug` output
- integer decoding failures do not silently become zero
- atomic ordering decode failures remain conservative
- pointer/provenance handling is unchanged
- existing MIR importer tests and CUDA regression examples continue to pass

uurl:
Closes #1174

## Summary

Removes the remaining semantic dependence on rustc `Debug` output from non-pointer MIR constant decoding.

This follows the float constant cleanup from #527 and keeps the legacy pointer/provenance path explicitly out of scope.

## Changes

- decode integer constants directly from MIR/allocation bytes
- remove the textual integer fallback and its silent `0` behavior
- replace enum discriminant `Debug` parsing with structured constant extraction
- return `None` when a discriminant cannot be decoded
- preserve `SeqCst` as the conservative atomic-ordering fallback
- replace `SharedArray` const-generic `Debug` parsing with `TyConst::eval_target_usize()`
- keep the existing pointer constant `Debug` handling isolated and unchanged

## Why

rustc `Debug` output is not a stable interface and should not participate in semantic constant decoding.

The previous integer fallback was also unsafe because malformed or unsupported representations could silently decode as zero.

Using structured MIR and allocation APIs makes these paths independent of formatting changes and allows decode failures to remain explicit.

## Scope

This PR intentionally does not change:

- pointer constants
- pointer provenance
- relocations
- address-space handling
- pointer/reference kind tracking

Those paths have different correctness constraints and should remain separate.

## Validation

```text
cargo fmt --all -- --check
cargo test -p mir-importer
git diff --check
```

Results:

* `mir-importer`: 1108 tests passed
* compile-only smoketest: 3/3 passed

  * `atomics`
  * `primitive_stress`
  * `scoped_atomic_load_store`
* `primitive_stress`: 6/6 runtime checks passed
* `primitive_stress` with `CUDA_OXIDE_NO_OPT=1`: 6/6 passed
* `atomics`: 20/20 runtime tests passed
* `scoped_atomic_load_store`: passed

The remaining semantic `Debug` parsing is confined to the legacy pointer/no-provenance path.

### DIFF
--- a/crates/mir-importer/src/translator/rvalue.rs
+++ b/crates/mir-importer/src/translator/rvalue.rs
@@ -2292,9 +2292,6 @@ pub fn translate_operand(
                 );
             }
 
-            // Debug parsing remains temporarily for non-floating constants.
-            let const_str = format!("{:?}", constant.const_);
-
             // Handle pointer-to-array constants (byte strings, typed arrays like [f64; 3], etc.)
             if is_ptr_to_array {
                 return translate_ptr_to_array_constant(
@@ -2365,6 +2362,11 @@ pub fn translate_operand(
                 .deref(ctx)
                 .is::<dialect_mir::types::MirPtrType>()
             {
+                // Pointer constants still have legacy Debug-based handling in the
+                // no-provenance path. Keep that compatibility code isolated here;
+                // non-pointer constants below must not depend on rustc Debug output.
+                let const_str = format!("{:?}", constant.const_);
+
                 // Pointer type constant - could be:
                 // 1. A raw pointer constant (like core::ptr::null()) - just bytes,
                 //    no provenance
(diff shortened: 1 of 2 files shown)


nihalpasham:
Closing as already-fixed. Right issue (#1174), right direction, but #1186 (1f0b065ac, merged Aug 28) rewrote every site this touches, more strictly:

- integer constants: typed allocation walk, hard error on provenance or uninit bytes (here `constant_bytes` zero-fills uninit and follows provenance, so a bad constant could still decode as a wrong number)
- atomic orderings: matched by variant name, hard error on unknowns (here a positional 0-4 map with a silent SeqCst fallback)
- SharedArray const generics: read positionally (here a filter_map that can read ALIGN as N)

The branch now conflicts on every hunk. #1174 closed citing #1186.

PR #1176, opened 2026-08-24
fix(mir-lower): guard shared-pointer aggregate cast fallbacks

uurl:
Refs #883

## Summary

Guard the remaining `emit_pointer_cast` aggregate memory fallbacks against shared memory (`addrspace(3)`) pointer bytes.

PR #1061 already legalized the canonical slice fat-pointer case in SSA by converting the data-pointer half semantically and preserving metadata. This PR keeps that path unchanged and closes the remaining unsafe fallback cases that still use an `alloca` + `store` + `load` round trip.

## What changed

- Add a single `emit_pointer_cast_via_memory` helper for pointer-cast memory fallbacks.
- Reject the fallback when either source or destination contains an `addrspace(3)` pointer, using `llvm_type_contains_pointer_in_address_space`.
- Preserve the existing scalar-pointer remedy through the 64-bit generic address space.
- Keep canonical slice fat-pointer casts on the existing SSA path from #1061.
  - shared pointers in source and destination aggregate positions;
  - array-based pointer-cast fallback;
  - existing slice fat-pointer behavior remaining unchanged.

## Why

A memory round trip exposes the physical representation of shared-memory pointers.

That representation is not a stable Rust-visible address:

- legacy PTX / NVVM uses 64-bit shared pointers;
- modern NVVM may use `p3:32`;
- the raw shared value is a space-local offset, so a valid first shared allocation can have raw value zero.

Scalar pointer conversions already avoid this by bridging through the CUDA generic address space. Aggregate pointer casts that fall back to memory must therefore fail closed instead of exposing raw shared-pointer bytes.

## Validation

```text
cargo oxide fmt --check
PASS

git diff --check
PASS

cargo test -p mir-lower
272 unit tests passed
129 lowering tests passed

PASS

cargo oxide run shared_ptr_transmute
PASS
```

The GPU regression confirms that scalar and slice-fat-pointer casts still preserve shared-pointer semantics.

### DIFF
--- a/crates/mir-lower/src/convert/ops/cast.rs
+++ b/crates/mir-lower/src/convert/ops/cast.rs
@@ -40,11 +40,11 @@
 //! | ptr → integer                                  | genericize named-space pointers, `ptrtoint` |
 //! | integer → ptr                                  | `inttoptr`                                  |
 //! | slice-shaped struct → slice-shaped struct     | fieldwise SSA; `addrspacecast` data as needed |
-//! | other struct → struct                          | `alloca` + `store` + `load`                 |
+//! | other struct → struct                          | guarded `alloca` + `store` + `load`         |
 //! | ptr → ptr (diff addrspace)                     | `addrspacecast`                             |
-//! | struct → integer, equal size                   | `alloca` + `store` + `load`                 |
+//! | struct → integer, equal size                   | guarded `alloca` + `store` + `load`         |
 //! | struct → integer, mismatched size              | cuda-oxide error (see issue #21)            |
-//! | array ↔ anything, equal size                   | `alloca` + `store` + `load`                 |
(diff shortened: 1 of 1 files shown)


nihalpasham:
Closing this PR because the AS3-only physical-shape guard does not establish a sound aggregate pointer-cast boundary.

- Treating any `{pointer, integer}` shape as a semantic slice is not sufficient proof.
- The equivalent AS7 cluster-shared fallback remains exposed to the same representation problem.
- The tests construct lower-level shapes directly and do not demonstrate a real Rust/MIR regression.

This needs a dedicated semantic slice and address-space-origin contract, with a real before/after case and named-address-space controls. That is the standard in the [Discord announcement](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

PR #1178, opened 2026-08-24
fix(mir-lower): legalize cluster-shared slice fat-pointer casts in SSA

### LINKED ISSUE #1177: mir-lower: legalize AS7 slice fat-pointer casts in SSA
## Context

#1060 / #1061 legalized semantic slice fat-pointer casts in SSA for shared-memory `addrspace(3)` pointers.

That legalization was intentionally kept narrow and explicitly excluded cluster-shared `addrspace(7)` pointers so that AS7 semantics could be audited separately.

The remaining AS7 path is reachable through the public cluster DSMEM API:

```rust
let mapped =
    cluster::map_shared_rank(addr_of!(SHMEM) as *const u32, rank);

let slice = core::ptr::slice_from_raw_parts(mapped, 2);
let recast = slice as *const [i32];
```

`map_shared_rank` produces a cluster-shared pointer represented as LLVM `addrspace(7)`. When that pointer becomes the data field of a slice fat pointer, a semantic slice-to-slice pointer cast currently bypasses the SSA legalization and falls through to the aggregate memory fallback.

The current lowering explicitly contains this exclusion:

```rust
if src_data_as == address_space::CLUSTER_SHARED
    || dst_data_as == address_space::CLUSTER_SHARED
{
    return Ok(None);
}
```

As a result, an otherwise canonical slice-shaped `{ pointer, integer metadata }` cast is lowered through `alloca` + `store` + `load` instead of converting its semantic components directly.

## Proposed lowering

Cluster-shared slice fat pointers have the same semantic decomposition used by the existing slice legalization:

```text
{ src_data, metadata }
        |
        +-- extract src_data
        |       |
        |       +-- addrspacecast when address spaces differ
        |
        +-- extract metadata unchanged
        |
        +-- reconstruct destination aggregate in SSA
```

The aggregate representation should never need to be materialized in memory.

This should apply to canonical slice-shaped pointer casts involving `addrspace(7)` in either direction while keeping the existing narrow shape checks:

* exactly two fields
* field 0 is a pointer
* field 1 is integer metadata
* source and destination metadata types match
* semantic pointer casts only

Trait-object-like `{ ptr, ptr }` aggregates and arbitrary aggregate casts remain outside this path.

## Reachability

A concrete end-to-end path exists through distributed shared memory:

```text
map_shared_rank
    ↓
addrspace(7) pointer
    ↓
slice_from_raw_parts
    ↓
slice fat-pointer cast
    ↓
ordinary dereference
```

The resulting dereference must retain cluster-shared semantics and lower to `ld.shared::cluster`.

## Acceptance criteria

* Remove the special-case exclusion for `addrspace(7)` slice fat pointers.
* Reconstruct canonical AS7 slice casts entirely in SSA.
* Preserve slice metadata unchanged.
* Emit `addrspacecast` only for the data-pointer field when address spaces differ.
* Emit no temporary `alloca` / `store` / `load` for the slice cast.
* Cover both AS7 → generic and generic → AS7 lowering.
* Add an end-to-end cluster regression using `map_shared_rank`.
* Verify that the post-cast remote dereference still emits `ld.shared::cluster`.
* Keep non-slice aggregate behavior unchanged.

Related: #883
Follow-up to: #1060, #1061

uurl:
Closes #1177 

## Summary

Extend the existing slice fat-pointer SSA legalization to cluster-shared `addrspace(7)` pointers.

#1061 introduced fieldwise lowering for canonical slice-shaped pointer casts, but intentionally excluded AS7 so that cluster-shared semantics could be audited separately.

This follow-up removes that exclusion after validating the path end to end through `map_shared_rank`.

A canonical slice fat pointer is lowered semantically as:

```text
{ data pointer, metadata }
        |
        +-- extract data pointer
        |       |
        |       +-- addrspacecast when required
        |
        +-- extract metadata unchanged
        |
        +-- reconstruct destination aggregate in SSA
```

No aggregate bytes are materialized in memory.

## Why

Cluster DSMEM mappings produced by `map_shared_rank` carry `addrspace(7)` semantics.

Those pointers can reach a slice fat-pointer cast through ordinary Rust code:

```text
map_shared_rank
    ↓
addrspace(7)
    ↓
slice_from_raw_parts
    ↓
*const [u32] → *const [i32]
```

Before this change, `try_emit_slice_fat_pointer_cast` explicitly rejected casts involving `addrspace(7)`, causing the canonical slice case to fall back to the generic aggregate `alloca` + `store` + `load` path.

There is no need to reinterpret a slice fat pointer as aggregate bytes. Its data pointer and metadata have independent semantics: the data pointer can be converted with `addrspacecast`, while the integer metadata is preserved unchanged.

Keeping the conversion in SSA also makes the cluster-shared address-space semantics explicit through lowering.

## Implementation

* Remove the `CLUSTER_SHARED` early return from `try_emit_slice_fat_pointer_cast`.
* Reuse the existing canonical `{ pointer, integer metadata }` legalization for AS7.
* Preserve the existing narrow shape and metadata checks.
* Replace the previous AS7 fallback regression with coverage that verifies:

  * no `alloca`
  * no `store`
  * no `load`
  * data and metadata extraction
  * SSA reconstruction
  * the required destination `addrspacecast`
* Extend the `cluster` example with an end-to-end AS7 slice fat-pointer test.
* Extend `verify-code-shape.sh` to require `ld.shared::cluster` after the fat-pointer cast.
* Document the new regression in the cluster example README.

## End-to-end regression

The cluster test exercises:

```text
map_shared_rank
    ↓
cluster-shared addrspace(7) pointer
    ↓
slice_from_raw_parts(..., 2)
    ↓
non-inlined slice element-type cast
    ↓
metadata extraction + data-pointer dereference
    ↓
remote DSMEM loads
```

On `sm_120`, each block successfully reads a two-element slice from its neighbor rank after the fat-pointer cast:

```text
Block 0: sum=8021, expected=8021, len=2 ✓
Block 1: sum=8041, expected=8041, len=2 ✓
Block 2: sum=8061, expected=8061, len=2 ✓
Block 3: sum=8001, expected=8001, len=2 ✓
```

The code-shape regression also confirms that the post-cast dereference emits `ld.shared::cluster`.

## Scope

This remains the same narrow slice legalization introduced by #1061.

It does not extend to:

* arbitrary aggregate shapes
* trait-object-like `{ ptr, ptr }` aggregates
* aggregate `Transmute`
* unrelated pointer-cast memory fallbacks

## Validation

```text
cargo fmt --all -- --check
PASS

git diff --check
PASS

cargo test -p mir-lower
270 unit tests passed
129 lowering tests passed

PASS

cargo oxide run cluster
PASS

test_dsmem_slice_fat_pointer_cast
Block 0: sum=8021, expected=8021, len=2
Block 1: sum=8041, expected=8041, len=2
Block 2: sum=8061, expected=8061, len=2
Block 3: sum=8001, expected=8001, len=2

crates/rustc-codegen-cuda/examples/cluster/verify-code-shape.sh
cluster code shape: PASS
```

Related: #883
Follow-up to: #1061

### DIFF
--- a/crates/mir-lower/src/convert/ops/cast.rs
+++ b/crates/mir-lower/src/convert/ops/cast.rs
@@ -59,7 +59,7 @@ use crate::convert::types::{
 use crate::helpers;
 use dialect_mir::attributes::MirCastKindAttr;
 use dialect_mir::ops::MirCastOp;
-use dialect_mir::types::{MirArrayType, MirPtrType, address_space};
+use dialect_mir::types::{MirArrayType, MirPtrType};
 use llvm_export::op_interfaces::{CastOpInterface, CastOpWithNNegInterface};
 use llvm_export::ops as llvm;
 use llvm_export::types::{FuncType, PointerType, PointerTypeExt};
@@ -495,8 +495,7 @@ fn slice_fat_pointer_fields(
 /// The data-pointer half carries address-space semantics and is converted with
 /// `addrspacecast` when the source and destination spaces differ. The integer
 /// metadata half is copied unchanged. No aggregate bytes are materialized in
-/// memory, so an addrspace(3) pointer is never exposed as its target-dependent
-/// physical representation.
+/// memory, so the pointer's address-space semantics remain explicit in SSA.
 fn try_emit_slice_fat_pointer_cast(
     ctx: &mut Context,
     rewriter: &mut DialectConversionRewriter,
@@ -528,13 +527,6 @@ fn try_emit_slice_fat_pointer_cast(
         .downcast_ref::<PointerType>()
         .expect("slice fat-pointer data field must be a pointer")
         .address_space();
-
-    // Cluster-shared pointers have distinct lowering semantics and are intentionally
-    // outside this legalization. Preserve the existing aggregate fallback for AS7.
-    if src_data_as == address_space::CLUSTER_SHARED || dst_data_as == address_space::CLUSTER_SHARED
-    {
-        return Ok(None);
-    }
 
     let extract_data = llvm::ExtractValueOp::new(ctx, val, vec![0])
         .map_err(|e| pliron::input_error_noloc!("slice pointer cast data extraction: {e}"))?;
@@ -1983,7 +1975,7 @@ mod tests {
     }
 
     #[test]
-    fn cluster_shared_slice_fat_pointer_cast_keeps_existing_fallback() {
+    fn cluster_shared_slice_fat_pointer_cast_rebuilds_in_ssa() {
         for cluster_shared_is_source in [true, false] {
             let mut ctx = make_ctx();
             let cluster_shared = dialect_mir::types::address_space::CLUSTER_SHARED;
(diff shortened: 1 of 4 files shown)


nihalpasham:
Closing this PR because it permits generic-to-AS7 `addrspacecast` without proving cluster-shared provenance.

- The physical recognizer can accept any exact two-field `{pointer, integer}` source/destination pair with matching metadata when marked `PtrToPtr`; it does not prove slice origin.
- The supplied source case genericizes the pointer before constructing the slice, so it does not exercise the changed AS7 fat-pointer path.
- A generic pointer may represent global or local memory; an address-space cast cannot establish cluster-shared origin.

A replacement needs a dedicated semantic slice and address-space-origin contract and a genuine fail-before/pass-after regression, consistent with the [Discord acceptance criteria](https://discord.com/channels/1515530041767759993/1515561621156663487/1540970771323551775).

PR #1179, opened 2026-08-24
codegen: accept `char` in device-extern signatures

### LINKED ISSUE #875: `char` is rejected in `#[device]` extern signatures

nihalpasham:
Closes #875.

This independently lands the sound ABI change proposed by @NgoQuocViet2001 in #1049 on a signed maintainer branch; the contributor fork does not allow maintainer edits.

Before, a `#[device] extern "C"` taking or returning Rust `char` was rejected. After, `char` uses the same plain 32-bit slot as CUDA C++ `unsigned int` in parameter, result, and pointee positions.

The device-FFI example now exercises parameter, return, and pointer forms. Its checked-in shape gate builds both the legacy `sm_90` and modern `sm_100` NVVM paths and rejects `signext`/`zeroext`, which the functional round-trip alone cannot detect.

Local checks: 27 rustc-codegen tests, strict clippy for the backend and example, formatting/SPDX/shell checks, and both exact NVVM-IR paths pass.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/device_ffi_test/README.md
+++ b/crates/rustc-codegen-cuda/examples/device_ffi_test/README.md
@@ -176,7 +176,24 @@ extern "C" __device__ float warp_reduce_sum(float val) {
 | `warp_reduce_sum`    | Warp-level sum          |
 | `warp_ballot`        | Warp ballot             |
 | `simple_add`         | Simple a + b            |
+| `char_to_upper`      | ASCII uppercase, `char` |
 | `clamp_value`        | Clamp to range          |
+
+### `char` in a `#[device]` extern
+
+Rust `char` is a 32-bit Unicode scalar value. It therefore lowers to a plain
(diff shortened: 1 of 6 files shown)


nihalpasham:
Reproduced the gap, and @NgoQuocViet2001 had the compiler mapping right in #1049: Rust `char` is a plain 32-bit device-extern slot, with no extension attribute.

I landed it independently on this signed maintainer branch because the contributor fork does not allow maintainer edits. I kept the parameter/return example and added the pointer form plus a checked-in dual-path IR gate. The deliberate `ZeroExtInteger(32)` mutation fails the legacy path and is rejected by the modern `sm_100` shape assertion.

Exact head `889737ff` passes every hosted check, including strict clippy, docs, CodeQL, the complete unit matrix, and the 22-minute GPU-less example compile. No GPU was available locally; the original parameter/return case was also exercised by the contributor on an RTX 5090.

This closes #875 and supersedes #1049.

PR #1181, opened 2026-08-24
fix(cargo-oxide): doctor probes two toolkit tools on PATH only

### LINKED ISSUE #1180: cargo oxide doctor reports the CUDA toolkit missing when it is not on PATH, right after finding four of its components
## What happens

On an install where the CUDA toolkit is not on `PATH` — the layout `doctor` itself tells you to use — `cargo oxide doctor` reports the toolkit missing and exits non-zero, immediately after reporting four components of that same toolkit as found:

```
CUDA headers (cuda.h)... ✓ /home/.../cuda-13.3/include/cuda.h
CUDA toolkit (nvcc)... ✗ nvcc not found
nvJitLink (libnvJitLink.so)... ✓ nvJitLink 13.3
libdevice (libdevice.10.bc)... ✓ /home/.../cuda-13.3/nvvm/libdevice/libdevice.10.bc
cuda-gdb (optional)... - not found (only needed for `cargo oxide debug`)
```

`nvcc` is present the whole time, at `$CUDA_HOME/bin/nvcc`, in the toolkit the four `✓` lines just resolved. `cuda-gdb` is reported missing by the same mechanism.

## Why

`doctor` probes three CUDA toolkit executables, by two different mechanisms:

| tool | how `doctor` finds it | how the command that uses it finds it |
|---|---|---|
| `compute-sanitizer` | `find_cuda_toolkit_executable` | `sanitize` — same helper |
| `nvcc` | `Command::new("nvcc")` — `PATH` only | *(nothing else invokes it)* |
| `cuda-gdb` | `Command::new("cuda-gdb")` — `PATH` only | `debug` — `find_cuda_toolkit_executable` |

`find_cuda_toolkit_executable` tries `PATH`, then the root from `CUDA_TOOLKIT_PATH` / `CUDA_HOME` / `cuda-oxide.toml`, then the standard install roots. The two bare probes see only `PATH`.

Two consequences:

- **`nvcc` is in the required tier**, so this sets `ok = false` and `doctor` exits 1. A working environment is reported as broken.
- **`cuda-gdb` contradicts `cargo oxide debug`.** `codegen_debug` resolves it with `find_cuda_toolkit_executable`, so `debug` finds a cuda-gdb that `doctor` — whose whole job is predicting whether `debug` will work — says is absent.

The toolkit-root contract is `doctor`'s own stated model. Its `cuda.h` failure says so verbatim:

```
Set CUDA_TOOLKIT_PATH or CUDA_HOME to a CUDA Toolkit install root;
when neither is set, /usr/local/cuda is used.
```

So `doctor` tells you to configure the toolkit that way, and then ignores it for two of the three tools it looks for.

`COMPUTE_SANITIZER_FALLBACK_PATHS` records the intended rule in its own doc comment — *"Shared by `sanitize` … and `doctor` so both use the same discovery order by construction."* That helper arrived with the compute-sanitizer probe in #537; the older `nvcc` and `cuda-gdb` probes were never moved onto it.

## Who this affects

Anyone whose CUDA toolkit is not on `PATH`, which includes the setup documented for building this project without a system-wide install. They are told their environment is broken when it is not, and `doctor`'s non-zero exit is the signal the issue template asks bug reporters to attach (#956).

## Scope

Exactly three toolkit executables are referenced anywhere in `cargo-oxide` (`nvcc`, `cuda-gdb`, `compute-sanitizer`); one is already correct and two are not. `ptxas` is *not* probed by `doctor` at all — a separate coverage question, not part of this.

I have a fix ready that routes all three through the existing helper, hoists the `cuda-gdb` fallback list into one constant shared by `doctor` and `debug`, and adds a source-level test so a fourth tool cannot arrive with a bare probe. Happy to open it, or to hold if you would rather the `nvcc` **required/optional classification** be settled first — I deliberately left that tier untouched, since whether `nvcc` should be required at all is a policy call rather than a discovery bug.

vyncint:
Closes #1180.

`cargo oxide doctor` reports the CUDA toolkit missing, and exits non-zero, on an install where the toolkit is not on `PATH` — immediately after reporting four components of that same toolkit as found.

## Before → after

Same machine, same environment, `CUDA_HOME` set, toolkit not on `PATH`:

```
before                                          after
────────────────────────────────────────────    ────────────────────────────────────
CUDA headers (cuda.h)... ✓ .../include/cuda.h   unchanged
CUDA toolkit (nvcc)... ✗ nvcc not found         ✓ Cuda compilation tools, release
                                                  13.3, V13.3.33 (.../bin/nvcc)
libNVVM (libnvvm.so)... ✓ libNVVM 2.0           unchanged
nvJitLink... ✓ nvJitLink 13.3                   unchanged
libdevice... ✓ .../libdevice.10.bc              unchanged
```

`nvcc` was present at `$CUDA_HOME/bin/nvcc` the whole time — in the very toolkit those four `✓` lines resolved.

## Root cause, across all three tools

`doctor` probes three CUDA toolkit executables by two different mechanisms:

| tool | how `doctor` finds it | how the command that uses it finds it |
|---|---|---|
| `compute-sanitizer` | `find_cuda_toolkit_executable` | `sanitize` — same helper |
| `nvcc` | `Command::new("nvcc")`, PATH only | *nothing else invokes it* |
| `cuda-gdb` | `Command::new("cuda-gdb")`, PATH only | `debug` — via the helper |

```
find_cuda_toolkit_executable:   PATH → CUDA_TOOLKIT_PATH/CUDA_HOME/cuda-oxide.toml → standard roots
Command::new(name):             PATH
```

Two consequences:

- `nvcc` sits in the **required** tier, so a working environment sets `ok = false` and `doctor` exits 1.
- `cuda-gdb` **contradicts `cargo oxide debug`**, which resolves it through the helper. Doctor calls absent a tool that the command it exists to predict would find.

The toolkit-root contract is doctor's own stated model — its `cuda.h` failure prints *"Set CUDA_TOOLKIT_PATH or CUDA_HOME to a CUDA Toolkit install root"*. And `COMPUTE_SANITIZER_FALLBACK_PATHS` states the rule the other two break, in its own doc comment: *"Shared by `sanitize` … and `doctor` so both use the same discovery order by construction."* That helper arrived with the compute-sanitizer probe in #537; the older two probes were never moved onto it.

Exactly three toolkit executables are referenced anywhere in `cargo-oxide`, so this is the whole affected path rather than the one symptom I noticed first.

## The change

All three probes go through `DOCTOR_TOOLKIT_TOOLS`, one table pairing each tool with its fallbacks, resolved by `doctor_toolkit_tool`. The `cuda-gdb` list becomes `CUDA_GDB_FALLBACK_PATHS`, shared by `doctor` and `codegen_debug`, so the two answer identically **by construction** rather than by coincidence — `codegen_debug`'s inline copy is gone.

**Deliberately out of scope:** the required/optional tiers. Whether `nvcc` should be required at all is a policy question, and mixing it in would make a discovery fix harder to judge. Say the word if you want it demoted and I will do that separately. `ptxas` is not probed by doctor at all — also a separate coverage question.

## Tests

| kind | what it pins |
|---|---|
| regression | every tool in the table resolves under a configured toolkit root with nothing on PATH — the broken case |
| unaffected control | the configured root is still preferred over the standard install roots, so discovery order is unchanged; only which function doctor calls changed |
| negative | absent from PATH, the configured root and the fallbacks is still `None`; the fix invents no path |
| source invariant | no CUDA toolkit executable may be probed with a bare `Command::new`, read out of the source rather than from the table, so a tool left *out* of the table is still caught |
| single list | `doctor` and `debug` reference one `cuda-gdb` fallback list; a second inline copy fails |

PATH precedence is asserted end to end against the real binary rather than by mutating the process `PATH` — that is `unsafe` in edition 2024 and racy across test threads, and I would rather not add a flaky test to prove a property the CLI can demonstrate directly.

Each was confirmed to bite by reverting the thing it guards:

- the whole `nvcc` probe restored to its original form compiles, fails `no_toolkit_executable_is_probed_by_bare_path`, and the CLI returns to `✗ nvcc not found`;
- re-adding an inline `cuda-gdb` list fails the single-list test;
- adding a bare `ptxas` probe fails the source invariant;
- a stub `nvcc` placed earlier on PATH still resolves from PATH, checked against the real binary.

## Verification

183 tests pass, fmt clean, `clippy --all-targets -- -D warnings` clean, `RUSTDOCFLAGS="-D warnings" cargo doc` clean, all guard scripts pass. Rebased onto `b22efa99`.

Doctor's exit status on my machine is still 1, for a genuinely absent `clang` — the fix does not paper over a real failure.

One file. No open PR touches `crates/cargo-oxide/src/commands.rs`; re-checked immediately before opening.

### DIFF
--- a/crates/cargo-oxide/src/commands.rs
+++ b/crates/cargo-oxide/src/commands.rs
@@ -2592,6 +2592,58 @@ const COMPUTE_SANITIZER_FALLBACK_PATHS: &[&str] = &[
     "/usr/bin/compute-sanitizer",
 ];
 
+/// Fallback locations probed for `nvcc`, in the same idiom and for the same
+/// reason as the list above. `nvcc` ships in the toolkit's `bin/` beside
+/// `compute-sanitizer` and `cuda-gdb`, so it has to be found the same way.
+const NVCC_FALLBACK_PATHS: &[&str] = &[
+    "/usr/local/cuda/bin/nvcc",
+    "/opt/cuda/bin/nvcc",
+    "/usr/bin/nvcc",
+];
+
+/// Fallback locations probed for `cuda-gdb`. Shared by `debug`
+/// (`codegen_debug`) and `doctor` so both use the same discovery order by
+/// construction -- doctor exists to predict whether `debug` will work, so the
+/// two answering differently is the bug this list prevents.
+const CUDA_GDB_FALLBACK_PATHS: &[&str] = &[
+    "/usr/local/cuda/bin/cuda-gdb",
+    "/opt/cuda/bin/cuda-gdb",
+    "/usr/bin/cuda-gdb",
+];
+
+/// Every CUDA toolkit executable `doctor` probes, with its fallbacks.
+///
+/// One table so a fourth tool cannot be added with a different discovery rule.
+/// Before this existed, `compute-sanitizer` resolved through the toolkit root
+/// while `nvcc` and `cuda-gdb` used a bare PATH lookup, so doctor reported the
+/// toolkit missing on an install where it had already found `cuda.h`, libNVVM,
+/// nvJitLink and libdevice under that same root.
+const DOCTOR_TOOLKIT_TOOLS: [(&str, &[&str]); 3] = [
+    ("nvcc", NVCC_FALLBACK_PATHS),
+    ("cuda-gdb", CUDA_GDB_FALLBACK_PATHS),
+    ("compute-sanitizer", COMPUTE_SANITIZER_FALLBACK_PATHS),
+];
+
+/// Resolve one of [`DOCTOR_TOOLKIT_TOOLS`] the way the command that uses it
+/// does: PATH first, then the configured toolkit root, then the standard
+/// install roots.
+/// The standard install roots to try for `name`, or none if the table does not
+/// know it. Split out so the pairing can be checked without a filesystem or an
+/// ambient environment, neither of which a test can control here: the PATH
+/// probe inside `find_cuda_toolkit_executable` shells out to `which` and has no
+/// injection point.
+fn toolkit_tool_fallbacks(name: &str) -> &'static [&'static str] {
+    DOCTOR_TOOLKIT_TOOLS
+        .iter()
+        .find(|(tool, _)| *tool == name)
+        .map(|(_, paths)| *paths)
+        .unwrap_or(&[])
+}
+
+fn doctor_toolkit_tool(ctx: &Context, name: &str) -> Option<PathBuf> {
+    find_cuda_toolkit_executable(ctx, name, toolkit_tool_fallbacks(name))
+}
+
 fn run_compute_sanitizer(
     ctx: &Context,
     example_dir: &Path,
@@ -3867,24 +3919,16 @@ pub fn codegen_debug(
         reject_interop_output_mode(false, &materialization);
     }
 
-    let cuda_gdb = find_cuda_toolkit_executable(
-        ctx,
-        "cuda-gdb",
-        &[
-            "/usr/local/cuda/bin/cuda-gdb",
-            "/opt/cuda/bin/cuda-gdb",
-            "/usr/bin/cuda-gdb",
-        ],
-    )
-    .unwrap_or_else(|| {
-        eprintln!("Error: cuda-gdb not found!");
-        eprintln!();
-        eprintln!("Make sure CUDA toolkit is installed and cuda-gdb is in your PATH");
-        eprintln!("or configured CUDA toolkit root:");
-        eprintln!("  export PATH=\"/usr/local/cuda/bin:$PATH\"");
-        eprintln!("  export CUDA_TOOLKIT_PATH=/usr/local/cuda");
-        std::process::exit(1);
-    });
+    let cuda_gdb = find_cuda_toolkit_executable(ctx, "cuda-gdb", CUDA_GDB_FALLBACK_PATHS)
+        .unwrap_or_else(|| {
+            eprintln!("Error: cuda-gdb not found!");
+            eprintln!();
+            eprintln!("Make sure CUDA toolkit is installed and cuda-gdb is in your PATH");
+            eprintln!("or configured CUDA toolkit root:");
+            eprintln!("  export PATH=\"/usr/local/cuda/bin:$PATH\"");
+            eprintln!("  export CUDA_TOOLKIT_PATH=/usr/local/cuda");
+            std::process::exit(1);
+        });
 
     let cgdb_path = if use_cgdb {
         Some(find_executable("cgdb", &[]).unwrap_or_else(|| {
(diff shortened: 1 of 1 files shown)


nihalpasham:
Fix verified, closes #1180:

- those were the last two bare toolkit probes in cargo-oxide (audited every `Command::new`)
- one table + the shared discovery helper is the right mechanism, and the cuda-gdb fallback list is now a single constant

The test-hermeticity blocker from the earlier version of this comment is resolved by your force-push: the real-name test is gone, the fabricated-name replacements match the sibling idiom, and the table coverage moved to source-based tests. Confirmed on a toolkit-on-PATH machine:

```text
old head:  every_doctor_toolkit_tool_resolves_under_the_configured_root
           FAILED (which nvcc -> /usr/local/cuda/bin/nvcc wins the probe)
new head:  cargo test -p cargo-oxide  185 passed, 0 failed
           fmt + clippy -D warnings   clean
```

One behavior change to note: a found nvcc whose `--version` fails now prints `✓ <path>` instead of failing doctor. That matches the compute-sanitizer arm, so we're keeping it. Labels added, CI approved on the new head. Mergeable as-is after CI.

vyncint:
Both points addressed; force-pushed, rebased onto `60cee76d`.

## The blocker — you are right, and I reproduced it

Putting a real-named `nvcc` on PATH and rerunning the old test:

```
assertion `left == right` failed: nvcc did not resolve under the configured toolkit root
  left: Some("/tmp/pathnvcc/nvcc")
 right: Some("/tmp/cargo_oxide_doctor_tool_.../bin/nvcc")
```

Worth stating plainly: **that test passed here only because this machine has no `nvcc` on PATH — the very configuration the fix is about.** The bug and my bad test shared a root cause in my environment, and I copied the sibling test's structure while changing the one thing that made it hermetic.

Rather than fabricate a name for the same filesystem test — which would then be a near-duplicate of `sanitizer_tool_lookup_uses_project_cuda_toolkit_root` — I split the lookup out and made the whole set hermetic:

| test | kind |
|---|---|
| `each_toolkit_tool_is_paired_with_its_own_fallback_paths` | pure — catches pairing `nvcc` with `CUDA_GDB_FALLBACK_PATHS` |
| `the_fallback_lookup_returns_each_tools_own_list` | pure — each tool gets its own list, unknown gets none |
| `every_doctor_toolkit_tool_call_names_a_tool_in_the_table` | source — an unknown name silently loses its install roots |
| `no_toolkit_executable_is_probed_by_bare_path` | source — unchanged |
| `doctor_and_debug_share_one_cuda_gdb_fallback_list` | source — unchanged |

No filesystem, no environment, so no host shape can move them. Toolkit-root resolution itself stays covered by the existing sibling test.

Suite is clean in all three shapes — toolkit on PATH, toolkit on PATH with `CUDA_HOME` pointing elsewhere, and no toolkit on PATH: **185 passed** each time.

Each still bites, confirmed by breaking what it guards: mispairing the table, probing a tool absent from it, making the lookup return one list for every tool, and re-adding an inline `cuda-gdb` list each fail their own test. Reverting a probe to `Command::new` no longer compiles; restored to its original form so it does, it fails the source invariant and the CLI returns to `✗ nvcc not found`.

## The behaviour change — now on the record

Recorded in the commit message under its own heading. One correction to the framing: it is slightly broader than a *discovered* nvcc. Before, a broken `nvcc` **on PATH** also fell to the catch-all and printed `✗ nvcc not found` — wrong, since it was found. After, both cases report `✓ <path>`, matching the `compute-sanitizer` arm. Demonstrated with a stub that exits 3:

```
CUDA toolkit (nvcc)... ✓ /tmp/brokencuda/bin/nvcc
```

## Manual doctor runs, both configurations

```
toolkit NOT on PATH, CUDA_HOME set:
  CUDA toolkit (nvcc)... ✓ Cuda compilation tools, release 13.3, V13.3.33 (.../cuda-13.3/bin/nvcc)

toolkit ON PATH:
  CUDA toolkit (nvcc)... ✓ Cuda compilation tools, release 13.3, V13.3.33 (.../cuda-13.3/bin/nvcc)
```

Same resolution either way. Exit status here is still 1, for a genuinely absent `clang` — the fix does not paper over a real failure.

fmt, `clippy --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc` and every guard script pass. Thanks for catching this one before it landed.

nihalpasham (approved):
Hermetic-test rework verified on a toolkit-on-PATH machine; CI green. Merging.

Issue #1183, opened 2026-08-25
Parse CUDA targets once and thread CudaArch through codegen


almightychang:
## Summary

PR #1105 gives CUDA target grammar and PTX-floor policy one dependency-free authority, `cuda-target-spec`, but target values still flow through codegen mostly as `&str`/`String`. I propose parsing each external target once into `CudaArch` and passing that value through internal APIs.

This is “parse, don't validate”: `CudaArch` carries proof that the grammar was accepted, while a boolean or string does not. In categorical terms, it is the normalized representation of the valid subobject of raw target spellings (`sm_80` and `compute_80` map to the same architecture). Parsing is the mediating arrow into that representation.

## Current flow

```text
      Str ──parse (partial)──▶ CudaArch ──floor (partial)──▶ u16
       ▲                          │
       └──────── render ─────────┘        law: parse ∘ render = id

  Today, consumers holding an already-proven value still take the loop:

  CudaArch ──render──▶ Str ──parse──▶ CudaArch      (retraction recomputed)


  boundaries                    internal consumers — each spoke re-parses
  ─────────────                 ──────────────────────────────────────────
  CLI / env ─────┐
  device hint ───┼───▶ ((Str)) ──parse──▶ resolver ──returns String──┐
  catalog ───────┘        ▲                                          │
                          └──────────────────────────────────────────┘
                          │
                          ├──parse──▶ arch_satisfies ────▶ bool   ← witness discarded
                          ├──parse──▶ LLVM support gate
                          ├──parse──▶ final-IR check
                          └──parse──▶ PTX-floor / feature selection

  the same target string is parsed 3+ times along one production path
```

Concrete examples:

- An explicit override is parsed at `target.rs:1810-1818`. `validate_target_features` then renders it with `sm()`, and `arch_satisfies` parses it again (`target.rs:1770-1775`, `1653-1658`, `1929-1934`). Generated-target validation gets another rendered string (`target.rs:1825-1831`).
- Resolution returns `String` (`target.rs:1802-1808`). The production path subsequently checks it during requirement merge (`ptx.rs:396-398`), parses it for the LLVM gate (`ptx.rs:409-410`; `target.rs:1878-1884`), parses it after final IR analysis (`ptx.rs:491-500`), and parses it again for PTX feature selection (`ptx.rs:534-536`; `target.rs:1850-1860`).
[...]

PR #1184, opened 2026-08-26
refactor(codegen): parse targets once, one vocabulary owner

almightychang:
Implements the direction proposed in #1183. The three boundary choices it raises -- parse at the resolution
boundary, direct import with the re-export kept, one migration rather than a staged one -- were confirmed
aligned in review.

Rebased onto main now that #1105 has landed, so the range below is this PR's own diff. The patch is
byte-identical to the pre-rebase range (same patch ID, `git range-diff` maps all six commits `=`).

## Problem 1 — targets flow as strings

```text
  BEFORE

  CLI/env ──┐
  hint ─────┼──▶ ((Str)) ──parse──▶ resolve ──returns String──┐
  catalog ──┘       ▲                                         │
                    └─────────────────────────────────────────┘
                    │
                    ├──parse──▶ feature check          the same target string
                    ├──parse──▶ LLVM gate              parsed 3+ times along
                    └──parse──▶ PTX selection          one production path

  AFTER

  CLI/env ──parse──┐
  hint ────parse───┼──▶   CudaArch   ──▶ every check (typed, no reparse)
  catalog ──parse──┘    (proof-carrying)     │
   (partial: once,                           └──render──▶ llc args /
    at the boundary)                                      diagnostics only
```

## Problem 2 — consumers re-encode leaf facts

```text
  BEFORE

  cuda-target-spec                          cuda-oxide-codegen / probe
  ┌──────────────────────┐                  ┌──────────────────────────────┐
  │ spelling set + order │ ═══ hand iso ══▶ │ PtxIsaRequirement enum       │
  │ recorded floor keys  │ ═══ hand copy ═▶ │ candidate string list + order│
  │ floor / render arrows│ ═ re-assembled ▶ │ compare+render, per crate ×2 │
  └──────────────────────┘                  └──────────────────────────────┘
        each ═▶ pinned by a hand-maintained test, correct until someone drifts

  AFTER

  cuda-target-spec                          cuda-oxide-codegen / probe
  ┌──────────────────────────────────┐      ┌──────────────────────────────┐
  │ PtxSpelling  (membership proven, │ ───▶ │ PtxIsaRequirement            │
  │   Ord = numeric, render total)   │      │   = Option<PtxSpelling>      │
  │ RECORDED_PTX_FLOORS order        │ ───▶ │ candidates derived from table│
  │ PtxSpelling::feature_beyond_floor│ ───▶ │ both consumers call it       │
  └──────────────────────────────────┘      └──────────────────────────────┘
        three pinning tests deleted: the properties hold by construction
```

## Changes

- Targets parse into `CudaArch` once at the resolution boundary; resolution returns the typed value plus its source label. Internal checks take `&CudaArch`; rendering happens only at llc arguments and diagnostics.
- `PtxSpelling` (new, in `cuda-target-spec`): construction proves vocabulary membership, `Ord` is numeric, feature rendering is total. `PtxIsaRequirement` wraps `Option<PtxSpelling>` — `.max()` merges hold by construction, and the enum variants had no external consumers.
- `RECORDED_PTX_FLOORS` order is now documented as the selection preference (existing order preserved exactly); the duplicate candidate list is deleted.
- Three pinning tests deleted because their properties became definitional (spelling round-trip, requirement ordering, candidate/key-set equality).

## Behaviour

Accepted inputs keep identical selected targets, feature strings, and error text.

Two rejection paths do change, both deliberately. `sm_05` was accepted before and is now rejected
(`compute capability must contain at least two digits`); `sm_` now reports `compute capability is not a valid
integer` where it previously reported a different message. The unchanged-error-text claim holds on the accept
paths only.

One legacy asymmetry is deliberately kept and pinned by a regression test: the PTX path ignores `compute_XX`
device hints while the NVVM path normalizes them. #1189 tracks that question rather than carrying it here.

## Size and verification

`+938 / −610` overall across six files.

- Full per-target backend probe on the rebased base: 2,365 configurations / 1,016 routes, evidence identical; generated artifacts byte-current.
- Crate suites 179 / 8 / 309 / 7 pass; clippy `-D warnings`; fmt; all 13 guard scripts -- all re-run after the rebase.
- CI-parity container (CUDA 13.1.115 + LLVM 21.1.8): `--all-targets` passes, both `sm_120` libdevice tests executed by name. Run before the rebase, against the byte-identical patch.
- Not verified locally: terminal probe revalidation (evidence pins CUDA 13.3 `ptxas`, not installed).

Can land as one PR or split at maintainers' preference — a per-topic stack exists.


nihalpasham:
Reviewed as a stacked draft. The direction is sound, and I re-derived the key equivalences:

- floor table matches the old `target_minimum_ptx_isa` everywhere except the intended sm_90a 7.8 -> 8.0 fix
- candidate list, its order, and the known-target set are unchanged

Before this leaves draft:

- **sequencing**: this bundles all of #1105; #1105 lands first, then this rebases down to its own diff
- **rebase** onto current main and regenerate the example lockfiles on the new base
- #1183 had no maintainer response when this opened, so "agreed" was premature. That said, the three boundary answers here (parse at resolution, direct import with the re-export kept, single migration) are the right ones. Consider them aligned now.

Small body fixes for the same pass:

- `sm_05` was accepted before and is rejected now, and `sm_` gets a different error message. Both fine, but the "same error text" claim only holds on accept paths.
- the `compute_XX` device-hint question is attributed to #1183, which never asks it; file it separately

We'll re-check the CUDA 13.3 ptxas evidence at validation. Labels added.

nihalpasham:
Gave this the full review it hadn't had yet (the earlier pass only judged direction, back when the diff still bundled #1105), then rebased and validated it ourselves since the conflicts were all fallout from our module split. Your six commits are intact under your authorship; our merge commit only re-homes your `target.rs` delta into the split `target/{arch,detect,features,select,...}` modules.

**The bar for a refactor is "provably no behavior change", so we proved it:**

```
main (8b41ac1b5):  build backend -> compile all 222 examples --compile-only
this branch:       build backend -> compile all 222 examples --compile-only
byte-compare:      612/612 device artifacts identical (.ll, .opt.ll, .ptx, .target)
probe gate:        2374 target configs / 1025 routes both ways, logs byte-identical
candidate order:   derived-from-floors list == the deleted 28-entry list, element for element
```

Plus the full battery: workspace tests, clippy `-D warnings`, all nine repo guards, and a vecadd run on the 5090.

**Two deliberate rejection-path tightenings ship with it** (unreachable from any supported producer, and better errors): `sm_05`-style single-digit capabilities now reject at parse, and bare `sm_` gets a clearer message. The `compute_XX` device-hint asymmetry between the PTX and NVVM paths is preserved as-is; that question lives in #1189.

**Two small things we added:**

- a crate-doc sketch stating the ownership contract:

  ```
  before:  "sm_90a" -> ptx path parses -> nvvm path parses -> probe parses
  after:   "sm_90a" -> CudaArch { capability: 90, suffix: 'a' }  (parse once)
  ```

- a guard test for something the refactor quietly made load-bearing: `RECORDED_PTX_FLOORS` entry order now IS the target-selection preference (first satisfying entry wins), and the old set-equality test died with the list it compared against. One `windows(2)` walk now freezes `[base asc][a asc][f asc]`, so an out-of-section insert fails the build instead of silently reshuffling which target wins.

Review verdict on the whole final delta: clean. Dropping `Depends`; CI + landing next, and #1183 gets closed as implemented.

almightychang:
@nihalpasham 
Thanks for the re-home and the byte-level proof — one heads-up: a8711d0f (test-only, pins the sm_05/sm_ reject paths) landed after your verdict, drop it if you'd rather take the reviewed tip; PR body now scopes the error-text claim to accept paths and points the compute_XX question at #1189.

nihalpasham (approved):

PR #1185, opened 2026-08-27
feat(cuda-device): add scalar bf16_to_f32

lucifer1004:
`cuda_device::convert` has `f32_to_bf16` but not the scalar inverse, so reading a raw bf16 means open-coding the high-half widen at each call site:

```rust
f32::from_bits(u32::from(bits) << 16)
```

## Added

| Function | Purpose |
|---|---|
| `bf16_to_f32(u16) -> f32` | inverse of `f32_to_bf16` for values already exact in bf16 |

This is the same bit-shift already used by the packed bf16 unpackers (`cvt_f32x2_bf16x2` and friends). The low 16 bits of the `f32` become zero.

## Why no generator / no GPU example

This is not a new PTX intrinsic. There is nothing for the generator to admit, and no new GPU example — the same as existing `f32_to_bf16`, which is also a host-side bit reinterpret.

## Tests

- **1.0** — `0x3F80` widens to `1.0`
- **Sign of zero** — `0x8000` widens to `-0.0` (bit-identical)
- **Round-trip** — `bf16_to_f32(f32_to_bf16(1.0)) == 1.0`

Host unit tests only; `fmt --check` clean.

### DIFF
--- a/crates/cuda-device/src/convert.rs
+++ b/crates/cuda-device/src/convert.rs
@@ -44,6 +44,16 @@ pub fn f32_to_bf16_rne(value: f32) -> u16 {
     (bits.wrapping_add(0x7FFF + retained_lsb) >> 16) as u16
 }
 
+/// Widens raw `bf16` bits to `f32` by placing them in the high 16 bits.
+///
+/// This is the inverse of [`f32_to_bf16`] for values that are already exact
+/// in bf16: the low 16 bits of the `f32` become zero.
+#[must_use]
+#[inline(always)]
+pub fn bf16_to_f32(bits: u16) -> f32 {
+    f32::from_bits(u32::from(bits) << 16)
+}
+
 /// Packs two raw `bf16` bit patterns into a `u32`, low half first.
 #[must_use]
 #[inline(always)]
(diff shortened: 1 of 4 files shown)


nihalpasham:
Sound and mergeable as-is. Verified:

- the widen is exact for all 65536 bf16 patterns (bf16 is literally the top half of an f32)
- same layer as nvcc's `__bfloat162float` and the `half` crate's `bf16::to_f32`
- the tests already catch the plausible wrong versions (numeric cast, missing shift, sign mangling)

Two small things I added at landing:

- an exhaustive round-trip test:

  ```rust
  for bits in 0..=u16::MAX {
      assert_eq!(f32_to_bf16(bf16_to_f32(bits)), bits);
  }
  ```

- pointing the three hand-rolled copies (`gemm_sol`, `gemm_sol_final`, `tcgen05_matmul`) at the new helper

CI is approved on the newest head. Labels added.

nihalpasham (approved):
Validation complete on this head (crate tests, per-example gates, compile-only smoketest); CI green. Merging.

PR #1195, opened 2026-08-29
ptx-schedule: classify the asynchronous proxy pipeline as schedule sites

### LINKED ISSUE #1113: ptx-schedule: the async proxy pipeline (cp.async / wgmma / tcgen05 commit-wait pairs) is not classified, so 227 ordering instructions are invisible to a campaign

vyncint:
Closes #1113.

`classify_instruction` covered the synchronous ordering primitives and nothing from the asynchronous proxy. An opcode that classifies as nothing is not a site of any kind, so it received no injection and never appeared in a report — the pipeline a Hopper or Blackwell kernel is built around was the part a campaign could not reach.

## Why none of the existing arms reached these

Each family misses for its own reason, which is why this was not a one-line oversight:

| instruction | why it fell through |
|---|---|
| `cp.async.bulk.tensor...mbarrier::complete_tx::bytes` | carries `mbarrier` in its qualifiers but spells it `mbarrier::`, so it never matches the barrier arm's `mbarrier.` |
| `wgmma.fence.sync.aligned` | a fence that does not start with `fence.` |
| `tcgen05.ld.sync.aligned...` | splits to a base of `tcgen05`, so the ordered-memory arm returns before the `ld`/`st` test |
| `clusterlaunchcontrol.try_cancel.async` | completes through an mbarrier like the bulk copy, and matches nothing at all |

Adds `SiteKind::AsyncProxy` and one prefix arm for the four families. Taking each family whole matches how `bar.`/`mbarrier.` and the warp collectives are already handled, and keeps the consumer halves (`tcgen05.ld`, `clusterlaunchcontrol.query_cancel`) as sites alongside the issues and waits they read.

## Two corrections to the issue, both making the gap larger

**372 instructions, not the 227 I filed.** That figure came from a line-anchored `grep`; 120 of these sit inside single-line inline-asm blocks, which the parser finds and the grep did not. Measured with `analyze_ptx` over the 167 `.ptx` the corpus builds:

```
before:  total_sites=2093   async_proxy_sites=0
after:   total_sites=2465   async_proxy_sites=372
```

Per example: `gemm_sol` 206, `gemm_sol_final` 118, `tcgen05_matmul` 14, `cp_async_zfill` 3, `wgmma` 3, `tma_copy` 2, `tma_multicast` 1.

**`clusterlaunchcontrol.*` is a fourth family** the issue did not name — 25 instructions, same pipeline. Splitting it out would have meant two PRs editing the same match arms, so it is folded in and the issue is updated.

## The perturbed PTX still assembles

The real risk in classifying these is emitting something `ptxas` rejects, and it is a risk I cannot answer by running the kernels: every affected example needs sm_80 or above, and the six Hopper/Blackwell ones cannot execute anywhere I have access to. So I assembled instead. Each of the 7 affected examples was perturbed at `intensity 1.0` under 3 seeds and fed to `ptxas` **for its own target**:

```
perturbed:  21/21 assembled   (sm_80, sm_90a, sm_100, sm_100a)
originals:   7/7  assembled   (control)
```

Negative control, so the 21/21 is not vacuous — corrupting one injected instruction makes `ptxas` reject it:

```
$ sed 's/nanosleep\.u32/nanosleep.BOGUS/' wgmma__7.ptx | ptxas -arch=sm_90a
ptxas line 36; error : Unknown modifier '.BOGUS'
ptxas line 36; error : Unexpected instruction types specified for 'nanosleep'
```

## Tests

Three, following `redux_is_a_warp_collective_and_red_is_still_a_reduction` — the closest precedent, since it also added a classification and had to prove it did not steal from a neighbour. Every opcode spelling in the fixture is taken verbatim from PTX the backend emits, not hand-written.

Mutations, each applied to the fixed tree:

| mutation | result |
|---|---|
| drop `cp.async` from the prefix list | **caught** |
| drop `wgmma.` | **caught** |
| drop `tcgen05.` | **caught** |
| drop `clusterlaunchcontrol.` | **caught** |
| add `ld.` (would swallow ordered memory) | **caught** |
| relax the *fence* arm to `contains()` (would steal `wgmma.fence`) | **caught** |

That last one is what the negative-control test is for. I checked the symmetric case too and am reporting it accurately rather than claiming more: relaxing the *barrier* arm to `contains()` changes nothing, because the bulk-tensor opcode spells the qualifier `mbarrier::` and so never contains `mbarrier.`. The test's comment says so.

## Scope

- The crate README carries the `SiteKind` table, so it gains the row in the same position; the two lists agree in order and content.
- `SiteKind` has no exhaustive `match` anywhere, so the new variant breaks no caller; it is additive in the serialized report.
- **No GPU evidence.** `cp_async_zfill` (sm_80) is the only affected example that could run on hardware I can reach, and g5 spot capacity was exhausted in both permitted regions throughout this session; the T4 alternative is sm_75, below what `cp.async` needs. Rather than claim a run I did not make, the validity argument rests on the `ptxas` assembly above, which covers all four affected targets — including the two no one here can execute.

## Test plan

- `cargo test -p ptx-schedule` — 19 passed (16 pre-existing + 3 new), plus the 3 binary tests.
- `cargo clippy -p ptx-schedule --all-targets -- -D warnings` — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p ptx-schedule` — clean.

### DIFF
--- a/crates/ptx-schedule/README.md
+++ b/crates/ptx-schedule/README.md
@@ -30,6 +30,7 @@ places where another thread's progress can be observed:
 | `Fence` | `fence.*` / `membar.*` ordering |
 | `OrderedMemory` | a load or store carrying an explicit ordering qualifier (`.volatile`, `.acquire`, `.relaxed`, ...) |
 | `WarpCollective` | `shfl`, `vote`, `match`, `redux` and friends |
+| `AsyncProxy` | the asynchronous proxy pipeline: `cp.async.*`, `wgmma.*`, `tcgen05.*` and `clusterlaunchcontrol.*` -- the bulk-copy and matrix issues and the commit/wait pairs that order them |
 | `Backedge` | a `bra` back to the same or an earlier block -- a conservative loop detector |
 
 Each site keeps its ordinal, enclosing callable, byte span, the instruction
--- a/crates/ptx-schedule/README.md
+++ b/crates/ptx-schedule/README.md
@@ -30,7 +30,7 @@ places where another thread's progress can be observed:
 | `Fence` | `fence.*` / `membar.*` ordering |
 | `OrderedMemory` | a load or store carrying an explicit ordering qualifier (`.volatile`, `.acquire`, `.relaxed`, ...) |
 | `WarpCollective` | `shfl`, `vote`, `match`, `redux` and friends |
-| `AsyncProxy` | the asynchronous proxy pipeline: `cp.async.*`, `wgmma.*`, `tcgen05.*` and `clusterlaunchcontrol.*` -- the bulk-copy and matrix issues and the commit/wait pairs that order them |
+| `AsyncProxy` | the asynchronous proxy pipeline: `cp.async.*`, `cp.reduce.async.*`, `wgmma.*`, `tcgen05.*` and `clusterlaunchcontrol.*` -- the bulk-copy, bulk-reduce and matrix issues and the commit/wait pairs that order them |
 | `Backedge` | a `bra` back to the same or an earlier block -- a conservative loop detector |
 
 Each site keeps its ordinal, enclosing callable, byte span, the instruction
--- a/crates/ptx-schedule/src/lib.rs
+++ b/crates/ptx-schedule/src/lib.rs
@@ -41,6 +41,7 @@ pub enum SiteKind {
     Fence,
     OrderedMemory,
     WarpCollective,
+    AsyncProxy,
     Backedge,
 }
 
@@ -333,6 +334,29 @@ fn classify_instruction(instruction: &Instruction<'_>) -> Option<SiteKind> {
     .any(|prefix| head.starts_with(prefix))
     {
         return Some(SiteKind::WarpCollective);
+    }
+
+    // The asynchronous proxy pipeline: the bulk-copy and matrix issues, and
+    // the commit/wait pairs that order them against the synchronous proxy.
+    // None of the arms above reach these. `cp.async.bulk.tensor...` carries
+    // `mbarrier::complete_tx::bytes` in its qualifier list but does not start
+    // with `mbarrier.`, so the barrier arm walks past it; `wgmma.fence` is a
+    // fence that does not start with `fence.`; and `tcgen05.ld` splits to a
+    // base of `tcgen05` rather than `ld`, so the ordered-memory arm below
+    // returns before seeing it. `clusterlaunchcontrol.try_cancel.async`
+    // completes through an mbarrier exactly as the bulk-tensor copy does, and
+    // its `query_cancel` reads that result, so the family is taken whole --
+    // the same shape as the three above.
+    //
+    // An opcode that classifies as nothing is not a site of any kind, so it
+    // receives no injection and never appears in a report. That left the
+    // pipeline a Hopper or Blackwell kernel is built around as the part a
+    // campaign could not reach.
+    if ["cp.async", "wgmma.", "tcgen05.", "clusterlaunchcontrol."]
+        .iter()
+        .any(|prefix| head.starts_with(prefix))
+    {
+        return Some(SiteKind::AsyncProxy);
     }
 
     let mut parts = head.split('.');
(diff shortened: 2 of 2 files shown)


nihalpasham:
Reviewed in depth; this is exactly the fix #1113 needed and the evidence work (corpus recount, 21/21 ptxas on perturbed output with a negative control, the mutation table) made it easy to trust. One gap found, and we patched it on your branch rather than round-tripping.

**The gap: `cp.reduce.async` was still half-covered**

mir-lower's TMA reduce path emits an instruction your prefix list can't match, while its ordering pair can:

```
cp.reduce.async.bulk.tensor.2d.global.shared::cta.add.tile.bulk_group
^^^^^^^^^^^^^^^ "cp.async" doesn't match (reduce sits before async)

cp.async.bulk.commit_group;   <- matches, becomes a site
cp.async.bulk.wait_group 0;   <- matches, becomes a site
```

An issue the fuzzer can't delay against a wait it can perturb is the same half-covered condition your issue describes. No example emits it today (counts unchanged), but the API is user-reachable, so it would have bitten the first kernel that used TMA reduce.

- `cp.reduce.async` added to the AsyncProxy prefix arm, with a comment bullet quoting the emitted opcode
- one fixture line spelled verbatim from mir-lower's inline-asm template, site-count assert 14 -> 15
- README AsyncProxy row extended

**Validation**

- `cargo test -p ptx-schedule` (19 unit + 3 CLI), clippy `-D warnings`, fmt: all clean
- live campaign on an RTX 5090: `fuzz-schedule cp_async_zfill`, report now shows `AsyncProxy=3, Barrier=2`, baseline + 5 seeds all pass

Two adjacent families our verification turned up that are deliberately NOT in this arm: `griddepcontrol.wait/launch_dependents` (PDL grid ordering) and `tensormap.replace.*` (tensormap proxy). Different ordering mechanisms, semantically wrong to fold into AsyncProxy; we'll file follow-up issues. CI approval + landing next.

nihalpasham (approved):

PR #1197, opened 2026-08-29
docs: repoint the three source references the module split invalidated

### LINKED ISSUE #1196: Three source references still point at files the module split removed
#1190 split several large files into module directories. Three places in the tree still send a reader to a path that no longer exists; following any of them lands nowhere.

| where | points at | now lives in |
|---|---|---|
| `crates/cuda-oxide-codegen/tests/spine_kernel_ptx.rs:21` | `crates/mir-lower/src/convert/types.rs` | `convert/types/func_abi.rs` |
| `crates/rustc-codegen-cuda/examples/ref_index_projections/README.md:133` | `crates/mir-importer/src/translator/rvalue.rs` | `translator/rvalue/expr.rs` |
| `crates/rustc-codegen-cuda/examples/subslice_projection/README.md:10` | `crates/mir-importer/src/translator/rvalue.rs` | the walkers under `translator/rvalue/` |

Nothing catches this: a path in prose is not compiled, and `check-book-api-names.sh` verifies device-API *names*, not file paths.

## Two of them need more than a new path

The sentence around the path had gone stale as well, so a `sed` would leave a reference that resolves and still misleads.

**`spine_kernel_ptx.rs`** cites `lowering.rs:132` for `propagate_kernel_attrs` on the next line. That function is at line 336. A line number in prose rots on the next edit of the file it names, so naming the function in its file is the form that survives.

**`ref_index_projections/README.md`** says, in the present tense, that the `Rvalue::Ref` arm "has five cases", and describes case 2 walking an inner loop that breaks on non-`Field` projections. The arm has two labelled cases today; five is what it had *before* the fix that README documents. The passage is a root-cause description of the pre-fix code, so it reads as a false statement about current code.

## Not included

`crates/rustc-codegen-cuda/examples/atomics/README.md` names `intrinsics/atomic.rs` and `ops/atomic.rs` in a pipeline diagram. Those are crate-relative shorthand, and both resolve (`crates/mir-importer/src/translator/terminator/intrinsics/atomic.rs`, `crates/dialect-nvvm/src/ops/atomic.rs`), so they are correct as written.

I also swept the rest of the tree for source paths in tracked docs and comments that do not exist. The only other hits are deliberate: synthetic fixture names inside tests (`intrinsics/probes/removed.ll`, `intrinsics/overlay/test.toml`), two paths asserted *absent* by `cuda-intrinsics-gen` render tests, and the `mktemp` canary in `check-reserved-prefixes.sh`.

vyncint:
Closes #1196.

#1190 split several large files into module directories. Three references still point at paths that no longer exist.

| where | pointed at | now |
|---|---|---|
| `cuda-oxide-codegen/tests/spine_kernel_ptx.rs` | `mir-lower/src/convert/types.rs` | `convert/types/func_abi.rs` |
| `examples/ref_index_projections/README.md` | `mir-importer/src/translator/rvalue.rs` | `translator/rvalue/expr.rs` (+ `rvalue/place_addr.rs` for the address walk) |
| `examples/subslice_projection/README.md` | `mir-importer/src/translator/rvalue.rs` | the walkers under `translator/rvalue/` |

## Two needed the sentence fixed, not just the path

Repointing alone would have left a reference that resolves and still misleads.

**A line number that had also moved.** The same doc comment cites `lowering.rs:132` for `propagate_kernel_attrs`; it is at line 336. Rather than swap one rotting number for another, this names the function in its file — the form that survives the next edit.

**A present-tense claim about code that changed.** `ref_index_projections/README.md` says the `Rvalue::Ref` arm "has five cases" and describes case 2's inner loop breaking on non-`Field` projections. It has two labelled cases today; five is what it had *before* the fix that README documents. The passage is now past tense, and points at where that arm and its address walk live.

## Verified, not assumed

- Every claim was re-read against the code before repointing: `is_kernel_func` still returns `true` iff the func op carries a `gpu_kernel` `StringAttr`, and `translate_place_addr_from_slot` (named in the README) still exists, at `rvalue/place_addr.rs:610`.
- Re-ran a tree-wide sweep for source paths in tracked docs and comments that do not exist. After this change the only remaining hits are deliberate: synthetic fixture names inside tests (`intrinsics/probes/removed.ll`, `intrinsics/overlay/test.toml`), two paths that `cuda-intrinsics-gen` render tests assert are *absent*, and the `mktemp` canary in `check-reserved-prefixes.sh`.

## Deliberately not changed

`examples/atomics/README.md` names `intrinsics/atomic.rs` and `ops/atomic.rs` in a pipeline diagram. Those are crate-relative shorthand and both resolve (`crates/mir-importer/src/translator/terminator/intrinsics/atomic.rs`, `crates/dialect-nvvm/src/ops/atomic.rs`), so rewriting them would add noise, not accuracy.

I did not add a guard for this. A check that every path-shaped string in the docs exists would need an allowlist for the four synthetic fixtures above, and an allowlist is where that kind of guard goes soft. Happy to build one if you want it, but I would rather agree the shape with you than ship a check that can be green while wrong.

## Test plan

- `cargo test -p cuda-oxide-codegen --test spine_kernel_ptx` — 1 passed (the edited file is that test's module doc).
- `scripts/check-spdx-headers.sh`, `scripts/check-book-api-names.sh` — pass.
- Confirmed all four newly-named paths exist.

### DIFF
--- a/crates/cargo-oxide/README.md
+++ b/crates/cargo-oxide/README.md
@@ -443,7 +443,7 @@ crates/cargo-oxide/
     ├── main.rs       # CLI definitions (clap) + dispatch
     ├── backend.rs    # Backend discovery + build logic
     ├── artifact_identity.rs  # Embedded artifact identity/provenance
-    └── commands.rs   # All command implementations
+    └── commands/     # Command implementations, one module per command group
 ```
 
 ## Future Commands
--- a/crates/cuda-intrinsics-gen/src/render/families/packed.rs
+++ b/crates/cuda-intrinsics-gen/src/render/families/packed.rs
@@ -380,7 +380,7 @@ pub(in crate::render) fn packed_conversion_constraint(record: &CatalogIntrinsic)
 
 /// Whether the conversion lowers through a typed NVVM intrinsic call rather
 /// than inline PTX. Only the scalar-`f32` pair narrowed to FP8 does; see the
-/// matching predicate in `resolve.rs` for why.
+/// matching predicate in `resolve/families/packed_conversion.rs` for why.
 fn packed_conversion_uses_typed_nvvm(record: &CatalogIntrinsic) -> bool {
     let conversion = record
         .packed_conversion
--- a/crates/cuda-oxide-codegen/tests/spine_kernel_ptx.rs
+++ b/crates/cuda-oxide-codegen/tests/spine_kernel_ptx.rs
@@ -18,11 +18,12 @@
 //! carrying a `gpu_kernel` *attribute* on the func op. There is no calling
 //! convention to set by hand and no naming convention. The chain is:
 //!
-//!   1. `mir-lower`'s `is_kernel_func` (crates/mir-lower/src/convert/types.rs)
+//!   1. `mir-lower`'s `is_kernel_func`
+//!      (crates/mir-lower/src/convert/types/func_abi.rs)
 //!      returns `true` iff the func op's `attributes` contain a `StringAttr`
 //!      under the identifier `gpu_kernel`. The *value* is not inspected, only
 //!      presence; the rest of the pipeline writes `"true"`.
-//!   2. During lowering (lowering.rs:132) `propagate_kernel_attrs` copies a
+//!   2. During lowering, `lowering.rs`'s `propagate_kernel_attrs` copies a
 //!      `gpu_kernel="true"` `StringAttr` onto the produced `llvm::FuncOp` (plus
 //!      any optional `cluster_dim_*`/`maxntid`/`minctasm` ints).
 //!   3. `llvm-export`'s `PtxExportConfig::emit_ptx_kernel_keyword()` is `true`,
--- a/crates/mir-importer/src/translator/layout.rs
+++ b/crates/mir-importer/src/translator/layout.rs
@@ -5,8 +5,9 @@
 //!
 //! Both the type importer (`translator/types.rs`, which records tag and
 //! field byte offsets on `MirEnumType`) and constant decoding
-//! (`translator/rvalue.rs`, which slices payload bytes out of constant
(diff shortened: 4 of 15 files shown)


nihalpasham:
Reviewed and finished this one off ourselves; we've been landing big refactors and the leftover work here traces straight back to them, so it felt wrong to bounce it back to you.

**Your three repoints: all verified correct**

- `is_kernel_func` really is at `convert/types/func_abi.rs:50`, `propagate_kernel_attrs` at `lowering.rs:336`, the `Rvalue::Ref` arm at `rvalue/expr.rs:497`
- swapping the rotted `lowering.rs:132` line citation for a function-in-file citation was the right form; nothing left to rot

**The gap: the sweep matched one spelling, the tree uses three**

```
what the sweep grepped:   crates/mir-lower/src/convert/types.rs   -> 3 hits, fixed
what the tree also says:  mir-lower/src/convert/types.rs          (crate-relative)
                          "see the place walker in `rvalue.rs`"   (bare basename)
```

Same split commit (40f03e1ea) deleted 16 files, and 14 more references to them survived in the other two spellings, two of them in the user-facing book. We stacked a commit on your branch repointing all 14 (your commit is untouched underneath), and re-swept full-path, crate-relative, and basename spellings of all 16 deleted files: zero dangles left. So the "Closes #1196" in the body is now genuinely true.

**Validation**

- every new pointer read against the tree before writing it (the cited function/table subject actually lives there)
- comment/markdown-only diff, `cargo fmt --check` and `cargo check` on the touched crates clean, book builds

We'll approve CI on the new head and land. One thought your PR itself proves: a shorthand-aware sweep is hard to do by hand, so if you still want to file that broken-path-check issue you offered, go for it.

nihalpasham (approved):

PR #1200, opened 2026-08-29
fix(finalizer): reject links that drop expected kernels

### LINKED ISSUE #1166: finalizer: reject successful links that drop expected kernels

ZenAlexa:
## Summary

Prevent a successful nvJitLink result from silently publishing an image that dropped a declared CUDA kernel. This continues #1167, preserving Raul Estrada's original authored commit, and closes the three cross-path gaps identified in that review.

## Changes

- root expected kernels with `-kernels-used` and structurally verify the linked PTX/cubin inventory
- use `ArtifactEntryKind` as the shared entry-kind source of truth
- carry a versioned, normalized kernel-entry sidecar through NVVM IR, LTOIR, cargo-oxide compatibility finalization, file loading, PTX bridging, and cache identity
- add a live negative nvJitLink regression that first reproduces a successful empty link, then proves the expected root is retained

## Testing

- [ ] `just check` passes (upstream CI in progress)
- [ ] live CUDA Toolkit regression passes (in progress)
- [x] `cargo test -p oxide-artifacts --all-targets --features object`
- [x] `cargo test -p cuda-artifact-finalizer --all-targets`
- [x] `cargo test -p cargo-oxide`
- [x] `cargo test -p mir-importer`
- [x] `cargo test --lib` in `crates/rustc-codegen-cuda`

## Checklist

- [x] SPDX headers on new source files (no new source files)

Fixes #1166.

### DIFF
--- a/crates/cargo-oxide/src/commands/artifacts.rs
+++ b/crates/cargo-oxide/src/commands/artifacts.rs
@@ -111,6 +111,7 @@ pub(super) const GENERATED_ARTIFACT_SUFFIXES: &[&str] = &[
     "ptx.identity",
     "target",
     "options",
+    "kernels",
     "cubin.target",
 ];
 
--- a/crates/cargo-oxide/src/commands/interop.rs
+++ b/crates/cargo-oxide/src/commands/interop.rs
@@ -314,9 +314,10 @@ pub(super) fn read_interop_recorded_target(ir_path: &Path) -> Result<String, Str
     if target.is_empty() {
         return Err(format!("recorded CUDA target {} is empty", path.display()));
     }
-    match (lines.next(), lines.next()) {
-        (None, None) => Ok(target.to_string()),
-        (Some(marker), None) if marker == oxide_artifacts::COMPILE_OPTIONS_TARGET_MARKER => {
+    let markers = lines.collect::<Vec<_>>();
+    match markers.as_slice() {
+        [] => Ok(target.to_string()),
+        [marker] if *marker == oxide_artifacts::COMPILE_OPTIONS_TARGET_MARKER => {
             let options_path = ir_path.with_extension("options");
             if !options_path.is_file() {
                 return Err(format!(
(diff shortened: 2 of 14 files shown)


ZenAlexa:
I could not reproduce the live negative that this draft needs. On CUDA 13.3, both `sm_86` and `sm_120a` fixtures with the observed CDP2 dependency fail at `nvJitLinkComplete` with `NVJITLINK_ERROR_PTX_COMPILE` (`Unresolved extern function '__cudaCDP2Malloc'`); removing `llvm.used` instead lets libNVVM drop the kernel before finalization. I'm closing this rather than proposing synthetic-only coverage without the issue's real regression.

PR #1203, opened 2026-08-29
Stop opt from building .global lookup tables of shared-memory addresses (invalid PTX under LLVM 23)

nihalpasham:
## The bug

A kernel `match` that picks a per-pipeline-stage shared buffer gets "optimized" into an object that cannot exist:

```text
match stage {                       opt's favorite trick: precompute the
    0 => &SMEM_A0,                  4 answers into a constant table and
    1 => &SMEM_A1,         ──►      index it. Except the answers are
    2 => &SMEM_A2,                  SHARED-MEMORY addresses: per-block
    3 => &SMEM_A3,                  runtime values. A compile-time table
}                                   of them is physically impossible.

the failure chain:
  opt (LLVM 23)   builds the table                       ← trigger
  llc             prints it, exit 0, says nothing        ← silence
  ptxas / JIT     "Variable used as initial value not
                   in .global or .const state space"
                   → CUDA_ERROR_INVALID_PTX (218)        ← too late,
                                                           on a B200
```

Both PTX encodings are invalid (`= {sym}` and `= {generic(sym)}` alike, verified against ptxas 13.0 and 13.3). Affected: `gemm_sol_final` (loud failure on sm_100) and `gemm_sol` (same poison, masked — see below).

## Why it appeared now

`opt` decides whether the table trick is worth it from its target GPU, and we pass no `-mcpu`, so LLVM's built-in default decides:

```text
LLVM 22 default: sm_30  → trick OFF  → we were safe by accident
LLVM 23 default: sm_75  → trick ON   → llvm/llvm-project PR #176021
```

The bug was always latent, not new: LLVM 22 with an explicit modern `-mcpu` builds the same bad tables (upstream's `validLookupTableConstant` never checks address spaces). Confirmed by A/B on byte-identical `.ll`: opt-22 → 0 tables, opt-23 → 8 (4 of them shared-space), and by an isolation matrix over {old,new} × {opt,llc} × {input IR} — the trigger is the opt default, not our IR and not llc.

## The fix

- **One flag** at the single choke point both frontends share (`optimization_args()`): `-switch-to-lookup=false`. Accepted identically by every supported `opt` (LLVM 21+); kept permanently since our LLVM floor spans the broken majors (upstream precedent: NVPTX already disables relative lookup tables, llvm#159748). Switches lower to `brx.idx` code-label tables instead — legal PTX, and *faster*: a few uniform ALU ops instead of a dependent `.global` load on the mbarrier/TMA hot path.
- **A safety net** independent of the flag: right after llc, the pipeline scans the PTX and **fails the build** if any `.global`/`.const` initializer references a `.shared` symbol (bare or `generic()`), printing the offending line. Whatever pass invents this object next, it becomes a compile error, not a runtime 218.

## The masked failure this uncovered (examples)

`gemm_sol` carried the same broken tables yet reported PASS — because its host treated *any* load failure with error 218 as "right PTX, wrong GPU" and exited 0 via PTX-only verification, even on the sm_100 machine that should have executed it. `tcgen05` and `tcgen05_matmul` shared the pattern.

- All three now check GPU capability *before* loading (the model `gemm_sol_final` already used); error 218 on a capable GPU is a hard failure.
- The stricter path immediately caught a second real bug: `tcgen05`'s MMA helper mixed `cta_group::1` and `cta_group::2` in one function, which ptxas 13.x rejects — that module was never assemblable for sm_100, masked by the same fallback. Split into `_cg1`/`_cg2`. Its PTX-only verification also now assembles against the module's real `.target` (was hardcoded `sm_120a` and never actually succeeded).

## Regression coverage

- New `shared_stage_select` example: the exact match-over-stage-buffers shape, runs on any GPU; with the flag removed it fails the build via the new scan.
- smoketest's GPU-gated lanes (tcgen05/wgmma/blackwell-mma) now **assemble** their PTX with ptxas when one is available, instead of only checking that llc produced a file — closing the hole that let invalid PTX ship: GPU-less CI never assembled, and only sm_100 hardware ever JIT-compiled these modules.

## Validation

```text
isolation matrix         opt-22 → 0 tables/PASS; opt-23 → 8 tables/FAIL;
                         old IR through opt-23 → same FAIL (trigger pinned)
ptxas 13.3 sm_100a       regenerated gemm_sol + gemm_sol_final: PASS
PTX sweep, all examples  0 shared-symbol initializers, 0 data tables
cargo test --workspace   5305 passed, 0 failed
full GPU smoketest       221/221 (new example included), ptxas gate active
B200 (sm_100) hardware   gemm_sol_final: 16.7M/16.7M exact BF16 matches,
                         1472 TFLOPS (90.5% of cuBLASLt) — no perf cost
                         from dropping the tables; shared_stage_select and
                         the unmasked gemm_sol/tcgen05 lanes all pass
```

## Note for reviewers

A separate perf regression in `gemm_views` (`naive raw` 7218 → 5708 GFLOPS) was found during this investigation. It is **not** caused by this PR (that example has no switch tables and its PTX is byte-equivalent with and without the flag) — it's an independent llc-23 loop-shape change that defeats ptxas's SASS unroller (register count drops 38 → 26). Tracked separately.


nihalpasham:
CI's examples-compile failure here was the new ptxas gate doing its job: it exposed that the job's llc-21 floor pin has been producing unassemblable tcgen05 PTX all along (llc-21 predates `llvm.nvvm.stmatrix.*` and leaks the dotted intrinsic name into the PTX as an `.extern .func`, exit 0, silently). Two additions on the branch:

- the floor gate moves to **llc-22** (true floor for the stmatrix family; 221/221 compile-only with the gate active under llc-22)
- a second post-llc scan rejects leaked dotted externs with the intrinsic **named** and the remedy stated, so an under-floor llc now fails at build time instead of ptxas dying with `Parsing error near '.nvvm'`

PR #1204, opened 2026-08-29
[do not merge] CodeQL PR-attribution probe

elibol:
Comment-only touches of the six example files whose pre-existing kernel code CodeQL flags on #1102. If this PR's CodeQL check passes, the alerts are already in main's baseline and #1102's failures are attribution artifacts of the large diff; if it fails with the same alerts, the PR gate attributes them to any PR touching these files. Will be closed once the check reports.


elibol:
Probe answered: comment-only touches of these six files raise no new alerts, so the #1102 CodeQL findings come from its changed macro-generated launch code being re-fingerprinted, not from merely touching the files. Closing.

PR #1205, opened 2026-08-30
Restore ptxas SASS loop unrolling under LLVM 23 (disable llc's late branch folding); assemble every compile-only PTX in CI

nihalpasham:
## The bug

After the LLVM 23 bump, `gemm_views`' naive raw kernel lost 26% (7218 → 5708 GFLOPS) while its safe twin stayed put. The PTX math was identical — only the *jump layout* of the loop changed:

```text
llc-22 (fast)                          llc-23 (slow)
$LOOP:                                 $LOOP:
   ...same loads and FMAs...              ...same loads and FMAs...
   setp.eq  %p, %r3, %r26;                setp.eq  %p, %r3, %r26;
   @%p  bra $EXIT;    ← if done, leave    @!%p bra $LOOP;  ← if NOT done,
   bra.uni  $LOOP;    ← jump back                            jump back;
$EXIT:                                                       exit falls
                                                             through
```

LLVM 23's NVPTX backend learned to reverse branch conditions (llvm/llvm-project #191889, follow-up #191890), letting branch folding + block placement collapse the two-jump idiom. Legal, one instruction smaller — but **ptxas's SASS unroller recognizes hot loops by that layout** and stops unrolling the folded form:

```text
what ptxas produced          llc-22 layout    llc-23 layout
  registers                      38               26
  FMAs in the loop               18                4
  loads in flight                35                7    ← the 26%
```

Bisected by hand-editing PTX (live-benched, bit-identical numerics): ptxas requires the old orientation at **two** sites — the loop guard's taken-branch entering the loop, and the latch's taken-branch exiting with `bra.uni` continue. Negated predicates themselves are innocent.

## The fix

Two ancient, generic, disable-only llc flags on every llc invocation:

```text
-disable-branch-fold -disable-block-placement
    accepted identically by llc-21 / 22 / 23
    only TURN OFF passes → no new behavior to be wrong
    cost: PTX +2.6% lines suite-wide (redundant jumps that
    ptxas discards when laying out SASS itself)
```

Result, live on an RTX 5090: naive raw back to **7201 GFLOPS** (README table refreshed; all four kernels match the pre-bump numbers again). Upstream has no scoped opt-out (the enabling PR shipped with a single LGTM and no discussion of ptxas loop-shape sensitivity); an LLVM issue and an internal NVBug for the ptxas side are queued separately.

## Is it safe for everything else?

- **Correctness**: full GPU smoketest **221/221** — every example executes and self-checks numerically with the flags on.
- **Codegen sweep**: 890 kernels across 197 modules, per-kernel `ptxas -v` registers compared with/without the flags. 45 kernels move, every one explained:

```text
class                                    examples                         meaning
the unroller waking back up              table-lookup scan loops now      the intended
(registers AND SASS body grow)           fully unroll (SASS 64→~400),     effect
                                         array_index, split_at_mut, ...
exact return to pre-bump registers       dpx_minmax_chains (10 kernels    restoration
                                         20→18 = old baseline), others
tiny scheduling deltas, SASS size flat   hashmap_v3 (48→46),              benign
                                         set_discriminant_niche (38→39)
```

  Zero unexplained movers, zero below pre-bump baselines.
- **Benched where benches exist**: gemm_views restored; the warp-reduction family measured at ±0.3% (noise) at example scale and marginally *faster* at compute-bound scale.
- Honest asterisk: a few table-lookup micro-examples now unroll *beyond* their pre-bump shape (regs → 40). Numerically verified on hardware; no perf gate exists for them; expected faster (unrolled scans), not benchmarked.

## Also in this PR: ptxas gate for every example in CI

The compile-only smoketest (CI's mode) now assembles **every** example's PTX with ptxas, not just the GPU-gated categories — closing the remaining "llc emits garbage silently" window for all 221 examples. Measured cost: 183 files assembled in +4.5s, ~3% of the compile-only run. Normal-mode verdicts keep their scope: on a GPU box, execution supersedes assembly.

## Separate finding, measured and closed: the warp-reduction family

LLVM 23 unrolls the warp-shuffle reduction loops less at the IR level (PTX 97→60 lines, before ptxas). Benched properly: **immaterial** — ±0.3% at example geometry, 0.06-0.14% *faster* at 25M-thread scale, bitwise-identical outputs. Documented, deliberately not "fixed": chasing the old unrolled shape would spend registers for nothing.

## Validation

```text
cargo test --workspace          5307 passed, 0 failed
full GPU smoketest              221/221 (321s)
compile-only + wide gate        221/221 (141s, gate cost +4.5s)
register sweep                  890 kernels, all movers accounted for
gemm_views live bench           naive raw 5708 → 7201 GFLOPS
fmt / clippy / lock + contract  clean
```

### DIFF
--- a/crates/cuda-oxide-codegen/src/ptx.rs
+++ b/crates/cuda-oxide-codegen/src/ptx.rs
@@ -273,9 +273,13 @@ const DISABLE_SWITCH_LOOKUP_TABLES: &str = "-switch-to-lookup=false";
 /// - Measured on gemm_views' sgemm_naive_raw (RTX 5090, live benches,
 ///   bit-identical numerics): folded layout drops 38 -> 26 registers and
 ///   FFMA 18 -> 4, costing 26% throughput (7218 -> 5708 GFLOPS). With
-///   these flags: 38 registers, 7226 GFLOPS. Zero register movement on
-///   the 16 other kernels across 6 examples; PTX grows 1-6% in lines,
-///   all redundant jumps ptxas discards.
+///   these flags: 38 registers, 7226 GFLOPS. Full-suite sweep (890
+///   kernels, 197 modules): 45 kernels change registers, every one the
+///   SASS unroller re-enabling (register counts and SASS bodies grow,
+///   e.g. the table-lookup scan loops now fully unroll) or an exact
+///   return to the pre-LLVM-23 baseline; zero unexplained movers, all
+///   affected examples numerically verified on hardware. PTX grows
+///   +2.6% in lines suite-wide, all redundant jumps ptxas discards.
 ///
 /// Permanent until ptxas learns to unroll both layouts or upstream adds
 /// an opt-out for the NVPTX folding; an internal NVBug and an LLVM issue
--- a/crates/cuda-oxide-codegen/src/ptx.rs
+++ b/crates/cuda-oxide-codegen/src/ptx.rs
@@ -8,7 +8,7 @@ use crate::generated::GeneratedModuleRequirements;
 use crate::llvm_tools::LlvmToolchain;
 use crate::options::BackendOptions;
 use crate::target::{
-    ModuleRequirements, detect_module_requirements_in_llvm_file,
+    ModuleRequirements, PtxIsaRequirement, detect_module_requirements_in_llvm_file,
     merge_generated_module_requirements, merge_generated_module_requirements_for_target,
     required_ptx_feature, resolve_ptx_target_with_generated, validate_ptx_isa_for_llvm_major,
     validate_target_features, validate_target_for_llvm_major,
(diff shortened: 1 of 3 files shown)


nihalpasham:
The examples-compile failure was the new all-examples ptxas gate making its fourth catch, and this one is a pre-existing bug on main, not from this branch: llc-22 (CI's floor pin) emits DWARF label-difference expressions (`.debug_pubnames` end minus start, a PTX ISA 7.5 feature) into full-debug modules while declaring the target's default `.version` (7.0 at sm_80), so ptxas rejects them. Nothing ever assembled full-debug PTX before this gate: GPU boxes target sm_120a where the default version is already high enough, and llc-23 shapes its debug sections differently.

Fix in `049db58d5`: full-debug modules raise the existing PTX-ISA requirement floor to 7.8 (nearest supported spelling at or above 7.5) via `required_ptx_feature`, which already refuses to downgrade newer targets. `debug` and `reborrow` now declare `.version 7.8` and assemble under llc-22; full compile-only smoketest under `CUDA_OXIDE_LLC=llc-22`: 221/221.

PR #1212, opened 2026-08-30
[WIP - Please do not review] feat(intrinsics): validate PTX floors from evidence

almightychang:
## Summary

- add a CUDA 13.2.51 `ptxas` floor probe that lowers each resolved intrinsic at its declared PTX floor, requires assembly success there, and requires rejection at the immediately preceding PTX spelling
- record the accepted/rejected boundary, target, rejection diagnostic, and SHA-256 identities of both `ptxas` and the Rust LLVM `llc`
- represent floors below the live toolchain boundary as `unverifiable` instead of manufacturing a verdict
- make normal catalog resolution require exact floor-evidence coverage and equality with every resolved policy floor; CI therefore checks committed evidence without needing `ptxas`

## Population and measurement

The 160 literal `minimum_ptx` declarations in `intrinsics/overlay/*.toml` split into 102 probeable and 58 unprobeable declarations. Resolution expands compact admission matrices, so the generator's actual validation population is larger:

- 1,025 resolved intrinsic policies total
- 942 measured: CUDA 13.2.51 accepted the declared floor and rejected the immediately preceding PTX spelling
- 83 explicitly unverifiable: their floor is below PTX 6.3

The measurement used:

- `ptxas` CUDA 13.2 V13.2.51, SHA-256 `c0a7c9752d1a1b29310ffbb453e8b895fb09e909dcf7436fef5c2c986c1a658e`
- Rust LLVM 23.1.0 from nightly 2026-08-28, `llc` SHA-256 `ca7af83e237af640fe337faff2eff49055988921933186ab5535b38addeb7b49`
- no GPU execution

CUDA 13.2 `ptxas` accepts no target below `sm_75`, and PTX 6.2 cannot name `sm_75`; PTX 6.3 can. The negative diagnostic is retained per record, including cases where instruction and target floors are coupled. This also means `PtxIsaRequirement::Ptx62` appears unreachable with this current assembler, but this PR does not change requirement lowering.

CI does not rerun this experiment. It follows the existing recorded-evidence model: the live toolkit produces committed evidence, while stock runners validate the resolved catalog against that evidence. `probe --all --skip-terminal --per-target` continues to work without `ptxas`.

## Positive control

I deliberately probed `abs_bf16x2` with a false PTX 7.1 declaration. The command failed because CUDA 13.2 accepted the PTX 7.0 negative probe:

```text
error: below-floor positive control failed: ptxas accepted abs_bf16x2 at PTX 7.0, below declared PTX 7.1
```

## Legacy barrier note

The NVIDIA PTX ISA states exactly:

> `bar.sync` without a thread count introduced in PTX ISA version 1.0.
>
> Register operands, thread count, and `bar.{arrive,red}` introduced in PTX ISA version 2.0.

It also states that those register/count/arrive/red forms require `sm_20` or higher. Source: [NVIDIA PTX ISA 9.3, bar/barrier notes](https://docs.nvidia.com/cuda/parallel-thread-execution/#parallel-synchronization-and-communication-instructions-bar-barrier).

Accordingly, the fixed-immediate `sync.toml` declaration appears correct at PTX 1.0, while counted `bar.sync` and `bar.arrive` declarations appear under-declared. They are in the unprobeable population, and this PR intentionally does not edit `counted_barrier.toml` or `sync.toml` because those files are being changed concurrently.

## Not verified here

- CUDA 13.3 is not installed on this machine, so the new evidence has its own CUDA 13.2.51 profile and does not modify or claim to refresh the existing CUDA 13.3 evidence.
- Floors below PTX 6.3 cannot be measured by this supported assembler/target combination and remain explicitly `unverifiable`.
- No runtime behavior was tested; the GPU is currently unavailable, and floor validation requires only PTX assembly.

## Verification

- `cargo test -p cuda-intrinsics-gen` — 311 passed
- `git diff --check`
- full live `probe-floors` run using CUDA 13.2.51
- false-floor positive control described above

### DIFF
--- a/crates/cuda-intrinsics-gen/src/main.rs
+++ b/crates/cuda-intrinsics-gen/src/main.rs
@@ -94,6 +94,26 @@ fn try_main() -> Result<()> {
             },
             ProbeInvocation::Candidate(options) => probe::run_candidate(&repo_root, options),
         },
+        "probe-floors" => {
+            let llc = take_option(&mut arguments, "--llc")?
+                .map(PathBuf::from)
+                .context("probe-floors requires --llc FILE")?;
+            let ptxas = take_option(&mut arguments, "--ptxas")?
+                .map(PathBuf::from)
+                .context("probe-floors requires --ptxas FILE")?;
+            let intrinsic_id = take_option(&mut arguments, "--intrinsic")?;
+            let declared_floor = take_option(&mut arguments, "--declared-floor")?;
+            reject_extra(arguments)?;
+            probe::run_floors(
+                &repo_root,
+                probe::FloorProbeOptions {
+                    llc,
+                    ptxas,
+                    intrinsic_id,
+                    declared_floor,
+                },
+            )
+        }
         "check-abi-history" => {
             let base_ref = take_option(&mut arguments, "--base-ref")?
                 .context("check-abi-history requires --base-ref REF")?;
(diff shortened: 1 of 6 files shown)


nihalpasham:
Please submit an issue prior to opening a PR directly for any discussion.

almightychang:
@nihalpasham 
My bad. Accidentally pushed the draft to upstream.

PR #1214, opened 2026-08-31
ptx-schedule: classify grid dependency control sites

### LINKED ISSUE #1207: ptx-schedule: griddepcontrol.wait / launch_dependents are unclassified ordering instructions
Found while verifying #1195. The backend emits programmatic dependent launch (PDL) control:

```
griddepcontrol.launch_dependents;   // let the dependent grid start
griddepcontrol.wait;                // wait for the prior grid's writes
```

Both order one grid's memory against another's, so they are schedule sites in the same sense as barriers: a delayed `wait` vs the prior grid's flush is a real interleaving a campaign should be able to force.

- `classify_instruction` matches neither today: no injection, never in a report
- they do NOT belong in AsyncProxy (#1195): that kind models the async proxy pipeline (bulk copy / wgmma / tcgen05 commit-wait chains); PDL is grid-to-grid ordering, a different mechanism
- open design question, same as #1113 had: fold into an existing kind or add a `GridDependency` kind

Grep `griddepcontrol` in crates/mir-lower/src to see the emit sites.

uurl:
Closes #1207

## Summary

Classify programmatic dependent launch control instructions as schedule-sensitive sites in `ptx-schedule`.

This adds a dedicated `SiteKind::GridDependency` for:

* `griddepcontrol.launch_dependents`
* `griddepcontrol.wait`

These instructions order execution and memory visibility between dependent grids, so schedule perturbation should be able to target them directly.

## Why a separate site kind

No example in the current corpus emits `griddepcontrol` today, so this change leaves the example-set site count unchanged (0 -> 0). The instructions are user-reachable; campaign value arrives with the first PDL example together with a host launch carrying the programmatic-stream attribute.

`griddepcontrol.*` represents grid-to-grid ordering and is distinct from the asynchronous proxy pipeline currently modeled by `SiteKind::AsyncProxy`.

Keeping it separate also avoids folding grid-level dependency semantics into thread/CTA barriers or memory fences.

## Tests

Added coverage that verifies:

* both backend-emitted `griddepcontrol` spellings are discovered;
* both are classified as `SiteKind::GridDependency`;
* grid dependency sites can be selected and perturbed with `nanosleep.u32`.

Validation:

```text
cargo fmt --all --check
cargo test -p ptx-schedule grid_dependency
cargo test -p ptx-schedule
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p ptx-schedule
```

All checks pass.


nihalpasham:
Worth landing. Nothing on main has touched ptx-schedule since this was filed, so it applies cleanly.

Checks out:
- `head` is the dotted opcode without the `;`, and no earlier prefix arm shadows `griddepcontrol.`, so both emitted spellings reach the new arm.
- The ISA has only two `griddepcontrol` forms, so the prefix covers the family.
- One kind per ordering mechanism is how the table is organised, and PDL (programmatic dependent launch: one grid ordering against another) is its own mechanism. Same call as #1195.
- Reports and the regress script look kinds up by name, so the variant is additive.
- I assembled the perturbed fixture with ptxas 13.3 at sm_90, sm_100 and sm_120; a `.target sm_80` copy is rejected as expected.

Before merge:
- Add one unaffected control to the fixture: an instruction from a neighbouring kind that must keep its kind, same discipline as the AsyncProxy fixture. `fence.acq_rel.gpu;` stays `Fence`, GridDependency count stays 2.
- Note in the PR body that no example emits `griddepcontrol` today (0 -> 0 sites in the example set). Same standing as `cp.reduce.async` in #1195: user-reachable, classify now, campaign value arrives with the first PDL example plus a host launch carrying the programmatic-stream attribute.

```text
primary:    launch_dependents; nanosleep; <tail work>          <- the sleep widens this overlap
secondary:                     <prologue> griddepcontrol.wait; <dependent work>
```

#1215 edits the same four spots (enum, arm, README row, tests). I'd take this one first; #1215 then needs a small rebase.

Next step: push the control line and I approve CI on the new head.

PR #1215, opened 2026-08-31
ptx-schedule: classify tensor-map mutations as schedule sites

### LINKED ISSUE #1208: ptx-schedule: tensormap.replace proxy ordering is unclassified
Found while verifying #1195. The generated TMA intrinsics emit tensormap mutation:

```
tensormap.replace.tile.global_address...;  // mutate the map (generic proxy)
fence.proxy.tensormap::generic...;         // publish it        <- classified (Fence)
cp.async.bulk.tensor...;                   // consume it        <- classified (AsyncProxy, #1195)
```

The middle and last lines are sites; the first is not, so a campaign can never delay the mutation against its publish fence, which is the one interleaving that would catch a missing/misplaced tensormap fence.

- does NOT belong in AsyncProxy: replace is a generic-proxy store into the map object, not part of the async-proxy commit/wait pipeline
- probably its own kind (or OrderedMemory), since its ordering contract is the tensormap proxy fence

Emit sites: convert_tensormap_replace in the generated TMA lowering (crates/mir-lower/src/convert/generated_intrinsics/tma.rs).

uurl:
Closes #1208.

## Summary

Classify `tensormap.replace.*` instructions as schedule-sensitive sites in `ptx-schedule`.

Tensor-map replacement operations mutate the tensor-map descriptor through the generic proxy. They must be observable by schedule perturbation independently from the `fence.proxy.tensormap::*` operation that publishes those mutations.

This adds a dedicated `SiteKind::TensorMapMutation` rather than classifying these instructions as `AsyncProxy` or `OrderedMemory`.

## Changes

* Add `SiteKind::TensorMapMutation`.
* Classify `tensormap.replace.*` instructions as `TensorMapMutation`.
* Keep `fence.proxy.tensormap::*` classified as `Fence`.
* Add coverage for emitted tensor-map replacement spellings.
* Add a perturbation test verifying that a tensor-map mutation can be delayed before its publish fence.
* Document the new site kind in the `ptx-schedule` README.

## Validation

Validated with:

```text
cargo fmt --all -- --check
cargo test -p ptx-schedule tensormap -- --nocapture
cargo test -p ptx-schedule
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p ptx-schedule
git diff --check
```

Results:

* 3 tensor-map-specific tests passed.
* 22 `ptx-schedule` unit tests passed.
* 3 CLI integration tests passed.
* Clippy passed with warnings denied.
* Rustdoc passed with warnings denied.

The original and perturbed PTX were also assembled successfully with `ptxas`.

`--list-sites` reports:

```text
tensormap.replace.*       -> TensorMapMutation
fence.proxy.tensormap::*  -> Fence
```

### DIFF
--- a/crates/ptx-schedule/README.md
+++ b/crates/ptx-schedule/README.md
@@ -31,6 +31,7 @@ places where another thread's progress can be observed:
 | `OrderedMemory` | a load or store carrying an explicit ordering qualifier (`.volatile`, `.acquire`, `.relaxed`, ...) |
 | `WarpCollective` | `shfl`, `vote`, `match`, `redux` and friends |
 | `AsyncProxy` | the asynchronous proxy pipeline: `cp.async.*`, `cp.reduce.async.*`, `wgmma.*`, `tcgen05.*` and `clusterlaunchcontrol.*` -- the bulk-copy, bulk-reduce and matrix issues and the commit/wait pairs that order them |
+| `GridDependency` | programmatic dependent-launch ordering: `griddepcontrol.launch_dependents` / `griddepcontrol.wait` |
 | `Backedge` | a `bra` back to the same or an earlier block -- a conservative loop detector |
 
 Each site keeps its ordinal, enclosing callable, byte span, the instruction
--- a/crates/ptx-schedule/README.md
+++ b/crates/ptx-schedule/README.md
@@ -31,6 +31,7 @@ places where another thread's progress can be observed:
 | `OrderedMemory` | a load or store carrying an explicit ordering qualifier (`.volatile`, `.acquire`, `.relaxed`, ...) |
 | `WarpCollective` | `shfl`, `vote`, `match`, `redux` and friends |
 | `AsyncProxy` | the asynchronous proxy pipeline: `cp.async.*`, `cp.reduce.async.*`, `wgmma.*`, `tcgen05.*` and `clusterlaunchcontrol.*` -- the bulk-copy, bulk-reduce and matrix issues and the commit/wait pairs that order them |
+| `TensorMapMutation` | `tensormap.replace.*` generic-proxy tensor-map descriptor mutation, published by `fence.proxy.tensormap::*` |
 | `Backedge` | a `bra` back to the same or an earlier block -- a conservative loop detector |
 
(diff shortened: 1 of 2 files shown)


nihalpasham:
This is the fix #1208 asked for, and a dedicated kind fits. In this crate `OrderedMemory` means an `ld`/`st` with an ordering qualifier (the arm and the README both say so). `tensormap.replace` is neither; the ISA treats it as a weak write to the whole 1024-bit descriptor. The kind only affects reporting (`--list-sites`, the `kinds=` summary); the injector ignores kind except for `Backedge`.

A site is a PTX instruction where another thread's progress can be observed; an arm is one prefix test in `classify_instruction`.

```text
instruction                                 before        after
tensormap.replace.tile.global_dim...        not a site    TensorMapMutation
fence.proxy.tensormap::generic.release...   Fence         Fence       (unchanged)
cp.async.bulk.tensor...                     AsyncProxy    AsyncProxy  (unchanged)
```

Checked on 7fd3800:
- Arm order: no earlier arm owns a prefix of `tensormap.replace.`, and `fence.proxy.tensormap::...` still hits the `fence.` arm first.
- The seed 7 / `global_dim` test is deterministic: I replayed the SplitMix64 draws. A focused hit always takes `after_ns` from `draw_long`, at least 2000 once selected, and seed 7 selects it.
- Nothing matches `SiteKind` exhaustively; the gemm_sol regress script reads only `AsyncProxy`.
- No example emits `tensormap.replace` (or a tensormap fence) today, so the fixture spelling comes from the mir-lower template. Fine for a classifier arm, as the #1195 fixture did for `cp.reduce.async`, but say so in the fixture comment.

Two changes:
1. #1214 touches the same six spots (enum, arm, README row, three test hunks) and conflicts with this branch. Rebase this one onto #1214's branch now, keeping both sides in each hunk, so #1214 lands first and this merges clean.
2. Add one field3 line so all three operand forms mir-lower emits are in the fixture, e.g. `tensormap.replace.tile.swizzle_mode.global.b1024.b32 [%rd0], 3;`.

Next step: push the rebase plus the fixture line and comment; I approve the CI run on that head and merge right after #1214.

PR #1222, opened 2026-09-02
llvm-export: encode enum debug tags as unsigned storage

### LINKED ISSUE #1220: full debug: llc -O0 emits `.b8 4294967295` that ptxas rejects (repr_u32_enum_stride, set_discriminant)
With `CUDA_OXIDE_DEBUG=full` two examples produce PTX that ptxas rejects:

```text
ptxas repr_u32_enum_stride.ptx, line 4956; error : Value out of range for type .b8 4294967295
ptxas set_discriminant.ptx,     line 2259; error : Value out of range for type .b8 4294967291
ptxas fatal : Ptx assembly aborted due to errors
```

- Both are `.b8` initializers carrying a 32-bit value: `4294967295` is `-1` as u32, `4294967291` is `-5`. Something emits an i8 constant through a 32-bit path when `llc` runs at `-O0` (release builds of the same examples assemble fine).
- Independent of the MIR level (fails at 0 and at 2), so it is on the LLVM/llc side or in how the exporter spells the constant.
- Both examples exercise enum discriminants; likely one root cause.

Repro:

```bash
CUDA_OXIDE_DEBUG=full cargo oxide build repr_u32_enum_stride --arch=sm_80
ptxas -arch=sm_80 crates/rustc-codegen-cuda/examples/repr_u32_enum_stride/repr_u32_enum_stride.ptx -o /dev/null
```

Found by the 2026-09-02 forced full-debug sweep; CI never builds these two with full debug.

ZenAlexa:
## Summary

- encode signed direct-enum discriminator metadata as an unsigned carrier of the same storage width
- preserve each `DebugEnumVariant.discriminant` raw bit pattern and ordinary signed-local metadata
- exercise high-bit i8 tags through full-debug export and real LLVM 21, 22, and 23 `llc -O0` NVPTX lowering

## Motivation

`DebugEnumVariant.discriminant` stores the physical tag bits in a `u64`. LLVM's NVPTX DWARF path derives extension behavior from the artificial discriminator carrier. A signed i8 carrier sign-extends tags `255` and `251`, producing `.b8 4294967295` and `.b8 4294967291`; `ptxas` rejects both records. The same-width unsigned carrier selects zero extension and emits `.b8 255` and `.b8 251`.

The conversion is scoped to the artificial enum discriminator member. Ordinary signed locals retain `DW_ATE_signed`.

## Validation

- `cargo test -p llvm-export`
- focused full-debug enum test with LLVM 21, 22, and 23 `llc -O0`
- signed-carrier control: `.b8 0xffffffffffffffff`
- unsigned-carrier treatment: `.b8 255`
- H100 80GB HBM3 integration: both examples built with Rust 1.100 nightly and LLVM 23.1, emitted `.b8 255` and `.b8 251`, and assembled with CUDA 13.1.115 `ptxas` at exit code 0
- `cargo clippy -p llvm-export --all-targets -- -D warnings`
- repository formatting and `git diff --check`

Fixes #1220

### DIFF
--- a/crates/llvm-export/src/export/debug.rs
+++ b/crates/llvm-export/src/export/debug.rs
@@ -825,9 +825,12 @@ impl<'a> ModuleExportState<'a> {
         self.debug_types.insert(ty.clone(), enum_type_id);
 
         let discriminator_member_id = discriminant.as_ref().map(|discriminant| {
-            // Variant `extraData` carries the physical tag bit pattern as a
-            // `u64`. An unsigned artificial carrier keeps llc from extending
-            // high-bit integer tags beyond the width of PTX `.b8` records.
+            // LLVM MC's NVPTX text writer prints negative one-byte DWARF
+            // constants as 64-bit hex literals, which ptxas rejects.
+            // `DW_FORM_data1` is signless; the discriminator type determines
+            // its interpretation, so an unsigned carrier preserves the tag
+            // bits and comparison semantics. Remove this conversion once all
+            // supported llc versions truncate the literal to the record width.
             let discriminator_ty = match discriminant.ty.as_ref() {
                 DebugLocalTypeKind::Basic {
                     size_bits,
--- a/crates/llvm-export/src/export/debug.rs
+++ b/crates/llvm-export/src/export/debug.rs
@@ -825,7 +825,22 @@ impl<'a> ModuleExportState<'a> {
         self.debug_types.insert(ty.clone(), enum_type_id);
 
         let discriminator_member_id = discriminant.as_ref().map(|discriminant| {
-            let base = self.ensure_debug_type(&discriminant.ty);
+            // Variant `extraData` carries the physical tag bit pattern as a
+            // `u64`. An unsigned artificial carrier keeps llc from extending
+            // high-bit integer tags beyond the width of PTX `.b8` records.
+            let discriminator_ty = match discriminant.ty.as_ref() {
+                DebugLocalTypeKind::Basic {
+                    size_bits,
+                    encoding,
+                    ..
+                } if *encoding == "DW_ATE_signed" => DebugLocalTypeKind::Basic {
+                    name: format!("u{size_bits}"),
+                    size_bits: *size_bits,
+                    encoding: "DW_ATE_unsigned",
+                },
+                ty => ty.clone(),
+            };
+            let base = self.ensure_debug_type(&discriminator_ty);
             let id = self.alloc_metadata_id();
             self.debug_nodes.push((
                 id,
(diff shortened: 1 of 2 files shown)


ZenAlexa:
H100 integration is complete. Both examples built with Rust 1.100 nightly and LLVM 23.1, emitted `.b8 255` and `.b8 251`, and assembled with CUDA 13.1.115 `ptxas` at exit code 0.

nihalpasham:
Worth landing. I reproduced the chain with a hand-written module through llc-21, llc-22 and the toolchain's llc 23.1:

```text
discriminator type   DW_AT_discr_value in PTX    ptxas
i8  (signed)         .b8 0xffffffffffffffff      Value out of range for type .b8 4294967295
u8  (unsigned)       .b8 255                     ok
```

The `4294967295` in #1220 is ptxas re-printing that hex literal, so both reports are one record.

- The real bug is upstream. LLVM's text-assembly writer (`MCAsmStreamer::emitIntValue`) creates the 1-byte DWARF constant with no size, and NVPTX has no signed data directive, so it prints a 64-bit hex literal. DwarfUnit's form choice is fine. I will file the LLVM issue.
- Unsigned is a legitimate encoding, not a hack. The discriminator (the hidden tag byte that says which variant is live) is compared as `DW_FORM_data1`, which carries no sign; the discriminator's type decides, so 255 == 255. rustc already does this for niche tags and keeps the sign on direct tags only because ELF object emission never touches the text printer. It holds for 16/32/64-bit tags too: unsigned picks a wider form, and NVPTX masks the split `.b8` pairs.
- Coverage checks out: direct tags are the only signed discriminators the importer produces, `discriminant.ty` has no other consumer in the exporter, and nothing else there emits a signed DWARF constant.
- On this head, locally: `cargo test -p llvm-export` passes with llc 18, 21, 22 and 23 all discovered; both #1220 examples build with `CUDA_OXIDE_DEBUG=full`, assemble with ptxas 13.3 (`.b8 255` / `.b8 251`), and run on an RTX 5090. Nothing landed on main since you branched.

One change before I merge: rewrite the comment at `debug.rs:828` to name what it routes around (LLVM MC prints a negative 1-byte DWARF constant as a 64-bit hex literal on NVPTX, ptxas rejects it), why unsigned is equivalent (`DW_FORM_data1` has no sign, the discriminator type decides), and when to drop it (once every llc I support truncates the literal). Without that, someone will "fix" it back to match rustc.

Optional: have `discover_llvm_tools` skip majors below 21 so the gate only runs on toolchains cuda-oxide supports. Can you push the comment change?

ZenAlexa:
Thanks for tracing the LLVM MC path. The comment now records the NVPTX text-emission failure, signless `DW_FORM_data1` semantics, and the supported-`llc` removal condition. Tool discovery stays unchanged to preserve the broader regression probe; LLVM 21 remains the supported floor. `cargo test -p llvm-export`, crate clippy, repository formatting, and `git diff --check` pass.

PR #1225, opened 2026-09-03
codegen: support grid-constant kernel parameters

### LINKED ISSUE #1223: Support grid-constant by-value kernel parameters for TMA descriptors

CUDA recommends passing a tensor map to a kernel as a read-only, by-value
`__grid_constant__` parameter. cuda-oxide currently exposes TMA operations over
`*const TmaDescriptor`, so callers that need a runtime descriptor must allocate
device memory for its 128 bytes, copy the encoded descriptor to that allocation,
and pass an ordinary device pointer. This adds storage and transfer work and
does not express the intended immutable kernel-parameter ABI.

**Proposed API / behavior**

Provide one Rust kernel-parameter declaration whose device ABI and generated
host launcher both derive from it. The exact spelling is open to design; for
example:

```rust
#[kernel]
pub unsafe fn copy(
    input_tma: GridConstant<TmaDescriptor>,
    // ...
) {
    cp_async_bulk_tensor_2d_g2s(
        dst,
        input_tma.as_ptr(),
        coord0,
        coord1,
        barrier,
    );
}
```

The implementation would:

- lower the aggregate kernel parameter to a pointer-typed LLVM parameter with
  `byval(TmaDescriptor)` and the required alignment;
- add that parameter index to the kernel's `grid_constant` entry in
  `!nvvm.annotations`;
- keep the device view read-only and addressable so it can be passed directly
  to TMA intrinsics;
- make `#[cuda_module]` marshal the descriptor bytes as one by-value kernel
  argument, rather than requiring a device allocation;
- derive both sides from the one parameter declaration rather than adding an
  independent launcher option.

Ordinary pointers to tensor maps in global memory should remain available for
the genuinely mutable/dynamically selected descriptor case.

**Alternatives considered**

- Allocate a 128-byte device slot, copy the descriptor H2D, and pass its
  pointer. This works today but adds storage/transfer work and a separate
  lifetime to retain.
- A constant-memory global is not per invocation and conflicts with concurrent
  kernels carrying different descriptors.
- Inline PTX cannot change an already-established kernel parameter ABI.

**Additional context**

- CUDA Programming Guide, TMA tensor maps:
  https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/async-copies.html
- CUDA `__grid_constant__` parameter semantics:
  https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-extensions.html#grid-constant
- NVVM IR `grid_constant` property: the annotated parameter must be a pointer
  with `byval`, and the property value lists 1-based parameter indices:
  https://docs.nvidia.com/cuda/nvvm-ir-spec/#supported-properties

cuda-oxide already has aggregate-by-value launch machinery and emits other
kernel properties through `nvvm.annotations`; this request is for the missing
end-to-end parameter contract, including host marshalling.

lucifer1004:
## What this adds

TMA descriptors now travel with the kernel launch, removing their separate GPU allocation/upload in `tma_copy`.

```text
before: launch carries a pointer to a descriptor in GPU memory
after:  launch carries the descriptor's 128 bytes
```

```rust
#[kernel]
pub fn copy(#[grid_constant] descriptor: &TmaDescriptor) {
    // Use the descriptor or pass it to a device helper.
}
```

The host passes a `Copy` descriptor by value; device threads borrow one read-only copy shared by the grid for that launch.

## What we fixed

- Legacy NVVM now preserves the 128-byte descriptor and its alignment, instead of emitting a one-byte parameter.
- Argument positions stay correct when slices expand or zero-sized arguments disappear. Generic helpers retain their ordinary pointer ABI.
- Invalid layouts, mutable parameter storage and invalid outer lifetimes fail compilation.
- The TMA example exposed an existing legacy address-conversion bug; the fix preserves cluster-shared addresses correctly.

## Launch contract

- Grid-constant **host launches require `unsafe`**, including prepared/async calls, numeric payloads and kernels declared safe.
- Copying a descriptor does not own its GPU allocation. The caller keeps referenced memory GPU-accessible, alive and correctly synchronized through completion.
- The blanket rule is a conservative API choice, not a CUDA requirement for numeric data.

<details>
<summary>Implementation</summary>

- Macro/MIR carry the entry parameter's type, index and Rust alignment. Checks inspect specialized types and original reference lifetimes.
- Lowering shares argument mapping with existing reference/alignment handling and rejects unsupported storage, including target-dependent shared-pointer fields.
- LLVM emits `byval` and NVVM `grid_constant` metadata. Legacy declarations, symbol references and entry adapters retain the complete pointee type.
- All generated grid host launchers share the unsafe policy, including both async forms.
- The legacy TMA fallback uses `cvta.to.shared::cluster`, avoiding unsupported LLVM address-space handling while preserving cluster semantics.

</details>

## Verification

Checked at `f27fd082`:

- ✅ `just check`: 5,393 tests/doctests, formatting, strict Clippy, guards and docs. Book passes with warnings as errors.
- ✅ All 53 hosted checks passed; PR book deployment was intentionally skipped.
- ✅ `tma_copy` memcheck/synccheck on RTX 5090 through LLVM NVPTX, modern libNVVM and legacy libNVVM: 4,096 copied values and all 256 pipeline threads verified.
- ✅ GPU/PTX checks for mixed arguments, generic helpers, descriptor layout and parameter positions.
- ✅ Compile-only probes reject all 15 unguarded grid launch variants and accept their explicit-unsafe counterparts.

## Caveats

- Native compiler only; the experimental CUTLASS exporter needs separate integration. No speedup has been measured.
- Use `&T` or `&'_ T` for the parameter; named or `'static` outer lifetimes are rejected.
- Legacy GPU runs used SM90 output and CUDA 13.0 for compatibility with the installed driver.
- Ordinary `KernelScalar: Copy` launches retain an existing nested-reference safety gap; this PR does not redesign that API.
- Multicast: modern libNVVM memcheck passes, but synccheck reports 32 existing divergent-barrier errors with both old and repaired lowering. Native LLVM passes; legacy remains blocked by unchanged cluster-register intrinsics.

Closes #1223.


nihalpasham:
The generic-kernel path needs to be part of the design before this lands:

- For a generic `#[kernel]`, the macro splits your fn into a generated `#[inline(always)]` helper that keeps the body and a concrete entry wrapper per instantiation that calls it. The marker lands in both. When rustc's `MIR` inliner folds the generated helper into the entry, `detect_grid_constant_params` hard-errors on the duplicate.
- Repro: 
```rust
#[kernel] 
pub fn k<T: Copy>(#[grid_constant] d: &Desc, out: *mut u32, _tag: T)  { 
   unsafe { *out = d.values[3] } 
 } 

``` 
fails with kernel contains duplicate grid-constant marker for source `parameter 0`. Bigger bodies only compile because they exceed the inliner's cost threshold.
- More importanly, the helper's marker leaks: a plain kernel 

```rust 
other(d: &Desc, out: *mut u32) { 
  k::<u8>(d, out, 0) 
} 
```
with no attribute comes out with `.param .align 64 .b8 other_param_0[128]` and a `grid_constant` annotation. Host passes an 8-byte pointer, device expects 128 bytes by value.
- `unchecked_indexing` already handles this by stripping its marker from the re-emitted helper, see `strip_unchecked_indexing_config_marker`. `grid_constant` needs the same in both generic paths. A duplicate-tolerant detector alone fixes the error but not the leak.

And we need tests in the PR exercise a generic kernel with `#[grid_constant]`.

**To land this:** the strip in both generic paths, a generic `#[kernel]` + `#[grid_constant]` test with a body small enough to be inlined, and a negative test that a non-opted kernel calling the helper still gets a plain pointer param.

nihalpasham (changes requested):
Changes requested, details in https://github.com/NVlabs/cuda-oxide/pull/1225#issuecomment-5557568274

lucifer1004:
Addressed the generic-helper ABI concern in f2b6b150.

The macro now captures the grid-constant marker for the generated entry and strips it from the re-emitted callable helper in both generic expansion paths. The contract example now covers both sides:

- an instantiated generic grid-constant entry retains the 128-byte by-value PTX parameter and launches successfully on sm_120;
- an ordinary kernel calling the same generic helper retains a `.param .u64` pointer ABI and does not acquire the 128-byte parameter.

Measured validation:
- `pixi run cargo test -p cuda-macros`
- strict clippy for cuda-macros and cuda_module_contract
- `pixi run cargo oxide run cuda_module_contract --arch sm_120 -- --verify-ptx`
- `pixi run cargo oxide run cuda_module_contract --arch sm_120`
- the full `just check` code/test portion passed; its cargo-deny guard initially hit the local NFS advisory-lock limitation, then all deny/license guards passed with the advisory DB placed on local `/tmp` storage.

nihalpasham (dismissed):
Reviewed `f2b6b150`: **changes requested**.

- The generic marker repair is present. The legacy NVVM path still loses the descriptor's pointee type:

```text
host payload: 128 bytes -> i8* byval -> PTX parameter: 1 byte
```

- Preserve the aggregate type through the legacy signature and body references, and add a real legacy NVVM control. The modern path passing does not cover this.
- Define the read-only/lifetime contract across entry routes. Give the generic GPU test one writer; it currently has 32 threads writing one element.
- Sign the revised stack and restore its missing DCO trailers.

Independent review and libNVVM output confirm the ABI blocker. #1223 remains open.

nihalpasham:
I added the needed follow-ups to the descriptor-by-value support:

- Fixed legacy NVVM loosing the descriptor's size: a 128-byte parameter was becoming one byte.
- Kept argument positions and alignment consistent when slices expand or zero-sized arguments disappear; generic helpers retain their normal pointer ABI.
- Added layout, read-only and lifetime checks so unsupported payloads fail compilation.
- Made grid-constant **host launches** `unsafe`, including prepared/async calls. The caller must keep referenced memory GPU-accessible, alive and properly synchronized. This conservative rule also covers numeric payloads and kernels declared safe.
- Switched `tma_copy` to pass descriptor bytes with the launch, removing the separate descriptor allocation/upload. Also fixed the existing legacy TMA address-conversion bug this exposed.
- Added regressions for mixed arguments, generic helpers, invalid payloads and the unsafe launch requirement.

At `f27fd082`, full local checks and all 53 CI checks passed. `tma_copy` also passed memcheck/synccheck through all three tested compiler paths.

This addresses #1223 for the native backend. CUTLASS integration is separate; no speedup has been measured yet.

PR #1229, opened 2026-09-04
fix(mir-importer): preserve call destination evaluation order

### LINKED ISSUE #1164: Support chained projected destinations for call and intrinsic results

uurl:
Closes #1164

## Summary

MIR call destinations must be evaluated before the callee executes. This matters for projected destinations such as:

```rust
RET.1[index]
(*ptr).field
array[index].field
```

where evaluating the destination after the call can observe state mutated by the callee and store the result at a different location.

This PR introduces a prepared destination-write path that materializes the destination before the producer executes, retains the typed storage information across the call or intrinsic, and performs the final store afterward.

## What changed

* Added a `PreparedDestinationWrite` abstraction for call results:

  * bare locals continue through `ValueMap::store_local`
  * ordinary projected destinations retain a precomputed address
  * canonical enum payload destinations retain their typed payload-store metadata and enclosing enum address

* Ordinary function-item, closure, and other call paths now follow:

```text
evaluate destination
    -> call callee
    -> write result through prepared destination
```

instead of evaluating the projected destination after the call.

* Generated intrinsic importers now prepare their destination before emitting the result-producing NVVM operation and use the same prepared-result epilogue.

* Regenerated the affected intrinsic importer sources from `cuda-intrinsics-gen`.

* Migrated hand-written intrinsic result paths to the same destination-write abstraction, including unit-returning paths such as `core::intrinsics::copy`, `typed_swap_nonoverlapping`, and compiler-hint no-op intrinsics.

* Removed the old result-store helpers and their independent single-projection grammar.

* Preserved the existing canonical payload storage path, including storage coercions such as boolean payload representation and pointer storage representation.

* Updated the intrinsic contributor documentation to reflect the new prepared-destination flow.

## Correctness regression

The original ordering bug is observable when the callee mutates a local used by the destination projection.

For example, conceptually:

```rust
_1 = 0;
Call(RET[_1] = bump(&mut _1), ...)
```

where `bump` changes `_1` to `2`.

The destination must resolve to `RET[0]` before entering the callee. Resolving it afterward incorrectly stores into `RET[2]`.

The regression fixture compares host and device behavior and now covers:

* mutation of an index local by the callee
* `Field -> Index`
* `Deref -> Field`
* `Index -> Field`
* bare-local destination control
* projected generated-intrinsic result destination
* existing deref, field, index, float, function-item, and `sincos` cases

It also includes a negative importer test ensuring that if the general place-address walker cannot represent a destination, translation fails with an unsupported-construct error instead of emitting a semantically incorrect store.

## Validation

```text
cargo fmt --all -- --check
PASS

git diff --check
PASS

generated intrinsic outputs are current

1145 passed; 0 failed

PASS

scripts/smoketest.sh --compile-only -o '^mir_projected_call_destination$'
PASS
ptxas sm_80 ok
```

Runtime host/device comparison:

```text
host folded:   0x935299a71fef4287
device folded: 0x935299a71fef4287

PASS: host/device agree on twelve call-destination cases and the generated SREG projection
```


nihalpasham:
Reviewed and validated:

- the destination address is now fixed before the callee runs, matching MIR; `RET[i] = f(&mut i)` lands in `RET[0]`, not `RET[2]`
- applies uniformly to plain calls, closure calls, hand-written and generated intrinsics
- closes #1164

**Layered three commits:**

- merge of main, the branch predated #1102
- rustfmt on the example's `through_index_field` signature
- the clc render test now asserts `prepare_destination_write` then `emit_prepared_result_and_goto`

**Verified here:** 12/12 host/device cases on the 5090, full smoketest 226/226, fork CI green.

Merging.

nihalpasham (approved):

PR #1232, opened 2026-09-04
refactor(llvm): use upstream GEP no-wrap flags

### LINKED ISSUE #1231: Replace private GEP inbounds metadata with upstream Pliron no-wrap flags
CUDA_OXIDE currently preserves the distinction between ordinary and wrapping pointer offsets using private LLVM-dialect metadata.

This was introduced in #612 to fix wrapping pointer-offset lowering: ordinary pointer offsets may carry the LLVM `inbounds` contract, while wrapping raw-pointer offsets must not be emitted as `getelementptr inbounds`.

At the moment, that distinction is represented in two layers:

* MIR keeps `mir_ptr_offset_inbounds`, which models the Rust-level semantic distinction.
* The LLVM dialect carries a private `cuda_oxide_gep_inbounds` marker consumed by `llvm-export`.

Pliron now has first-class support for LLVM GEP no-wrap flags through pliron-org/pliron#244:

* `inbounds`
* `nusw`
* `nuw`
* canonical `inbounds => nusw` semantics
* `GetElementPtrOp::new_with_no_wrap_flags`
* preservation across LLVM IR conversion

CUDA_OXIDE should adopt that upstream representation and remove the private LLVM-level shim.

### Scope

* Keep the MIR-level `mir_ptr_offset_inbounds` marker because it represents Rust source semantics.
* Lower ordinary pointer offsets to GEPs carrying `GepNoWrapFlags::INBOUNDS`.
* Lower wrapping pointer offsets to GEPs with no no-wrap flags.
* Replace existing production GEP sites that previously relied on CUDA_OXIDE's implicit `inbounds` behavior with explicit upstream flags.
* Keep dynamic shared-memory byte-offset GEPs unflagged: the `[0 x i8]` declaration does not establish LLVM's allocated-object `inbounds` contract for non-zero runtime offsets.
* Teach `llvm-export` to emit Pliron's `INBOUNDS`, `NUSW`, and `NUW` flags.
* Remove `cuda_oxide_gep_inbounds` and its associated helpers.
* Add regression coverage for the supported flag combinations and pointer-offset semantics.

### Acceptance criteria

* No remaining uses of the private `cuda_oxide_gep_inbounds` representation.
* Ordinary pointer offsets continue to export as `getelementptr inbounds`.
* Wrapping pointer offsets export as plain `getelementptr`.
* Dynamic shared-memory addressing remains plain GEP.
* `plain`, `nusw`, `nuw`, `inbounds`, and `inbounds nuw` export correctly.
* Workspace tests and Clippy pass.
* Full compile-only smoketest suite passes.
* `ptr_offset_from` and `dynamic_smem` pass on real GPU execution.

uurl:
Closes #1231 

## What this is

Replace CUDA_OXIDE's private LLVM GEP `inbounds` metadata with the first-class GEP no-wrap representation added upstream in pliron-org/pliron#244.

The Rust-level distinction introduced for wrapping pointer offsets remains unchanged:

```text
ordinary pointer offset  -> GEP INBOUNDS
wrapping pointer offset  -> plain GEP
```

What changes is where that information is represented after MIR lowering.

Before:

```text
MIR mir_ptr_offset_inbounds
        |
        v
private cuda_oxide_gep_inbounds metadata
        |
        v
llvm-export conditionally prints `inbounds`
```

After:

```text
MIR mir_ptr_offset_inbounds
        |
        v
Pliron GepNoWrapFlags
        |
        v
llvm-export emits the corresponding LLVM GEP flags
```

This removes the CUDA_OXIDE-specific LLVM shim and uses the upstream dialect model directly.

## GEP semantics

`mir_ptr_offset_inbounds` stays in MIR because it represents a Rust semantic distinction rather than an LLVM implementation detail.

During MIR lowering:

* ordinary pointer offsets use `GepNoWrapFlags::INBOUNDS`;
* wrapping pointer offsets use `GepNoWrapFlags::empty()`.

Production GEP sites that previously depended on CUDA_OXIDE's implicit `inbounds` behavior now request `INBOUNDS` explicitly.

Dynamic shared-memory addressing deliberately remains unflagged.

CUDA dynamic shared memory is represented through an external `[0 x i8]` object in shared address space. Although the launch supplies backing storage at runtime, that declaration does not establish LLVM's allocated-object `inbounds` contract for arbitrary non-zero runtime byte offsets. Those GEPs therefore continue to use plain `GetElementPtrOp::new`.

## LLVM export

`llvm-export` now consumes Pliron's `GepNoWrapFlags` directly and emits:

```text
plain
nusw
nuw
inbounds
inbounds nuw
```

`inbounds` implies `nusw` in Pliron, matching LLVM semantics, but the implied `nusw` is not redundantly printed.

The private:

```text
cuda_oxide_gep_inbounds
```

attribute and its getter/setter helpers are removed.

## Why this Pliron revision

This PR moves the Pliron dependency from:

```text
823d9d8087652cd3f2d006a7ea534df921a23564
```

to:

```text
8e4500e16d96926a91751a505e3d8b4ef9c4f7c8
```

The revision range contains five upstream commits:

* pliron-org/pliron#241: preserve LLVM data layout and target triple;
* pliron-org/pliron#242: LLVM metadata support;
* pliron-org/pliron#243: upgrade the optional LLVM bridge to LLVM 23;
* pliron-org/pliron#244: add GEP no-wrap support;
* `8e4500e`: address the GEP no-wrap review comments for pliron-org/pliron#244.

Only the GEP no-wrap support requires CUDA_OXIDE source changes in this PR. The preceding commits are rides-along from the upstream revision range. CUDA_OXIDE continues to consume `pliron-llvm` with `default-features = false`, so its `llvm-sys` bridge remains disabled.

The merged pliron-org/pliron#244 revision is intentionally not used here. pliron-org/pliron#249 landed on Pliron `master` before pliron-org/pliron#244 was merged, so the eventual pliron-org/pliron#244 merge also carries the unrelated `ConstantOp` API changes from pliron-org/pliron#249. Adopting that revision would require a separate CUDA_OXIDE-wide compatibility migration.

Pinning `8e4500e` isolates the GEP migration from that unrelated `ConstantOp` refactor. The full CUDA_OXIDE workspace tests and the 222/222 compile-only example sweep pass with this revision.

## Tests

Added/updated regression coverage verifies:

* plain GEP export;
* `nusw`;
* `nuw`;
* `inbounds`;
* `inbounds nuw`;
* ordinary pointer offsets retain `INBOUNDS`;
* wrapping pointer offsets lower without `INBOUNDS`;
* unrelated address-of export tests do not depend on GEP no-wrap semantics.

## Validation

```text
cargo fmt --all -- --check
    PASS

cargo test --workspace
    PASS

    PASS

scripts/smoketest.sh --compile-only
    222 / 222 PASS

targeted compile-only smoke:
    dynamic_smem       PASS
    ptr_offset_from    PASS

targeted GPU execution:
    dynamic_smem       PASS
    ptr_offset_from    PASS
```

`dynamic_smem` was also exercised through its alignment and partitioning cases on real hardware, and `ptr_offset_from` verified pointer distance, wrapping, and provenance behavior.

The private GEP metadata is no longer present in the tree.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1026,7 +1026,7 @@ checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 [[package]]
 name = "pliron"
 version = "0.17.0"
-source = "git+https://github.com/pliron-org/pliron?rev=823d9d8087652cd3f2d006a7ea534df921a23564#823d9d8087652cd3f2d006a7ea534df921a23564"
+source = "git+https://github.com/pliron-org/pliron?rev=8e4500e16d96926a91751a505e3d8b4ef9c4f7c8#8e4500e16d96926a91751a505e3d8b4ef9c4f7c8"
 dependencies = [
  "awint",
  "combine",
@@ -1053,7 +1053,7 @@ dependencies = [
 [[package]]
 name = "pliron-derive"
 version = "0.17.0"
-source = "git+https://github.com/pliron-org/pliron?rev=823d9d8087652cd3f2d006a7ea534df921a23564#823d9d8087652cd3f2d006a7ea534df921a23564"
+source = "git+https://github.com/pliron-org/pliron?rev=8e4500e16d96926a91751a505e3d8b4ef9c4f7c8#8e4500e16d96926a91751a505e3d8b4ef9c4f7c8"
 dependencies = [
  "combine",
  "convert_case",
@@ -1067,7 +1067,7 @@ dependencies = [
 [[package]]
 name = "pliron-llvm"
 version = "0.17.0"
-source = "git+https://github.com/pliron-org/pliron?rev=823d9d8087652cd3f2d006a7ea534df921a23564#823d9d8087652cd3f2d006a7ea534df921a23564"
+source = "git+https://github.com/pliron-org/pliron?rev=8e4500e16d96926a91751a505e3d8b4ef9c4f7c8#8e4500e16d96926a91751a505e3d8b4ef9c4f7c8"
 dependencies = [
  "bitflags",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,14 +48,14 @@ repository = "https://github.com/NVlabs/cuda-oxide.git"
 # Examples are in: crates/rustc-codegen-cuda/examples/
 
 [workspace.dependencies]
-# Pliron MLIR framework, pinned to the upstream revision containing explicit LLVM struct layouts (#232).
-# pliron-llvm uses default-features = false to keep llvm-sys (LLVM 22) out:
+# Pliron MLIR framework, pinned to upstream struct-layout and GEP no-wrap semantics.
+# pliron-llvm uses default-features = false to keep llvm-sys (LLVM 23) out:
 # cuda-oxide consumes the dialect modeling only, not the FFI bridge.
 # `crates/rustc-codegen-cuda/Cargo.toml` MUST stay on this same rev, or pliron
 # resolves to two distinct crates and types stop unifying.
-pliron = { git = "https://github.com/pliron-org/pliron", rev = "823d9d8087652cd3f2d006a7ea534df921a23564" }
-pliron-derive = { git = "https://github.com/pliron-org/pliron", rev = "823d9d8087652cd3f2d006a7ea534df921a23564" }
-pliron-llvm = { git = "https://github.com/pliron-org/pliron", rev = "823d9d8087652cd3f2d006a7ea534df921a23564", default-features = false, features = [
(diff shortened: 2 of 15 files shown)


nihalpasham:
Reviewed and validated:

- the private `cuda_oxide_gep_inbounds` attribute is gone; every production GEP now carries pliron-llvm's no-wrap flags with the same effective flag it had before
- the extern_shared byte-offset GEP drops an `inbounds` it never earned, the old attribute read as inbounds when unset; that branch is importer-unreachable today since `offset(N)` goes through `mir.ptr_offset`, tracking that separately
- closes #1231

**Layered one commit:**

- two mir-lower tests locking that GEP to plain `getelementptr`, with the zero-offset control

**Verified here:** `dynamic_smem` + `ptr_offset_from` on the 5090, full smoketest 222/222, fork CI green.

Merging.

nihalpasham (approved):

PR #1236, opened 2026-09-05
feat(wmma): add shared-u32 ldmatrix forms

lucifer1004:
## Summary

Add `.shared::cta.u32` address forms for the classic `ldmatrix` APIs, so callers can convert a generic shared pointer once and keep byte-offset/swizzle arithmetic in the 32-bit PTX state-space address domain.

This is the first implementation slice of #1235. It does not add the counted register-MMA carrier described there.

## Changes

- add `cvta_generic_to_shared_u32(*const u8) -> u32` alongside the existing u64 descriptor-address conversion
- add x1/x2/x4 normal and transposed `ldmatrix_*_shared_u32(u32)` source forms
- declare those source aliases in the ldmatrix catalog, so importer dispatch, collector recognition, ABI markers, references, and generated tests remain one projection
- allow the existing `nvvm.ldmatrix` operation to carry either a compatible pointer or a `.shared::cta.u32` state-space address
- pass u32 directly to the libNVVM inline-PTX route; materialize an address-space-3 pointer only at the LLVM-NVPTX intrinsic boundary
- extend the existing generated-ldmatrix GPU oracle and lowering tests

The large generated-file count is the repository's normal catalog-digest propagation; the semantic additions remain in the ldmatrix overlay/generator and handwritten device stubs.

## Testing

- [ ] `just check` passes (all formatting, Clippy, unit tests, and structural guards passed; the final `cargo deny --locked check` could not run because this environment has no `cargo-deny` subcommand)
- [x] `cargo oxide run generated_ldmatrix --arch sm_120a`
- [x] `cargo oxide run generated_ldmatrix --emit-nvvm-ir --arch sm_120a`
- [ ] New example added (not applicable; extended the existing `generated_ldmatrix` oracle)

Both GPU runs passed all 128 x4 fragment checks and both 32-lane dynamic-index checks on sm_120 after the API was renamed to the explicit `_shared_u32` spelling.

Additional targeted release tests passed for `cuda-intrinsics-gen`, `cuda-device`, `dialect-nvvm`, `mir-importer`, and `mir-lower`. Targeted Clippy passed with `-D warnings`, including the example in the CUDA 13.3 environment.

## Checklist

- [x] All commits signed off (`git commit -s`)
- [x] SPDX headers on new source files (no new handwritten source file; generated files retain generated SPDX headers)

Issue #1235, opened 2026-09-05
Keep counted register-MMA shared addressing in the PTX u32 domain


lucifer1004:
**What problem does this solve?**

CTA-local shared-memory instruction addresses are a 32-bit hardware domain, but cuda-oxide currently exposes pointer-form `ldmatrix` APIs and `cvta_generic_to_shared_offset -> u64`. Pointer indexing therefore naturally uses Rust `usize`/u64 until each `ldmatrix` use truncates the shared pointer. In a register-heavy counted BF16 HMMA loop this leaves avoidable wide address recurrences and conversions, while mechanically rewriting the source loop can make libNVVM replicate the loop body early and spill the accumulator set.

The u64 value remains appropriate for WGMMA/tcgen descriptor construction. The missing contract is the `.shared::cta.u32` state-space address operand used by instructions such as `ldmatrix`, plus a compiler-owned way to keep a proven counted register-MMA loop in one lowering lifetime.

**Proposed API / behavior**

The first layer follows the PTX carrier directly:

```rust
let base = unsafe { cvta_generic_to_shared_u32(smem.cast()) };
let address = base.wrapping_add(swizzle_offset_u32);
let a = unsafe { ldmatrix_x4_shared_u32(address) };
let b = unsafe { ldmatrix_x2_shared_u32(other_address) };
```

Semantics:

- `cvta_generic_to_shared_u32` produces a raw `.shared::cta.u32` state-space address.
- Generic/shared pointer conversion happens once. Byte-offset and swizzle arithmetic can remain u32.
- Existing pointer-form `ldmatrix_*` APIs remain available.
- The existing WGMMA/tcgen u64 descriptor-address API does not change.
- Pointer and shared-u32 forms converge on the same `nvvm.ldmatrix` semantic operation rather than introducing duplicate dialect truth. The LLVM-NVPTX route may materialize an address-space-3 pointer at its intrinsic boundary; the libNVVM route can pass the u32 address directly to inline PTX.

A follow-up compiler layer should recognize only a canonical counted register-MMA loop and replace it with an internal value-form carrier that owns the accumulator recurrence, u32 shared-address recurrence, ldmatrix loads, loop control, and HMMA issue order in one convergent lowering lifetime. Unsupported or unproven loops remain unchanged. This is analogous in ownership to the existing counted WGMMA carriers, but for synchronous register MMA; it should not expose an application-specific MxN kernel API.

I propose implementing this as two reviewable PRs:

[...]

PR #1237, opened 2026-09-05
draft(mma): investigate counted register-MMA preservation

lucifer1004:
## Status

Stacked investigation draft. Depends on #1236 and tracks the possible second slice of #1235. The head currently contains only the dependency stack; no counted-HMMA implementation has been added yet.

The current evidence does **not** yet justify a carrier-formation pass. This draft remains open while the engine-side x4 schedule is exhausted.

## Corrected problem statement

The original Rust kernel encoded one N8 BF16 B fragment as one `ldmatrix.x2`. That was the wrong abstraction boundary, not a compiler combine failure. Every admitted general-BF16 topology has an even number of N8 fragments per warp, so adjacent fragments can be loaded as x4 pairs with no x2 tail.

Matched NCU disassembly for M8192/N1024 had shown:

- cuBLAS `nvjet_sm120_tss_mma_128x176x64_2_32x88x64_tmaAB_alignCD4_bz_TNNN`: 30 static `LDSM.16.M88.4`, no x2, and 88 HMMA;
- old Oxide BM128/BN128/S2: 8 static `LDSM.16.M88.4`, 32 `LDSM.16.M88.2`, and 64 HMMA.

After pairing adjacent B fragments in ordinary Rust, all 20 registered general-BF16 variants build with x4-only SASS. Representative static bodies are:

- BM128/BN64: 16 x4, 0 x2;
- BM128/BN128: 24 x4, 0 x2;
- BM64/BN64: 12 x4, 0 x2;
- BM64/BN128: 20 x4, 0 x2;
- BM32/BN64: 8 x4, 0 x2.

All 20 variants passed the repository patterned-BF16 correctness/clock case. At M8192/N1024, the best BN128 median improved from the previous approximately 207.232 us to 196.608 us. At M8192/N64, the best BN64 median was 32.544 us versus the previous approximately 32.384 us, which is not a meaningful change.

Changing the x4-only body from pointer addresses to the #1236 `.shared::cta.u32` API did not materially change timing. It preserved the no-spill resource shape for BM32/BM64 and BM128/BN64. BM128/BN128 remains on a register-allocation cliff: x4-only pointer lowered to REG168/STACK56; x4-only shared-u32 lowered to REG168/STACK48. Despite that spill, the x4 schedule is faster than the old x2-heavy control.

The earlier isolated two-x4/eight-x2-per-step spike proves only that a u32 address carrier can survive both compiler routes. It is not a production micro-schedule and is no longer an acceptance target.

## Remaining question

Can the BM128/BN128 x4-only Rust schedule be reordered or represented more directly so it returns to zero spill without a new compiler carrier?

Only if source-level scheduling cannot remove that final spill while preserving the measured x4 performance should this PR introduce a compiler facility. Such a facility would preserve a kernel-authored counted register lifetime through libNVVM; it would not select x4/x2 grouping, tile geometry, HMMA issue order, or tuning policy.

## Non-goals

- restoring or admitting the old x2-heavy schedule
- a public application-specific GEMM entry point
- automatic kernel fusion
- a general scheduling DSL or arbitrary PTX region
- tuning policy or hardware-specific candidate selection
- replacing the existing individual register-MMA and ldmatrix operations
- changing WGMMA counted-loop behavior

## Evidence required before any implementation direction

- exhaust the small set of source-level x4 issue-order/live-range alternatives on BM128/BN128;
- retain x4-only SASS and repository correctness;
- compare REG/STACK/local load-store counts and matched repository medians;
- demonstrate that an ordinary Rust representation cannot meet the no-spill gate before adding a pass or compound carrier.

Draft discussion will revise or close this PR based on that evidence.


nihalpasham:
Closing as a duplicate of #1236: both branches point at 4284c72, so nothing here is not already there, and the carrier pass it was opened for is not written. Slice-2 discussion goes on #1235.

PR #1241, opened 2026-09-05
docs(book): put the book's Rust blocks on rustfmt's import order

### LINKED ISSUE #1240: Book Rust blocks still use the pre-#1193 import order, so copying one fails `cargo oxide fmt --check`
Filing the follow-up invited in the #1193 review.

#1193 made `cargo oxide new` emit rustfmt-clean projects, because a scaffolded project failing its own first `cargo oxide fmt --check` is a bad first five minutes. The book's hand-written Rust blocks were out of scope there and still carry the old import order, so a reader who copies one into a project gets the same diff on the same command.

## Measured

Over every tracked `cuda-oxide-book/**/*.md`, taking each ```` ```rust ```` block that contains a `cuda_*` import and running the pinned `rustfmt` on it at the book's edition:

```
blocks containing a cuda_* import                 96
  blocks rustfmt would change                     38   (19 files)
  blocks rustfmt cannot parse (partial snippets)  49
```

The 38 break down as: crates not sorted (`cuda_device` / `cuda_host` ahead of `cuda_core`, and `cuda_async` where present), and `DisjointSlice` and friends sorted after the lowercase items inside the braces, which is not where style edition 2024 puts them.

**Every one of the 38 diffs is imports-only.** rustfmt does not touch a single non-import line in any of them, so this cannot reflow deliberately-formatted teaching code — I checked that specifically before filing, because that was the risk that would have made this not worth doing.

| file | blocks |
|---|---|
| `appendix/api-quick-reference.md` | 4 |
| `gpu-programming/launching-kernels.md` | 4 |
| `advanced/cluster-programming.md` | 3 |
| `gpu-programming/closures-and-generics.md` | 3 |
| `gpu-programming/memory-and-data-movement.md` | 3 |
| `advanced/shared-memory-and-synchronization.md` | 2 |
| `advanced/tensor-memory-accelerator.md` | 2 |
| `advanced/warp-level-programming.md` | 2 |
| `getting-started/hello-gpu.md` | 2 |
| `gpu-programming/error-handling-and-debugging.md` | 2 |
| `gpu-programming/virtual-memory-and-peer-access.md` | 2 |
| `gpu-safety/the-safety-model.md` | 2 |
| seven more files | 1 each |

## Why nothing catches it

A markdown code block is compiled by nothing and formatted by nothing. `scripts/check-book-api-names.sh` verifies that every device-API *name* the docs show exists, which is a different property and passes today.

## Scope

The 49 blocks rustfmt cannot parse are partial snippets with no enclosing item, and are left alone: reformatting them would mean inventing a wrapper, and the ordering question does not arise for most of them anyway.

vyncint:
Closes #1240. This is the follow-up invited in the #1193 review.

#1193 made `cargo oxide new` emit rustfmt-clean projects. The book's hand-written blocks kept the old import order, so a reader who copies one into a project gets the same diff on the same first `cargo oxide fmt --check`.

## Scope, measured

Over every tracked book page, taking each ```` ```rust ```` block that contains a `cuda_*` import and running the pinned rustfmt at the book's edition:

| | blocks |
|---|---|
| already rustfmt-clean | 58 |
| **differ in imports only — fixed here** | **15** |
| differ in other formatting — left alone | 23 |

After the change: **73 clean, 0 differing in imports**, and the same 23 differing in other formatting as before.

The 15 are the two orderings #1193 fixed in the templates: crates not sorted (`cuda_device`/`cuda_host` ahead of `cuda_core`, and `cuda_async` where present), and `DisjointSlice` and friends after the lowercase items inside the braces instead of before them under style edition 2024.

## What is deliberately not touched

The 23 differ for reasons unrelated to imports — one parameter per line in a function signature, `{ /* ... */ }` split across two lines, a wrapped call collapsed. Reformatting deliberately laid-out teaching code is a different change with a different argument, so those blocks are byte-identical here.

**Only rustfmt's import region is spliced back into each block, not its whole formatted output.** That distinction is load-bearing: taking the whole block reflows the teaching code above, and I only found that out by diffing rather than trusting my own classifier. Verified afterwards by grepping the resulting diff for anything that is not a `use` statement or one of its continuation lines — nothing.

## A correction I owe on my own issue

The scope figures in #1240 as originally filed were wrong: it said 38 blocks and claimed every diff was imports-only. My measurement filtered rustfmt's `--check` output with `startswith("-")`, but that output is colourised, so every changed line begins with an ANSI escape, nothing matched, and the "is every changed line an import?" test was vacuously true. 38 was really "blocks that differ from rustfmt for any reason". Re-measured with `--color=never`, corrected on the issue, and the table above is the corrected version.

## Why nothing catches this

A markdown code block is compiled by nothing and formatted by nothing. `scripts/check-book-api-names.sh` verifies that the device-API *names* the docs show exist, which is a different property and passes today, before and after.

## Test plan

- `sphinx-build -W --keep-going -b html cuda-oxide-book` — build succeeded.
- `scripts/check-book-api-names.sh`, `scripts/check-book-catalog-stamp.sh`, `scripts/check-spdx-headers.sh` — pass.
- Re-ran the block measurement on the result: 0 blocks differ from rustfmt in imports.
- Confirmed the diff touches only `cuda-oxide-book/` and only import lines.
- No device code, so there is no GPU evidence to attach.

### DIFF
--- a/.github/workflows/status-guard.yml
+++ b/.github/workflows/status-guard.yml
@@ -112,6 +112,28 @@ jobs:
       - name: Verify every device call the book shows resolves
         run: bash scripts/check-book-api-names.sh
 
+  # A book block is pasted into an edition-2024 `cargo oxide new` project, where
+  # the first `cargo fmt` reorders its `use` lines (`{DisjointSlice, kernel,
+  # thread}`, crates sorted). The code is right either way, so nothing fails and
+  # the book drifts from the toolchain and from itself: #1241 found 47 blocks
+  # on the old order across 18 pages, after #1193 had fixed the scaffold
+  # templates with a test that runs the real rustfmt. Only the top-level `use`
+  # region of each block is formatted, so teaching layout in signatures and
+  # bodies, and partial snippets with no enclosing item, never trip it.
+  book-import-order:
+    name: book import order
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      # rustfmt comes from rust-toolchain.toml's component list through the
+      # rustup proxy, as in fmt.yml, so the order checked is the pinned
+      # nightly's. No cargo, no CUDA toolkit, no network beyond the toolchain.
+      - name: Verify the book's Rust blocks use rustfmt's import order
+        run: bash scripts/check-book-import-order.sh
+
   # The dialect-nvvm book page prints the catalog SHA-256 and states its own
   # freshness rule: if the stamp no longer matches, every count on the page
   # predates the catalog being read. Nothing enforced that rule, so the page
--- a/Justfile
+++ b/Justfile
@@ -213,6 +213,7 @@ check-guards:
     bash scripts/check-toolchain-parity.sh
     bash scripts/check-cli-doc-coverage.sh
     bash scripts/check-book-api-names.sh
+    bash scripts/check-book-import-order.sh
     bash scripts/check-reserved-prefixes.sh
     bash scripts/check-device-only-build.sh
     bash scripts/check-test-matrix-coverage.sh
--- a/cuda-oxide-book/advanced/cluster-programming.md
+++ b/cuda-oxide-book/advanced/cluster-programming.md
@@ -46,7 +46,7 @@ supports up to 8 blocks per cluster.
 In cuda-oxide, you annotate the kernel with `#[cluster_launch]`:
 
 ```rust
-use cuda_device::{kernel, cluster_launch, thread, cluster, SharedArray};
+use cuda_device::{SharedArray, cluster, cluster_launch, kernel, thread};
 
 #[kernel]
 #[cluster_launch(4, 1, 1)]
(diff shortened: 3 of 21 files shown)


nihalpasham:
The splice mechanism is right and kept: rustfmt's import region back into the block, nothing else moved. After your commit, 30 blocks still carried the old order, including the landing block in index.md.

```text
                blocks  pages
your commit       17     14
stacked here      31     13   (one page shared)
book now          47     18   -> 0 blocks differ from rustfmt in imports
```

- 627150a: the 30 remaining blocks (16 whole-item, 14 partial snippets whose `use` lines parse on their own) plus the missing `DisjointSlice` import in the `reduce_dynamic` block; every changed line is a `use` or a brace-list continuation
- 714d0bd: `scripts/check-book-import-order.sh` and a status-guard job, the pinned `rustfmt --edition 2024 --check` over each block's import region. Fails on your head (lists the 30), passes now. Same idea as the #1193 test: the real rustfmt decides, nobody re-derives its rules.

**Validation**: book builds under `sphinx-build -W`, all status guards pass. Merged, and with the full sweep "Closes #1240" holds. One correction to the body for the record: 17 blocks in 14 files, and "58 clean" is 9 rustfmt-clean plus 49 blocks rustfmt cannot parse whole.

PR #1242, opened 2026-09-05
docs(tma_multicast): gate on the ISA floor, not on this build target

vyncint:
Addresses the Hopper half of #966, which stays open for the hardware run.

The example refused to run below compute capability 10 and said, in ten places, that TMA multicast "requires sm_100a". Multicast is **sm_90+ in the PTX ISA** — `sm_90a` is advised there for performance, not required for legality, which is what the sm_90+ catalog row already records. What actually keeps an H100 out is this example's own build target, not the instruction.

## Separating the two facts

| | before | after |
|---|---|---|
| host gate | `major < 10` | `major < 9` |
| skip text | "TMA multicast requires sm_100a" | names the declined build target, points at `--arch=sm_90a` |
| README Hopper section | "Not supported" | "Legal per the ISA, hardware run pending" |
| `tma_copy` cross-reference | "sm_100a only" | "sm_90+ per the ISA; that example ships as sm_100a" |

Moving the gate does not make the default build run on Hopper — it ships `.target sm_100a`, the driver declines it, and **the existing load-failure arm reports that as its own clean skip**, so the smoketest contract is unchanged on every GPU that has one. What the gate change does is stop turning Hopper away *before the loader has an opinion*, so an `--arch=sm_90a` build can JIT there.

## Verified on an A10G (sm_86, driver 595.71.05)

```
build --arch=sm_90a    ->  .target sm_90a, 1 multicast instruction
ptxas -arch=sm_90a     ->  ACCEPTED            (ptxas 13.2)
build --arch=sm_100a   ->  .target sm_100a
ptxas -arch=sm_100a    ->  ACCEPTED
run on sm_86           ->  "skipping: TMA multicast requires sm_90 or newer"
                           "   Your GPU is sm_86. Use: cargo oxide run tma_copy"
```

The last line is the new gate firing on real hardware, not a claim about it.

Negative control, so the ptxas results are not vacuous — the same sm_90a PTX offered as `-arch=sm_80`:

```
ptxas fatal : SM version specified by .target is higher than default SM version assumed
```

## What this does not do

**It does not run the example on Hopper.** There is no H100 here and the sm_90a build has never executed. #966 stays open for exactly that run, the README says so in those words, and the Hopper transcript is explicitly labelled as the shape of the message the host emits rather than a capture — I would rather label it than invent one. Its lines were checked against the `println!`s they mirror.

## Deliberately deferred: the `--arch=sm_90a` smoketest entry

Your third item is not in this PR, and I raised the reason on the issue before starting rather than guessing. `tma_multicast` is in none of the category arrays, so it is `standard` today: the smoketest builds **and runs** it, with the example's own gate making that a clean skip. There is no existing hook for "also compile this one at a second architecture" — `SM100_COMPILE_EXAMPLES` and `BLACKWELL_COMPILE_EXAMPLES` make an example compile-*only*, which here would trade away runtime coverage it currently has.

So it needs either a new per-example "extra compile arch" mechanism in `smoketest.sh`, or a decision to leave the sm_90a build outside CI. Inventing the mechanism unasked seemed the wrong call; say which you want and I will follow up.

## Test plan

- `cargo fmt --check` in the example and across the workspace — clean.
- `check-spdx-headers.sh`, `check-error-example-status.sh`, `check-example-license-policy.sh`, `check-reserved-prefixes.sh`, `check-book-api-names.sh` — pass.
- Both architectures build and assemble on hardware, with the negative control above.
- The new gate exercised on a real sm_86 GPU.
- README transcript lines checked line-for-line against the `println!` calls they show.

### DIFF
--- a/crates/rustc-codegen-cuda/examples/tma_copy/README.md
+++ b/crates/rustc-codegen-cuda/examples/tma_copy/README.md
@@ -5,7 +5,7 @@
 Demonstrates TMA for efficient tensor data movement. TMA offloads memory copies
 to a dedicated hardware unit, enabling overlap with computation.
 
-For TMA multicast (sm_100a only), see the [`tma_multicast`](../tma_multicast/) example.
+For TMA multicast (sm_90+ per the ISA; that example ships as sm_100a), see the [`tma_multicast`](../tma_multicast/) example.
 
 ## What This Example Does
 
--- a/crates/rustc-codegen-cuda/examples/tma_copy/README.md
+++ b/crates/rustc-codegen-cuda/examples/tma_copy/README.md
@@ -5,7 +5,7 @@
 Demonstrates TMA for efficient tensor data movement. TMA offloads memory copies
 to a dedicated hardware unit, enabling overlap with computation.
 
-For TMA multicast (sm_90+ per the ISA; that example ships as sm_100a), see the [`tma_multicast`](../tma_multicast/) example.
+For TMA multicast (also sm_90+), see the [`tma_multicast`](../tma_multicast/) example.
 
 ## What This Example Does
 
--- a/crates/rustc-codegen-cuda/examples/tma_copy/src/main.rs
+++ b/crates/rustc-codegen-cuda/examples/tma_copy/src/main.rs
@@ -12,7 +12,7 @@
 //! - mbarrier: Barrier-based completion tracking
 //!
 //! Note: This example requires Hopper (sm_90) or newer GPUs.
-//! For TMA multicast (sm_100a), see the `tma_multicast` example.
+//! For TMA multicast (also sm_90+), see the `tma_multicast` example.
 //!
 //! Build and run with:
 //!   cargo oxide run tma_copy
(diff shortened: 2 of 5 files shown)


nihalpasham:
Gate change is right and kept. The text around it said the example ships as `.target sm_100a`, and it does not: `cargo oxide run` forwards the detected GPU as a hint and rustc-codegen-cuda builds for it whenever the kernel's features run there.

```text
run on a 5090    hint sm_120a -> .target sm_120a -> loads, runs   (transcript re-captured)
run on an H100   hint sm_90a  -> .target sm_90a  -> expected to run, not yet captured (#966)
no usable GPU    -> sm_100a fallback
```

- banner and load-failure arm name no target; the arm prints the CUDA driver's error instead
- README Hopper section, hardware table, "Which GPUs Run This?" rewritten on that model; "rebuild with --arch=sm_90a" dropped
- swept the sites the diff missed: Cargo.toml comment, tma_copy main.rs and README, the architecture table row

**Validation**: `cargo oxide run tma_multicast` on the 5090, smoketest for tma_multicast and tma_copy, fmt, host-api-paths and lock guards. Merged; #966 stays open for the H100 run.

Issue #1098, opened 2026-08-20
Emit LLVM `nonnull` / `align` / `dereferenceable` for proven Rust references


uurl:
## Motivation

Part of #413.

After #1062 preserves Rust pointer/reference kind through dialect MIR and #1089
adds conservatively proven `noalias` / `readonly` parameter attributes, the
compiler has the information needed to emit additional LLVM validity attributes
for Rust references.

Rust references carry stronger validity guarantees than raw pointers. In
particular, for sized pointees the compiler can conservatively preserve:

- non-nullness
- pointee alignment
- a dereferenceable byte count when the reference is known to remain valid for
  the required extent

These facts should be derived while the original rustc `Ty`, layout, and trait
information are still available rather than inferred later from LLVM pointer
shape.

## Proposed semantics

Use a conservative subset of rustc's reference parameter policy:

| Rust parameter | Condition | Additional LLVM attributes |
| --- | --- | --- |
| `&T` | `T: Sized` | `nonnull`, `align N` when `N > 1` |
| `&mut T` | `T: Sized` | `nonnull`, `align N` when `N > 1` |
| `&T` | `T: Sized + Freeze`, `size_of::<T>() > 0` | `dereferenceable(size_of::<T>())` |
| `&mut T` | `T: Sized + Unpin`, `size_of::<T>() > 0` | `dereferenceable(size_of::<T>())` |
| unsized reference | — | no validity attributes in this change |
| raw pointer | — | no reference-derived validity attributes |

`align 1` is omitted as redundant.

Zero-sized pointees never produce `dereferenceable(0)`. An over-aligned ZST may
therefore carry `nonnull align N` without `dereferenceable`.

## Implementation outline

1. During MIR import, inspect the original rustc parameter type while type,
   layout, `Freeze`, and `Unpin` information is available.
2. Record only proven per-source-argument validity facts on `MirFuncOp`.
3. Verify that reference-derived facts are attached only to source arguments
   represented as Rust references.
4. Preserve the facts through MIR lowering.
5. Remap source arguments to physical LLVM pointer parameters after ABI
   flattening.
6. Make `llvm-export` mechanically serialize the already-proven attributes.

The downstream layers must not infer Rust validity semantics from pointer
mutability, address space, LLVM pointer shape, or `DisjointSlice`.

## Conservative cases

This change intentionally does not emit validity attributes for:

- raw pointers
- erased/compiler-generated pointers
- unsized references
[...]

PR #1244, opened 2026-09-06
perf(llvm): emit proven nonnull and align for kernel refs

uurl:
## What this fixes

Emit `nonnull` and `align` for kernel reference parameters when typed Rust layout facts and the host launch packet establish those guarantees. This completes the currently unblocked portion of #1098.

## Shape of the fix

- Carry reference kind and alignment through MIR and the existing argument mapping.
- Check nonempty device addresses before enqueue. Null or insufficiently aligned pointers panic at the shared packet boundary.
- Give empty and zero-sized slices an aligned, non-null sentinel.

```text
Rust reference -> typed layout fact -> kernel pointer -> LLVM attribute
                                             ^
                                  checked host launch packet
```

## Guardrails

- Synchronous, borrowed async, owned async and row-width paths share the pointer check.
- Lengths and raw pointers receive no reference attributes; metadata names use the shared symbol owner.
- `dereferenceable`, `noalias` and `readonly` remain outside this change. #1098 stays open for that work; no speedup is claimed.

## Validation

- Workspace and affected compiler tests, strict Clippy, formatting, docs and the 228-case hosted example sweep passed.
- Native, modern/legacy NVVM and materialized-cubin controls passed; valid GPU launches reported zero memcheck errors.
- Combined integration with #1247 and #1258 passed.
- All seven hosted workflows passed on `c5e94c51`; PR-only Pages deployment was intentionally skipped.


nihalpasham (approved):
Reviewed and validated `c5e94c51`: **ready after maintainer repair**.

- Kept the typed reference facts and added the host-side alignment check before enqueue.
- Empty/ZST slices retain their aligned sentinel; sync and both async routes share the guarantee.
- The example now runs on GPU. Metadata naming and example pins also pass their CI guards.

```text
align 8192 required + address 0x1000 -> reject before launch
empty/ZST slice                    -> aligned packet sentinel
```

Workspace/example checks, native and legacy/modern NVVM, GPU memcheck and all seven hosted workflows pass. Author/DCO are preserved and commits are signed. This partially addresses #1098; broader alias attributes remain open.

PR #1245, opened 2026-09-06
feat(fuzzer): generate device static operands

### LINKED ISSUE #1162: Rustlantis: support Static and StaticMut operands in custom MIR

uurl:
## What this adds

Rustlantis can generate scalar `static` and `static mut` allocations and use the custom-MIR `Static` / `StaticMut` operands. The production adapter preserves those declarations in the generated CPU/GPU case.

```text
before: generated programs cannot obtain pointers to static allocations
after:  generated programs can read, write and dump scalar static values
```

`static_count` defaults to zero. Enabled statics use persistent allocations with global-root provenance in the place graph.

## Alias model

Static reacquisition reuses persistent global-root provenance rather than creating an independent tag. Acquisition is rejected when an incompatible live alias blocks access.

Reference validity follows memory-tag validity, so invalidated static-derived references cannot subsequently be selected or copied as function arguments.

Raw pointers derived through a shared reference preserve the inherited provenance across copies, overwrites and call/return cleanup. Retiring the direct shared-reference graph edge therefore cannot make a derived raw pointer writable. Raw pointers obtained directly from `StaticMut` retain the mutable static-root provenance.

Call-frame protection for static-derived references is tracked across calls and released at frame exit without perturbing the ordinary zero-static generation path.

## Verification

Validated at `99b2ca4ec` against main `b9847e95`:

* `generate`: 34/34 tests pass.
* Rustlantis workspace check and tests pass.
* Added graph regressions covering:

  * repeated static acquisition;
  * live reference arguments;
  * frame-exit protection;
  * shared aliases invalidated by a static-root write;
  * a shared-derived raw pointer remaining non-writable after its source reference is overwritten;
  * a shared-derived raw pointer retaining those permissions across call/return;
  * a raw pointer returned directly from `StaticMut` remaining writable.
* Added an argument-selection regression proving an invalidated static-derived reference is not selectable for a call.
* With the checked-in full config and `static_count = 0`, seeds 0–19 are byte-identical to main, including seed 15.
* With the checked-in full config and `static_count = 8`, seeds 0–199 generate successfully.
* Seed 0 with statics enabled completes the `StaticMut` read/write/read workload with matching CPU/GPU traces.
* The generated PTX contains `.global` device symbols and global-memory load/store operations (`ld.global` / `st.global`) for the static workload.

Partially addresses #1162.
Supersedes #1163.


nihalpasham (changes requested):
Reviewed `878a0bcf`: **changes requested**.

- The scalar workload works: seed 0 matches CPU/GPU traces, and the zero-static-count control preserves the old seed output.
- Static reacquisition needs a consistent alias rule. It currently creates a fresh tag that bypasses an existing exclusive borrow:

```text
live &mut global passed into a call
  +-- independent StaticMut(global) -> graph permits read/write
```

- Model global-root provenance across calls and derived aliases, and reject unsupported combinations. Add graph controls for repeated acquisition, live reference arguments and frame exit.
- Rebase onto current main and sign the revision.

This is needed before generated differences can be trusted as compiler findings. #1162 remains partial.

uurl:
@nihalpasham updated in `078845f64` and rebased onto current `main`.

The static aliasing model now keeps stable global-root provenance across reacquisition and calls, with coverage for repeated acquisition, live reference arguments, and frame exit.

`generate` passes 26/26, seed 0 still matches CPU/GPU, and seed 33 with `static_count = 0` remains byte-identical to `main`.

Ready for re-review.

nihalpasham (changes requested):
The persistent global-root provenance addresses the earlier reacquisition problem, but `078845f6` still needs changes before the generated programs can be trusted.

- A write through the static root removes a shared alias's memory tag, but `is_ref_valid()` still accepts that alias because its pointee is initialized. The argument selector can then copy it into a call, where `protect_tag()` panics with `tag has registered pointee ranges`.
- This also affects the default path: with the checked-in config and `static_count = 0`, main completes seeds 0–19; this revision panics on 15, and the five successful outputs all change. The tiny seed-33 control misses the call paths.

Please make reference validity and tag invalidation agree across copies, arguments and frame exit, and add regressions for invalidated aliases and the full zero-static configuration. Removing the assertion alone would hide the inconsistent state.

#1162 remains partially addressed.

uurl:
@nihalpasham updated in `d1b2adca7`.

Reference validity now follows memory-tag invalidation, with regressions for the invalidated-alias and argument-selection paths.

I also reran the full-config controls: seeds 0–19 with `static_count = 0` are byte-identical to main, including seed 15, and 200/200 seeds with `static_count = 8` generate without the call-protection panic.

`generate` is 28/28 and the Rustlantis workspace checks/tests pass.

Ready for re-review.

nihalpasham (changes requested):
I rechecked `d1b2adca7`. The invalidated-reference and full-config zero-static regressions from the last review are fixed.

The new tag cleanup still allows a write through a raw pointer derived from a shared reference. For scalar statics `static mut S: i32` and `static T: i32`, this custom-MIR sequence is enough:

```rust
p = StaticMut(S);
s = &(*p);
q = core::ptr::addr_of_mut!(*s);
s = Static(T);
*q = 9;
```

- Before reassigning `s`, the model correctly rejects writes through `q`. Afterwards, [`remove_edge`](https://github.com/NVlabs/cuda-oxide/blob/d1b2adca7646962a4f78bf8691e82d22628a12d3/crates/fuzzer/rustlantis/generate/src/pgraph.rs#L1530-L1535) removes the shared tag because its final direct graph alias disappeared. The derived raw pointer survives, becomes writable, and is selectable by the ordinary assignment generator.
- The same permission change occurs when a callee returns such a raw pointer and frame cleanup removes its shared local. Returning a raw pointer obtained directly from `StaticMut` works and needs to remain supported.

Creating a mutable raw pointer in custom MIR does not give it permission to write through shared-derived provenance. Reassigning the shared local or ending its frame cannot grant that permission. See Rust's [pointer provenance contract](https://doc.rust-lang.org/std/ptr/index.html#provenance).

Please preserve inherited permissions when retiring graph aliases, consistently across pointer derivation, copies, overwrites and calls/returns. Add regressions for both cases above and retain the direct-`StaticMut` return control. Keeping every old tag indefinitely would also retain obsolete exclusive borrows, so that alone is not a complete repair.

The broader generation sweep passed 3,290 configuration/seed cases; zero-static output matches both baselines. After local mechanical integration with current main, 46 Rustlantis tests pass, but both new alias regressions still fail. This is the generator accepting an invalid write, not evidence of a CUDA backend miscompile.

Please also rebase onto current main, adapt the new calls to the checked `Place` API, and sign the revision. #1162 remains partially addressed.

uurl:
@nihalpasham updated in `99b2ca4ec` and rebased onto main `b9847e95`.

Raw pointers derived through shared static references now preserve the inherited provenance across overwrites and call/return cleanup. I added regressions for both cases, plus a positive control confirming that a raw pointer returned directly from `StaticMut` remains writable.

`generate` is 34/34 and the Rustlantis workspace checks/tests pass. Full-config `static_count = 0` seeds 0–19 remain byte-identical to main, including seed 15; `static_count = 8` generates 200/200 seeds successfully; and seed 0 still matches CPU/GPU with the expected PTX global-memory operations.

Ready for re-review.

nihalpasham (changes requested):
The persistent static roots and adapter preservation are useful, but the shared-reference permissions still have a hole in the `Ref` path.

- `generate_ref` can select `&mut (*shared)` using only read permission. `set_ref` gives it a fresh exclusive tag.
- Overwriting `shared` then removes its shared tag, so the derived `&mut` becomes writable. Overwriting a local cannot grant write permission to a reference derived through shared access.
- I added an isolated graph check for that sequence; it fails on this head. The same custom-MIR sequence fails Miri Tree Borrows at the write. All 500 generation seeds completed, which does not establish that their programs are valid.

Please carry the source permissions through reference creation, copies and call/return cleanup, with regressions for each. This partially addresses #1162; the alias model needs fixing before it can land.

Issue #1238, opened 2026-09-05
TMA G2S lowering makes ptxas ignore setmaxnreg via an extern compatibility call


lucifer1004:
**Description**

A kernel that combines `setmaxnreg` warpgroup register reallocation with `cp_async_bulk_tensor_2d_g2s` loses both register-reallocation instructions during PTX-to-SASS compilation. The emitted PTX retains `setmaxnreg.dec 40` and `setmaxnreg.inc 232`, but CUDA 13.3 ptxas reports C7506 and the TMA kernel's SASS contains an address-compatibility `CALL.ABS` followed by `UTMALDG.2D`, with no `USETMAXREG`. An otherwise identical no-TMA control retains both `USETMAXREG` instructions.

**Minimal reproducer**

At cuda-oxide revision `f381885651b10e7c46cf036dab71622470ca02b8`, create a standalone example with `cuda-device`, `cuda-host`, and `cuda-core = "0.3.1"` dependencies:

```rust
use cuda_device::{
    SharedArray,
    barrier::{Barrier, mbarrier_init},
    kernel, launch_bounds, thread,
    tma::{TmaDescriptor, cp_async_bulk_tensor_2d_g2s},
};
use cuda_host::cuda_module;

const THREADS: u32 = 384;
const MATH_THREADS: u32 = 256;

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    #[launch_bounds(THREADS, 1)]
    pub unsafe fn setmaxnreg_control(output: *mut u32) {
        let thread = thread::threadIdx_x();
        if thread >= MATH_THREADS {
            unsafe { thread::setmaxnreg_dec::<40>() };
            return;
        }
        unsafe { thread::setmaxnreg_inc::<232>() };
        if thread == 0 {
            unsafe { output.write(1) };
        }
    }

    #[kernel]
    #[launch_bounds(THREADS, 1)]
    pub unsafe fn setmaxnreg_with_tma(tensor_map: *const TmaDescriptor, output: *mut u32) {
        static mut TILE: SharedArray<u8, 128, 128> = SharedArray::UNINIT;
        static mut READY: Barrier = Barrier::UNINIT;

        let thread = thread::threadIdx_x();
        if thread == 0 {
            unsafe { mbarrier_init(&raw mut READY, 1) };
        }
        thread::sync_threads();
        if thread >= MATH_THREADS {
            unsafe { thread::setmaxnreg_dec::<40>() };
[...]

PR #1249, opened 2026-09-06
fix(tma): select CTA-local G2S on SM120/SM121

lucifer1004:
## What this changes

Select CTA-local TMA global-to-shared loads on SM120/SM121, including their `a` and `f` targets. For local destinations, this avoids the compatibility call that conflicts with register reallocation. Partially addresses #1238.

## Shape of the change

- Select local inline PTX for non-multicast loads of ranks 1–5.
- Retain the cluster route for other targets and multicast operations.
- Reject explicitly cluster-addressed destinations on the local route.
- Fix the target hint before lowering so automatic and explicit target selection agree.

```text
SM120 local load -> direct TMA instruction -> register reallocation retained
```

## Validation

- Target/rank controls and native, real NVVM and ptxas checks passed.
- Valid local GPU examples passed memory and synchronization checks.
- SM121 hardware and peer-CTA operations were not exercised.

## Remaining work

The existing API permits cluster-shared memory. A GPU target alone does not prove that the destination and barrier belong to the executing CTA (thread block), and a helper call can carry these addresses through a generic pointer.

Prove locality for both operands or expose an explicit local API while preserving the cluster contract. The successful local controls do not resolve this API gap.


nihalpasham (changes requested):
Reviewed `5620feea`: **changes requested**.

- The local path removes the compatibility call and retains register reallocation; rank/target controls and local GPU memory/synchronization checks pass.
- The existing API permits cluster-shared destinations. A GPU target alone does not prove that both destination and barrier belong to the executing CTA (thread block).

```text
same GPU target:
  local destination + barrier -> CTA form is valid
  peer-CTA destination        -> preserve the cluster contract
```

- Establish locality or introduce an explicit local API. Checking only explicitly cluster-typed pointers misses generic pointers across helper calls. Cover both operands and preserve existing cluster behavior.
- Rebase and sign the revision.

#1238 remains partial.

nihalpasham (changes requested):
Rechecked `5620feea`: the locality blocker remains.

- The target/rank tests still pass, but a new control with **both destination and barrier passed as generic pointers** also lowers to `shared::cta` on both backends. Neither operand's CTA ownership is established.
- The existing API promises cluster-shared destinations. A pointer passed through a helper cannot safely acquire a narrower contract because the selected GPU is SM120/SM121.

I would not land this target-only conversion. Please preserve the cluster route and either prove both operands are local or introduce an explicit local API, with generic-helper regressions. The B200 cannot supply SM120/SM121 runtime validation. #1238 remains partial.

Issue #1251, opened 2026-09-06
Expose source-derived launch signatures for concrete #[cuda_module] kernels


lucifer1004:
**What problem does this solve?**

`#[cuda_module]` already parses each kernel's ordered source arguments to
generate typed launchers and marshal them into the CUDA launch packet.

That information is not currently available to heterogeneous kernel
registries or graph compilers. Such consumers must repeat the entry name,
argument names, order, and launch carrier kinds in a second handwritten
descriptor. This can drift when several arguments have the same Rust type:
the typed launcher still compiles, while a name-addressed dynamic binder may
silently bind the wrong resource.

The requested capability is static metadata generated from the same
`CudaModuleKernel` model that already drives launcher generation. It is not
runtime reflection or cubin introspection.

**Proposed API / behavior**

For a concrete kernel:

```rust
#[cuda_module]
mod kernels {
    #[kernel]
    pub unsafe fn transform(
        scale: f32,
        input: *const u16,
        output: *mut u16,
    ) {
        // ...
    }
}
```

the macro would expose a host-only static descriptor, with an API along these
lines:

```rust
pub struct CudaKernelSignature {
    pub entry: &'static str,
    pub arguments: &'static [CudaKernelArgument],
}

pub struct CudaKernelArgument {
    pub name: &'static str,
    pub kind: CudaKernelArgumentKind,
}
```

For example:

```rust
let signature = &kernels::TRANSFORM_CUDA_SIGNATURE;

assert_eq!(signature.entry, "transform");
assert_eq!(signature.arguments[0].name, "scale");
```

The exact namespace is open for discussion: a module-level constant, an
associated constant on the generated kernel marker, or a generated accessor
would all satisfy the use case.

The descriptor should represent source launch arguments rather than a
flattened PTX parameter list. For example, a Rust slice remains one source
argument whose kind records that it expands to pointer and length launch
slots.

The entry name should reuse the existing `CudaKernel::PTX_NAME` authority.
Argument names, order, and carrier classification should come from the same
parsed parameter model used by the generated launcher.

Initial scope can be limited to non-generic concrete entries. Generic
specializations can be designed separately if a real consumer requires them.

The metadata should:

- only be emitted with the host surface;
- follow the kernel's visibility and `cfg` attributes;
- support nested `#[cuda_module]` modules;
- describe existing scalar, raw-pointer, and slice argument forms;
- compose with grid-constant arguments once #1225 lands;
[...]

PR #1252, opened 2026-09-06
feat(cuda-module): expose concrete kernel signatures

lucifer1004:
## What this adds

Expose host-side kernel metadata for registries and launch binders: the PTX entry name, ordered source arguments and their launch carriers. Partially addresses #1251; this draft depends on #1225.

## Shape of the change

- Generate static descriptors for concrete, non-generic kernel entries.
- Group flattened pointer/length and row-width carriers under their source argument.
- Reuse the launcher's PTX name and retain visibility, configuration and nested-module handling.

```text
source slice argument -> one descriptor -> pointer + length carriers
```

## Validation

Ordinary host and macro suites passed. Fresh, separate builds of the parent and this head confirmed the type, lifetime and name controls below.

## Remaining work

- Use resolved scalar identity or report the type as opaque. With `type f32 = u32`, the launcher accepts `u32` while the descriptor incorrectly reports `F32`.
- Preserve lifetime scope: `PhantomData<&'a ()>` is valid on the parent but the generated module constant loses `'a` here.
- Avoid collisions between distinct names such as `step` and `STEP`.
- Resolve the parent PR's legacy ABI and reference-contract gaps before treating this metadata as a launch ABI contract.


lucifer1004:
Downstream integration validation is now complete. I rebuilt the local aggregate with this PR\x27s exact signature delta as `5bf361d3`, migrated the heterogeneous registry consumer to `entry` / `arguments` / `CudaKernelArgumentKind`, and rebuilt the release kernel pack. The cubin remained byte-identical at `65441e62870ec7ed723b82597aa07ebbb517a26b5c4fd23955f1f110d107bcac`; the full downstream fmt, test, release-Clippy, and crate-fence gate passed, followed by all 54 ignored GPU kernel regressions.

nihalpasham (changes requested):
Reviewed draft `64333820`: **changes requested on the metadata changes**.

- Scalar identity cannot come from the identifier text:

```text
type f32 = u32;
kernel(value: f32)
  launcher: u32       descriptor: F32
```

- Use resolved type identity, or opaque metadata where identity is unproven. Equal-size aliases also fail, so a size check is insufficient.
- Preserve lifetime scope and distinct names: `PhantomData<&'a ()>` and `step` / `STEP` compile on the parent but fail with the new module constants.
- Keep the draft blocked on #1225, and sign/complete DCO on the revised stack.

Ordinary host/macro tests pass; independent host-only controls confirm these three boundaries. #1251 remains partial.

nihalpasham (dismissed):
I fixed the metadata blockers in `54afbfa4` and brought the branch onto the merged #1225 implementation:

- Ambiguous scalar names now report opaque layout metadata. Qualified `core::primitive`/`std::primitive` paths retain their scalar identity, so a shadowed `f32` cannot be reported as a float.
- Lifetime-only signatures retain valid layout queries.
- Generated signature names preserve case; `step` and `STEP` no longer collide.
- Added semantic regressions for aliases, lifetimes, raw names, configuration and nested modules.

The host and macro suites pass with default and all features, along with strict Clippy and formatting. The remaining diff is metadata only; current grid-constant ABI and unsafe launch contracts are preserved. This is a partial answer to #1251, not a complete device ABI or a safe dynamic launch binder.

nihalpasham (approved):
I fixed the metadata blockers in `54afbfa4` and brought the branch onto the merged #1225 implementation:

- Ambiguous scalar names now report opaque layout metadata. Qualified `core::primitive`/`std::primitive` paths retain their scalar identity, so a shadowed `f32` cannot be reported as a float.
- Lifetime-only signatures retain valid layout queries.
- Generated signature names preserve case; `step` and `STEP` no longer collide.
- Added semantic regressions for aliases, lifetimes, raw names, configuration and nested modules.

The host and macro suites pass with default and all features, along with strict Clippy and formatting. The remaining diff is metadata only; current grid-constant ABI and unsafe launch contracts are preserved. This is a partial answer to #1251, not a complete device ABI or a safe dynamic launch binder.

At `54afbfa4`, hosted tests, lint/format checks, guards, docs/book and example compilation passed. CodeQL was still running at the final review check.

Issue #1217, opened 2026-09-02
smoketest: code-shape gates are not applicable to full-debug builds (11 examples)


nihalpasham:
Forcing `CUDA_OXIDE_DEBUG=full` on every example (not a CI configuration) fails 11 of them for one reason: their code-shape gates assert on **optimized** output, and full debug removes the optimizer by design.

```text
 full debug          normal build
 mem2reg   skipped   run          ← gates that grep for "no local memory" fail
 opt -O2   skipped   run          ← gates that grep for fused / vectorized forms fail
 llc       -O0       O2
```

| example | gate asserts | needs |
|---|---|---|
| aligned_field_loads / aligned_field_stores | vectorized field-pair `ld` / `st` | `opt` |
| cluster | exact 6-read, 2-mad, 2-mul formula | `opt` |
| copy_aggregate_borrow | no `.local` / `ld.local` / `st.local` | mem2reg |
| dpx_minmax_chains | min/max chains fused into `vimax3` | `opt` |
| redux_f32 | `redux.sync` f32 forms | `opt` |
| generated_intrinsics / generated_intrinsics_blackwell / tcgen05 | instruction-form counts of the optimized route | `opt` |
| array_constants | `test -s array_constants.opt.ll` | the `.opt.ll` artifact, which does not exist without `opt` |
| disjoint_slice_len | one `load {ptr,i64}` of the receiver feeding `extractvalue ... 1` (inline check in smoketest.sh) | the SSA shape mem2reg produces |

None of these is a compiler bug, and debugging any of these examples under cuda-gdb with full debug is still a legitimate use. Proposal: when full debug is on (`--device-debug` or `CUDA_OXIDE_DEBUG=full`), smoketest skips code-shape gates with a note instead of reporting the example as failed.

Measured 2026-09-02 on `fix/full-debug-keeps-mir-optimizations`; the same 11 fail on main at MIR level 0.

vyncint:
Taking this, building your proposal. Two things I found while mapping it that are worth agreeing before I write code.

[...]

PR #1253, opened 2026-09-07
smoketest: refuse an ambient full-debug CUDA_OXIDE_DEBUG instead of half-honouring it

vyncint:
## What this fixes

An ambient full-debug setting no longer starts a smoketest run whose checks expect optimized device code. The script directs the caller to the existing explicit full-debug census.

```text
CUDA_OXIDE_DEBUG=2 scripts/smoketest.sh              -> explain and stop
CUDA_OXIDE_DEBUG=2 scripts/smoketest.sh --full-debug -> run the debug census
```

## What we fixed

- The hidden `cargo oxide __debug-policy` helper uses the compiler's shared Rust parser, including aliases, case and Unicode whitespace.
- Preflight consumes the helper's exact result and rejects failed or malformed replies.
- Explicit full-debug mode and non-full policies retain their existing behavior. Build, artifact and debug checks remain in their established paths.
- Added executable preflight controls for both modes and failed helper replies.

## Verification

Checked locally at `853102ce`:

- 229 cargo-oxide tests and 33 finalizer tests passed; six optional finalizer tests remained ignored.
- 57 real CLI parser controls and 114 real-parser smoketest controls passed across three locales.
- All 22 permanent preflight controls passed and rejected both unchecked-status and output-truncation mutations.
- Formatting, strict Clippy, warnings-denied docs, shell syntax and affected guards passed.

## Caveats

This changes preflight only; it does not alter device compilation or execute kernels. No GPU result is claimed for this head. An ambient full-debug setting requires `--full-debug`; the script does not silently switch modes.

Partially addresses #1217 alongside the full-debug census added in #1259.


vyncint:
Correcting my own PR body: it said the T4 was on driver 580.95. It was **595.71.05** -- the evidence log records `Tesla T4, 15360 MiB, 595.71.05, 7.5`, and I wrote the other number from nothing. Body fixed; no other figure in it came from anywhere but the captured runs, and the compute capability (7.5) and the pass/fail counts are as measured.

vyncint:
Filed the pre-existing failure this PR surfaced as #1254, and it is broader than the one example I mentioned: it is the whole `ltoir` category, all nine, every one with the same diagnosed refusal (`legacy LLVM 7 NVVM IR for sm_86 does not yet support cuda-oxide debug metadata`). Grepping all nine logs for `error: [rustc_codegen_cuda]` returns that one string and nothing else.

Recording why it stays out of this PR: those nine never reach a shape gate, the build itself refuses. This change skips gates for a named list and never skips a build, so it leaves them untouched by design. #1254 has the options.

nihalpasham (changes requested):
Reviewed `bf8ef607`: **changes requested**.

- The submitted code still returns before the second libNVVM build in `generated_intrinsics`, `generated_intrinsics_blackwell` and `tcgen05`. The description says this was fixed, but exact-function controls show those builds are skipped.
- Rework forced-environment handling around main's `--full-debug` mode from #1259, with one policy for applicable checks.

```text
build PTX -> build/verify NVVM -> check debug/artifact invariants
                                skip only optimized-shape checks
```

- Test with fresh artifacts and a deliberately failing second build. Retain every build and propagate its failure; report skipped checks clearly.

#1217 remains partial.

vyncint:
Your "three branches" finding is right, and the reason is worse than a missed case: **the commit I pushed does not contain the fix my own evidence describes.** Correcting that before you spend any more time on it.

What happened. I hit exactly the failure you describe during development — `generated_intrinsics_blackwell` came back `FAIL (libNVVM compile produced no fresh LTOIR)` because the guard returned before the second build. I fixed it by having the guard run the libNVVM route itself, verified it, and the `11 / 11` block in the PR body is a real capture of that tree. Then, while checking something unrelated (whether `device_ffi_test`'s failure was pre-existing), I reverted `scripts/smoketest.sh` to upstream and restored it with `git checkout HEAD -- scripts/smoketest.sh` — which restored the *commit*, discarding the uncommitted fix. I pushed after that.

So on `bf8ef607`:

```
nvvm_args invocations inside a shape_gate_exempt guard: 0
```

All three guards are the plain `CARGO_EC=0; return` form, and `blackwell-compile`, `tcgen05` and `generated_intrinsics` each skip a second `cargo oxide` below them. The PR body's after-state describes a tree that is not what you reviewed. That is my error, not an ambiguity in the report.

It also means one of my own checks was hollow. I verified "nothing builds after this guard" with a script whose block-end regex failed on the `tcgen05` branch, so it scanned an empty range and answered "none" — the same green-because-nothing-was-scanned shape I have been careful about elsewhere and still walked into. Re-run properly, the second builds sit at lines 1035/1038, 1354/1357 and 1628/1631.

Taking the whole review rather than just restoring the lost hunk:

- rebase onto `main` and reconcile the forced-environment path with `--full-debug` from #1259, so there is one mode rather than two spellings of it
- no branch skips a build: every second libNVVM route runs, and its failure still propagates to the verdict
- controls for both of those — fresh artifacts rather than stale ones, and a deliberately failed second build that must still be reported

I will re-verify on hardware and post the capture from the tree that is actually pushed, with the commit SHA next to it so the two cannot drift again.

vyncint:
Force-pushed a rewrite rather than a patch, and retitled.

Your pointer to #1259 made the old approach redundant: `--full-debug` already keeps the optimized shape gates out of its lane and allowlists the one invariant contract, which is a better design than my eleven-name exempt list. Measured it against #1217's eleven on an A10G before rewriting — 11/11 pass under `--full-debug`, so there was nothing left for the exempt list to do.

What was left is the gap that mode does not cover: `smoketest.sh` never reads `CUDA_OXIDE_DEBUG`, so an ambient `full` silently produces the hybrid. This now refuses it and names the flag. That also disposes of your other two points rather than answering them — nothing here touches `run_cargo`, so no branch skips a build and there is no second-build failure to propagate; the fresh-artifact checking is #1259's and the run above exercises it.

Branch reset onto `main`, so this is one commit against `6abfaa09`.

nihalpasham (changes requested):
Re-reviewed `5305b634`: **changes still needed**.

- The rewrite removes the skipped-build problem and uses the existing `--full-debug` mode.
- One policy gap remains: this guard recognizes only literal `full`. The shared `DebugPolicy` parser also treats `2`, `FULL` and padded `Full` as full debug.

```text
CUDA_OXIDE_DEBUG=full -> rejected
CUDA_OXIDE_DEBUG=2    -> accepted by sweep, full debug in compiler
```

- Use the canonical debug-policy interpretation at preflight, including aliases, case and whitespace. Keep explicit-mode and non-full controls.
- A 30-case script harness and the real parser checks confirm the mismatch.

#1217 remains partial.

vyncint:
Right, and the gap was wider than `2`: `FULL`, `Full`, `" full "` and a `\r`-terminated value all selected full debug in the compiler and sailed past a literal comparison.

Fixed by not having a table here at all. The preflight reads the aliases out of `parse_env_override`'s own arm, so the two cannot drift: dropping `2` from the parser stops the preflight refusing `2`, and adding a `dbg` alias makes it refuse `dbg`, both with no edit to the script. An arm shape the extraction cannot read is a hard error rather than a fallback, since a silent fallback is the same failure in a new place.

One case I did not expect to matter and then did: `str::trim` trims the Unicode White_Space set, and glibc deliberately keeps U+00A0 out of `[[:space:]]`. So `CUDA_OXIDE_DEBUG=$' full'` — what pasting a command out of rendered docs produces — was full debug to the compiler and an ordinary value to a POSIX trim, which fails in the under-firing direction. The set is spelled out now. That was 2 mismatches out of 39 before the fix and 0 after; the full parity table and the five mutations are in the body.

Also in the body, and the thing I owed you from last time: the hardware capture is tied to the pushed tree by `sha256sum scripts/smoketest.sh`, taken on the box and compared against `git show <head>:scripts/smoketest.sh`. Both read `69691c56…`. The refusal matrix and #1217's eleven at 11/11 are from that run, A10G / CC 8.6 / driver 595.71.05. I am not quoting the smoketest header's commit line, because it comes from the box's own clone metadata rather than from the synced tree — which is the kind of thing that let my evidence and my commit disagree last time.

Rebased onto current `main`, still one commit.

nihalpasham (changes requested):
Reviewed `c7b86376`: **changes still needed**.

- The rebase is clean, and the earlier skipped-build problem is resolved.
- Two gaps remain in the new preflight: Bash normalization depends on locale, and source-text extraction can lose accepted values after valid Rust formatting changes.

```text
LC_ALL=C, CUDA_OXIDE_DEBUG=2 or FULL
preflight -> normal sweep
compiler  -> full debug
```

- Use the shared Rust policy result directly. Another whitespace table or source parser would keep two implementations of the same policy.
- On this head, 16 focused script controls reproduce both locale failures; explicit-mode and non-Full controls pass. The real Rust parser and independent source-layout controls confirm the underlying mismatch.

#1217 remains partially addressed. The preflight needs one policy implementation before approval.

vyncint:
Both reproduced here before changing anything, and you were right that the shape of the fix was the problem, not the details.

```
LC_ALL=C   CUDA_OXIDE_DEBUG=2      preflight: allowed   compiler: full debug
LC_ALL=C   CUDA_OXIDE_DEBUG=FULL   preflight: allowed   compiler: full debug
```

`${v,,}` and `[[:space:]]` both answer to the locale, and my `$' '`-style whitespace set degrades to bytes under `LC_ALL=C`, which is what corrupts the comparison. The source-extraction half fails independently: splitting the arm across lines, which rustfmt accepts, turns the old preflight into a hard error.

| parser arm | old preflight | new |
|---|---|---|
| `"2" \| "full" => Some(Self::Full),` | REFUSED | REFUSED |
| same arm reformatted across lines | **hard error: cannot read the aliases** | REFUSED |

## One implementation

Taken as you asked — the script no longer models the policy, it asks for the result. `cargo oxide __debug-policy` prints what `parse_env_override` makes of `CUDA_OXIDE_DEBUG` in the caller's own environment: `unset`, `none`, `line-tables`, `full`, or `unrecognized`. Hidden, following `__materializer-handshake`, which exists for the same reason; `check-cli-doc-coverage.sh` reports `2 hidden one(s) correctly excluded`, so it adds no documented surface.

The staleness question answers itself, which I had not realised until I went looking: `.cargo/config.toml` defines `oxide = "run --package cargo-oxide --"`, so inside this repository `cargo oxide` always resolves to the tree under test rather than to an installed binary. The helper cannot drift from the sweep. A cargo-oxide that cannot answer at all is an error rather than a guess — it would otherwise run the hybrid this exists to prevent:

```
error: 'cargo oxide __debug-policy' reported no usable debug policy
```

## Controls

Every value below, in `LC_ALL=C`, `C.UTF-8` and `en_US.UTF-8` — 24 runs, all three locales agreeing:

```
2  FULL  full  " full "  " full"    -> REFUSED
line-tables  off  garbage                -> allowed
```

Explicit mode: `2`, `FULL`, `" full "` and `full` each passed with `--full-debug` — allowed in all four, in every locale. A new unit test pins the token mapping over 24 spellings including `2`, case, padding and the NBSP form, next to the existing `parse_env_override` tests. `cargo test -p cargo-oxide` 229 passed, fmt clean, clippy `-D warnings` clean on both crates touched, `check-cli-doc-coverage.sh` passes.

No overlap with #1268, which touches the same three cargo-oxide files in different regions (`codegen_env.rs` ~548/~644 against my ~183, `tests.rs` ~3425 against ~4480, `main.rs` ~768/~804 against the command enum).

## On the hardware capture

It is from `c7b86376`, the immediate parent — `sha256 fcd1e9e6…`, A10G / CC 8.6 / driver 595.71.05, refusal matrix as above and #1217's eleven at 11/11 in 158s. This head is `sha256 9fe31a88…`, and the delta is exactly the preflight rewrite plus the new hidden subcommand, so the census would be re-running #1259's lane unchanged. I stopped the box rather than spend another on-demand hour on that; say the word and I will re-run the eleven against this head.

nihalpasham (commented):
The shared Rust policy now fixes the alias, whitespace and locale mismatch from the earlier review.

I added a small follow-up so preflight also rejects a helper that exits unsuccessfully or prints extra output, with executable controls for normal and explicit full-debug modes. The existing build and debug-check paths stay intact.

At `853102ce`, the cargo-oxide/finalizer suites, 114 real-parser smoketest controls, formatting, strict Clippy and docs passed. The new controls also catch both helper-error regressions. Hosted CI also passes on this head.

This partially addresses #1217 alongside #1259's explicit full-debug census.

nihalpasham (approved):

PR #1256, opened 2026-09-09
fix(codegen): reject invalid device architecture hints on both backends

### LINKED ISSUE #1189: PTX and NVVM paths disagree on `compute_XX` device hints
## Problem

`CUDA_OXIDE_DEVICE_ARCH` is an advisory device-architecture hint, but its accepted spelling depends on which artifact path the pipeline takes. A `compute_XX` hint is ignored by the PTX path and accepted by the NVVM path. Issue #1183 covers parsing CUDA targets once and threading `CudaArch` through codegen, but does not decide whether these two paths should treat the same hint consistently. PR #1184 preserves the existing behavior and was asked to move that policy question to a separate issue.

```text
CUDA_OXIDE_DEVICE_ARCH=compute_120
                 │
                 ▼
       BackendOptions.device_arch_hint
                 │
        ┌────────┴────────┐
        │                 │
        ▼                 ▼
   PTX output         NVVM IR output
   sm_-only gate      CudaArch::from_str
        │                 │
        ▼                 ▼
 hint ignored        hint accepted
        │                 │
        ▼                 ▼
 feature default     normalized target
     sm_80               sm_120
```

## Current behavior

The rustc codegen entry point reads `CUDA_OXIDE_DEVICE_ARCH` without parsing it and places the string in `PipelineConfig.device_arch_hint` (`crates/rustc-codegen-cuda/src/device_codegen.rs:1101-1119`). `BackendOptions::from_env` provides the other environment-driven entry point and does the same (`crates/cuda-oxide-codegen/src/options.rs:100-105`). `cargo oxide run` normally populates this variable from detected hardware through `apply_device_arch_hint` (`crates/cargo-oxide/src/commands.rs:6304-6319`); that producer documents and emits `sm_XX` spellings (`crates/cargo-oxide/src/commands.rs:6330-6337`). A user or another frontend can nevertheless supply `compute_XX` directly.

After lowering, the shared pipeline forwards the hint to one of two resolvers:

- The PTX path copies `BackendOptions.device_arch_hint` and passes it to `resolve_ptx_target_with_generated` (`crates/cuda-oxide-codegen/src/ptx.rs:372`). That resolver considers only hints beginning with `sm_`; other spellings, including valid `compute_XX`, fall through to feature-based target selection: the hint is gated by `arch_satisfies` (`crates/cuda-oxide-codegen/src/target.rs:1815`), which resolves the string through `arch_compute_capability_and_suffix` (`target.rs:1640`, `target.rs:1910-1916`) and rejects anything not starting with `sm_`. PR #1184 lifts that condition into an explicit filter and pins it with a new `resolve_ptx_target_ignores_compute_spelled_device_hints` regression test.
- The NVVM path passes the same hint to `resolve_nvvm_target_with_generated` (`crates/cuda-oxide-codegen/src/pipeline.rs:379-384`). That resolver parses the hint as `CudaArch` and uses it when it satisfies the detected and generated requirements (`crates/cuda-oxide-codegen/src/export.rs:99-150`; the no-automatic-feature branch does the same at lines 158-163). `CudaArch::from_str` accepts both `sm_XX` and `compute_XX`, and its renderers produce the spelling required by each downstream consumer (`crates/cuda-target-spec/src/lib.rs:15-19,67-74,88-94`).

The observable result is not limited to diagnostics. For example, with a basic kernel and `CUDA_OXIDE_DEVICE_ARCH=compute_120`, the PTX path ignores the hint and selects its feature default, currently `sm_80`; the NVVM path accepts the hint as capability 120 and reports/exports target `sm_120`. With `compute_80`, both paths currently arrive at `sm_80`, but the PTX path reports the source as `feature requirement` rather than `detected GPU`. Invalid hints also differ: the PTX path silently ignores them, while the NVVM resolver returns a target-selection error when it attempts to parse them.

| Advisory hint (basic kernel) | PTX path | NVVM path | Observable difference |
|---|---|---|---|
| `sm_120` | selects `sm_120` from `detected GPU` | selects `sm_120` | none in target capability |
| `compute_120` | ignores hint; selects default `sm_80` from `feature requirement` | accepts and normalizes to `sm_120` | different compiled target |
| `compute_80` | ignores hint; selects default `sm_80` from `feature requirement` | accepts and normalizes to `sm_80` | same target, different provenance |
| malformed spelling | silently falls back | returns `TargetSelection` when parsed | fallback versus error |

## Provenance

The hint behavior was introduced by commit `9955a3ef` (`fix(codegen): treat detected GPU arch as a hint, not a hard target`). Its compatibility helper was explicitly documented to accept examples such as `sm_120a` and to extract the capability from an `sm_…` string. The normal `cargo oxide run` producer also emits `sm_` spellings. I found no commit message, test, or comment from that change identifying different `compute_XX` behavior between PTX and NVVM as an intended policy.

PR #1184 makes preservation deliberate during its typed-target refactor: the resolver comment says it preserves the legacy PTX behavior, and the regression test above records it. That is evidence that changing the behavior was intentionally kept out of the refactor; it is not evidence that the original asymmetry was deliberately designed. Whether the original omission was accidental is therefore unverified.

## Why decide this

The same advisory input currently selects different compilation targets solely because the pipeline emits PTX or NVVM IR. It can also turn a malformed input into either a silent fallback or an error. This makes target provenance and the generated artifact depend on output mode rather than only on the requested/detected architecture and module requirements.

## Decision options

1. **Accept and normalize both spellings on both paths.** Parse a device hint once as `CudaArch`, then render `sm_XX` for LLVM/PTX consumers and `compute_XX` where libNVVM requires it. This makes `compute_XX` preserve the user's advisory capability across output modes and matches the shared type's current grammar. It changes the PTX result for callers that currently pass `compute_XX` and receive a feature-default target.

2. **Restrict device hints to `sm_XX` on both paths.** Treat `compute_XX` as invalid at the shared input boundary. This matches the spelling produced by `cargo oxide run` and keeps the hint described as a device architecture rather than a libNVVM target. It would remove NVVM's current acceptance of `compute_XX`; a decision is also needed on whether invalid advisory hints should fail or be ignored.

3. **Keep the path-specific behavior and document it.** This has the lowest compatibility risk, but retains output-mode-dependent target selection and different malformed-input handling. The reason for exposing two accepted vocabularies would need to be made explicit because no such rationale is currently documented.

This issue is requesting a maintainer decision on the input contract and invalid-hint policy before an implementation change is proposed.

almightychang:
Fixes #1189

A device hint identifies the GPU in this machine. Being advisory means a valid hint may yield to the feature floor, not that invalid input should be ignored.

Previously, `CUDA_OXIDE_DEVICE_ARCH=compute_120` selected sm_80 on the PTX path but sm_120 on the NVVM path.

### Before

```text
CUDA_OXIDE_DEVICE_ARCH
          │
      raw string
          │
     ┌────┼──────────────────┐
     ▼    ▼                  ▼
    PTX  NVVM            IKET present
     │    │                  │
 compute_120          promote raw hint
     │    │           to explicit target
     ▼    ▼                  │
   sm_80 sm_120               ▼
 ignored accepted      placeholder error
```

### After

```text
CUDA_OXIDE_DEVICE_ARCH
          │
       parse once
          │
    DeviceArchHint
          │
 shared pipeline boundary
     ┌────┴─────┐
     ▼          ▼
  Invalid   Valid(DeviceArch)
     │          │
 one error  validated backend view
                │
       ┌────────┼─────────┐
       ▼        ▼         ▼
      PTX      NVVM      IKET
                          │
                   promote proven hint
                   to explicit target

     No repeated hint validation
```

### Changes

- Parse hints into a typed `DeviceArch`, retaining invalid input for one shared `TargetSelection` diagnostic.
- Construct validated backend options before all IKET handling. Both resolvers receive only the proven hint.
- Accept `sm_<capability>` with an optional `a` suffix. Reject `f` hints because they select a GPU-family compatibility set rather than identify the detected GPU.
- Preserve explicit `compute_` and family targets, valid-hint fallback, no-hint feature floors, and every in-tree producer.

| Input | Before: PTX / NVVM | After: both |
|---|---|---|
| hint `sm_120` | sm_120 / sm_120 | unchanged |
| hint `compute_120` | sm_80 / sm_120 | identical TargetSelection error |
| hint `compute_80` | sm_80 / sm_80, different provenance | identical TargetSelection error |
| hint `sm_` or `foo` | ignored / error | identical TargetSelection error |
| hint `sm_120f` | accepted / accepted | identical TargetSelection error |
| no hint, Basic features | sm_80 / sm_80 | unchanged |
| explicit `compute_80` | sm_80 / sm_80 | unchanged |

On the tested baseline, IKET promotes `compute_120` to a hard target and then fails its placeholder-contract lookup. This change rejects it at the shared input boundary, before promotion.

### Validation

- Relevant unit and integration suites, including codegen `--all-targets` and the separate rustc backend workspace.
- Clippy with `-D warnings`, formatting, repository guards, and generated intrinsic checks.
- CUDA 13.1 container: **226/226** compile-only examples passed.
- Artifact comparison: **623/624** files byte-identical. The remaining PTX differs only in global declaration order; replaying unchanged baseline IR through the same llc reproduced both hashes across 12 runs.
- RTX 5080: `vecadd`, `math_tan`, `iket_trace`, and `interop_cubin_identity` executed successfully.
- An additional IKET run verified automatic `sm_120a` detection, target promotion, and GPU execution without `--arch`.

The full example corpus was compiled. GPU execution covered the four examples listed above; explicitly ignored live tests were not run.


nihalpasham (approved):
I kept the single typed hint boundary and added a regression for caller-supplied hints overriding the environment. Invalid hints now fail before either backend or IKET promotion, while explicit compilation targets keep their existing behavior.

The codegen, target and importer suites, strict Clippy and docs pass on the rebased branch. The IKET example also passes on RTX 5090 with automatic device detection. All seven hosted workflows passed at `617d13f4`.

This addresses #1189.

PR #1247, opened 2026-09-06
deps(pliron): update to merged GEP no-wrap revision

uurl:
## What this fixes

Use Pliron's merged GEP revision in both dependency graphs. Closes #1246.

## Shape of the fix

- Update the workspace and backend pins and lockfiles to `06c586b`.
- Adapt constant construction and borrowed-value access to the typed attribute API.
- Update the generator and its generated source together.

```text
constant value + width -> typed attribute -> same LLVM constant
```

## Guardrails

All four upstream commits were reviewed. Constant values and widths, GEP no-wrap behavior and existing volatile handling remain unchanged.

## Validation

- 5,269 workspace tests, 33 backend tests, strict Clippy, formatting and documentation checks passed.
- 2,374 compiler target configurations and the 227-case compile/NVVM/ptxas sweep passed.
- Vector, constant, half-precision and packed-aggregate GPU controls passed with zero memcheck errors on RTX 5090. tcgen05 paths were compile-only.
- Combined integration with #1244 and #1258 passed.
- All seven hosted workflows passed on `3ab9fa29`; PR-only Pages deployment was intentionally skipped.

PR #1258, opened 2026-09-10
feat(cargo-oxide): add build debug assertions flag

### LINKED ISSUE #1255: First party debug flag for cargo oxide build

I have a Maturin-based (pyo3) scientific pipeline where I occasionally need to compile conditional `gpu_printf!` macros diagnostics into kernels. `cargo oxide build` currently uses the release-like policy, so it always applies `-Copt-level=3 -Cdebug-assertions=off`, including when preparing the environment for `maturin develop`. 

**Proposed API / behavior**

`--debug-assertions = -Copt-level=3 -Cdebug-assertions=on -Coverflow-checks=off`
seems like the sensible choice of flag for the issue. It keeps the whole example suite working while overflow checks stay a separate decision. 

Nihal's perspective

- **more MIR** : debug_assert! checks and the panic paths behind them, plus your cfg(debug_assertions) code. All of that translates fine today; a failing assert becomes a device trap.
- **overflow checks come along for free**: rustc turns -Coverflow-checks on whenever debug-assertions is on. Every + and * grows a check and a trap. That changes the MIR enough to break at least one pattern-sensitive lowering (the WGMMA accumulator-fusion path stops recognising its K-loop).

```
flag                                     
examples compiling
debug-assertions=on, overflow follows    225 / 227
debug-assertions=on, overflow off        227 / 227
```

**Alternatives considered**

For my current project I am using a Cargo feature as a local workaround. I wanted to check whether exposing the independent rustc setting fits the intended profile model before potentially creating an issue.

`cargo oxide build --device-cfg gpu_diag plus #[cfg(gpu_diag)]` also exists today but seems like a workaround instead of having a toggle.

uurl:
## What this adds

`cargo oxide build --debug-assertions` enables conditional diagnostics while retaining release optimization and wrapping arithmetic. Closes #1255.

## Shape of the change

One shared profile applies to ordinary builds, Cargo passthrough and metadata-declared device builds.

```text
opt-level        = 3
debug-assertions = on
overflow-checks  = off
```

## Guardrails

- The option defaults off; `cargo oxide test` keeps Cargo's selected profile.
- Device-debug precedence and full-debug mode retain their existing behavior, including combinations with assertions.
- Ordinary and Cargo passthrough builds apply the profile to host crates in the same Cargo invocation too.

## Validation

- 228 CLI tests, strict Clippy and formatting passed.
- Assertion on/off and full-debug controls passed through all three build routes on RTX 5090. Device `cfg(debug_assertions)` changed as requested and overflow still wrapped.
- Metadata interop cubins returned the expected values and retained full-debug metadata.
- Combined integration with #1244 and #1247 passed.
- All seven hosted workflows passed on `0c4a62c3`; PR-only Pages deployment was intentionally skipped.


angus-forrest-uk:
#1255 Checked against my issue and it works within our pipeline. Thank @uurl for looking at this.
👍

nihalpasham (approved):
Reviewed and validated `0c4a62c3`: **ready after maintainer rebase**.

- Preserved main's device-debug handling while adding the build-only assertion profile.
- Ordinary, passthrough and metadata device builds share the policy; test profiles remain Cargo-owned.

```text
optimization 3 + assertions on + overflow checks off
```

All 228 CLI tests, strict Clippy, GPU on/off/full-debug controls across the three routes, and all seven hosted workflows pass. Combined integration with #1244/#1247 also passes. Author/DCO are preserved and both commits are signed.

This completes #1255's implementation.

PR #1260, opened 2026-09-11
fix(codegen): bound instruction modifier searches to the current token

paulzhng:
Avoid scanning the remaining LLVM module for each escaped whitespace delimiter. Preserve the first token boundary while eliminating quadratic feature detection on large generated modules.

## Summary

Reduces CUDA compilation time for large modules by avoiding quadratic string searches during target-feature detection. A production crate’s release build dropped from 185.5s to 24.7s, with byte-for-byte identical generated LLVM IR and PTX.

## Changes

- Bound escaped-whitespace searches to the first ordinary token delimiter instead of scanning the remaining LLVM module for every candidate instruction.
- Preserve token-boundary semantics and all target-feature checks.
- Add regression tests for ordinary and escaped delimiters, Unicode whitespace, large modules, and late feature matches.

## Testing

- [X] `just check` passes (the local mirror of CI: fmt, clippy, tests, guards, docs)
- [ ] `cargo oxide run <example>` passes, or `scripts/smoketest.sh -o '^<example>$'` - not ran; checked byte-equivalence on a production device-library build

## Checklist

- [X] All commits signed off (`git commit -s`)
- [X] SPDX headers on new source files


nihalpasham (approved):
I rebased this and checked the token-boundary behavior independently. All 6,084 delimiter/near-match cases agreed with the old implementation, and the large-module scan now scales linearly in the measured cases.

All 1,261 codegen tests, strict Clippy and docs pass. All seven hosted workflows passed at `4953e1e1`.

No issue is closed.

Issue #1211, opened 2026-08-30
`just check` cannot pass inside the recommended `nix develop` shell: the probe pins `llc` by SHA-256, nixpkgs rewrites it


almightychang:
After #1209 / #1210 makes the flake usable again, `just check` still cannot pass in the `nix develop` environment recommended by `CONTRIBUTING.md:123`. The intrinsics probe requires the rust-toolchain `llc` binary to match the catalog's SHA-256, but nixpkgs rewrites that ELF.

In the fixed shell, `just check` passes `fmt-check`, `clippy`, `test`, `test-cuda`, and all of `check-guards`. It then fails in `check-intrinsics` at Justfile line 131, on `cargo run -p cuda-intrinsics-gen -- probe --all --skip-terminal`:

```text
error: validate probe backend for abs_bf16x2: rust-toolchain llc SHA-256 mismatch: selected evidence records 2e5bc688625ca7d98bca9b5f0f0d14525d2eb59e8a9842964e480df3ffd1f921, found ca7af83e237af640fe337faff2eff49055988921933186ab5535b38addeb7b49; use an explicit `--llc` only for a comparison probe
```

`intrinsics/catalog.json` records profile `rust-llvm-23.1.0-16696adc`, version `LLVM version 23.1.0-rust-1.100.0-nightly`, and SHA-256 `2e5bc688...f921`. The `nightly-2026-08-28` rustup `llc` hashes to `2e5bc688625ca7d9...` and is byte-identical to the cataloged binary.

The Nix shell supplies the same nightly and LLVM version. Its `llc` symlinks into `/nix/store/...-llvm-tools-preview-1.100.0-nightly-2026-08-28-x86_64-unknown-linux-gnu/...` and reports `LLVM version 23.1.0-rust-1.100.0-nightly`. Nixpkgs rewrites its interpreter to `/nix/store/fjkx1l5cnskzrqacf08z7i8z17256w0j-glibc-2.42-61/lib/ld-linux-x86-64.so.2` and adds an RPATH into `/nix/store/...`, producing `ca7af83e...` instead.

The compiler identity matches, but the bytes cannot. Every `nix develop` shell therefore fails this validation, and neither a lock nor toolchain bump can align the hashes. The failure is only reachable after the flake fix because the current `main` shell does not start. The diagnostic also reserves `--llc` for comparison probes, so another binary is not the intended escape.

[...]

PR #1261, opened 2026-09-13
fix(intrinsics): validate authored PTX floors against hardware introductions

### LINKED ISSUE #1213: Derive intrinsic PTX floors through the existing target vocabulary owner

`minimum_ptx` is written in both the intrinsic overlay and Rust recipe. Validation only compares the copies ([validation](https://github.com/NVlabs/cuda-oxide/blob/5944e11/crates/cuda-intrinsics-gen/src/resolve/families/execution_control.rs#L573-L594)). Nothing downstream checks it. `ptx_isa_requirement_for_floor` returns `Default` for every encoded floor `<= 60` ([codegen](https://github.com/NVlabs/cuda-oxide/blob/5944e11/crates/cuda-oxide-codegen/src/target/generated_requirements.rs#L97-L112)), so a wrong floor through PTX 6.0 changes only the generated reference.

The [#1138 review](https://github.com/NVlabs/cuda-oxide/pull/1138#issuecomment-5468085826) named this shape:

> The recorded libNVVM evidence was a hand-built fixture rather than real lowering output, which is why nothing caught it.

Where `minimum_sm` is a hardware requirement, the effective floor can be derived instead of copied.

**Proposed API / behavior**

```text
AS-IS                              TO-BE

recipe  "1.0" --.                  minimum_sm
overlay "1.0" --+ ensure!(equal)       |
                |                      v
                | no external      TargetPtxFloor   (in-tree)
                | floor input          |
                v                      v
          catalog.json            max(target default,
             /      \                 instruction requirement)
    encoded<=60   verbatim             |
        |            |                 v
     Default     "1.0" in         catalog.json
 1.0 = ... = 6.0  reference
```

`TargetPtxFloor` / `RECORDED_PTX_FLOORS` already store the per-target floor ([cuda-target-spec](https://github.com/NVlabs/cuda-oxide/blob/5944e11/crates/cuda-target-spec/src/lib.rs#L154-L223)). [#1105](https://github.com/NVlabs/cuda-oxide/pull/1105) introduced them, and [#1184](https://github.com/NVlabs/cuda-oxide/pull/1184) routed target parsing and PTX spelling through them. Intrinsic `minimum_ptx` still does not use them.

```toml
# before
minimum_ptx = "1.0"
minimum_sm  = "sm_20"

# after — floor derived as PTX 2.0
minimum_sm  = "sm_20"
```

This rule applies only when `minimum_sm` is a hardware requirement. Legacy rows such as `sm_20 -> 2.0` would be recorded as pinned ISA history. Instructions needing more than their target floor would keep a written value.

**Alternatives considered**

Correcting each floor separately leaves the self-check unchanged. [#519](https://github.com/NVlabs/cuda-oxide/issues/519) fixed `trap`/`breakpoint`/`pmevent`; the 1.0 / 3.2 / `sm_20` shape remains in counted `bar.arrive` ([overlay](https://github.com/NVlabs/cuda-oxide/blob/5944e11/intrinsics/overlay/counted_barrier.toml#L150-L159)), from merged [#721](https://github.com/NVlabs/cuda-oxide/pull/721).

**Additional context**

Of 130 explicit overlay routes, 104 meet or exceed the target floor. The 26 below are `lib_nvvm` at `sm_75`: 16 at PTX 6.0, 7 at 4.2, 2 at 1.0, and 1 at 6.2. Here `sm_75` records evidence collection, not instruction hardware. The #519 thread explains:

> backend-codegen evidence must sit exactly at the hardware floor and CUDA 13 cannot probe older targets; the PTX floor alone was the blocker.

`minimum_sm` has two route meanings. That distinction must be resolved before or with floor derivation; no fix is proposed here.

Does anything depend on the per-entry declaration that I have not seen? Should this cover merged declarations or new entries only? [#1189](https://github.com/NVlabs/cuda-oxide/issues/1189) records a second consumer left outside the same owner.

#1212 opened this as a PR before any discussion and was closed for that; this issue is that discussion.

almightychang:
Resolves #1213 

Counted `bar.sync` and `bar.arrive` declare PTX 1.0 even though their thread-count operands require PTX 2.0 and sm_20. The overlay and execution-control recipe agree on the wrong value, so their equality check accepts it. Correct both declarations and regenerate the catalog, reference, and managed outputs.

Add a lower-bound check on top-level authored PTX floors using the PTX version that first named the hardware SM. The authored value remains necessary for LLVM predicate validation, evidence validation, and codegen fallback; instructions added later to an existing SM can require more than its introduction version.

```text
AS-IS

  counted bar.sync / bar.arrive

  Overlay: PTX 1.0 -------- equality -------- Recipe: PTX 1.0
                              |
                            ACCEPT
                              |
                    Catalog / reference: 1.0

  ISA: PTX 2.0 / sm_20 --------X-------- missing correction


TO-BE

  Known defect: overlay + recipe -> PTX 2.0 / sm_20 -> regenerate

  Top-level entry
       |
       +-- authored minimum_ptx ------------------------ A
       |                                                 ^
       +-- native_minimum_sm, otherwise minimum_sm       |
                            |                            | <=
                            v                            |
                    ARCH_INTRODUCTION_PTX --------------- J

                    Require J <= A; preserve A

       sm_80 -> J = 7.0, A = 1.0 -> REJECT
       sm_80 -> J = 7.0, A = 7.8 -> ACCEPT lower-bound check

  Backend-profile route floors remain outside this comparison.
```

The introduction table is separate from `RECORDED_PTX_FLOORS`, which records LLVM defaults. Its 13 rows are the hardware floors currently consumed by this check: sm_70/75/80/86/89/90/100 and historical sm_11/20/30/53/60/61. The historical rows matter even though cuda-oxide emits for newer GPUs: packed f16x2, for example, uses native sm_53 rather than the project's admitted sm_70 floor; packed atomic f16x2 likewise uses native sm_60. An unrecorded hardware floor produces a diagnostic instead of a guessed fallback.

The counted `barrier.sync` and `barrier.arrive` forms retain PTX 6.0 / sm_30. Backend route floors, schema, and toolchain evidence remain unchanged. The new check alone would not find the original counted `bar.*` defect because those entries had no `minimum_sm`; the explicit correction follows the [ISA notes](https://docs.nvidia.com/cuda/parallel-thread-execution/#parallel-synchronization-and-communication-instructions-bar-barrier).

Validated in `nix develop` on main `6abfaa091`:

- `cargo test -p cuda-target-spec -p cuda-intrinsics-gen`: 324 unit tests and two CLI integration tests passed. The negative test changes `redux_sync_add` on sm_80 from PTX 7.0 to 1.0 and asserts exit code 1 plus `minimum_ptx 1.0 is below PTX 7.0, which first named hardware floor sm_80`. The control runs `check` on the current catalog, including native sm_53 packed operations and unchanged libNVVM profile floors.
- `check-abi-history --base-ref origin/main`: passed.
- Existing `probe --all --skip-terminal --per-target`, using `llc` from the Nix Rust sysroot: 2,374 target configurations across 1,025 intrinsic routes passed. This was comparison lowering with the same recorded LLVM version and a different binary SHA; terminal assembly and recorded evidence reproduction were not performed.

The recipe-floor regression tests explicitly accept the introduction bound
before asserting the original recipe rejection. Their SM-only mutations
preserve the authored PTX, so the new check cannot mask the stricter recipe
validation. Revalidated with `nix develop -c cargo test -p cuda-intrinsics-gen`:
315 unit tests and both CLI integration tests passed.

Regeneration updates 1,312 managed outputs; 1,307 change only their catalog SHA-256. The five substantive generated outputs are the catalog, reference, raw ABI documentation, and execution-control target records/tests.

Addresses #1213. This is a preparation draft, not ready for review or merge. Per the sequencing requested in #1213, review and merge should follow #1189 / #1256 (#1211 has landed). Once #1256 lands, this branch will be updated to the resulting main and regenerated/revalidated before being marked ready.


nihalpasham (approved):
I checked the authored floors separately from LLVM defaults and backend-profile evidence. The counted-barrier correction and the native packed-operation floors now agree with that distinction.

All 326 generator/target tests, ABI-history checks, strict Clippy and docs pass on the current-main rebase. All seven hosted workflows passed at `85e0d080`; this should follow #1256 as agreed in the issue.

This addresses #1213.

PR #1262, opened 2026-09-13
fix(rustlantis): drop the removed box_patterns feature gate

sachinsharma3191:
## What this fixes

The Rustlantis generator builds again on the pinned `nightly-2026-08-28`. That nightly removed `box_patterns`; the vendored MIR crate still enabled it even though it contains no box patterns.

```text
before: cargo build -p generate -> E0557: feature has been removed
after:  generator builds; mir_generator.py emits the seed-1 case
```

## What changed

Removed the unused feature gate. I rebased and signed the contributor commit, preserving its author and sign-off.

## Verification

Checked at `eea15821`:

- Reinstating the gate reproduces E0557; removing it restores the build.
- All 24 MIR/generator tests pass, formatting is clean, and the real Python adapter generates seed 1.

Strict Clippy in this vendored workspace still stops in unchanged `abi/src/size.rs` on `manual_div_ceil`. This change does not resolve that existing lint debt.

Related to the toolchain update in #1201. #1263 shares this prerequisite commit.


nihalpasham (commented):
I kept the one-line build fix and rebased and signed the contributor commit.

Restoring the removed feature gate reproduces E0557 on the pinned nightly. With it removed, all 24 MIR/generator tests pass and the Python adapter generates seed 1. Strict Clippy still stops on the existing `abi/src/size.rs` lint.

This restores the generator after the toolchain update in #1201 and supplies the prerequisite for #1263.

Checked at `eea15821`. Hosted CI also passes on this head.

nihalpasham (approved):

PR #1263, opened 2026-09-13
fix(rustlantis): validate place projections against LocalDecls

### LINKED ISSUE #1160: rustlantis: validate projection chains when constructing places

sachinsharma3191:
## What this adds

Rustlantis now checks projection chains when it constructs a place. An invalid array index or field path is reported by the generator, before it can become a misleading backend compilation failure.

```text
before: arr[index_i32] becomes custom MIR and fails in rustc
after:  place construction reports NonUsizeIndexLocal
```

This is defensive hardening. No existing generated seed has been shown to require it.

## What changed

- Construction uses the current body's `LocalDecls`, including the existence and `usize` type of index locals.
- One type-projection implementation checks pointers, tuple/struct fields, array constant indices and enum variant fields.
- Failed extensions leave the original place intact. Generator graph paths use the checked constructor.
- I rebased and signed the contributor commits, preserving authorship and sign-offs. The first commit is shared with #1262.

## Verification

Checked at `21d460fe`:

- All 38 tests in the affected Rustlantis workspace packages pass; formatting is clean.
- Disabling just index-local validation makes both real-generator-path rejection tests fail. The valid `usize` control passes with validation enabled.
- Generated output is byte-identical before/after for seeds 0–199 under both the tiny adapter config and the full example config.

The workspace test run excludes `difftest`, whose external compiler/Miri configurations are separate. Existing vendored Clippy and rustdoc warnings remain; strict Clippy stops in unchanged `abi/src/size.rs`.

Closes #1160 as defensive construction validation. Depends on #1262.


nihalpasham (commented):
I kept the validation against the body's actual local declarations and rebased and signed the contributor commits.

The projection rules have one owner, and a failed extension leaves the original place unchanged. All 38 affected workspace tests pass; disabling index-local validation makes both generator-path regressions fail. Seeds 0–199 also stay byte-identical under both configs. Existing vendored lint warnings remain.

This addresses #1160 as defensive validation: it catches malformed generator state earlier, without claiming an existing generated miscompile. #1262 is the build prerequisite.

Checked at `21d460fe`. Hosted CI also passes on this head.

nihalpasham (approved):

PR #1264, opened 2026-09-13
fix(mir-lower): lower redux.sync results to signless integers

sachinsharma3191:
## What this fixes

Device helpers can return integer `redux.sync` results directly. The lowerer previously kept the result's Rust signedness after converting the function signature to LLVM's signless integer type.

```text
before: fn reduce(v: u32) -> u32 { redux_sync_add(FULL_MASK, v) }
        -> return-type verification failure
after:  the helper compiles and returns the reduction
```

## What we fixed

- Convert the reduction result through the existing type converter, shared by all 16 reduction operations.
- Keep the real helper-return shape in `redux_minmax`.
- I extended the lowering regression to all 16 operations and checked the exact intrinsic, value/mask order, integer result types and unchanged float types.
- I rebased and signed the contributor commit, preserving its author and sign-off.

## Verification

Checked at `5a3653e3`:

- All 293 unit and 136 lowering tests pass, along with formatting, strict Clippy for the crate and example, and rustdoc with warnings denied.
- Restoring the unconverted result makes the regression fail at return-type verification.
- `redux_minmax` passes on RTX 5090, SM120, through LLVM NVPTX and modern libNVVM. Helper results are `-16` and `4294967295`; memcheck and synccheck both report zero errors.
- Generated PTX contains the helper's signed-min and unsigned-max reduction instructions.

This removes the reported lowering blocker in #811. The architecture selection and `warp_reduce` API work remain separate; #811 remains open for that work. Float operations were checked in lowering, not on hardware.

## Known backend limit

Legacy libNVVM still emits unresolved `llvm.nvvm.redux.sync.*` calls. The existing direct-reduction kernels fail the same CUDA 13.0 PTX assembly check after removing the new helpers; this is an existing intrinsic-support gap, separate from the return-type fix. LLVM NVPTX and modern libNVVM execute correctly.


nihalpasham (commented):
I kept the shared result-type conversion and added a few checks around it:

- Covered all 16 reduction operations, including unchanged float results.
- Checked the exact intrinsic and value/mask order so a different call cannot satisfy the regression.
- Rebasing and signing preserved the contributor's original commit and sign-off.

All 429 MIR-lower tests, formatting, strict Clippy and docs pass. `redux_minmax` also passes on RTX 5090 with zero memcheck/synccheck errors, including both helper-returned results.

This removes the return-type blocker reported in #811; architecture selection and `warp_reduce` remain separate. Modern libNVVM also runs correctly. Legacy libNVVM retains its existing unresolved-redux-intrinsic failure, reproduced with the original direct kernels.

Checked at `5a3653e3`. Hosted CI also passes on this head.

nihalpasham (approved):

Issue #1234, opened 2026-09-04
`DeviceAtomicF32::fetch_add` expands to a CAS loop instead of `atom.global.add.f32`


mxk:
**Description**

`DeviceAtomicF32::fetch_add` expands to a CAS loop instead of the native `atom.global.add.f32` instruction. `DeviceAtomicU32::fetch_add` is unaffected. The problem was introduced in #1201.

**Minimal reproducer**

```rust
use cuda_device::atomic::{AtomicOrdering, DeviceAtomicF32};
use cuda_device::{kernel, launch_bounds};
use cuda_host::cuda_module;

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    #[launch_bounds(256)]
    pub fn add_f32(total: &[f32]) {
        let slot: *const f32 = &raw const total[0];
        // SAFETY: one live `f32`, reached only through this atomic.
        let total = unsafe { DeviceAtomicF32::from_ptr(slot.cast_mut()) };
        total.fetch_add(1.0, AtomicOrdering::Relaxed);
    }
}

fn main() {}
```

`Cargo.toml` dependencies:

```toml
cuda-device = { git = "https://github.com/NVlabs/cuda-oxide.git", rev = "8ac82ec761e950594db5f47322330b231f3d80be" }
cuda-host = { git = "https://github.com/NVlabs/cuda-oxide.git", rev = "8ac82ec761e950594db5f47322330b231f3d80be" }
cuda-core = "0.3.1"
```

```
cargo oxide inspect
```

**Expected behavior**

One native instruction as on `224ae8de` (nightly-2026-04-03):

```ptx
// %bb.1:                               // %bb2
	atom.global.add.f32 	%r1, [%rd1], 0f3F800000;
	ret;
```

**Actual behavior**

A CAS loop from `fe616639` (nightly-2026-08-28) onward:

```ptx
// %bb.1:                               // %bb2
	ld.relaxed.gpu.global.b32 	%r3, [%rd1];
$L__BB0_2:                              // %atomicrmw.start
                                        // =>This Inner Loop Header: Depth=1
	add.f32 	%r2, %r3, 0f3F800000;
	atom.relaxed.gpu.global.cas.b32 	%r1, [%rd1], %r3, %r2;
	setp.ne.b32 	%p2, %r1, %r3;
	mov.b32 	%r3, %r1;
	@%p2 bra 	$L__BB0_2;
```

**Environment**

[...]

PR #1267, opened 2026-09-14
docs(atomic): fix stale DeviceAtomicF32 fetch_add claims

thearyanahmed:
## Summary

Refs #1234

Global-scope `f32 fetch_add` lowers to a CAS loop on LLVM 23, not the native `atom.global.add.f32` these three spots still claim: `DeviceAtomicF32`'s doc comment, the atomics README test table, and its op-support matrix. Root
cause is an upstream NVPTX regression, not cuda-oxide's codegen (see #1234 ). **This PR only fixes the docs**.

## Changes

Only docblock. 

## Testing

1. grepped for other `atom.add.f32` references (none).
2. `cargo fmt -p cuda-device --check` 
3. `cargo doc -p cuda-device -D warnings`

## Checklist


vyncint:
Those are the three, and your grep matches mine — when I filed #1234 I swept the tree for the claim in any form and found `crates/cuda-device/src/atomic.rs:491`, `examples/atomics/README.md:82` and `:193`, and nothing else.

Two things from that investigation that may save you a round of review questions:

- **Docs-only is the right scope, and not just conservatism.** There is no IR spelling that gets the instruction back on LLVM 23 — `atomicrmw fadd` at every scope and ordering, generic or `addrspace(1)`, and `llvm.nvvm.atomic.load.add.f32.p1` all lower to `atom.relaxed.gpu.global.cas.b32`. Anything that restores it today has to be inline PTX, which is a memory-model decision rather than a mechanical fix.
- **It is specifically global/generic `f32`.** Shared `f32` still lowers to `atom.relaxed.gpu.shared.add.f32`, and global `f64` and `i32` are both native. So if any wording near those spots generalizes to "f32 atomics" or "atom.add.f32/f64", the shared and f64 halves are still true and only the global-f32 claim needs to move.

Whether the CAS loop is correct is not in question — it is, just slower.

nihalpasham:
Thanks, @thearyanahmed — the stale native-instruction claims needed correcting.

I completed the follow-up in #1319 because this fork disables maintainer edits. I kept your commits and DCO credit, scoped the note to our pinned LLVM NVPTX defaults, and explained why subnormal handling matters: the native global add flushes those values, while the default CAS loop preserves them.

Warning-denied rustdoc and the independent PTX controls pass; hosted checks are running on the signed continuation. Both PRs remain open for the final decision.

This partially addresses #1234; it does not close the code-generation issue.

nihalpasham:
Thanks, @thearyanahmed — your documentation correction has landed in #1319, including the follow-ups on the pinned compiler and subnormal handling. Your contributor commits and DCO credit are retained in that continuation.

Closing this PR as superseded by the merged change. The remaining code-generation work is still tracked in #1234.

PR #1268, opened 2026-09-14
fix(cargo-oxide): warn when build uses backend default arch

### LINKED ISSUE #1266: cargo oxide build defaults to sm_80 while cargo oxide run auto-detects the local GPU
## Summary

`cargo oxide build` compiles for `sm_80` by default, while `cargo oxide run` auto-detects the local GPU's actual architecture. Building a binary (rather than running via `cargo oxide run`) silently ships Ampere-targeted code even on newer hardware, with no warning.

## Repro

On a Blackwell box (GB10, sm_121):

```bash
cargo oxide build -q
strings target/release/<bin> | grep '^\.target'
# -> .target sm_80
```

```bash
cargo oxide build --arch sm_121a -q
strings target/release/<bin> | grep '^\.target'
# -> .target sm_121a
```

`cargo oxide run` (no explicit `--arch`) correctly detects and targets sm_121a on the same machine — only `build` falls back to sm_80.

## Why this matters

For anyone building a release binary rather than iterating with `cargo oxide run`, this is an easy silent trap: the build succeeds, the binary runs (compatible PTX), but it's not using the actual device's ISA/features, and there's no diagnostic pointing at this. `cargo oxide doctor` would be a natural place to warn when a build's target arch doesn't match the detected local GPU.

## Suggested fix

Either:
- Make `cargo oxide build`'s architecture detection match `cargo oxide run`'s (auto-detect the local GPU when `--arch` isn't given), or
- If a fixed default is intentional for reproducible/portable builds, have `cargo oxide doctor` (or the build output itself) warn when the resolved target differs from the local GPU's compute capability.

## Environment

- DGX Spark (GB10, sm_121), CUDA 13.0, driver 580.173.02
- cuda-oxide 0.2.1

YunusMutlu:
## What this adds

`cargo oxide build` warns when no architecture is configured and the first GPU reported by `nvidia-smi` can be queried. The warning explains automatic backend selection and points to `--arch`.

```text
build with no architecture -> warning; ordinary example still targets sm_80
build --arch sm_120        -> no warning; example targets sm_120
```

## Implementation

Both regular and Cargo passthrough builds use the existing architecture probe. CLI, environment and project configuration suppress the probe when they already choose a target. The diagnostic does not set target environment variables or change the build fingerprint.

I rebased and signed the contributor commit on current main, preserving its author and sign-off.

## Verification

Checked at `e02a086b` against main `b631e5d7`:

- All 231 cargo-oxide tests, formatting, strict Clippy and warnings-denied rustdoc pass.
- Real `array_index` builds with two-GPU-shaped probe output retain `.target sm_80` without an override and emit `.target sm_120` with an explicit target.
- The ordinary and Cargo passthrough routes emit the warning; an explicit architecture suppresses it.

## Caveats

`nvidia-smi` enumeration does not identify the active CUDA device, so the message names the first reported GPU and gives generic target-selection guidance. GPU-less builds stay quiet. Backend feature requirements can still raise the selected architecture.

Closes #1266. The separate doctor diagnostic in #1278 is related; this PR covers the build command itself.


nihalpasham (commented):
The warning preserves the build command's target-selection behavior and correctly describes the first GPU reported by `nvidia-smi`.

I rebased and signed the commit on current main. At `e02a086b`, all 231 cargo-oxide tests and strict checks pass. Real default, explicit-target and Cargo passthrough builds confirm the warning behavior: the ordinary example stays at `sm_80` until `--arch sm_120` is supplied.

Hosted CI also passes on this head. This closes #1266 once merged; #1278's doctor diagnostic is related.

nihalpasham (approved):

PR #1271, opened 2026-09-15
docs: clarify kernel reference comments

nihalpasham:
## What this changes

Clarify three comments added by #1244:

- Describe the slice packet's data pointer and its canonical empty-slice value directly.
- Say that the exporter emits LLVM attributes after validating the reference facts.

## Validation

- `cargo fmt --all -- --check` and `git diff --check` passed.
- An exact comparison confirmed three comment substitutions across two files; every other byte is unchanged.


nihalpasham:
I checked the three wording changes against the launch-packet helper and LLVM exporter. “Pointer value” and “emitting attributes” describe what those paths do more directly, and the `DisjointSlice` comment still makes clear that its packet helper does not establish reference-validity attributes.

The patch changes only those comments. Formatting and diff checks pass, integration with current main is clean, and all 53 hosted checks passed. The PR remains a draft.

This is related to #1244; it does not close #1098.

PR #1273, opened 2026-09-15
refactor(llvm): use upstream volatile load/store support

### LINKED ISSUE #1272: llvm: replace local volatile load/store shim with upstream Pliron support
## Problem

CUDA Oxide currently carries a local `cuda_oxide_op_volatile` attribute and helper functions to preserve volatile load/store semantics through LLVM lowering and export.

Pliron now provides first-class volatile support for LLVM `LoadOp` and `StoreOp` through `VolatilityOpInterface`, so the local compatibility shim is no longer necessary.

This cleanup follows #1247, which moved CUDA Oxide onto the canonical merged Pliron GEP baseline.

## Scope

* Update Pliron to the revision containing upstream volatile load/store support.
* Replace the local volatile attribute with `VolatilityOpInterface`.
* Preserve MIR volatile semantics through lowering and LLVM export.
* Migrate volatile-specific tests to the upstream interface.
* Remove the local volatile helper functions and attribute key.

MIR-level volatile attributes should remain unchanged because they encode MIR semantics independently of the LLVM representation.

## Acceptance criteria

* No remaining uses of `cuda_oxide_op_volatile`, `OP_VOLATILE_KEY`, `set_op_volatile`, or `op_volatile`.
* Volatile MIR loads/stores lower to LLVM operations carrying the upstream volatile attribute.
* LLVM export emits `volatile` based on `VolatilityOpInterface`.
* Existing volatile tests are migrated to the upstream interface.
* `volatile_memory` compile-only smoketest passes.
* Formatting, workspace check, Clippy, and `git diff --check` pass.

uurl:
## What this changes

Volatile LLVM loads and stores now use Pliron's upstream `VolatilityOpInterface`, removing cuda-oxide's duplicate LLVM attribute and helpers.

```text
before: MIR volatile flag -> private LLVM attribute -> volatile load/store
after:  MIR volatile flag -> upstream interface     -> volatile load/store
```

The MIR representation and generated memory operations retain their existing semantics.

## What we added

- Synchronized both Cargo workspaces on Pliron `edd41fe`, which contains volatile support and is also used by #1275. The two changes combine without a pin conflict.
- Added absent, true and explicitly cleared false controls for both loads and stores through modern and legacy LLVM exporters. Each test verifies the dialect and requires exactly one memory operation.
- Rebased onto current main and signed both commits while preserving contributor authorship and sign-off.

## Verification

Checked at `1933b81f` against `b631e5d7`:

- 5,353 workspace/backend tests and 33 doctests pass; formatting, strict Clippy and warning-free docs pass.
- `volatile_memory` builds; LLVM and PTX retain its two volatile reads and two volatile writes.
- RTX 5090 memcheck verifies 128 round-trip values with zero errors.
- Independent review checked the exact upstream dependency changes and the boolean value semantics.

Closes #1272.


nihalpasham (commented):
I kept the upstream volatile interface and added the missing negative controls:

- Both loads and stores now cover an absent flag, `true`, and `true` reset to `false`, through modern and legacy export.
- The tests require one memory operation and verify that an explicitly false attribute stays non-volatile.
- I aligned the Pliron pin with #1275, rebased and signed the commits while keeping your authorship and sign-off.

At `1933b81f`, workspace/backend tests, doctests, strict Clippy, formatting and docs pass. `volatile_memory` retains the expected LLVM/PTX operations and passes RTX 5090 memcheck for 128 values.

This addresses #1272. Hosted CI also passes on this head.

nihalpasham (approved):

PR #1275, opened 2026-09-15
refactor(llvm): use upstream global constant semantics

### LINKED ISSUE #1274: llvm: use upstream GlobalOp constant semantics
## Summary

CUDA Oxide currently represents LLVM global immutability with a private `cuda_oxide_global_immutable` attribute attached through `GlobalOpExt`.

Pliron now provides first-class support for LLVM global constants through `GlobalOp::set_constant()` and `GlobalOp::is_constant()`.

CUDA Oxide should use the upstream representation instead of maintaining the local LLVM-side marker.

## Changes

* update the Pliron revision to one containing upstream global constant support;
* lower MIR immutable globals to `llvm::GlobalOp::set_constant()`;
* make the textual LLVM exporter use `GlobalOp::is_constant()`;
* remove the private `cuda_oxide_global_immutable` attribute and its `GlobalOpExt` helpers;
* update the affected lowering and exporter tests.

## Scope

The MIR-level `MirGlobalAllocOp` immutability attribute remains unchanged.

That attribute represents Rust/MIR semantics and is still used when deciding whether a promoted global may be treated as never-written. This issue only replaces the duplicate LLVM-dialect representation with the upstream Pliron one.

The observable LLVM output should remain unchanged: compiler-promoted immutable globals continue to export as `constant`, while ordinary mutable globals continue to export as `global`.

uurl:
## What this changes

Promoted immutable storage now uses Pliron's upstream `GlobalOp` constant property. This removes the private LLVM immutability marker and its duplicate helpers.

```text
promoted immutable table -> LLVM constant
host-writable symbol    -> LLVM global
```

MIR still owns the decision that storage is never written. The upstream property carries that decision through LLVM export.

## What we added

- Added a true-to-false reset control alongside the unmarked writable global, through modern and legacy exporters. A present false property must not make storage constant.
- Rebased onto current main and signed both commits, preserving contributor authorship and sign-off. Its Pliron `edd41fe` pin now agrees with the repaired #1273 branch.

## Verification

Checked at `699a312e` against `b631e5d7`:

- 5,353 workspace/backend tests and 33 doctests pass; formatting, strict Clippy and warning-free docs pass.
- `const_table_lookup` builds with promoted AS1 tables still `constant` and host-writable AS4 symbols still `global`.
- All seven example kernels are bit-correct on RTX 5090; memcheck reports zero errors.
- Independent review checked the dependency diff and confirmed that immutability remains an explicit storage property.

Closes #1274.


nihalpasham (commented):
Using the upstream constant property removes the duplicate LLVM marker while keeping the MIR storage guarantee in one place.

I added a reset-to-false control through both LLVM exporters: a global whose constant property was cleared must stay writable, just like an unmarked global. I also rebased and signed the commits, preserving your authorship and sign-off; the Pliron pin agrees with the repaired #1273 branch.

At `699a312e`, workspace/backend tests, doctests, strict Clippy, formatting and docs pass. All seven `const_table_lookup` kernels are bit-correct on RTX 5090 with a clean memcheck run.

This addresses #1274. Hosted CI also passes on this head.

nihalpasham (approved):

Issue #1265, opened 2026-09-14
cuda-bindings picks the highest-versioned /usr/local/cuda-13.* dir instead of the complete one, ignoring which is actually usable


bernardogv:
## Summary

`cuda-bindings`' host-side toolkit discovery picks the *highest-versioned* `/usr/local/cuda-13.*` directory rather than honoring `CUDA_TOOLKIT_PATH`/`CUDA_HOME`, so a partial toolkit installed for an unrelated purpose silently wins over a complete one and breaks the build.

## Repro

My machine has two CUDA 13 installs under `/usr/local`:

- `/usr/local/cuda-13.0` — full toolkit, includes `curand.h`
- `/usr/local/cuda-13.3` — a *partial* toolkit pulled in by an unrelated apt package (nvcc, cudart, cccl, crt only — no full `lib64`, no `curand.h`)

`update-alternatives` ranks `cuda-13.3` higher by priority, and cuda-oxide's build resolves the toolkit to that dir regardless of `CUDA_TOOLKIT_PATH`/`CUDA_HOME`:

```
$ ls /usr/local/cuda-13.3/include/curand.h
ls: cannot access '/usr/local/cuda-13.3/include/curand.h': No such file or directory
$ ls /usr/local/cuda-13.0/include/curand.h
/usr/local/cuda-13.0/include/curand.h

$ cargo oxide build
...
curand.h not found
```

Explicitly exporting both env vars works around it:

```bash
export CUDA_TOOLKIT_PATH=/usr/local/cuda-13.0
export CUDA_HOME=/usr/local/cuda-13.0
```

but without that, the build fails with a confusing missing-header error that gives no hint it picked the wrong of two present toolkits.

## Why this is worth fixing over just documenting the workaround

This isn't a case of the env vars being ignored outright (that's #93/#16, already closed) — they *are* honored when set. The bug is the fallback when they're *not* set: it should probably prefer the toolkit with the more complete/expected layout (or at least fail with a clearer "found N candidate toolkits, none has curand.h" message) rather than silently taking whichever versioned dir sorts highest.

## Environment

- DGX Spark (GB10), Ubuntu 24.04, CUDA 13.0 (full) + CUDA 13.3 (partial, from a tileiras-only package), driver 580.173.02
- cuda-oxide 0.2.1

vyncint:
[...]

PR #1276, opened 2026-09-15
fix(doctor): resolve the CUDA toolkit the way the build script does

vyncint:
Step one of #1265, and the part that is ours regardless of how the other two questions there are answered.

`cargo oxide doctor` keeps a hand-written mirror of the toolkit discovery in the shared `cuda-bindings` build script — its own comment says *"Mirrors BY HAND … If that discovery changes, mirror it here."* The mirror had drifted to "first set variable among `CUDA_TOOLKIT_PATH`/`CUDA_HOME`, else `/usr/local/cuda`". I read the pinned `cuda-bindings 0.3.1` build script out of the cargo registry rather than infer from the symptom; it is wrong in three ways.

| | build script (`resolve_cuda_toolkit` / `find_default_cuda_toolkit` / `validate_cuda_toolkit`) | old mirror |
|---|---|---|
| neither variable set | fixed list: `/usr/local/cuda-13.3` → `-13.2` → `-13` → `/usr/local/cuda`, first that validates | `/usr/local/cuda` |
| accepting a root | must hold a `cuda.h` whose `CUDA_VERSION` ≥ 13.0 | no check at all |
| variable set to `""` | an error | stepped over, fall through |

The middle row is the one that makes doctor wrong rather than merely incomplete: an explicit `CUDA_HOME` pointing at a CUDA 12 toolkit reported `✓ cuda.h` for a build that rejects it on the floor.

## Why this is the bug behind #1265

The reporter has a complete 13.0 and a partial 13.3 installed. Two details the issue did not have, both from the build script:

- **`/usr/local/cuda-13.0` is not a candidate at all.** It is not in that list, so the complete toolkit is only reachable through the `/usr/local/cuda` symlink — which `update-alternatives` points at 13.3. No ordering makes it win on its own name.
- **The partial 13.3 passes validation**, because validation wants `cuda.h` and a version, and it has both. But `wrapper.h` includes *both* `<cuda.h>` and `<curand.h>`, and only the first is part of the acceptance test — which is why the failure arrives late, from inside `cuda-bindings`, naming a header instead of a toolkit choice.

So a machine with two toolkits gets a green doctor and a failing build, which is the one thing a diagnostic must not do.

## What changed

`resolve_cuda_toolkit` now implements the real rule and returns how the root was chosen plus what was turned down on the way. Doctor prints the rejections, because that is the part a user needs — a root can be skipped for a missing header or for its version, and from outside two installed toolkits look identical:

```
CUDA headers (cuda.h)... ✓ /opt/ctk/include/cuda.h (via CUDA_TOOLKIT_PATH)

CUDA headers (cuda.h)... ✗ no usable CUDA toolkit; nearest root probed was `/opt/old`
  Probed for cuda.h:
    /opt/old/include/cuda.h
    /opt/old/targets/x86_64-linux/include/cuda.h
  Turned down, in order:
    CUDA_TOOLKIT_PATH=/opt/old is invalid: CUDA toolkit 12.8 is too old; the host crates require CUDA 13.0+
```

`validate` is injected, so precedence and ordering are tested without a filesystem; the acceptance rule itself is tested against real files.

## Tests

`cargo test -p cargo-oxide` — 232 pass. Five cover this directly:

- variable precedence, `CUDA_TOOLKIT_PATH` before `CUDA_HOME`
- a set-but-invalid variable is answered, **not** stepped over — the case that would otherwise read a typo in `CUDA_HOME` as a healthy default install
- an empty value is its own error
- discovery stops at the first candidate that validates, reporting the ones ahead of it in order — asserted by index into the candidate list, so the test is about the order rather than the literals
- acceptance against real files: 13.3 accepted, 12.8 rejected as too old, `cuda.h` removed, root not a directory

Exercised end to end against synthetic roots (complete / too-old / header-less / empty variable); all four report correctly.

## Also corrected, same rule stated elsewhere

- `cuda-oxide-book/getting-started/installation.md` asserted the `/usr/local/cuda` default as fact. It now gives the real order and says plainly that with two toolkits installed the order decides — not the highest version, and not the symlink, which is last.
- A codegen test's doc comment claimed to mirror this contract while implementing a simpler `bin/ptxas` probe. Finding a binary needs no header and no floor, so it now says that instead of claiming a mirror it does not implement.

fmt, clippy `-D warnings` on the touched crate, all 17 `scripts/check-*.sh`, and `sphinx-build -W` on the book all pass. No GPU claim here: nothing in this change reaches device code.

Still open on #1265, and deliberately not in this PR: whether doctor should report the full resolution in its normal output, and whether the mirror should be guarded against the pinned `cuda-bindings = "0.3.1"` so a version bump fails until someone re-reads the build script.


nihalpasham (approved):
I kept the discovery rules from your change and added two follow-ups so doctor agrees with the pinned build script:

- Explicit toolkit paths retain their original bytes, including non-Unicode paths.
- `CUDA_TOOLKIT_TARGET_DIR` selects only that target tree; a healthy top-level header cannot hide a missing or outdated override.

I rebased the branch and added filesystem regressions. All 237 cargo-oxide tests, formatting, strict Clippy and rustdoc pass at `c8be8d2d`; all seven hosted workflows passed on this final head.

This partially addresses #1265. The dependency's incomplete-toolkit/`curand.h` behavior remains separate.

PR #1278, opened 2026-09-16
Doctor warns when resolved target is different

thearyanahmed:
## Testing

- [ ] `just check` passes (the local mirror of CI: fmt, clippy, tests, guards, docs)
- [ ] `cargo oxide run <example>` passes, or `scripts/smoketest.sh -o '^<example>$'`

## Checklist

- [ ] All commits signed off (`git commit -s`)

---


nihalpasham (changes requested):
I wouldn't land this version. #1268 already addresses #1266 at the build command, and the extra doctor advice disagrees with target selection: `run` keeps an explicit configured architecture, while this message says it would choose the detected GPU. The same-capability message also promises that `sm_86a` builds and loads, although that suffix is not supported there.

I suggest leaving the existing build warning as the solution. A separate doctor report would need a concrete additional use and should describe the effective configuration without promising compatibility from capability equality alone.

This does not add another fix for #1266; that issue is already resolved.

PR #1279, opened 2026-09-16
Add generated debug info testing with cuda-gdb (basic sanity)

dpolyanitsa-nvidia:
## What this adds

Adds a local cuda-gdb testsuite contributed by the NVIDIA debug-tools team. It builds a Rust example with full debug information, runs debugger commands, and checks what the debugger reports.

```text
debug-tests Rust example
  -> build with CUDA_OXIDE_DEBUG=full
  -> cuda-gdb commands and output checks
  -> pytest pass/fail results
```

The Rust program lives in `crates/rustc-codegen-cuda/examples/debug-tests/`; the 13 Python tests live in `scripts/debug-tests/`.

## What it covers

- Primitive values, pointers, arrays, structs, tuples and enums.
- Global, shared and constant memory.
- Source/function breakpoints, stepping and frame navigation.
- Physical and inlined call frames.
- Assertions, traps, bounds checks, invalid memory access and GPU coredumps.

These tests complement the existing `scripts/debug-smoketest.sh`.

## Running the suite

Requires the repository's Rust build prerequisites, an NVIDIA GPU and driver, a CUDA toolkit containing cuda-gdb, and the Python packages in `scripts/debug-tests/requirements.txt`.

From the repository root, with those requirements installed:

```sh
CUDA_TOOLKIT_PATH=/path/to/cuda pytest scripts/debug-tests
```

Use `-s` for annotated debugger transcripts and `-k NAME` to select a test. `CUDA_OXIDE_CUDA_GDB` overrides the debugger executable; `CUDA_OXIDE_TARGET` overrides the GPU architecture.

<details>
<summary>Implementation</summary>

- The Rust example selects one scenario per invocation. Five scenarios run normally; the other five exercise deliberate faults or debugger stops.
- The session fixture builds the example with `CUDA_OXIDE_DEBUG=full`.
- Each pytest test records a sequence of commands and runs one cuda-gdb session.
- Output markers associate assertions with individual commands. Failed checks are annotated in the captured transcript.
- The coredump test generates a GPU dump and opens it for offline inspection.

</details>

## Verification

Checked at `6b5bd607`:

- Dmitry reported **9 passed, 4 failed** on an RTX 5070 with CUDA 13.3.
- Maintainer validation reproduced the same **9 passed, 4 failed** on an RTX 5090 with CUDA/cuda-gdb 13.3.
- ✅ All five normal example scenarios passed. Compute Sanitizer memcheck reported zero errors for `memory_spaces`.
- ✅ Focused Rust formatting, diff whitespace checks, Python compilation and collection of all 13 tests passed.
- ❌ Local strict Clippy fails on the trap fixture's unreachable assignment and unused binding.
- ❌ The example lock consistency check reports 20 dependency-version mismatches. The ordinary GPU smoketest also invokes the new example without its required scenario.
- Hosted workflows are awaiting approval; no hosted-CI pass is claimed.

The four failing debugger tests are:

```text
test_execution.py::test_values_named_and_inlined_frames
test_execution.py::test_mixed_physical_and_inlined_callstack
test_inspection.py::test_values_source_inspection_and_exit
test_inspection.py::test_memory_spaces_inspection
```

## Caveats

- The pytest suite is intentionally outside CI. As agreed, remaining debugger limitations may stay as documented failing tests; CI integration is a follow-up.
- Some current failures come from test assumptions about symbol spelling, selected frames or breakpoint formatting. The qualified global and kernel locals were readable in separate debugger probes.
- The `deep_middle` inline frame is visible, but its `scaled` local was unavailable in the tested configuration.
- The [review](https://github.com/NVlabs/cuda-oxide/pull/1279#pullrequestreview-5244464267) requests tighter value checks, debugger exit-status handling and the agreed known-failure note in the suite itself.
- The new Rust example is already discovered by existing example checks, even though pytest is outside CI.
- This PR adds tests; it does not change the compiler implementation.

Partially addresses #226.


crates/rustc-codegen-cuda/examples/debug-tests/Cargo.lock
```diff
@@ -0,0 +1,734 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 4
+
+[[package]]
+name = "aho-corasick"
+version = "1.1.5"
```
nihalpasham: Please align this Cargo.lock with the other examples. `scripts/sync-example-locks.sh --check` reports 20 version mismatches introduced here; for example, this pins `aho-corasick` to 1.1.5 while the other examples use 1.1.4.

The existing examples-compile workflow runs that check, even though the new pytest suite is outside CI. `scripts/sync-example-locks.sh --align crates/rustc-codegen-cuda/examples/debug-tests` is the existing way to align a new example; please rerun `--check` afterwards.

crates/rustc-codegen-cuda/examples/debug-tests/src/main.rs
```diff
@@ -0,0 +1,670 @@
...
+    // Generates CUDA_ERROR_ILLEGAL_INSTRUCTION;
+    // cuda-gdb sees $GDB_SIGNAL_CUDA_WARP_ILLEGAL_INSTRUCTION.
+    // -------------------------------------------------------------------------
+    #[kernel]
+    pub fn debuginfo_trap_fail(mut out: DisjointSlice<i32>) {
+        let idx = thread::index_1d();
+        if let Some(slot) = out.get_mut(idx) {
+            // <<<GDB BREAK TRAP>>> cuda-gdb catches WARP_ILLEGAL_INSTRUCTION here
+            debug::trap();
+            *slot = 0;
```
nihalpasham, line 355: I hit two warnings in strict Clippy here: `debug::trap()` never returns, so the assignment is unreachable and `slot` is unused.

If this shape is intentional for the debugger fixture, please add a narrowly scoped `#[allow(unreachable_code, unused_variables)]` with a short comment explaining why. The existing examples Clippy check uses `-D warnings`, so it currently fails on this example.

crates/rustc-codegen-cuda/examples/debug-tests/src/main.rs
```diff
@@ -0,0 +1,670 @@
...
+// =============================================================================
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let args: Vec<String> = std::env::args().collect();
+    let Some(kernel) = args.get(1).map(String::as_str) else {
+        eprintln!(
+            "usage: debug-tests <kernel>\n\nkernels: values, aggregates, loop, memory_spaces, \
+             deep_stack, assert_fail, trap_fail, breakpoint, oob_index, null_deref"
+        );
+        std::process::exit(2);
```
nihalpasham: The ordinary GPU smoketest discovers this example automatically and runs `cargo oxide run debug-tests` without a scenario. That reaches this branch and exits 2; I reproduced the same exit with the built binary.

Please give the example a safe default such as `values`, or have the smoketest explicitly pass that scenario. The intentional crash modes can stay explicit.

scripts/debug-tests/test_inspection.py
```diff
@@ -0,0 +1,173 @@
...
+        check.matches(r"^\$[0-9]+ = ", "TILE is printable")
+    with gdb("print GLOBAL_COUNTER") as check:
+        check.matches(r"^\$[0-9]+ = [0-9]+", "GLOBAL_COUNTER is accessible")
+    with gdb("print coeff_val") as check:
+        check.matches(r"^\$[0-9]+ = [0-9.]", "coeff_val is printable")
+    with gdb("print COEFF") as check:
+        check.not_matches(
+            r"No symbol .COEFF. in current context",
+            "COEFF is not accessible by name",
+        )
```
nihalpasham, line 200: Please make the inspection checks require a readable, correct value. I fed this test a transcript with `TILE = <optimized out>` and `print COEFF` returning `Cannot access memory at address 0x0`; the whole test still passed. The COEFF check only excludes one missing-symbol message.

The same issue appears elsewhere in this file: the `1.5` check accepts `1.59999`, the `20` check accepts `200`, and the tuple/enum checks accept optimized-out values. Please tighten these checks and add small harness regressions using those incorrect outputs. These were deliberately substituted outputs, not values produced by the GPU.

scripts/debug-tests/conftest.py
```diff
@@ -0,0 +1,490 @@
...
+            if command is None:
+                trailing_failures.append(message)
+            else:
+                inline_failures.setdefault(command.label, []).append(message)
+
+        if timed_out:
+            record_failure("cuda-gdb timed out")
+        for command in self.commands:
+            if not command.captured:
+                continue
```
nihalpasham, line 427: Please check cuda-gdb's exit status as well as its output. `returncode` is saved above but never used to fail the test. I ran a checked command successfully, then deliberately made cuda-gdb exit 17 in the final unchecked command; the harness reported success.

The debugger process needs to finish successfully even when the program under it deliberately triggers a CUDA exception. Please cover abnormal termination with a small harness regression, while still allowing the normal `quit` command used by the coredump test.

Please also verify that every required command ran, including commands without assertions; allow the intentional final `quit` to omit its trailing marker.

scripts/debug-tests/test_inspection.py
```diff
@@ -0,0 +1,173 @@
...
+    with gdb("run") as check:
+        check.matches(r"CUDA thread hit Breakpoint", "source breakpoint hits")
+        check.matches(
+            f"src/main\\.rs:{lines.values}",
+            f"stop is shown at line {lines.values}",
+        )
+
+    gdb("disable 1")
+    with gdb("info locals") as check:
+        check.matches(r"^x =", "scalar local x is visible")
```
nihalpasham, line 36: Please select and verify the kernel frame before inspecting its variables. In my run, this source breakpoint stopped in an inline `wrapping_add` frame, so `info locals` reported no locals. Selecting the existing `debuginfo_values` caller showed `x = 1`, `y = 1.5`, the array, pointer locals and kernel arguments.

The values are available here; the test is inspecting a different function. Select by the intended function rather than assuming it is always frame 0.

scripts/debug-tests/test_inspection.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+    with gdb("info line") as check:
+        check.matches(r"starts at address 0x[0-9a-f]+", "info line shows start address")
+        check.matches(r"ends at 0x[0-9a-f]+", "info line shows end address")
+        check.matches(r"src/main\.rs", "info line references src/main.rs")
+    with gdb("info breakpoints") as check:
+        check.matches(
+            f"[0-9]+.*breakpoint.*src/main\\.rs:{lines.values}",
+            "breakpoint table lists source breakpoint",
+        )
```
nihalpasham, line 109: Please handle a breakpoint with multiple locations here. In the actual output, `breakpoint` is on the parent row `1`, while `src/main.rs:183` appears on child rows `1.1` and `1.2`. This expression requires both on one line, so it rejects a valid breakpoint table.

Please check the location belonging to the requested breakpoint and cover both single-location and multiple-location output.

scripts/debug-tests/test_execution.py
```diff
@@ -0,0 +1,139 @@
...
+            r"INVALID_PTX|JIT compilation failed|No device code|CUDA_ERROR_INVALID_PTX",
+            "PTX JIT error present",
+        )
+
+    gdb("disable 1")
+    gdb("break cuda_device::DisjointSlice::get_mut")
+    with gdb("continue") as check:
+        check.matches(
+            r"CUDA thread hit Breakpoint.*get_mut",
+            "fully-qualified get_mut breakpoint fires",
```
nihalpasham: This name stays pending in my run. The actual name `cuda_device::DisjointSlice::<'_, i32>::get_mut` resolves and hits. At entry, cuda-gdb can select an inline `ThreadIndex::get` frame above `get_mut`, which also shifts the fixed frame 0/1 checks below.

Please use the emitted name for the baseline named-breakpoint check, then locate and verify the `get_mut` and kernel frames. Keep this as a named-breakpoint test. If resolving the shortened name is intentional new coverage, please separate it as a documented known failure.

scripts/debug-tests/test_inspection.py
```diff
@@ -0,0 +1,173 @@
...
+    with gdb("run") as check:
+        check.matches(r"CUDA thread hit", "GPU breakpoint fires")
+    gdb("disable 1")
+    with gdb("info locals") as check:
+        check.matches(r"^TILE =", "TILE is visible in info locals")
+        check.matches(r"^i =", "scalar local i is visible")
+    with gdb("print TILE") as check:
+        check.matches(r"^\$[0-9]+ = ", "TILE is printable")
+    with gdb("print GLOBAL_COUNTER") as check:
+        check.matches(r"^\$[0-9]+ = [0-9]+", "GLOBAL_COUNTER is accessible")
```
nihalpasham: I could read this global with `print debug_tests::GLOBAL_COUNTER`: it returned 7, and `ptype` reported `@global u64`. Only the unqualified lookup failed.

Please use the qualified name for the memory-inspection check. If unqualified lookup is a feature we want to test separately, keep that as a named known limitation so this doesn't report that global debug information is missing.

scripts/debug-tests/test_execution.py
```diff
@@ -0,0 +1,139 @@
...
+    gdb(f"break src/main.rs:{lines.deep_leaf_bp}")
+    with gdb("run") as check:
+        check.matches(r"CUDA thread hit Breakpoint", "deep_leaf breakpoint fires")
+    gdb("disable 1")
+    with gdb("backtrace") as check:
+        check.matches(r"debug_tests::deep_leaf", "backtrace shows deep_leaf")
+        check.matches(
+            r"debug_tests::deep_middle",
+            "backtrace shows inlined deep_middle",
+        )
```
nihalpasham: The inline frame is present in my backtrace as `deep_middle`, between `debug_tests::deep_leaf` and `debug_tests::deep_outer`. This check rejects it because the namespace isn't printed.

Please verify the intended frame and its position while allowing the observed spelling. The missing `scaled` local is a separate failure: I also couldn't inspect it, and that can remain documented under the scope we agreed on.

scripts/debug-tests/README.md
```diff
@@ -0,0 +1,17 @@
+# cuda-gdb debug-info tests
+
+This suite exercises cuda-oxide debug information with `cuda-gdb`. For now,
+these tests are intended to be run locally; they are not plugged into CI yet.
```
nihalpasham, line 4: Please add the known-failure note here and in the suite's top-level comments, as we discussed. At the moment, only the PR description tells someone that failures are expected.

After correcting the test assumptions, list the remaining failing tests, the missing behavior and the tested GPU/toolkit version. Keeping those tests outside CI is fine; we can fix the remaining debugger limitations and wire the suite into CI separately.

scripts/debug-tests/test_exceptions.py
```diff
@@ -0,0 +1,112 @@
...
+        check.not_matches(
+            f"src/main\\.rs:{lines.oob_fn}",
+            f"backtrace points to function signature line {lines.oob_fn}",
+        )
+    gdb("info registers pc errorpc")
+    with gdb("info locals") as check:
+        check.matches(
+            r"No symbol table info available|No locals",
+            "trap attribution site has no DWARF locals",
+        )
```
nihalpasham: One non-blocking follow-up: please avoid requiring locals to be unavailable. This accepts `No locals.` but would fail if a future debugger improvement correctly exposed `tid = 4`.

The trap, source line and thread are the behavior we want to preserve here. More useful local-variable information should also be accepted.

nihalpasham (changes requested):
Thanks Dmitry — the stepping, exception and coredump cases give us useful coverage.

I reviewed `6b5bd607` and left targeted changes below:

- Align the example's Cargo.lock and keep the ordinary smoketest invocation working.
- Add narrow lint allowances if the trap fixture's unreachable code is intentional.
- Tighten the value checks and reject abnormal cuda-gdb exits.
- Correct the frame selection, symbol names and breakpoint-output assumptions in the baseline tests.
- Add the known-failure note to the suite itself. The missing `scaled` local can remain documented; keeping pytest outside CI is still fine.

On an RTX 5090 with CUDA 13.3, I reproduced the same 9 passing and 4 failing tests. All five normal example scenarios passed, and `memory_spaces` passed memcheck. Local strict Clippy and the example lock check fail as described inline; hosted workflows are still awaiting approval.

Please also sign the revised commit. The DCO sign-off is present, but GitHub reports the current commit as unsigned and main requires a verified signature.

This partially addresses #226. The existing compiler/debugger limitations and CI integration can be follow-ups.

crates/rustc-codegen-cuda/examples/debug-tests/Cargo.lock
dpolyanitsa-nvidia: Done in the next revision

crates/rustc-codegen-cuda/examples/debug-tests/src/main.rs
dpolyanitsa-nvidia: Done in the next revision

crates/rustc-codegen-cuda/examples/debug-tests/src/main.rs
dpolyanitsa-nvidia: Done in the next revision

scripts/debug-tests/test_inspection.py
dpolyanitsa-nvidia: Done in the next revision - value checks are more robust now

scripts/debug-tests/conftest.py
dpolyanitsa-nvidia: Done in the next revision

scripts/debug-tests/test_inspection.py
dpolyanitsa-nvidia: Done in the next revision

scripts/debug-tests/test_exceptions.py
dpolyanitsa-nvidia: This check is generally useless, I've removed it

scripts/debug-tests/README.md
dpolyanitsa-nvidia: I've decided to mark the currently failing tests with pytest's xfail and added two messages: one in README.md with the list of failing tests, and one in conftest.py (the "root" of the testsuite) that mentions the usage of xfail. Is this sufficient?

scripts/debug-tests/test_execution.py
dpolyanitsa-nvidia: I think it's debatable whether allowing not fully qualified names for inlined frames is correct, but I think this can be discussed separately, so allowed both variants in the test for now.

scripts/debug-tests/test_inspection.py
dpolyanitsa-nvidia: Interesting - I expected it to work, but testing the behavior with the host Rust code, it's the same. Agreed and changed the test to use the qualified name.

scripts/debug-tests/test_execution.py
dpolyanitsa-nvidia: That's fair - agreed and changed.

scripts/debug-tests/test_inspection.py
dpolyanitsa-nvidia: Done in the next revision - I've missed it was an inline frame.

I'm leaving a hard-coded "up" in the test because with the current harness it's pretty hard to switch to another frame by function name. One way to do that would be using cuda-gdb's Python interpreter, but not all builds have it, so I decided against that. This can be changed in the future, as necessary.

dpolyanitsa-nvidia:
I think I've addressed all comments, please let me know if I've missed anything of if you want me to change anything else.

nihalpasham (commented):
Thanks Dmitry — the revised checks address the earlier review, and the exception and coredump cases give us useful coverage.

I added a few follow-ups:

- Kept the missing inline local as the only expected failure; wrong frames, values and debugger exits still fail.
- Made command errors stop the debugger session and tied inspection to the native binary just built.
- Initialized all inspected shared memory through raw pointers, with checks for every output thread.
- Added regressions for those failure paths.

At `370a548e`, the local suite reports 36 passed and one expected missing-local failure on an RTX 5090 with CUDA 13.3. The ordinary scenarios and memory sanitizers pass. Hosted compilation, tests, lint, docs and guards pass. Final CodeQL verification is with the maintainer.

This partially addresses #226. CI integration and the missing inline local remain separate.

PR #1083, opened 2026-08-20
Add multi-value debug location infrastructure

uurl:
Closes #1082 
Part of #226 

## Summary

Add generic multi-value debug-location support to the MIR-to-LLVM debug pipeline.

This provides the infrastructure needed for source-variable locations that depend on more than one runtime SSA value, such as a future runtime-indexed projection:

```text
*(base + index * stride)
```

## Changes

* add `mir.dbg_value_list` for ordered multi-value debug operands
* lower it to a new `llvm.dbg_value_list` marker
* add a typed `DebugValueExpression` representation
* preserve debug expressions through MIR-to-LLVM lowering
* export multi-value locations through `llvm.dbg.value` using `DIArgList`
* support:

  * `DW_OP_LLVM_arg`
  * `DW_OP_constu`
  * `DW_OP_plus`
  * `DW_OP_plus_uconst`
  * `DW_OP_mul`
  * `DW_OP_deref`
  * `DW_OP_stack_value`
* validate `DW_OP_LLVM_arg` operand indices during export
* add lowering coverage for operand ordering and expression preservation
* add LLVM-export coverage for `DIArgList`
* parse the emitted debug module with `llvm-as`

For example, the expression:

```text
*(base + index * 4)
```

is represented as:

```text
DW_OP_LLVM_arg, 0
DW_OP_LLVM_arg, 1
DW_OP_constu, 4
DW_OP_mul
DW_OP_plus
DW_OP_deref
```

with the base and index values carried in the corresponding `DIArgList`.

## Why

Existing projected debug locations use static byte offsets. That is sufficient for fields and constant indices, but not for runtime-dependent projections.

A dynamic array index requires both the base address and the runtime index to remain available to the debugger. LLVM's `DIArgList` plus `DW_OP_LLVM_arg` model provides the required representation without materializing a synthetic address computation in executable IR.

## Non-goals

This PR only introduces the generic multi-value location machinery.

It does not yet translate `ProjectionElem::Index` from rustc MIR. That will be implemented as a follow-up using this infrastructure.

It also does not change existing single-value `dbg.value` behavior or migrate the exporter to LLVM debug records.

## Validation

Validated locally with:

```text
cargo fmt --all -- --check
cargo test -p llvm-export full_debug_metadata_emits_diarglist_for_multi_value_locations -- --nocapture
cargo test -p llvm-export --all-targets
cargo test -p mir-lower --all-targets
cargo clippy --workspace --all-targets -- -D warnings
```

[...]

PR #1282, opened 2026-09-17
fix(debug): support runtime-indexed fixed-array references

### LINKED ISSUE #1280: Full debug: support runtime-indexed statement debug locations
## Summary

Runtime-indexed source locations require both a base address and a runtime index.

With the rustc toolchain used by cuda-oxide, `ProjectionElem::Index` is not permitted in ordinary `VarDebugInfoContents::Place`. Runtime-indexed debug places instead arise through statement-level debug information such as `StmtDebugInfo::AssignRef`.

#1259 already handles the narrow runtime-indexed form:

```text
&slice[index]
```

by reconstructing the reference value and storing it in the destination local's existing debug stack slot.

This issue extends that statement-debug bridge to the corresponding fixed-size array case:

```text
&array[index]
```

for one runtime `usize` index.

## Why not `mir.dbg_value_list`?

#1083 added the generic multi-value debug-location infrastructure needed to express locations such as:

```text
base + index * stride
```

using `mir.dbg_value_list`, LLVM `DIArgList`, and `DW_OP_LLVM_arg`.

That representation was evaluated for the fixed-array runtime-index case.

The metadata survived far enough for CUDA-GDB to recognize the projected reference, but CUDA-GDB did not evaluate the dynamic index contribution correctly: the base pointer was visible while dereferencing the resulting projected reference produced the wrong value.

The fixed-array implementation therefore follows the already validated #1259 strategy instead:

```text
StmtDebugInfo::AssignRef
    -> runtime-indexed fixed-array Place
    -> existing MIR place-address translation
    -> reconstructed reference
    -> destination debug stack slot
    -> CUDA-GDB
```

This introduces address materialization only for the full-debug statement-debug bridge, matching the existing #1259 fallback strategy.

## Supported scope

Support a single runtime `ProjectionElem::Index` for a direct fixed-size array reference when:

* the destination is a reference debug local with an existing stack slot
* the base is a fixed-size array
* the runtime index local has type `usize`
* the place shape is directly representable by the existing address walker
* no slice/DST metadata reconstruction is required

The existing `&slice[index]` support from #1259 remains unchanged.

Unsupported shapes continue to fail closed rather than emitting approximate debug locations.

## Non-goals

This issue does not include:

* ordinary `VarDebugInfo` runtime-index projections
* `ProjectionElem::Subslice`
* slice/DST fat-pointer reconstruction beyond the existing #1259 case
* multiple runtime indices
* repeated dereferences
* arbitrary dereference-plus-index chains
* trait-object metadata
* general fat-pointer debug types
* replacing the #1259 stack-home strategy with `DIArgList`

## Validation

The implementation includes:

* importer coverage for the accepted fixed-array runtime-index shape
* fail-closed importer coverage for multiple runtime indices
* preservation of the existing #1259 slice-index behavior
* full workspace formatting and test coverage
* documentation for the supported runtime-index subset
* CUDA-GDB validation on real hardware

Validated on `sm_120` with:

```text
runtime_index = 3
projected_runtime = valid *mut u32
*projected_runtime = 55
```

The full `compiler_features` debug smoke test passes with runtime-index, static projection, enum projection, dereference projection, closure, caller-frame, and helper-frame coverage.

Part of #226.

Related: #1083, #1259.

uurl:
## What this adds

CUDA-GDB can inspect an immutable reference selected by one runtime index into a direct fixed-size array in full-debug builds.

```rust
let selected = &values[index];
```

```text
before: the optimized-away reference has no debugger location
after:  its reconstructed address is stored in the debugger stack slot
```

The existing runtime-indexed slice path stays available.

## How it works

The importer extends the existing statement-debug bridge from #1259 and uses the ordinary place-address walker. It accepts only a debug-only destination, initialized array and `usize` index, matching element type, and an existing stack slot. Unsupported shapes omit the binding before emitting address calculations.

I added an example follow-up that exercises all four runtime indices and kept index 3 first for the CUDA-GDB smoke test. The branch is rebased and signed, preserving contributor authorship and sign-off.

## Verification

Checked at `77f0b73e` against `b631e5d7`:

- 5,355 workspace/backend tests and 33 doctests pass; formatting, strict Clippy, warning-free docs and the book pass.
- The new array stack-home regression fails with the pre-PR shape predicate and passes on this head.
- The full CUDA-GDB smoke test passes on RTX 5090, including existing slice, caller/helper, closure, enum and projection checks.
- A separate debugger probe reads `(index, pointee)` as `(3, 55)`, `(0, 13)`, `(1, 21)`, `(2, 34)`.
- `compiler_features` passes memcheck with zero errors.
- Independent review checked preflight, definite initialization and the shared address walker.

## Caveats

This covers direct immutable fixed-array references with one runtime index. Mutable array references, multiple indices and surrounding field/dereference chains remain unsupported. Address materialization is confined to full-debug mode.

Closes #1280. Partially addresses #226; related to #1083 and #1259.


nihalpasham (commented):
I kept this on the existing statement-debug stack-slot path and added coverage for every runtime array index.

- The four lookups return `13`, `21`, `34`, and `55`; a separate CUDA-GDB probe confirms each reconstructed reference points to the corresponding element.
- The existing preflight still requires initialized inputs and a debug-only destination. Unsupported projections omit the binding before adding executable operations.
- I rebased and signed the commits while preserving your authorship and sign-off.

At `77f0b73e`, workspace/backend tests, doctests, strict Clippy, formatting, docs and the book pass. The full CUDA-GDB smoke test and `compiler_features` memcheck also pass on RTX 5090.

This addresses #1280's direct immutable array-reference case and partially addresses #226. Hosted CI also passes on this head.

nihalpasham (approved):

PR #1285, opened 2026-09-17
feat(cuda-host): load embedded modules from the binary that maps the artifact anchor

### LINKED ISSUE #1284: `load()` reads the process executable, so a `#[cuda_module]` compiled into a shared object cannot find its own artifact
## Summary

The `load()` / `load_named()` that `#[cuda_module]` generates call `cuda_host::load_embedded_module`, which reads the `.oxart` section of `std::env::current_exe()`. That is the right binary only while the module was compiled into the executable itself. When the crate holding the module is built as a shared object and loaded into another process, the artifact is embedded in that shared object, `current_exe()` names the host executable, and `load()` fails with

```
embedded CUDA module '<name>' was not found
```

Anything built as a `cdylib` hits this: `dlopen`ed plugins, PyO3 extension modules, evcxr cells. The crate builds normally and the failure only appears at `load()`.

## Reproduction

Built with `cargo oxide build` at `7ce30ec` on a Tesla T4 (CUDA 13.4, `CUDA_OXIDE_TARGET=sm_75`).

`plugin/src/lib.rs`, a `cdylib` that depends on `cuda-device`, `cuda-host` and `cuda-core`. It loads the module and, if that works, launches the kernel once:

```rust
use cuda_core::simt::LaunchConfig;
use cuda_core::{CudaContext, DeviceBuffer};
use cuda_device::{DisjointSlice, cuda_module, kernel, thread};

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    pub fn fill(mut output: DisjointSlice<u32>) {
        let idx = thread::index_1d();
        let raw = idx.get();
        if let Some(out) = output.get_mut(idx) {
            *out = raw as u32;
        }
    }
}

#[unsafe(no_mangle)]
pub extern "C" fn plugin_load() -> i32 {
    let ctx = match CudaContext::new(0) {
        Ok(ctx) => ctx,
        Err(error) => {
            eprintln!("plugin: CudaContext::new(0) failed: {error}");
            return 2;
        }
    };
    let loaded = match kernels::load(&ctx) {
        Ok(loaded) => loaded,
        Err(error) => {
            eprintln!("plugin: kernels::load() failed: {error}");
            return 1;
        }
    };
    println!("plugin: kernels::load() ok");

    const N: usize = 256;
    let result = (|| {
        let mut out = DeviceBuffer::<u32>::zeroed(&stream, N)?;
        unsafe { loaded.fill(&stream, LaunchConfig::for_num_elems(N as u32), &mut out) }?;
        out.to_host_vec(&stream)
    })();
    match result {
        Ok(values) if values[..8] == [0, 1, 2, 3, 4, 5, 6, 7] && values[N - 1] == (N - 1) as u32 => {
            println!("plugin: fill kernel launched, output[..8] = {:?}", &values[..8]);
            0
        }
        Ok(values) => {
            eprintln!("plugin: unexpected output {:?}", &values[..8]);
            3
        }
        Err(error) => {
            eprintln!("plugin: launch failed: {error}");
            3
        }
    }
}
```

`host/src/main.rs`, a plain binary depending on `cuda-core` and `libloading`. It lists the bundles of its own executable and of the plugin, then `dlopen`s the plugin and calls `plugin_load`:

```rust
use cuda_core::embedded::artifact_bundles_from_binary_path;
use std::path::Path;

fn bundle_names(path: &Path) -> String {
    match artifact_bundles_from_binary_path(path) {
        Ok(bundles) => format!(
            "{:?}",
            bundles.iter().map(|b| format!("{} ({})", b.name, b.target)).collect::<Vec<_>>()
        ),
        Err(error) => format!("error: {error}"),
    }
}

fn main() {
    let plugin = std::env::args().nth(1).expect("usage: host <path to libplugin.so>");
    let exe = std::env::current_exe().unwrap();
    println!("host: current_exe = {}", exe.display());
    println!("host: bundles in current_exe: {}", bundle_names(&exe));
    println!("host: bundles in {}: {}", plugin, bundle_names(Path::new(&plugin)));

    let library = unsafe { libloading::Library::new(&plugin) }.expect("dlopen libplugin.so");
    let plugin_load: libloading::Symbol<unsafe extern "C" fn() -> i32> =
        unsafe { library.get(b"plugin_load\0") }.expect("plugin_load symbol");
    let code = unsafe { plugin_load() };
    println!("host: plugin_load returned {code}");
    std::process::exit(code);
}
```

Output of `target/release/host target/release/libplugin.so`:

```
host: current_exe = /content/repro-upstream/target/release/host
host: bundles in current_exe: []
host: bundles in /content/repro-upstream/target/release/libplugin.so: ["plugin (sm_75)"]
plugin: kernels::load() failed: embedded CUDA module 'plugin' was not found
host: plugin_load returned 1
```

So the bundle is in `libplugin.so`, and `load()` never looks there. evcxr runs into the same thing, since every cell is a shared object loaded into one long-lived process.

## Proposed fix

`load_named()` already references the artifact anchor symbol that the backend defines inside the artifact object (the keep-alive handshake from #72), and the linker puts that anchor in the same binary as the `.oxart` section. `dladdr(anchor)` gives the path of that binary, and `artifact_bundles_from_binary_path` can read bundles from a path.

- `cuda_host::load_embedded_module_from_address(ctx, name, addr)`: locate the binary that maps `addr`, read its bundles, then the existing payload handling (cubin, PTX, NVVM IR, LTOIR). When the loader reports no absolute path for the binary (glibc names the main program by `argv[0]`), read `current_exe()` exactly as today.
- `#[cuda_module]`: record the anchor address in `load_named()` and pass it to the new function. Modules that reference no anchor (not built by cargo, crate excluded by the owner filter, only generic kernels, or every concrete kernel cfg'd out) keep calling `load_embedded_module`.

The user-facing API does not change. `load_all_ptx_bundles_merged` (generic kernels) is not part of this and could use the same lookup later.

With this change the reproduction above prints `plugin: kernels::load() ok`, then `plugin: fill kernel launched, output[..8] = [0, 1, 2, 3, 4, 5, 6, 7]`, and `plugin_load` returns 0.

xavierforge:
Closes #1284.

`#[cuda_module]`'s `load()` read the artifact bundle from `std::env::current_exe()`. When the module was compiled into a shared object (a `cdylib` plugin, a PyO3 extension, an evcxr cell) the bundle is in that shared object, so `load()` failed with `ModuleNotFound`.

`load_named()` already references the artifact anchor symbol, and the linker puts that anchor in the same binary as the `.oxart` section. This PR:

- adds `cuda_host::load_embedded_module_from_address(ctx, name, addr)`. `dladdr` gives the path of the executable or shared object that maps `addr`, the bundles are read from that path, and the payload handling (cubin, PTX, NVVM IR, LTOIR) is the same code as `load_embedded_module`. When the loader has no absolute path for the binary (glibc names the main program by `argv[0]`) it reads `current_exe()` as before.
- makes the generated `load_named()` keep the anchor address in an `Option` and pass it to the new function. When no anchor is referenced (not built by cargo, crate excluded by the owner filter, only generic kernels, or every concrete kernel cfg'd out) it calls `load_embedded_module` as before.
- declares `dladdr` and `Dl_info` in `cuda-host` instead of adding the `libc` crate. A new dependency of cuda-host changes the lockfile of every example workspace, about 230 files. If you would rather take the `libc` dependency, I can switch.

`load_all_ptx_bundles_merged` (generic kernels) is untouched and still reads the executable. It could use the same lookup later.

## Tests

- `cuda-host` unit tests: an address in the test executable resolves to `current_exe()`. An address obtained with `dlsym` from the C library resolves to that shared object. Bundles read through an address in the executable are the same as those of `current_exe()`.
- `cuda-macros`: expansion tests for the new loader, with and without generic kernels. The real expansion gets compiled by `crates/cuda-host/tests/cuda_module_api.rs` and the trybuild suite.
- On a Tesla T4 (CUDA 13.4, `cargo oxide build`, backend at `7ce30ec`) I ran the reproduction from the issue, a `cdylib` with one `#[cuda_module]` loaded with `dlopen` from a plain binary. With cuda-host from `main`, `kernels::load()` fails with `embedded CUDA module 'plugin' was not found`. With this branch it loads and the `fill` kernel launches from the loaded module (`output[..8] = [0, 1, 2, 3, 4, 5, 6, 7]`). In both runs the host executable has no bundle and `libplugin.so` has `plugin (sm_75)`.

The repository CI (fmt, clippy, unit-tests, cargo-deny, examples-compile) passes on this branch in my fork.


nihalpasham (dismissed):
I kept the containing-image approach and added the follow-ups needed for shared libraries:

- Replaced the raw-address interface with a borrowed anchor byte, then locate its mapping through procfs. A relative `dlopen("./plugin.so")` still works after `chdir`.
- Validate the mapped file's device/inode and read from the same handle. Deleted, replaced or unavailable images fail explicitly, and unusual filenames keep their native bytes.
- Reused the existing artifact parser and kept executable loading on platforms without anchor discovery.
- Added shared-library, missing-mapping, file-identity and macro-routing regressions.

At `5a64eb58`, all 231 host/macro tests passed, along with strict Clippy, formatting, docs and the 233-example lock check. The original plugin fails with `ModuleNotFound`; the repaired relative-load and changed-directory cases each verified all 256 GPU outputs on RTX 5090. The ordinary executable control also passed, and plugin memcheck reported zero errors. All seven hosted workflows passed, with 54 successful checks and the PR book deployment intentionally skipped.

Merge #1291 first; this branch includes its final signed commits. This addresses #1284 for non-generic Linux/Android modules with accessible procfs and mapped files. Generic PTX merging remains separate.

xavierforge:
Thanks for the follow-up. I've read through 5a64eb58: the procfs lookup with the dev/inode check is a clear improvement over dladdr for the relative dlopen + chdir case, and dropping the current_exe fallback makes the failure mode explicit instead of silently reading the wrong binary. Happy with the rename to `load_embedded_module_from_anchor` as well.

Note for anyone reading later: the PR description above describes the original dladdr version; the final implementation is the one in 5a64eb58.

nihalpasham (dismissed):
Reconfirming approval after the prerequisite merge changed the comparison base. The reviewed head remains `5a64eb58`, all seven PR workflows pass, and merging it into current main produces the exact tree already validated in the batch integration.

nihalpasham (approved):
Reconfirming approval after the prerequisite merge changed the comparison base. The reviewed head remains `5a64eb58`, all seven PR workflows pass, and merging it into current main produces the exact tree already validated in the batch integration.

PR #1286, opened 2026-09-17
feat: accept any kernel slice arg in generated launchers

drbh:
this pr lets the generated sync launchers take any `impl KernelSliceArg` / `impl KernelSliceArgMut` where they used to require a `DeviceBuffer`, matching what the async launchers already accept. launcher half of https://github.com/NVlabs/cuda-oxide/pull/1081 and #968

previously the traits were gated behind `async` and the sync path was hardcoded to `DeviceBuffer`, so using memory from another crate meant copying it in or faking a `DeviceBuffer` with `from_raw_parts`. any type implementing the traits now works in every slice position, `DeviceBuffer` still does so nothing changes for existing callers

the `cudarc_slice` example is the whole point, a newtype over a cudarc `CudaSlice` goes straight into the launcher

```rust
struct Cudarc<T>(CudaSlice<T>);

unsafe impl<T> KernelSliceArg for Cudarc<T> {
    type Elem = T;
    fn cu_deviceptr(&self) -> CUdeviceptr {
        self.0.device_ptr(self.0.stream()).0
    }
    fn len(&self) -> usize {
        self.0.len()
    }
}
unsafe impl<T> KernelSliceArgMut for Cudarc<T> {}

let input = Cudarc(stream.clone_htod(&[1.0f32, 2.0, 3.0, 4.0])?);
let mut output = Cudarc(stream.alloc_zeros::<f32>(4)?);
unsafe { module.double(&ctx.default_stream(), LaunchConfig::for_num_elems(4), &input, &mut output) }?;
```

note: no runtime changes, the launcher only reads ptr/len and never frees anything. overlaps with #1251/#1252 in `model.rs`


nihalpasham (changes requested):
I would reject this version. The interop goal is useful, but safe prepared launches need more than a pointer and length:

- These launchers enqueue work and return; the Rust borrow ends before the GPU necessarily finishes. The slice traits only promise validity for that borrow.
- The cudarc adapter drops its event guard before the launch and uses the read accessor for writable arguments. cudarc therefore cannot track oxide's reads/writes or wait for them before freeing on another stream.
- The example uses one default stream, which orders that example correctly. The generated safe API also accepts other streams, where that ordering is absent.

This needs a stream-aware use/lifetime contract, with read/write dependencies recorded around submission and storage kept valid through completion. An explicitly unsafe interop API could also be scoped separately. I haven't patched around the issue because it changes the abstraction itself.

The host and macro suites pass with default and all features. The example also needs a supported cudarc/toolkit selection: its normal check fails on CUDA 13.4 and passes with the CUDA 13.0 binding override.

PR #1288, opened 2026-09-17
fix(atomics): support core AtomicPtr lowering

### LINKED ISSUE #1287: core::sync::atomic::AtomicPtr<T> fails to compile in CUDA kernels
**Description**

`core::sync::atomic::AtomicPtr<T>` operations fail during MIR atomic lowering because raw pointer element types are not recognized by the core atomic type extractor.

The same lowering path already handles supported integer and floating-point `core::sync::atomic` types. However, when an `AtomicPtr<T>` method such as `load` is compiled, its intrinsic element type is a raw pointer and type extraction fails before NVVM or PTX can be produced.

**Minimal reproducer**

Using the dependencies from the existing `atomics` example, the following complete `main.rs` reproduces the issue:

```rust
use core::sync::atomic::{AtomicPtr, Ordering};
use cuda_host::cuda_module;

#[cuda_module]
mod kernels {
    use super::*;

    #[kernel]
    pub fn atomic_ptr_load(storage: &[usize], mut out: DisjointSlice<u32>) {
        let gid = thread::index_1d();

        if gid.in_bounds(storage.len()) {
            let index = gid.get();

            let atomic_ptr = unsafe {
                &*(storage.as_ptr().add(index) as *const AtomicPtr<u8>)
            };

            let loaded = atomic_ptr.load(Ordering::Relaxed);

            if let Some(out_elem) = out.get_mut(gid) {
                *out_elem = loaded.is_null() as u32;
            }
        }
    }
}

fn main() {}
```

Build with:

```bash
cargo oxide build atomics --arch sm_86
```

**Expected behavior**

The kernel should compile successfully.

On the currently supported `nvptx64` target, generic-address-space pointer values use the target's 64-bit pointer representation. Pointer-valid atomic operations such as `load`, `store`, `swap`, and `compare_exchange` should preserve the pointer type through lowering and use 64-bit pointer registers.

Address-space-specific pointer values whose representation may differ from generic pointers do not need to be accepted as part of this behavior.

**Actual behavior**

Compilation fails during core atomic type extraction. The diagnostic contains:

```text
Unsupported construct: unsupported core atomic operation `load`: no element type in its generics, and it is not one of the ordering-only fences
```

The failure occurs before LLVM IR or PTX is produced.

The core atomic type extractor recognizes supported integer and floating-point MIR types but returns no type information for `RigidTy::RawPtr`, which is the element type used by `AtomicPtr<T>` intrinsics.

**Environment**

Tested against upstream commit:

```text
7ce30ec798c490e6ed772dfcbad193b3733fe55a
```

<details><summary><code>cargo oxide doctor</code></summary>

```text
cargo-oxide environment check
=============================

Rust nightly toolchain... ✓ rustc 1.100.0-nightly (e457a7b0d 2026-08-27)
rust-toolchain.toml... ✓ channel nightly-2026-08-28
Pinned toolchain active... ✓ nightly-2026-08-28-x86_64-unknown-linux-gnu (overridden by environment variable RUSTUP_TOOLCHAIN)
Required rustup components... ✓ rust-src, rustc-dev, rust-analyzer, clippy, rustfmt, llvm-tools
Codegen backend... ✓ /home/yunus/cuda-oxide/crates/rustc-codegen-cuda/target/x86_64-unknown-linux-gnu/debug/librustc_codegen_cuda.so
Project config (.cargo/cuda-oxide.toml)... - not present (using defaults)
Shared cache (external projects)... ✓ /home/yunus/.cargo/cuda-oxide/librustc_codegen_cuda.so
CUDA headers (cuda.h)... ✓ /usr/local/cuda/include/cuda.h
CUDA toolkit (nvcc)... ✓ Cuda compilation tools, release 13.3, V13.3.73 (/usr/local/cuda-13.3/bin/nvcc)
nvJitLink (libnvJitLink.so)... ✓ nvJitLink 13.3
libdevice (libdevice.10.bc)... ✓ /usr/local/cuda/nvvm/libdevice/libdevice.10.bc
llc (LLVM)... ✓ LLVM version 23.1.0-rust-1.100.0-nightly (/home/yunus/.rustup/toolchains/nightly-2026-08-28-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llc)
clang / libclang resource dir... ✓ /usr/lib/llvm-21/lib/clang/21
NVIDIA driver / GPU... ✓ NVIDIA GeForce RTX 3050 Ti Laptop GPU (compute capability 8.6, driver 610.62)
cuda-gdb (optional)... ✓ exec: /usr/local/cuda-13.3/bin/cuda-gdb-minimal --version (/usr/local/cuda-13.3/bin/cuda-gdb)
compute-sanitizer (optional)... ✓ Version 2026.2.1.0 (build 38334959) (public-release) (/usr/local/cuda-13.3/bin/compute-sanitizer)

✅ Environment looks good!
```

</details>

**Additional context**

The core atomic importer currently reconstructs supported scalar result types from width and signedness. Pointer atomics need to retain their translated MIR pointer type instead of being reconstructed as an integer with the same width.

A narrow implementation should:

- recognize raw pointer element types for core atomics;
- preserve pointer types through MIR and LLVM lowering;
- allow pointer-valid operations: `load`, `store`, `swap`, and `compare_exchange`;
- reject arithmetic and bitwise pointer RMW operations;
- accept only generic-address-space pointer values where the backend can rely on the 64-bit `nvptx64` representation;
- continue rejecting address-space-specific pointer values whose representation may differ from generic pointers.

No `.ll` or `.ptx` artifact is produced by the reproducer because compilation fails during MIR import.

YunusMutlu:
## What this fixes

Ordinary `core::sync::atomic::AtomicPtr<T>` load, store, swap and compare-exchange now compile for generic pointer values on nvptx64. Pointer values retain their types through the importer, dialects and LLVM lowering.

```rust
let slot = AtomicPtr::new(core::ptr::null_mut());
slot.store(pointer, Ordering::Release);
let old = slot.swap(other, Ordering::AcqRel);
```

Previously pointer operands were rejected. The first implementation also exposed invalid PTX when the atomic object itself lived in thread-local storage.

## What we fixed

- Kept the contributor's pointer representation and added shared storage handling for all atomic operations. Local storage uses ordinary typed accesses; a generic storage pointer is tested with `isspacep.local`. The other branch retains its generic address, including cluster-shared memory.
- Extended the existing legacy NVVM legalization to pointer exchange and CAS with scoped PTX and pointer-typed operands/results. CAS preserves both success and failure ordering requirements, including Release/Acquire and Relaxed/SeqCst.
- Repaired the example's writable-storage access and four divergent barriers. Added local pointer/integer cases, failed CAS, wrapping arithmetic and an actual pointee readback. Contended counters use raw writable pointers and `from_ptr`; the host waits before reading or releasing their allocations.
- Added storage-space, old-value, signedness, floating-point, ordering and rejection controls. Rebased and signed the commits on current main, preserving contributor authorship and sign-offs.

## Verification

Checked at `ba63a7f1` against main `b631e5d7`:

- 5,368 workspace/backend tests and 33 doctests pass; 11 optional tests and 157 documentation examples remain ignored. Formatting, strict workspace/backend/example Clippy, warnings-denied rustdoc and Sphinx, and affected guards pass.
- All 22 runtime atomic tests pass on RTX 5090 through native LLVM23 and modern libNVVM, with zero memcheck and synccheck errors.
- An independent 256-thread probe selects local, shared or global storage at runtime and checks pointer readback, CAS outcomes, integer boundaries and floating controls. Native, full device-debug and modern libNVVM runs pass; focused pointer/integer operations also pass through CUDA 13.0 legacy LLVM7 NVVM and driver JIT. Memcheck reports zero errors throughout.
- 64 independent optimized/unoptimized LLVM/libNVVM controls assemble successfully, including cluster-shared addresses. Removing the private-storage branch reproduces the invalid `.local.cas` instruction and fails assembly.
- Combining this head with #1273 and #1275 is conflict-free; 661 affected/backend tests and the backend library build pass against their shared Pliron revision.
- Independent compiler and source-safety reviews found no remaining counterexample. Hosted CI also passes on this head.

## Caveats

Legacy NVVM retains its existing restrictions on floating exchange and explicit cluster-shared storage; the focused legacy run does not claim those features. Cluster-shared preservation was checked in IR/PTX and assembly, without a multi-CTA cluster hardware claim. The dispatch adds a conditional path for an unknown storage address; no performance improvement is claimed.

Closes #1287.


nihalpasham (commented):
I kept the pointer-typed lowering and filled in the storage and backend cases it exposed:

- Local atomic objects now use ordinary thread-private accesses. Generic pointers retain the existing global/shared/cluster path when they are not local.
- Legacy pointer swap/CAS uses the existing scoped PTX legalizer, preserving pointer results and both CAS ordering requirements.
- I repaired the example's storage borrows and divergent barriers, then added local operations, failed CAS, integer boundaries and a real pointee readback.

At `ba63a7f1`, all 5,401 workspace/backend tests and doctests, strict checks, docs and affected guards pass. The 22-kernel suite passes native LLVM and modern libNVVM on RTX 5090 with zero memcheck/synccheck errors. Independent dynamic-address probes also pass in device-debug and through the legacy pointer/integer path; removing the private-storage fix reproduces invalid local CAS PTX.

Legacy floating exchange and explicit cluster-shared NVVM support retain their existing limits. Cluster preservation was checked through compilation and assembly. Hosted CI also passes on this head.

This closes #1287 once merged.

nihalpasham (approved):

PR #1291, opened 2026-09-17
fix(cuda-host): use libc constants for the cache's no-symlink open

### LINKED ISSUE #1290: ltoir_cache: the no-symlink open uses x86-64 flag values, inert on aarch64
`read_regular_file_exact` in cuda-host/src/ltoir_cache.rs opens cache artifacts with

    options.custom_flags(0o400000 | 0o4000);  // O_NOFOLLOW | O_NONBLOCK

but those are x86-64 values. On aarch64 Linux O_NOFOLLOW is 0o100000, and 0o400000 is O_DIRECT. So on arm64 (Grace, or any arm64 container) the open follows symlinks — the thing the flag exists to prevent — and every cache read carries an accidental O_DIRECT with its buffer-alignment rules.

The crate's own test already catches it on that arch:

    cache_reads_do_not_follow_artifact_symlinks
    assertion failed: read_regular_file_exact(&link, 13).is_err()

Fix is a one-liner: `libc::O_NOFOLLOW | libc::O_NONBLOCK`, plus `libc::ELOOP` for the errno check below it. Have it ready with the test passing on aarch64 — happy to open the PR.

navneet83:
Fixes #1290.

`read_regular_file_exact` opened cache artifacts with hardcoded flags `0o400000 | 0o4000` for `O_NOFOLLOW | O_NONBLOCK`. Those are x86-64 values: on aarch64 Linux `O_NOFOLLOW` is `0o100000` and `0o400000` is `O_DIRECT`, so the symlink guard was inert there and every cache read carried an accidental `O_DIRECT` with its buffer-alignment rules. The errno literal `40` beside it happens to be `ELOOP` on both architectures but is the same class of literal.

Change: `libc::O_NOFOLLOW | libc::O_NONBLOCK` and `libc::ELOOP`, with `libc` added to cuda-host scoped to the cfg that uses it (already recorded in dependency-licenses.csv).

Before/after on aarch64 Linux (repo devcontainer image, linux/arm64):

```text
before:  ltoir_cache::tests::cache_reads_do_not_follow_artifact_symlinks ... FAILED
         assertion failed: read_regular_file_exact(&link, 13).is_err()
after:   test result: ok. 74 passed; 0 failed; 1 ignored
```

The existing test is the regression case: it asserts the guard on exactly the flag this fixes, and only compiles on Linux, so it bites on any arm64 Linux dev box or CI runner. No behavior change on x86-64, where the literals and the constants agree.

The 226 example `Cargo.lock` files each gain the single line `+ "libc",` under cuda-host's dependency list: every example workspace path-depends on cuda-host and CI resolves them `--locked`, so the new dependency edge has to be recorded in each lock. `libc 0.2.180` was already the pinned version throughout; `scripts/sync-example-locks.sh --check` passes.


nihalpasham (approved):
I kept the libc-based flag fix and rebased it onto current main.

- Added the dependency edge to the three example lockfiles introduced since this branch.
- Corrected the aarch64 explanation: the old value misses `O_NOFOLLOW`; it is not aarch64's `O_DIRECT` value.

At `7f3b5f88`, all 103 cuda-host tests/doctests passed, including the normal cache-hit and symlink-rejection checks. Strict Clippy, formatting, docs and all 233 example-lock consistency checks passed. I verified the aarch64 constants in libc; this machine's runtime checks are x86-64. All seven hosted workflows passed, with 54 successful checks and the PR book deployment intentionally skipped.

This closes #1290 once landed.

PR #1295, opened 2026-09-18
feat(intrinsics-gen): admit plain sparse FP8 MMA for SM89

0z5a:
Phase 3 of #337: the four plain SM89 sparse FP8 MMA forms with F32 accumulation.

```
mma.sp.sync.aligned.m16n8k64.row.col.f32.e4m3.e4m3.f32
mma.sp.sync.aligned.m16n8k64.row.col.f32.e4m3.e5m2.f32
mma.sp.sync.aligned.m16n8k64.row.col.f32.e5m2.e4m3.f32
mma.sp.sync.aligned.m16n8k64.row.col.f32.e5m2.e5m2.f32
```

| catalog id | operation key | ABI | floor |
|---|---|---|---|
| `mma_sp_m16n8k64_f32_e4m3_e4m3_f32` | `matrix.mma.sp.m16n8k64.row.col.f32.e4m3.e4m3.f32.not_applicable.standard_metadata` | `i1026` | PTX 8.4 / sm_89+ |
| `mma_sp_m16n8k64_f32_e4m3_e5m2_f32` | `…f32.e4m3.e5m2.f32…` | `i1027` | PTX 8.4 / sm_89+ |
| `mma_sp_m16n8k64_f32_e5m2_e4m3_f32` | `…f32.e5m2.e4m3.f32…` | `i1028` | PTX 8.4 / sm_89+ |
| `mma_sp_m16n8k64_f32_e5m2_e5m2_f32` | `…f32.e5m2.e5m2.f32…` | `i1029` | PTX 8.4 / sm_89+ |

## These could not be admitted through the overlay alone

Worth stating up front, because the issue describes this phase as an
`intrinsics/overlay/*.toml` + `intrinsics/abi-v1.toml` change. That is not
sufficient here, and I would rather explain why than leave a reviewer guessing.

The pinned toolchain is **not** the blocker: all four records are already in
`intrinsics/imported.json` (LLVM rev `16696adc`, LLVM 23.1.0) with
`ImmArg`/`Range<arg13,0,1>` properties and an effective floor of sm_89 / PTX 8.4.
Measured coverage: 138 register-form `NVVM_MMA_SP` intrinsics imported, 125
admitted, and of the 16 unadmitted exactly these 4 are the plain sparse FP8
forms. The gap was entirely in the admission layer.

The `sparse_mma` resolver treated "FP8 operands" plus an F16/F32 accumulator as
implying the `kind::f8f6f4` type-kind **and** `sp::ordered_metadata`. PTX keeps
those axes separate, so all three TOML routes dead-ended:

* the `scalar_contract` F16/F32 arm required `metadata == Ordered`, so a plain
  `standard` declaration became `"requests an unsupported sparse-MMA variant"`;
* `sparse_mma_identity` routed any FP8 element down the `f8f6f4` branch
  unconditionally, so choosing `ordered` emitted the wrong instruction *and*
  collided with the existing ordered records on both id and operation key;
* `sparse_mma_minimum_ptx` and `sparse_mma_hardware` derived their answer from
  the accumulator class alone, giving the Blackwell block-scale floor and target
  list instead of PTX 8.4 / `sm_89`;
* the imported-predicate check pinned every F16/F32-accumulator declaration to
  `["Subtarget->hasMMABlockScale()"]`, which is not the contract these records
  carry.

The `[sparse_mma_integer]` route was never viable either: it hardcodes an `s32`
accumulator and `unreachable!()`s on floating-point elements.

## What changed

**Generator (hand-edited)**

* `model/overlay.rs` — new `SparseMmaFp8F32Admission` and a `sparse_mma_fp8_f32`
  field on `OverlayShardFile`. That struct is `deny_unknown_fields`, so this has
  to land before any TOML can name the table.
* `resolve/families/mma/sparse.rs` — the `f8f6f4` identity branch is now gated on
  the metadata mode actually requested; a Standard/F32 branch emits plain typed
  operands; `sparse_mma_recipe` accepts `NotApplicable + Standard` for FP8
  carriers at F32 while still requiring `Ordered` for the `kind::f8f6f4` form;
  minimum PTX, target set and accepted predicate set are derived from the form
  being admitted rather than from the accumulator class.
* `resolve/overlay.rs` — take site, `compact_mma_count`, expand dispatch,
  `SPARSE_MMA_FP8_F32_SHARD_SCHEMA = 64` and `OVERLAY_SHARD_SCHEMA` 63 → 64.

**Reviewed inputs** — `intrinsics/overlay/sparse_mma_fp8_f32.toml` (new, indexed
in sorted position), `intrinsics/abi-v1.toml` (+`i1026`–`i1029`, append-only, no
existing id/symbol/signature touched), and two evidence files.

**Generated** — a second commit carries the regenerated outputs: 1316 files from
`generate`, of which every pre-existing file differs only in the catalog SHA-256
in its header (spot-checked: `git diff --stat` on an unrelated generated file is
1 line). The four new families add their device wrappers, NVVM dialect ops,
importer arms, lowering arms, codegen targets and probe modules.

## Verification

| Check | Result |
|---|---|
| `cuda-intrinsics-gen generate` | 1316 outputs written |
| `cuda-intrinsics-gen check` | byte-current |
| `cuda-intrinsics-gen probe --all --skip-terminal --per-target` | 2378 target configurations across 1029 routes, 0 errors |
| `cuda-intrinsics-gen check-abi-history --base-ref …` | 1025 base entries preserved, 1029 total |
| candidate probe ×4 (llc nightly-2026-08-28 / LLVM 23.1.0-rust; ptxas CUDA 13.3 V13.3.73) at sm_89 / +ptx84 | all 4 succeeded with exact PTX |
| `cargo test -p cuda-intrinsics-gen --all-targets` | 316 passed / 0 failed |
| `cargo test -p mir-lower --test lowering_test` | 146 passed / 0 failed |
| `cargo test -p dialect-nvvm` | 57 passed / 0 failed |
| `cargo test -p cuda-intrinsics` | 1029 passed / 0 failed |
| `cargo fmt --all -- --check` (pinned toolchain) | clean |
| GPU oracle `sparse_mma_fp8` on NVIDIA L20 sm_89 (driver 570.86.10) | 4 forms × 6 metadata codes × {C=0, C≠0}, 32 lanes × 4 accumulators/lane, 0 mismatches |
| GPU layout sweep `sparse_mma_fp8 -- --probe` | 13,824 candidates, exactly one wiring fits |

The lowering test covers all four forms against both backends, asserting the
exact inline-asm template, the `=f,=f,=f,=f,f,f,f,f,r,r,r,r,r,r,r,r,r,n`
constraint string, and `AsmKind::Convergent`.

**The libNVVM route for `compute_89` is accepted.** A real harness against
`libnvvm.so.4` and `libnvJitLink.so.13` compiles all four forms to exact PTX
(7,193 bytes, `.version 9.3`, `.target sm_89`, all four instructions retained),
then to LTOIR, and links to a cubin with both ptxas and `nvJitLink`. No
capability error for any form.

## What this does not do

* **The catalog still says `unexecuted`, but the family now has a GPU oracle.**
  `crates/rustc-codegen-cuda/examples/sparse_mma_fp8` runs all four forms on an
  **NVIDIA L20 (sm_89, driver 570.86.10)** and matches an exact integer host
  GEMM: 4 forms × 6 metadata codes (`0x4 0x6 0x8 0x9 0xc 0xe`) × {`C = 0`,
  `C != 0`}, every one of the 32 lanes and 4 f32 accumulators per lane, and 0
  mismatches. Operands are integers in `1..=8`, which both FP8 formats represent
  exactly, so the comparison is equality rather than a tolerance. Run it with
  `cargo run -p cargo-oxide -- run sparse_mma_fp8`; it prints `skipping:` below
  sm_89 instead of failing. `cargo run -p cargo-oxide -- run sparse_mma_fp8 --
  --probe` re-derives the layout: a 13,824-candidate sweep executed on the device
  fits exactly one wiring, and that wiring is what the example encodes.
  `runtime_validation` nevertheless stays `unexecuted` in the shard and both
  evidence files, because the generator refuses the claim:
  `crates/cuda-intrinsics-gen/src/resolve/families/mma/sparse.rs:1043` is a hard
  `ensure!(admission.runtime_validation == RuntimeValidation::Unexecuted, …)`
  and `resolve/tests/mma.rs:2340` asserts that refusal. Flipping it would also
  require a hand-written `runtime` stage in the evidence, and **no tool in this
  repository produces one** — `EvidenceStageKind::Runtime`
  (`model/evidence.rs:207`) is only ever validated
  (`resolve/evidence.rs:773`, `:1129`) and rendered
  (`render/common.rs:223`); nothing writes into `intrinsics/evidence/`, the
  `probe` subcommand stops at PTX assembly, and no workflow references a GPU.
  Marking it `executed` by hand would assert a runtime stage that never ran, so
  the field is left alone: making the claim honestly needs a repo-native way to
  record an executed runtime stage, and this repository does not have one today.
  Filed as #1298 with the file:line detail and the question of which direction the
  maintainers want — a producer for the `Runtime` stage, or a guard that requires
  that record instead of rejecting the value outright.
* **`cargo test -p mir-lower --all-targets` does not compile**, on this revision
  or on the pristine baseline: `crates/mir-lower/src/convert/ops/control_flow.rs`
  (test-only) calls `llvm_export::ops::op_volatile`, which does not exist. Both
  that file and the whole `crates/llvm-export` crate are byte-identical to
  baseline. Left alone as unrelated scope; `--test lowering_test`, which covers
  this change, passes 146/146.
* **The evidence profile is `cuda-13.3-libnvvm-13.3.73`, not `…13.3.33`.** This
  host has V13.3.73 with different tool hashes, and naming the profile after a
  version the evidence does not record would make the profile contradict its own
  `backend_version`.
* **No SASS disassembly.** The single-`.func` probe modules carry no `.entry`, so
  ptxas dead-code-eliminates the body into a code-free stub — byte-identical to
  what the already-admitted dense FP8 family produces, so it is a probe-harness
  artifact rather than a property of these intrinsics. Real codegen is evidenced
  separately by an `.entry`-wrapped four-form fixture producing a 3,368-byte
  cubin with a 384-byte `.text.sparse_fp8_all_probe`. No disassembler for the
  CUDA 13.3 "Std ELF" cubin format was available, so "real SASS" is established
  by a non-empty `.text`, not by decoding the opcode.
* ptxas prints an advisory for these plain forms suggesting the
  `sp::ordered_metadata` modifier. It is informational, not a rejection.

## Note for reviewers

I made no repository or global configuration changes to work around this
environment; `github.com` is unreachable from the build host, so the pinned
`pliron` git dependency was reached through an ephemeral `GIT_CONFIG_*`
`insteadOf` rewrite for that shell only.

Refs #337 — this completes Phase 3's four entries and does not close the
tracking issue, which still has open Phase 1 items.


nihalpasham (dismissed):
I fixed the current-main integration and strengthened the GPU check:

- Updated two lowering-test calls for pliron's current inline-asm API.
- Changed the test inputs so swapping B fragment registers is detectable, and added a regression for all six register swaps.
- Covered all twelve legal standard-metadata codes, corrected the example lockfile, and fixed strict Clippy failures.

At `0b462e34`, the generator, ABI, 2,378 PTX routes, lowering and dialect checks pass. Real libNVVM accepts `compute_89`. On B200, all 12,288 results match the host GEMM exactly; memcheck and synccheck report zero errors.

This is ready from my local review. Catalog runtime evidence remains separate under #1298; no performance claim is made.

nihalpasham (dismissed):
I fixed the current-main integration and strengthened the GPU check:

- Updated two lowering-test calls for pliron's current inline-asm API.
- Changed the test inputs so swapping B fragment registers is detectable, and added a regression for all six register swaps.
- Covered all twelve legal standard-metadata codes, corrected the example lockfile, and fixed strict Clippy failures.

At `0b462e34`, the generator, ABI, 2,378 PTX routes, lowering and dialect checks pass. Real libNVVM accepts `compute_89`. On B200, all 12,288 results match the host GEMM exactly; memcheck and synccheck report zero errors.

This is ready from my local review. Catalog runtime evidence remains separate under #1298; no performance claim is made.

nihalpasham (approved):
I fixed the current-main integration and strengthened the GPU check:

- Updated two lowering-test calls for pliron's current inline-asm API.
- Changed the test inputs so swapping B fragment registers is detectable, and added a regression for all six register swaps.
- Covered all twelve legal standard-metadata codes, corrected the example lockfile, and fixed strict Clippy failures.

At `0b462e34`, the generator, ABI, 2,378 PTX routes, lowering and dialect checks pass. Real libNVVM accepts `compute_89`. On B200, all 12,288 results match the host GEMM exactly; memcheck and synccheck report zero errors.

The docs-only follow-up `a6dffecd` refreshes the catalog stamp and entry count after recomputing every operation-table row. Documentation guards and the full Sphinx build with warnings as errors pass. The tested code is unchanged from `0b462e34`.

This is ready from my local review. Catalog runtime evidence remains separate under #1298; no performance claim is made.

At `a6dffecd`, hosted tests, lint/format checks, guards, docs/book and example compilation passed. CodeQL was still running at the final review check.

PR #1297, opened 2026-09-18
test(atomics): cover shared-memory AtomicPtr paths

YunusMutlu:
## Summary

Adds regression coverage for shared-memory `core::sync::atomic::AtomicPtr` paths introduced by the storage-aware atomic lowering in #1288. The existing implementation already supports these paths, but coverage did not exercise shared-memory-origin pointer values, source-level `AtomicPtr` storage backed by shared memory, or the lower-level combination of typed AS3 storage carrying generic pointer values.

## Changes

- Add an end-to-end `AtomicPtr` test that round-trips a pointer originating in `SharedArray` through load/store/swap/compare_exchange and verifies the loaded pointer remains dereferenceable.
- Add an end-to-end test with `AtomicPtr` backing storage physically located in shared memory.
- Add `mir-lower` regression coverage for typed shared (AS3) atomic storage carrying generic pointer values through load/store/xchg/compare_exchange.
- Assert that RMW/CAS storage remains in AS3 while pointer values/results remain generic (AS0), preserving the existing atomic value invariant.
- Update the atomics example documentation and expected output from 22 to 24 runtime tests.

## Testing

- [ ] `just check` passes (the local mirror of CI: fmt, clippy, tests, guards, docs)
  - Currently blocked by an existing `upstream/main` failure while compiling `mir-lower` tests:
    `crates/mir-lower/src/convert/ops/control_flow.rs:1059` references the missing
    `llvm_export::ops::op_volatile` helper.
  - The same call is present in the current `upstream/main`, independently of this change.
- [x] `cargo oxide run atomics` passes, or `scripts/smoketest.sh -o '^<example>$'`
  - `=== SUCCESS: All 24 runtime atomic tests passed! ===`

Additional validation:

- `cargo test -p mir-lower --test lowering_test shared_storage_preserves_generic_pointer_atomic_values -- --nocapture`
  - 1 passed
- `cargo test -p mir-lower --test lowering_test atomic_storage:: -- --nocapture`
  - 8 passed
- `just test-cuda`
  - passed
- `just fmt-check`
  - passed
- `just check-guards`
  - passed
- `git diff --check`
  - passed

## Checklist

- [x] SPDX headers on new source files (not applicable; no new source files added)

---


nihalpasham (approved):
I kept the shared-pointer coverage and rebased it onto current main. The two source cases distinguish a pointer originating in shared memory from an atomic slot stored there; the lowering check also keeps AS3 storage separate from AS0 pointer values.

All 441 lowering tests pass. All 24 atomic GPU checks pass through native LLVM and modern libNVVM on RTX 5090, with zero native memcheck errors. All seven hosted workflows pass on this exact head.

This adds coverage for #1288; it does not change the atomic implementation.

PR #1299, opened 2026-09-18
fix(mir-lower): use the upstream volatility interface in the unroll test

sachinsharma3191:
## Summary

`main` does not compile its tests, and both the `clippy` and `cargo test -p mir-lower` jobs have been red since #1293 merged:

```text
error[E0425]: cannot find function `op_volatile` in module `llvm_export::ops`
    --> crates/mir-lower/src/convert/ops/control_flow.rs:1059:39
     |
1059 |             assert!(llvm_export::ops::op_volatile(&ctx, store.get_operation()));
```

Two PRs that were green on their own bases crossed:

- #1273 (merged 01:10) replaced the local `llvm_export::ops::op_volatile` helper with upstream Pliron's `VolatilityOpInterface`, deleting the helper and its attribute key.
- #1293 (merged 07:12, written earlier) added `unrolled_loop_preserves_assertions_and_stores`, which still calls the deleted helper.

Neither PR's own CI run could see the other's change, so the break only appears on the merge result.

## Changes

- Query the store through `VolatilityOpInterface::is_volatile`, the same way `convert/ops/memory/tests/access.rs` already does, and import that interface in the test module.

The assertion itself is unchanged: the volatile stores #1293 checks for are still required.

## Testing

From the repository root on `b9847e95` plus this commit:

```text
cargo test -p mir-lower                                 295 + 145 passed, 7 doc ignored
cargo clippy -p mir-lower --all-targets -- -D warnings  clean
cargo fmt -p mir-lower -- --check                       clean
```

`cargo test -p mir-lower` fails to compile without this change, so #1293's new test runs here for the first time; it passes.

## Checklist


nihalpasham (commented):
Thanks, @sachinsharma3191 — this is the right replacement, and it keeps the volatile-store assertion intact.

The exact change is already on main in `b0f961df` (`fix(tests): use the upstream volatility interface`). I compared the affected file against current main and there is no remaining difference to apply. I'll leave the PR open for manual disposition.

nihalpasham:
Closing as already-fixed

PR #1301, opened 2026-09-18
fix(codegen): warn when llc's LLVM major differs from rustc's

### LINKED ISSUE #1300: llc is the pipeline anchor, but nothing checks it against the LLVM that wrote the IR
## Problem

`llc` is the pipeline's anchor. `LlvmToolchain::resolve` picks it first, reads its major from `llc --version`, and then requires `opt` and `llvm-link` to share that major, reporting a prominent diagnostic when an explicit override disagrees (`crates/cuda-oxide-codegen/src/llvm_tools.rs`).

Nothing checks the anchor itself against the LLVM that *produces* the IR `llc` consumes. `rustc` writes the module; its LLVM major is visible in `rustc -vV` (`LLVM version:`), and `cargo-oxide` already reads that output for the backend cache fingerprint (`crates/cargo-oxide/src/backend.rs`). Tool selection never consults it.

So an `llc` from an unrelated toolchain is accepted silently. `resolve_llc` prefers `sysroot_tool("llc")`, which resolves through the `rustc` on `PATH`; when that is a different toolchain from the one compiling the crate, the pipeline writes IR with one LLVM and lowers it with another, with no warning.

## Why it matters

This is what made #1234 reproduce for some people and not others. From that thread:

> rustc was correctly the pinned `nightly-2026-08-28`, but `llc` resolved out of a different toolchain that happened to be the rustup default. So the IR was written by LLVM 23 and lowered by LLVM 22, and LLVM 22 still selects the instruction — which is exactly why the bug vanished for me.

The same reproducer emitted a CAS loop under one `llc` and `atom.global.add.f32` under another, so "does this reproduce?" was answered by the `llc` major alone. Triage spent a round on it, and the mismatch was invisible in every build log.

Credit to @vyncint, who diagnosed this while investigating #1234 and offered to file it separately.

## Proposal

Record a selection diagnostic when the chosen `llc`'s major differs from `rustc`'s, naming both sides and the remedy that applies (unset `CUDA_OXIDE_LLC`, or install `llvm-tools` for the pinned toolchain).

A warning rather than a hard error: mixing majors is frequently tolerable, and failing closed would break setups that work today. If maintainers prefer it fail-closed, that is a one-line change in the same place.

## Follow-up, deliberately not bundled

`cargo oxide doctor` prints `llc`'s version and enforces the LLVM 21 floor, but makes no comparison either, so `doctor` still reports a mismatched setup as healthy. Doing that without copying the parser needs `cargo-oxide` to depend on `cuda-oxide-codegen` (it does not today), or a shared home for the helper — a dependency call for maintainers rather than something to slip into the fix.

sachinsharma3191:
Closes #1300.

## Summary

`llc` anchors the LLVM pipeline: `LlvmToolchain::resolve` reads its major and then requires `opt` and `llvm-link` to match, warning when an explicit override disagrees. Nothing checked the anchor against the LLVM that writes the IR it lowers, so an `llc` from an unrelated toolchain was accepted in silence and changed what the backend emitted.

`resolve_llc` prefers `sysroot_tool("llc")`, resolved through the `rustc` on `PATH`. When that differs from the toolchain compiling the crate, IR written by LLVM 23 is lowered by LLVM 22 with no diagnostic.

This is how #1234 was diagnosed twice: the same reproducer emitted a CAS loop under one `llc` and `atom.global.add.f32` under another, so the `llc` major alone decided whether the bug appeared. Credit to @vyncint for finding it there.

## Changes

- `rustc_llvm_mismatch_warning`: names both sides and the remedy that applies — `unset CUDA_OXIDE_LLC` when `llc` came from the override, otherwise install `llvm-tools` for the pinned toolchain.
- `LlvmToolchain::resolve` pushes it into the existing `diagnostics` channel, next to the `llvm-link` mismatch warning, so library code still never writes to stderr itself.
- `parse_rustc_llvm_major`: `rustc -vV` spells it `LLVM version: 23.1.0`, with a colon the existing `--version` banner parser does not accept, so the rustc banner gets its own parser.

A warning, not an error: mixing majors is often tolerable and failing closed would break setups that work today. Trivial to flip in one place if you would rather it be fatal.

## Testing

```text
cargo test -p cuda-oxide-codegen --lib llvm_tools     15 passed (3 new)
cargo fmt -p cuda-oxide-codegen -- --check            clean
cargo clippy -p cuda-oxide-codegen --lib -- -D warnings   clean
```

New tests, all pure-function, following the existing style in that module:

- `parse_rustc_llvm_major_reads_the_vv_line` — parses a real `rustc -vV`, and pins that neither parser accepts the other's spelling, so they cannot silently swap.
- `rustc_llc_mismatch_warns_and_names_both_sides` — both binaries appear, and each remedy matches how `llc` was selected.
- `rustc_llc_match_or_unknown_version_is_quiet` — equal majors stay silent, and an unparseable banner on either side never invents a mismatch.

Not run here: this machine has no CUDA toolkit, so `tests/libdevice_linking.rs` fails on `libdevice.10.bc was not found in any CUDA installation`, before and after this change alike. The rest of the crate's 1235 tests pass.

## Follow-up, deliberately not bundled

`cargo oxide doctor` prints `llc`'s version and enforces the LLVM 21 floor but makes no comparison, so it still calls a mismatched setup healthy. Wiring it up without copying the parser needs `cargo-oxide` to depend on `cuda-oxide-codegen` (it does not today), which is a dependency decision for you rather than something to slip into this fix.

## Checklist


nihalpasham (changes requested):
The warning needs the identity of the compiler actually running the backend. Spawning `rustc -vV` reads PATH again, so it can agree with the same wrong toolchain that supplied `llc`.

I compiled a resolver probe with our pinned LLVM 23 toolchain, then ran it with the installed LLVM 22 toolchain on PATH: it selected LLVM 22 `llc` and returned no diagnostics. The standalone codegen API also writes its own LLVM text, so PATH rustc is not its producer.

Please carry the relevant identity from the owning frontend and cover both routes through the resolver. This does not close #1300 yet.

PR #1305, opened 2026-09-19
feat(cuda-device): derive an SM floor cfg and use redux.sync in warp_reduce

vyncint:
Implements #811 as you specced it in option 1. **Stacked on #1304 — review only the second and third commits.**

Device code had no way to ask what target it was compiled for, so `warp_reduce` always paid five shuffles plus five combines even when the build was certainly Ampere. #811 measured that at 4.74x for the primitive (159.4 us against 33.6 us per launch on an A10G). This is the cfg that lets it take the one-instruction form.

## What it does

`cuda-device` gains a build script that resolves the floor — `CUDA_OXIDE_TARGET`, else `CUDA_OXIDE_DEVICE_ARCH`, else sm_80, your table exactly — and emits `cuda_oxide_sm_at_least = "XY"` for every rung at or below it, cumulative because `cfg` compares strings and not numbers. The rule lives in `cuda-target-spec`, shared with the backend rather than restated, so the cfg floor and the selected target read the same two variables through the same function and cannot disagree. A named target that does not parse fails the build: guessing high emits instructions the target cannot run, guessing low silently drops the faster form.

Rust has no specialization, so the reduction reaches a form that exists for only some `(Op, T)` pairs through a provided `ReduceOp::reduce_full_warp` hook plus a `const HAS_FULL_WARP_FORM`. `warp_reduce` consults it only at `N == 32`; everything else collapses to the butterfly at compile time.

## PTX, per kernel, at both floors

The measurement you asked for. Counts are per `.visible .entry`, since the controls legitimately keep their shuffles:

| kernel | sm_80 | sm_75 |
|---|---|---|
| `reduce_u32_all_ops` (6 integer ops) | **redux.sync 6**, shfl 0 | redux 0, **shfl 30** |
| `reduce_i32_signed_ops` (3 signed ops) | **redux.sync 3**, shfl 0 | redux 0, **shfl 15** |
| `reduce_f32_control` | redux 0, shfl 15 | redux 0, shfl 15 |
| `reduce_subwarp_control` (`WarpTile<16>`) | redux 0, shfl 4 | redux 0, shfl 4 |

Six ops × five rounds = 30 shuffles at sm_75 become six instructions at sm_80. The two controls are identical at both floors, which is the point: `f32` has no `redux` before sm_100, and the sub-warp tile is gated on `N == 32`.

## Correctness on hardware

A10G, compute capability 8.6, driver 595.71.05.

```
auto-detected (sm_86, so the fast path is live)
  [sum, min, max, and, or, xor] = [496, 0, 31, 0, 4294967295, 4294967295]   ✓
  [sum, min, max] i32           = [-16, -16, 15]                            ✓
  [sum, min, max] f32           = [496.0, 0.0, 31.0]                        ✓
  sub-warp tiles                                                            ✓
forced --arch sm_75 (below the floor, butterfly)
  all four                                                                  ✓
smoketest -o '^(warp_reduce_redux|redux_minmax|coop_groups_demo)$'   3 / 3
```

The signed values straddle zero on purpose: `redux.sync.min.s32` and `.u32` disagree on exactly those inputs, and only the signed answer is `-16`.

## The example is new, and it had to be

There was nowhere to read this. The only caller of the generic `warp_reduce` in the tree is `coop_groups_demo`, which also uses thread-block clusters, so it cannot be built below sm_90 at all — my first attempt at PTX evidence came back `CUDA target sm_80 cannot lower detected feature Cluster`. `warp_reduce_redux` is one warp and nothing else, so it builds at both floors and carries the `f32` and `WarpTile<16>` controls from your spec.

## Two things to flag

**The lock churn is real.** The build-dependency puts the first `cuda-device` → `cuda-target-spec` edge in the graph, and `cuda-device` is in every kernel crate's, so 227 example locks plus the `cuda-macros` device-only lock each gain one line. `check-example-license-policy.sh` and `check-device-only-build.sh` fail on a stale lock, so it is mandatory rather than cosmetic: exactly one added line per file, no version moves, and a build-dependency reaches only the host side so no kernel graph changes. If you would rather not carry that, the alternative is restating the floor ladder inside `build.rs` — which is the second-implementation-of-one-policy mistake you had me remove from #1253, so I did not take it on my own.

**Why this is stacked.** At `--arch sm_80` the module was rejected for `ReduxF32` with no f32 redux in it: the feature scan splits LLVM IR on `;`, which is a comment marker there, so `redux.sync` and `.f32` matched 547 lines apart. That is #1303, fixed in #1304. This branch cannot build at its own target floor without it.

On "the same floor goes into target selection as a minimum": both sides read the same two variables through the same function, so they agree in all three rows of your table, and a module emitting `redux.sync` is raised to sm_80 by the requirement scan regardless. I pinned that with tests rather than adding a clamp in the backend; say the word if you want the clamp as belt-and-braces.

`cargo test -p cuda-target-spec -p cuda-device` passes at every floor, fmt and clippy `-D warnings` clean, all 17 `scripts/check-*.sh` pass.

Note #1256 appends to `cuda-target-spec/src/lib.rs` right after `spelling_at_least`, where this also adds — different intent, but whichever lands first the other rebases.


nihalpasham (changes requested):
The integer reduction fast path is useful, but the cfg floor still needs to be part of target selection before this lands.

`resolve_sm_floor` is only called by `cuda-device/build.rs`. With no environment target, it advertises 80 while an sm_75-feature module selects sm_75. With a detected sm_120 hint and WGMMA, it advertises 120 while the backend selects sm_90a. I reproduced both through the actual resolver; the lower-bound assertion fails.

Please carry one resolved floor into native PTX and NVVM selection, handle incompatible architecture families explicitly, and add cross-stage regressions for these cases and invalid/blank inputs. The current tests only verify the new helper's own table.

This partially addresses #811; #1304 remains a prerequisite.

vyncint:
You were right that the lower bound did not hold, and chasing it turned up that the rule itself — the middle row of the table on #811 — is the unsound part, not just my wiring of it.

## Why a detected device cannot raise the floor

`CUDA_OXIDE_DEVICE_ARCH` is advisory by contract, said three times in the tree: *"the backend builds for the detected GPU only when that GPU can actually run the kernel"*, and otherwise for the arch the kernel requires (`apply_device_arch_hint`, `detect_run_target_arch`, and the fallback in `resolve_ptx_target_with_generated`). A lower bound taken from a value the backend is documented to override is not a lower bound, which is exactly your second case: detected sm_120 with WGMMA advertises 120 and lands on sm_90a.

So the rule is now asymmetric, and that is what makes the assertion hold:

| | floor |
|---|---|
| explicit target | that target, exactly — it is what gets built |
| detected device | **lowers** the floor, never raises it |
| neither | `DEFAULT_SM_FLOOR` (80) |

The hint says nothing about how *high* selection goes. It does bound how *low*, because a device that can run the kernel is selected as-is. That direction matters: `cargo oxide run` on a Turing GPU must keep producing sm_75, and a blanket clamp to 80 would have built PTX that GPU cannot load. Pulling the floor down instead keeps device code honest about what it will get.

## Carried into both selectors

Both `resolve_ptx_target_with_generated` and `resolve_nvvm_target_with_generated` now resolve the floor through the same `resolve_sm_floor` the build script uses — not a second copy of the rule — and raise a feature-selected target that lands below it. A floor that cannot host the module is reported rather than resolved to something else; nothing reaches that arm today, and it is there so a future floor cannot pick an architecture family silently.

Your two cases, now:

```
no env, ldmatrix-only module     PTX sm_80   NVVM sm_80    (was sm_75)
hint sm_120a, WGMMA module       PTX sm_90a  NVVM sm_90a   floor 80, so no contradiction
hint sm_75, basic module         PTX sm_75   (from "detected GPU"), floor 75
```

The regression you asked for is the property itself, not those three rows: a matrix over explicit × hint × features asserting `selected.capability() >= resolve_sm_floor(explicit, hint)` through **both** resolvers. Plus the blank and malformed inputs, where the two stages deliberately differ — a blank `CUDA_OXIDE_TARGET` is unset to the floor rule and a parse error to the backend, and the strict side is the one that decides whether anything is built; a malformed *hint* is ignored by both, matching the `.parse().ok()` filter in the PTX path.

## One behaviour change, stated because it reaches builds that never use this cfg

With nothing pinned, a module whose features ask only for sm_75 is now built for sm_80. That is the documented Ampere+ minimum and the same default a featureless module already got, and `--arch sm_75` still selects sm_75. It changed one existing test, which pinned the old semantics; I split it so it still asserts the feature's own floor is sm_75 and separately records the selected target.

## Hardware

A10G, compute capability 8.6. Per-kernel PTX at both floors, unchanged from the last revision:

| kernel | sm_80 | sm_75 |
|---|---|---|
| `reduce_u32_all_ops` | **redux.sync 6**, shfl 0 | redux 0, **shfl 30** |
| `reduce_i32_signed_ops` | **redux.sync 3**, shfl 0 | redux 0, **shfl 15** |
| `reduce_f32_control` | redux 0, shfl 15 | redux 0, shfl 15 |
| `reduce_subwarp_control` | redux 0, shfl 4 | redux 0, shfl 4 |

Run on the A10G with nothing pinned (hint sm_86, floor 80, fast path live): all four checks correct. `generated_ldmatrix` unpinned now reports `.target sm_80`, the behaviour change above. Smoketest 4/4 over `warp_reduce_redux`, `redux_minmax`, `redux_f32`, `generated_ldmatrix`.

1218 codegen tests, plus `cuda-device` and `cuda-target-spec`; clippy `-D warnings` clean on all three; fmt; all 17 guards. Still stacked on #1304 — review the second and third commits.

nihalpasham (changes requested):
The earlier floor-contract blocker is fixed. Both selectors now use the shared floor: no-env `ldmatrix` stays at least sm_80, and an sm_120 hint with WGMMA advertises floor 80 while selecting sm_90a. I shortened the comments and corrected the outdated cfg and reduction-hook descriptions.

At `db5c23f8`, all 1,394 tests/doctests across cuda-device, cuda-target-spec and cuda-oxide-codegen passed, along with strict Clippy and formatting. B200 runs of `warp_reduce_redux` passed for sm_80 and sm_75 through LLVM NVPTX. The sm_80 fast path also passed memcheck and synccheck with 0 errors; float and sub-warp controls passed in both builds. No new performance measurement was made.

This still depends on #1304, whose current detector misses valid float-redux asm and misreads comments/ordinary symbols. Please fix those lexical cases there and update this stack before landing. The remaining request is the scanner prerequisite, not the repaired floor invariant.

One minor distinction: malformed hints are ignored by the shared floor resolver and PTX selector, but the existing NVVM selector rejects them. The floor guarantee is still fail-closed.

vyncint:
Thanks for the B200 runs and the memcheck/synccheck pass — that is evidence I could not have produced on an A10G.

Rebased onto the updated #1304 (`279ffe0f`). Your `docs: clarify the shared SM floor contract` is carried through unchanged: all five files it touches are the same blobs as in `db5c23f8`, checked by object hash rather than by reading the diff.

On the minor distinction you noted: you are right, and I had it wrong in two places. My last comment here said a malformed hint "is ignored by both", and the test comment said "both stages ignore it" — true of the floor resolver and the PTX selector, not of the NVVM selector, which rejects it. One small commit on top, `96ea0b56`, makes the comment name all three readings and asserts the NVVM rejection beside the PTX fall-through the test already checked, so the difference is a stated contract rather than something a later change could erase unnoticed. No behaviour changes; as you say, both outcomes are fail-closed.

Across the stack: `cuda-target-spec` 15, `cuda-device` 114, `cuda-oxide-codegen` 1266 — 1395, one more than your 1394 at `db5c23f8`, which is the one test function #1304 gained. Clippy `-D warnings` clean on all three, fmt clean, all 17 guards pass.

Stack is now `279ffe0f` (#1304) ← `6dcbf00d` ← `df4ac829` ← `c87f5f7e` (yours) ← `96ea0b56`.

nihalpasham (changes requested):
Checked `96ea0b56`: the floor logic, reduction hooks and all five documentation-cleanup files are unchanged from the tested `db5c23f8`. The new test correctly pins NVVM's rejection of malformed hints beside PTX's fall-through.

All 1,395 tests/doctests across cuda-device, cuda-target-spec and cuda-oxide-codegen pass from an isolated build directory, along with strict Clippy and formatting. The earlier B200 sm_80/sm_75 runs and fast-path memcheck/synccheck covered the unchanged reduction implementation; they were not repeated for this scanner rebase.

The original four #1304 cases are fixed, but its PTX scan still misses legal whitespace/comment-separated modifiers and mistakes a quoted `.file` path for an instruction. Those cases are accepted by LLVM and ptxas, and the actual detector chooses the wrong targets. Please address that remaining prerequisite and update this stack. The floor invariant itself remains clear.

vyncint:
Rebased onto #1304's `6b4ac7d0`. The rebase changed nothing in this stack except #1304's two files — `git diff 96ea0b56 cf4c55c3` touches only `detect.rs` and its test — and your five cleanup files are still the blobs you tested at `db5c23f8`.

`cuda-target-spec` 15, `cuda-device` 114, `cuda-oxide-codegen` 1267: 1396, your 1395 plus the one test #1304 gained. Clippy `-D warnings` clean on all three, fmt clean.

PR #1307, opened 2026-09-19
fix(cargo-oxide): probe versioned clang binaries in doctor

### LINKED ISSUE #1302: cargo oxide doctor's clang check fails after following its own 'apt install clang-21' remedy
## Summary

`cargo oxide doctor`'s clang check probes the unversioned `clang` binary, but the remedy it prints installs a *versioned* one. Follow doctor's own instruction and doctor still fails — with an environment that is, by then, working.

## Repro

```
$ cargo oxide doctor
clang / libclang resource dir... ✗ clang not found
  Host `cuda-bindings` uses bindgen, which needs clang + its resource headers.
  Install with: sudo apt install clang-21
  (or at minimum `libclang-common-21-dev` alongside your libclang)

$ sudo apt install clang-21        # exactly what it said

$ cargo oxide doctor
clang / libclang resource dir... ✗ clang not found     # unchanged
```

## Why

`doctor.rs:645` runs `Command::new("clang")`. On Debian/Ubuntu the `clang-21` package installs `/usr/bin/clang-21` and does not provide `/usr/bin/clang` — that comes from the unversioned `clang` metapackage, or from `update-alternatives`. So the check and its own remedy disagree about the binary's name.

On this machine after the install:

```
command -v clang     -> (none)
command -v clang-21  -> /usr/bin/clang-21
clang-21 -print-resource-dir
                     -> /usr/lib/llvm-21/lib/clang/21
/usr/lib/llvm-21/lib/clang/21/include/stddef.h   exists
```

## This is a false negative, not just a cosmetic one

The thing doctor is actually guarding is bindgen's ability to find clang's own `stddef.h` — and that works here. Before the install, a host build died with

```
/usr/include/stdlib.h:32:10: fatal error: 'stddef.h' file not found
```

After installing `clang-21`, that error is gone and the build proceeds to the next (unrelated, toolkit-completeness) failure. So the resource headers doctor is looking for are present and usable, while doctor reports them missing. A user who trusts the check reinstalls a package they already have.

Worth noting bindgen never runs the `clang` binary at all — it loads `libclang.so`. The binary is a proxy for "are the resource headers there", which is why the versioned/unversioned distinction is invisible to everything except this check.

## Suggested fix

Probe `clang`, then the versioned names (`clang-21`, `clang-20`, …) rather than only the bare one, and name the binary that answered in the ✓ line so the resolution is visible. The resource-dir check underneath it stays as it is. Happy to write it — note it touches `doctor.rs`, which #1276 also touches (different function; a rebase, not a conflict of intent), and #1278 is in there too.

## Environment

Ubuntu, clang-21 from apt, cuda-oxide at `b9847e95`.

letv1nnn:
## Summary

cargo oxide doctor's clang check only ever ran the bare clang binary, so on Debian/Ubuntu, where installing clang-21 gives you only a clang-21 binary and no unversioned clang, the check's own suggested remedy could never satisfy the check. This falls back through a list of versioned clang names (clang-21 down to clang-15) and reports which binary actually answered, so the check reflects reality instead of a single hardcoded name.

Closes #1302 

## Changes

- `crates/cargo-oxide/src/commands/doctor.rs`: clang/libclang resource-dir check now probes clang, then clang-21...clang-15 in order, stopping at the first binary that responds to --version.

## Testing

- [x] cargo test -p cargo-oxide — 232 passed, 0 failed
- [x] cargo clippy -p cargo-oxide --all-targets -- -D warnings — clean

## Checklist


letv1nnn:
@vyncint, would appreciate the review

nihalpasham (dismissed):
I tightened the fallback so a compiler that answers `--version` but fails to report its resource directory does not hide a usable versioned Clang. Empty responses are skipped too, and the candidates now include `clang-22`.

The executable-probe regressions cover those failures and preserve first-success precedence. At `bbdcbe81`, all 234 cargo-oxide tests and strict Clippy passed. Full repository checks and hosted CI have not been rerun here.

nihalpasham (approved):
I tightened the fallback so a compiler that answers `--version` but fails to report its resource directory does not hide a usable versioned Clang. Empty responses are skipped too, and the candidates now include `clang-22`.

The executable-probe regressions cover those failures and preserve first-success precedence. At `bbdcbe81`, all 234 cargo-oxide tests and strict Clippy passed.

At `bbdcbe81`, hosted tests, lint/format checks, guards, docs/book and example compilation passed. CodeQL was still running at the final review check.

PR #1309, opened 2026-09-19
refactor(llvm): use upstream inline asm semantics

### LINKED ISSUE #1308: Migrate LLVM inline asm semantics to upstream Pliron attributes
## Context

cuda-oxide currently carries private inline-asm semantic attributes for concepts that are now modeled directly by `pliron-llvm`.

Pliron PR pliron-org/pliron#253 added upstream LLVM inline-asm support for:

* `llvm_inline_asm_side_effects`
* `llvm_inline_asm_attrs`
* generic LLVM call-site attributes such as `convergent`

This makes the cuda-oxide-specific representations redundant.

There is also a semantic compatibility hazard when updating Pliron: the final `bool` argument of `InlineAsmOp::new` now represents `side_effects`, whereas the previous API used that position for `convergent`. Updating the dependency without migrating the call sites can therefore compile while silently changing semantics.

## Proposed change

Migrate cuda-oxide inline asm modeling to the upstream Pliron representation:

* keep `AsmKind` only as a cuda-oxide builder/convenience classification;
* stop persisting `cuda_oxide_asm_kind`;
* use `llvm_inline_asm_side_effects` for LLVM `sideeffect`;
* use `llvm_inline_asm_attrs` with the `convergent` unit attribute for convergent inline asm;
* update the LLVM exporter to consume the upstream attributes directly;
* update inline PTX lowering and tests to use the upstream generated accessors;
* remove the private `cuda_oxide_inline_asm_sideeffect` representation.

The Pliron dependency update also requires mechanical compatibility changes for APIs that changed in the same upstream revision, including generated LLVM attribute accessors and `AllocaOp::new`.

## Acceptance criteria

* No production use of `cuda_oxide_asm_kind`.
* No production use of `cuda_oxide_inline_asm_sideeffect`.
* No use of the old inline-asm generated accessors:

  * `get_attr_inline_asm_template`
  * `get_attr_inline_asm_constraints`
  * `get_attr_inline_asm_convergent`
* Existing inline PTX semantics are preserved for all four `AsmKind` combinations:

  * convergent + side-effecting
  * convergent + pure
  * non-convergent + side-effecting
  * non-convergent + pure
* LLVM export continues to emit the correct `sideeffect` and `convergent` semantics.
* Workspace tests, Clippy, formatting, and `just check` pass.

uurl:
Closes #1308 

## Summary

Migrate cuda-oxide inline asm modeling to the upstream semantics provided by `pliron-llvm`.

The Pliron dependency is updated to revision `06d99a3df938394583715aacf5eadb65b4791c09`, which includes the upstream inline-asm and generic LLVM attribute support from Pliron PR pliron-org/pliron#253.

## What changed

* Removed the private `cuda_oxide_asm_kind` attribute.
* Removed the private `cuda_oxide_inline_asm_sideeffect` attribute.
* Kept `AsmKind` as a cuda-oxide convenience classification only.
* Mapped `AsmKind` onto upstream LLVM semantics:

  * side effects via `llvm_inline_asm_side_effects`;
  * convergence via `llvm_inline_asm_attrs` containing the `convergent` unit attribute.
* Updated `InlineAsmOpExt::build` to construct upstream-compatible inline asm.
* Updated `asm_kind` / `asm_kind_opt` to derive the cuda-oxide classification from upstream attributes instead of private state.
* Updated inline PTX lowering so the final `InlineAsmOp::new` argument is `sideeffect`, with `convergent` applied separately through the upstream call-site attribute representation.
* Updated the LLVM exporter to consume upstream inline-asm attributes directly.
* Updated tests to use the new `llvm_`-prefixed generated accessors and presence-based `convergent` semantics.

## Pliron 0.18 compatibility

The dependency bump also requires several mechanical updates introduced by the same upstream revision range:

* generated LLVM attribute accessors now use the `llvm_` prefix;
* `AllocaOp::new` now takes an explicit address space;
* `clone_skip_outlined` now takes the context;
* dictionary/static key handling was updated for the newer Pliron API;
* volatile test code now uses `VolatilityOpInterface` directly.

These changes are included with the dependency bump because an intermediate Pliron update would not be semantically safe for inline asm.

In particular, the final `bool` argument of `InlineAsmOp::new` changed meaning upstream from `convergent` to `side_effects`. Keeping the dependency update and call-site migration atomic avoids a state that compiles but models inline asm incorrectly.

## Validation

* `cargo fmt --all`
* `cargo test -p llvm-export --all-targets --locked`
* `cargo test -p mir-lower --all-targets --locked`
* `cargo clippy -p dialect-mir --all-targets -- -D warnings`
* `just check`
* `git diff --check`

Additional migration checks:

```text
get_attr_inline_asm_(template|constraints|convergent)
cuda_oxide_asm_kind
cuda_oxide_inline_asm_sideeffect
```

produce no remaining matches in the migrated code.


nihalpasham (approved):
I kept the dependency and constructor migration together, and added one exporter check for the new generic attribute container. The four existing inline-asm combinations remain supported; attributes such as `noduplicate` now produce a clear error instead of silently disappearing, and `convergent` requires its unit form.

The added tests exercise upstream constructors and verifier-valid unsupported attributes. `just check` passes with 5,425 tests/doctests. The inline-PTX and barrier examples also pass on RTX 5090, with clean memcheck/synccheck. All seven hosted workflows passed at `85eb881c`.

This addresses #1308 within the existing inline-asm attribute scope.

PR #1311, opened 2026-09-20
fix(cuda-device): handle partial warps in block collectives

### LINKED ISSUE #1310: cooperative_groups: block_reduce/block_scan mishandle partial final warps
## Summary

`block_reduce` and `block_scan` in `cuda_device::cooperative_groups` do not correctly handle thread blocks whose size is not a multiple of the physical warp size.

For example, a 48-thread block contains one full 32-lane warp and one 16-lane final warp. The current implementation partitions the block into `WarpTile<32>` groups and uses the regular full-warp reduction/scan path for every warp. As a result, the final partial warp can execute shuffle collectives with a full `0xffffffff` member mask even though only the low 16 lanes participate.

This violates the shuffle participation contract and can produce incorrect results depending on the GPU / generated code.

There is also a related shared-memory sizing issue: the runtime number of warps is `ceil(block_threads / 32)`, so scratch storage must have capacity for that many warp totals before any `SharedArray` indexing occurs.

## Reproduction

A minimal reduction kernel is:

```rust
#[kernel]
pub fn repro(mut out: DisjointSlice<u32>) {
    static mut SMEM: SharedArray<u32, 2> = SharedArray::UNINIT;

    let gid = thread::index_1d();
    let block = this_thread_block();

    let total = block_reduce::<u32, Sum, 2>(&block, 1u32, &raw mut SMEM);

    if gid.in_bounds(out.len()) {
        unsafe {
            *out.get_unchecked_mut(gid.get()) = total;
        }
    }
}
```

Launch it with:

```text
block_dim = (48, 1, 1)
grid_dim  = (1, 1, 1)
```

Every thread should receive `48`.

On an RTX 3050 Ti (`sm_86`), I observed the pre-fix/current upstream implementation produced `64` instead.

A partial-warp `block_scan` can sometimes still produce the correct output despite using an invalid full-warp mask, so an output-only scan test does not reliably expose the contract violation on every GPU.

## Shared-memory capacity

For a block with `block_threads` threads, the runtime scratch requirement is:

```text
runtime_warps = ceil(block_threads / 32)
```

`NUM_WARPS` therefore needs to represent shared-memory capacity rather than requiring an exact block-size relationship:

```text
runtime_warps <= NUM_WARPS
```

Extra capacity should be valid and should not be read.

This check also needs to happen before any per-warp `SharedArray` indexing. Otherwise a launch whose runtime warp count exceeds the supplied capacity can index outside the declared shared-memory object.

## Expected behavior

- Full 32-lane warps should keep using the existing full-warp fast path.
- A partial final warp should execute collectives only across its live contiguous lane prefix.
- `NUM_WARPS` should be treated as scratch capacity and must be at least `ceil(block_threads / 32)`.
- Extra scratch capacity should be allowed.
- An undersized scratch allocation should be rejected before any shared-memory indexing.
- `block_reduce` and `block_scan` should work for valid block sizes that are not multiples of 32.

I can follow up with a patch and regression coverage for normal partial tails, a single partial warp, non-power-of-two tails, and a one-lane final warp.

YunusMutlu:
Fixes #1310

## Summary

Fix `cooperative_groups::block_reduce` and `block_scan` for thread blocks with a partial final warp. The previous implementation used full-warp shuffle masks for every `WarpTile<32>`, which could include nonexistent lanes in the final warp and violate the shuffle participation contract. This also makes `NUM_WARPS` explicitly represent shared-memory capacity and validates that capacity before any per-warp scratch access.

## Changes

- Add a private contiguous-prefix warp group for partial final warps, using the live-lane mask for `shfl_up`.
- Keep the existing full-warp reduction/scan paths unchanged for complete 32-lane warps.
- Handle partial-warp reduction totals using the last live lane.
- Avoid a second full-warp collective for single-warp `block_scan`.
- Treat `NUM_WARPS` as shared-memory capacity and require `ceil(block_threads / 32) <= NUM_WARPS` before any scratch indexing.
- Restrict scratch reads/writes to the runtime warp count so extra capacity remains unused.
  - a normal partial tail (`48` threads),
  - a non-power-of-two tail (`45` threads),
  - a single partial warp (`17` threads),
  - and a one-lane final warp (`33` threads).
- Update `cooperative_groups` documentation, the supported-features table, and the cooperative-groups example documentation to describe the new capacity semantics and partial-warp support.

## Testing

- [x] `cargo oxide run warp_reduce --arch sm_86` passes on an RTX 3050 Ti
- [ ] New example added (not applicable; regression coverage was added to the existing `warp_reduce` example)

Additional validation:

- `cargo oxide build warp_reduce --arch sm_86`
- `cargo test -p cuda-device`
- `(cd crates/rustc-codegen-cuda && cargo clippy --all-targets -- -D warnings)`
- `cargo oxide fmt`
- `git diff --cached --check`

The runtime regression cases produced the expected results for `block_reduce` with 48, 17, and 33 threads and for `block_scan` with 45, 17, and 33 threads.

## Checklist

- [x] SPDX headers on new source files (no new source files added)

---


nihalpasham (approved):
I kept the partial-warp algorithm and added the scratch ownership follow-up:

- Raw element accesses avoid overlapping mutable borrows of the shared allocation.
- Both block collective APIs now require `unsafe`, with allocation, participation, and reuse requirements documented; all callers are updated.
- The regressions cover 77 launch shapes, including 3D blocks, two CTAs, maximum capacity, and scratch reuse. Undersized scratch traps.

The GPU cases, memcheck, racecheck, and synccheck pass on RTX 5090. Local device checks pass; all seven hosted workflows pass on this exact head.

This closes #1310 and changes the call sites to require `unsafe`.

PR #1314, opened 2026-09-21
fix(dialect-mir): fold shifts whose amount has a different width

midagedev:
## Summary

Constant-folding a shift whose amount has a different integer width from the shifted value panics the compiler:

```text
Internal compiler error in device codegen: assertion `left == right` failed: APInt::shl: bitwidth mismatch (32 vs 64)
```

Rust allows the shape (`u32 << usize`), and `convert_shift` in `mir-lower` already widens or narrows the amount when lowering. The folder did not: `MirShlOp`/`MirShrOp::check_fold` passed both operands straight to `APInt::shl`/`lshr`/`ashr`, which assert equal widths. `#[unroll]` is the easy way to reach it, since the pass materialises a `usize` loop counter as a 64-bit constant and then runs SCCP:

```rust
let mut i: usize = 0;
#[unroll]
while i < 1 {
    *o = 1u32 << i;
    i += 1;
}
```

A `u32` counter, or `i as u32`, avoids it; the fold is the only place that does not accept the mixed pair.

## Changes

- `dialect-mir/src/const_fold.rs`: `shift_in_range` becomes `shift_amount`, which returns the amount re-expressed at the shifted value's width, or `None` when it is `>= width` (unchanged: a Rust shift overflow is not folded). The range check reads the amount at its own width first, so a wide amount such as `1 << 32` is refused rather than truncated to `0` and folded.
- `dialect-mir/tests/const_fold.rs`: one test — wider amount for `shl`, logical and arithmetic `shr`; narrower amount (`u64 << u8`); the two refusals.

The fix mirrors the lowering rather than touching pliron's `APInt` asserts, so the folder and `convert_shift` agree on what a legal shift is.

## Testing

- The new test fails before the fix with exactly the message above and passes after.
- End to end on an RTX 3090 (sm_86, CUDA 13.0, nightly-2026-08-28), selecting the backend with `CUDA_OXIDE_BACKEND`: built from `b0f961d`, the kernel above fails with the ICE (exit 101); built from this branch it compiles and runs, all 32 elements `== 1`; a four-trip variant (`acc |= 1u32 << (2 * i)`) gives `== 85`, so the folded values are right, not merely accepted.
- `scripts/smoketest.sh --compile-only` over all 230 examples, patched backend: 227 pass, 3 fail (`gemm_sol`, `gemm_sol_final`, `tcgen05_matmul`). The `b0f961d` backend gives the same 227/3 with the same three, all for one reason: the machine's LLVM 21 `llc` does not know `llvm.nvvm.stmatrix.sync.aligned.m8n8.x2.b16.p3`. `unroll_smoke` and `unroll_bounds_check` executed on the 3090: pass.
- `just check` minus `fmt-check`, run recipe by recipe on the same machine (no `just` there): clippy, test (5359 tests, 0 failed), test-cuda, check-guards, doc-check all pass. `cargo oxide fmt --check` reports the same set on unpatched `main` and on this branch (two example manifests need a nightly Cargo feature; not touched here); the two files this PR changes are rustfmt-clean.
- [x] `cargo test -p dialect-mir`, `cargo clippy -p dialect-mir --all-targets -- -D warnings` pass
- [ ] `just check` as one command: not run (see above for the per-recipe results)
- [ ] New example: none — the dialect test carries the regression, and the kernel above is the runtime check

## Checklist

- [x] SPDX headers on new source files (none added)

The reproducer was reduced and this patch drafted with AI assistance; every result above was re-run by hand on the machine named.


nihalpasham (commented):
Thanks, @midagedev — this fixes a real compiler crash in unrolled `u32 << usize` code.

I kept your implementation and added 128-bit and invalid-count checks, exact result-type assertions, and a mixed-width `unroll_smoke` regression. The added tests catch a truncated-count bug that the original cases miss; both regressions fail with the old folder.

At `b024b5fd`, `just check` passes all 5,426 tests/doctests. Independent boundary checks pass, and the GPU regression passes through LLVM and libNVVM with zero memcheck errors. I found no introduced correctness or soundness issue for supported Rust integers. Hosted workflows await approval to run; final CodeQL monitoring remains with the maintainer.

No linked issue is closed.

nihalpasham (approved):
All 51 non-CodeQL checks passed at `b024b5fd`. I also checked the clean integration with current main (`ecbf6335`, Pliron 0.18): 1,885 affected compiler tests and the GPU regression pass, with zero memcheck errors.

Merging with the test follow-ups above. CodeQL can finish separately as agreed. No linked issue is closed.

